PR #102, opened 2015-04-20
Integrate zim reader

peter-x:
Basic support for .zim files: Title search (case sensitive!) and article decoding.
I hope I did not break evopedia archive support.

It would be great to get feedback on the performance on real FFOS hardware.


peter-x:
I had to switch from jQuery to Q for the deferred/promise implementation, since jQuery's caused "too much recursion" errors. Q seems to never directly call callbacks for already resolved promises but always waits for the "next tick".

www/js/lib/abstractBackend.js
```diff
@@ -0,0 +1,75 @@
...
+        } else {
+            var archive = new evopediaArchive.LocalArchive();
+            archive.initializeFromDeviceStorage(storage, path);
+            return archive;
+        }
+    };
+    function loadArchiveFromFiles(files) {
+        var archive = new evopediaArchive.LocalArchive();
+        archive.initializeFromArchiveFiles(files);
+        return achive;
```
mossroy: There's certainly a missing r at this line :
return archive;

www/js/lib/archive.js
```diff
@@ -388,21 +368,26 @@ define(['normalize_string', 'geometry', 'title', 'util', 'titleIterators', 'jque
...
+                        callbackFunction(titles, maxSize);
+                        return 1;
+                    }
                     titles.push(title);
                     return addNext();
                 });
             }
             return addNext();
-        }).then(callbackFunction, errorHandler);
+        }).then(function(){}, errorHandler);
```
mossroy: Is it on purpose that you replaced the call to callbackFunction by an empty function?

www/js/lib/abstractBackend.js
```diff
@@ -0,0 +1,75 @@
...
+    };
+    /**
+     *  Scans the DeviceStorage for archives
+     *
+     * @param storages List of DeviceStorage instances
+     * @param callbackFunction Function to call with the list of directories where archives are found
+     */
+    function scanForArchives(storages, callbackFunction) {
```
mossroy: I did not manage to make this function work.
It behaves as if it never finds any archive (of any format).
But when I put a breakpoint in it, the archives are in fact found, AFTER the GUI said it did not find anything.
So I suspect an issue about the Promises. Maybe because of the mix between jQuery and Q libraries?

mossroy:
Thank a lot for this code. It is VERY promising.
I like the way you did a generic interface for both backends (Evopedia and ZIM file formats).

I had to hardcode the location of my zim archive to make it work (see my comments in the code : the archive scanning process seems broken), but I managed to display some articles from a ZIM archive on a real Firefox OS device!

Regarding performance, it seems faster to read an article with a ZIM backend than with Evopedia archives. But it might be a side effect of many things : I tested with a very small ZIM archive (the simple english without pics : ~125MB), and some articles seem truncated (I only have the beginning) so I suspect it does not decompress the whole articles for now.

www/js/lib/archive.js
peter-x: I think I was unsure if Q can handle forwarding multiple arguments. `callbackFunction` is explicitly called inside.

mossroy: You're right : I had not noticed

mossroy:
I have merged this pull request, and fixed the archive scanning (by reverting to jQuery API, for now).
It's really amazing to be able to use both archive formats. I thought it would take many months to reach this result. So thanks a lot peter-x!
Of course, there are still many bugs to fix, and features to implement. I'll try to list them in the github bugtracker, for v2.0 milestone

PR #118, opened 2015-06-07
Handle redirects in ZIM archives.

peter-x:


mossroy:
Excellent !
I've also commited a fix to complete this pull request (so that redirection info are stored in a serialized title)
And special mention for the unit test ;-)

PR #128, opened 2015-11-29
Zim title search

### LINKED ISSUE #104: ZIM backend : title search should only return articles starting with the prefix
It always displays the same list of results

### LINKED ISSUE #122: ZIM backend : some titles can not be found
Example in the Ray Charles archive : "You got the right one baby" article can not be found if you put "y" or "you" as the prefix.
But it is in the ZIM file, and it appears if you put "u" as the prefix (until #104 is fixed...)

peter-x:
Fixed a bug about titles not being found because ZIM sorts empty titles to the end (this includes all auxiliary files like images).
Furthermore, added the check whether the search prefix is actually a prefix of the title.

Fixes #122
Fixes #104

### DIFF
--- a/www/js/lib/zimArchive.js
+++ b/www/js/lib/zimArchive.js
@@ -115,13 +115,13 @@ define(['zimfile', 'zimDirEntry', 'util'],
                 return prefix < dirEntry.title ? -1 : 1;
             });
         }, true).then(function(firstIndex) {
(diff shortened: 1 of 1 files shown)


mossroy:
Great!
Thanks a lot peter-x.

Does that mean that we now can read images from a ZIM file? Would you have an example on the Ray Charles ZIM file, so that we might create a unit test on it?

PR #140, opened 2016-01-07
Fix for finding some files in ZIM.

chriseth:
The namespace has to be taken into account when titles are compared.

### DIFF
--- a/tests/tests.js
+++ b/tests/tests.js
@@ -19,8 +19,8 @@
  * You should have received a copy of the GNU General Public License
  * along with Evopedia (file LICENSE-GPLv3.txt).  If not, see <http://www.gnu.org/licenses/>
  */
-define(['jquery', 'title', 'archive', 'zimArchive', 'zimDirEntry', 'util', 'geometry'],
- function($, evopediaTitle, evopediaArchive, zimArchive, zimDirEntry, util, geometry) {
+define(['jquery', 'title', 'archive', 'zimArchive', 'zimDirEntry', 'util', 'geometry', 'utf8'],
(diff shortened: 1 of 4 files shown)


tests/tests.js
```diff
@@ -172,7 +172,7 @@ define(['jquery', 'title', 'archive', 'zimArchive', 'zimDirEntry', 'util', 'geom
                 equal(title._name, "Diego_Velázquez", "Name of the title is correct");
                 start();
             };
-            localEvopediaArchive.getTitleByName("Diego_Velázquez").then(callbackFunction);
+            localEvopediaArchive.getTitleByName("A/Diego_Velázquez").then(callbackFunction);
```
mossroy: You seem to prefix Evopedia titles with ZIM namespaces : are you sure? I suppose it should be applied only on ZIM titles?

mossroy: If I do that, the unit tests seem to run fine, so I'll merge your pull request and fix that myself on master

www/js/lib/zimDirEntry.js
```diff
@@ -54,7 +54,7 @@ define([], function() {
         this.redirectTarget = dirEntryData.redirectTarget;
         this.cluster = dirEntryData.cluster;
         this.blob = dirEntryData.blob;
-        this.url = dirEntryData.url;
+        this.url = dirEntryData.namespace + '/' + dirEntryData.url;
```
mossroy: @peter-x  : I know this is already merged, but are you sure it's a good idea to modify this.url in the constructor?
Shouldn't we keep it as dirEntrydata.url?
I find it a bit confusing that one property 'url' might sometimes contain the namespace, and sometimes not.
And I think I already saw cases where a URL had two namespaces, like A/A/name.html (I'll have to check)

PR #142, opened 2016-01-07
Jquery fallback

mossroy:
I had to disable javascript & css again in the serviceworker, because the content received by the page still does not seem right.
I'm looking into it : the issue might be in the serviceworker, and not in the backend

### DIFF
--- a/service-worker.js
+++ b/service-worker.js
@@ -75,7 +75,7 @@ function(util) {
     var regexpCSS = new RegExp(/\.css$/i);
 
     var regexpContentUrlWithNamespace = new RegExp(/\/(.)\/(.*[^\/]+)$/);
-    var regexpContentUrlWithoutNamespace = new RegExp(/^(.*[^\/]+)$/);
+    var regexpContentUrlWithoutNamespace = new RegExp(/^([^\/]+)$/);
     var regexpDummyArticle = new RegExp(/dummyArticle\.html$/);
     
     function fetchEventListener(event) {
--- a/service-worker.js
+++ b/service-worker.js
@@ -113,7 +113,19 @@ function(util, utf8) {
                     else if (regexpCSS.test(titleName)) {
                         contentType = 'image/css';
                     }
-                    reject("temporarily refuse javascript and css dependencies");
+                    var responseInit = {
+                        status: 200,
+                        statusText: 'OK',
+                        headers: {
+                            'Content-Type': contentType
+                        }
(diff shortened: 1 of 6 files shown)


service-worker.js
```diff
@@ -91,21 +74,35 @@ function(util) {
...
             event.respondWith(new Promise(function(resolve, reject) {
-                var regexpResult = regexpContentUrl.exec(event.request.url);
-                var nameSpace = regexpResult[1];
-                var titleName = regexpResult[2];
+                var nameSpace;
+                var titleName;
+                var titleNameWithNameSpace;
                 var contentType;
+                if (regexpContentUrlWithoutNamespace.test(event.request.url)) {
+                    // When the request URL is in the same folder,
```
peter-x: So `event.request.url` could be a relative url? If yes, we have to make it an absolute url relative to a base url, because this will not work correctly if there are subdirectiories in the zim file.

www/js/app.js
```diff
@@ -271,6 +254,144 @@ define(['jquery', 'abstractBackend', 'util', 'cookies','geometry','osabstraction
...
+            if (!isServiceWorkerReady()) {
+                $('#serviceWorkerStatus').html("ServiceWorker API available : trying to register it...");
+                navigator.serviceWorker.register('../service-worker.js').then(function (reg) {
+                    console.log('serviceWorker registered', reg);
+                    serviceWorkerRegistration = reg;
+                    refreshAPIStatus();
+                    
+                    // We need to wait for the ServiceWorker to be activated
+                    // before sending the first init message
+                    var serviceWorker;
```
peter-x: `serviceWorker = reg.installing || reg.waiting || reg.active;` should also work

www/js/lib/zimArchive.js
```diff
@@ -208,6 +210,10 @@ define(['zimfile', 'zimDirEntry', 'util', 'utf8'],
      */
     ZIMArchive.prototype.getTitleByName = function(titleName) {
         var that = this;
+        // If no namespace is mentioned, it's an article, and we have to add it
```
peter-x: This is not totally true, because we could have `A/A/abc` - better to always specify the namespace.

peter-x:
I think this is fine to merge without changes, but then we should create issues.

www/js/app.js
mossroy: You're right, I just changed that

Issue #147, opened 2016-01-08
ZIM backend can't read javascript content in the Ray Charles archive


mossroy:
At least it can't find -/j/head.js or -/j/body.js in the ZIM file (getTitleByName returns null)

mossroy:
In fact, this content does not seem to be available in the Ray Charles ZIM file : there is a -/j/local.js content, but no -/j/head.js or -/j/body.js
So I close this : it's not a bug in the implementation

PR #148, opened 2016-01-08
Unit tests improvements, and add a unit test on a javascript content.

mossroy:
+ minor error handling improvements in app.js

The unit test on javascript currently fails : see #147


mossroy:
I modified the unit test on javascript content, so that it now succeeds (it was looking for a content that did not exist in the test ZIM file...)

peter-x:
Looks good!

PR #150, opened 2016-01-08
Decode URI encoding for dependencies (images, JS, CSS etc).

mossroy:
It's necessary for some of them that have special characters in it.

For example the main image of article "Modern Sounds in Country and Western Music" in the Ray Charles ZIM file : the filename has a & in it, which was encoded as %26. We need to decode it as & before passing it to the backend.


peter-x:
Good to go!

PR #153, opened 2016-01-08
Modify the ZIM backend, so that findTitlesWithPrefix only returns articles

### LINKED ISSUE #151: Title search should only display articles, not other content types
It's the reason of the empty lines when you don't search with no prefix

mossroy:
And not other types of contents (javascript etc).
Fixes #151 

plus a small fix in title search, for the case where an article title is exactly the prefix


peter-x:
Looks good!

PR #155, opened 2016-01-08
Add a spinner, in Service Worker mode, when browsing to another article.

### LINKED ISSUE #135: Display a spinner when the user clicks on a link inside an article (in ServiceWorker mode)
It can take a few seconds to display the new article : the user needs to have a visual feedback that his click is being processed.

One approach to test is to catch the onclick event on all the iframe, and then find out if the target of this click event is inside a `<a href>` tag. In this case, the spinner should be visible (and disappear when the new article appears)

mossroy:
So that the user knows his click is being processed.
Fixes #135


www/js/app.js
```diff
@@ -254,11 +254,33 @@ define(['jquery', 'abstractBackend', 'util', 'cookies','geometry','osabstraction
...
         checkSelectedArchiveCompatibilityWithInjectionMode();
     });
+    $('#articleContent').on('load',onIFrameLoad);
+    function onIFrameLoad() {
+        $('#centeredspinner').hide();
+        $('#articleContent').contents().on('click', function (event) {
+            if (contentInjectionMode === 'serviceworker') {
+                // In Service Worker mode, we want to display a spinner
+                // when the user clicks on a link
+                // We only do that when the click is somewhere inside a <a> tag
```
peter-x: Is that really a safe way? We could wait for a message from the service worker (that is sent as soon as the service worker starts loading the article).

peter-x:
I think this is good to merge for now, perhaps turn the comment above into an issue.

www/js/app.js
mossroy: You're right : it's not a safe way at all. It was just an easy way to do a first version.
In fact, I would have expected the browser to inform the user by itself that a ServiceWorker is working. Just like it informs the user when some content is downloading.
It seems to be the case on Chromium, Firefox dev edition, but not on Firefox OS.

I should file a bug for Mozilla to implement it on Firefox OS, instead of trying to do it by hand.

mossroy:
I close this old PR. It was not well tested, and I'm not sure if it was FirefosOS-specific.
The code is still in the branch and might be reused if necessary

PR #157, opened 2016-01-08
Compatibility with absolute URLs (from a ZIM) in jQuery mode.

mossroy:
For example for wikipedia_fr_all_09_2012.zim


mossroy:
This is finally not crucial, as the ZIM files generated since the end of 2014 all have relative URLs now

www/js/app.js
```diff
@@ -955,9 +957,14 @@ define(['jquery', 'abstractBackend', 'util', 'cookies','geometry','osabstraction
                     // It's a link to another article
                     // Add an onclick event to go to this article
                     // instead of following the link
+                    
                     if (url.length>=2 && url.substring(0, 2) === "./") {
                         url = url.substring(2);
                     }
+                    // Remove the initial slash if it's an absolute URL
+                    else if (url.length>=1 && url.substring(0, 1) === "/") {
```
peter-x: Length check is not needed

www/js/app.js
```diff
@@ -978,7 +985,11 @@ define(['jquery', 'abstractBackend', 'util', 'cookies','geometry','osabstraction
                     });
                 } else {
                     // It's a standard image contained in the ZIM file
-                    var imageMatch = image.attr("src").match(regexpImageUrl);
+                    // We try to find its name (from an absolute or relative URL)
+                    var imageMatch = image.attr("src").match(regexpImageAbsoluteUrl);
```
peter-x: We could use one regex for both cases.

mossroy:
I implemented your suggestions and merged the pull request

PR #160, opened 2016-01-09
On title search, try some upper/lower case variants.

### LINKED ISSUE #124: ZIM backend: Title search should try some variants (at least upper/lower case)
Titles are currently stored in a case-sensitive way in a ZIM archive.
But Kiwix tries to workaround that by trying some variants. Cf Reader::searchSuggestions and Reader::searchSuggestionsSmart in https://sourceforge.net/p/kiwix/kiwix/ci/master/tree/src/common/kiwix/reader.cpp

We should try to do the same, until the OpenZIM format evolves to store normalized titles : #117

mossroy:
Fixes #124.

On the other hands, it really slows down the title search.
When there are not many results, it launches 3 case insensitive searches before giving the result, which takes a few seconds.

### DIFF
--- a/www/js/lib/util.js
+++ b/www/js/lib/util.js
@@ -33,6 +33,78 @@ define(['q'], function(q) {
     }
     
     /**
+     * Returns the same String with the first letter in upper-case
+     * @param {String} string
+     * @returns {String}
+     */
+    function ucFirstLetter(string) {
+        if (string) {
+            if (string.length >= 1) {
+                return string.charAt(0).toLocaleUpperCase() + string.slice(1);
+            } else {
+                return string;
+            }
+        } else {
+            return string;
+        }
+    }
+    
+    /**
+     * Returns the same String with the first letter in lower-case
+     * @param {String} string
+     * @returns {String}
+     */
+    function lcFirstLetter(string) {
+        if (string) {
+            if (string.length >= 1) {
+                return string.charAt(0).toLocaleLowerCase() + string.slice(1);
+            } else {
+                return string;
+            }
+        } else {
+            return string;
+        }
+    }
+    
+    /**
+     * Returns the same String with the first letter of every word in upper-case
+     * @param {String} string
+     * @returns {String}
+     */
+    function ucEveryFirstLetter(string) {
+        if (string) {
+            return string.replace( /\b\w/g, function (m) {
+                return m.toLocaleUpperCase();
+            });
+        } else {
+            return string;
+        }
+    }
+    
+    /**
+     * Generates an array of Titles, where all duplicates have been removed
+     * (it also sorts the titles)
+     * 
+     * @param {Array.<Title>} array of Titles
+     * @returns {Array.<Title>} same array of Titles, without duplicates
+     */
+    function removeDuplicateTitlesInArray(array) {
+        array.sort(function(titleA, titleB) {
+            if (titleA.title < titleB.title) return -1;
+            if (titleA.title > titleB.title) return 1;
+            return 0;
+        });
+        for(var i = 1; i < array.length; ){
+            if(array[i-1].title === array[i].title){
+                array.splice(i, 1);
+            } else {
+                i++;
+            }
+        }
+        return array;
+    }
+    
+    /**
      * Read an integer encoded in 4 bytes, little endian
      * @param {Array} byteArray
      * @param {Integer} firstIndex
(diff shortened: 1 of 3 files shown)


www/js/lib/util.js
```diff
@@ -33,6 +33,78 @@ define(['q'], function(q) {
     }
     
     /**
+     * Returns the same String with the first letter in upper-case
+     * @param {String} string
+     * @returns {String}
+     */
+    function ucFirstLetter(string) {
+        if (string) {
```
peter-x: What about
```
if (string && string.length >= 1)
  return string[0].toLocaleUpperCase() + string.slice(1);
else
  return string;
```

www/js/lib/zimArchive.js
```diff
@@ -109,14 +109,62 @@ define(['zimfile', 'zimDirEntry', 'util', 'utf8'],
...
      * @param {String} prefix
      * @param {Integer} resultSize
-     * @param {type} callback
-     * @returns {callbackTitleList}
+     * @param {callbackTitleList} callback
      */
     ZIMArchive.prototype.findTitlesWithPrefix = function(prefix, resultSize, callback) {
         var that = this;
+        that.findTitlesWithPrefixCaseSensitive(prefix, resultSize, function(titles) {
+            if (titles.length < resultSize) {
```
peter-x: It's a matter of style, but I think it is better readable to reverse the cases:
```
if (titles.length >= resultSize) {
  callback(titles);
  return;
}
...
```
(and have to explicit else case to reduce the indentation).

www/js/lib/zimArchive.js
```diff
@@ -109,14 +109,62 @@ define(['zimfile', 'zimDirEntry', 'util', 'utf8'],
...
+                // Let's add results with first letter upper-case
+                var ucPrefix = util.ucFirstLetter(prefix);
```
peter-x: You could check whether `prefix !== ucPrefix`

www/js/lib/zimArchive.js
```diff
@@ -109,14 +109,62 @@ define(['zimfile', 'zimDirEntry', 'util', 'utf8'],
...
+                that.findTitlesWithPrefixCaseSensitive(ucPrefix, resultSize, function(ucTitles) {
+                    titles.push.apply(titles, ucTitles);
+                    titles = util.removeDuplicatesInArray(titles);
```
peter-x: If `prefix !== ucPrefix`, there should be no duplicates

www/js/lib/zimArchive.js
```diff
@@ -109,14 +109,62 @@ define(['zimfile', 'zimDirEntry', 'util', 'utf8'],
...
+                    if (titles.length < resultSize) {
+                        // Let's add results with first letter lower-case
+                        var lcPrefix = util.ucFirstLetter(prefix);
```
peter-x: Typo: should be `lcFirstLetter`

www/js/lib/zimArchive.js
```diff
@@ -109,14 +109,62 @@ define(['zimfile', 'zimDirEntry', 'util', 'utf8'],
...
      * 
      * @param {String} prefix
      * @param {Integer} resultSize
-     * @param {type} callback
-     * @returns {callbackTitleList}
+     * @param {callbackTitleList} callback
      */
     ZIMArchive.prototype.findTitlesWithPrefix = function(prefix, resultSize, callback) {
         var that = this;
+        that.findTitlesWithPrefixCaseSensitive(prefix, resultSize, function(titles) {
```
peter-x: What about the following:
```
var prefixVariants = util.removeDuplicatesInArray([prefix, util.ucFirstLetter(prefix), util.lcFirstLetter(prefix), util.ucEveryFirstLetter(prefix)]);
var titles = [];
function searchNextVariant() {
  if (prefixVariants.length == 0 || titles.length >= resultSize) { callback(titles); return; }
  var prefix = prefixVariants[0];
  prefixVariants = prefixVariants.slice(1);
  that.findTitlesWithPrefixCaseSensitive(prefix, resultSize - titles.length, function(newTitles) {
    titles.push.apply(titles, newTitles);
    searchNextVariant();    
  });
  searchNextVariant();
}
```

mossroy: Yes, your suggestion is much better! Thanks a lot.
It should be a bit faster, too.

PR #162, opened 2016-01-09
Fix big memory overhead due to Base64 encoding of images in src attribute, in jQuery mode

### LINKED ISSUE #161: Probable memory leak when opening some articles
Example : in wikipedia_fr_all_2015-11.zim, if you open the Article "Japon", it eats up a lot of memory (2 GB in my test) and CPU, and ends up crashing the browser.
It's the same behavior on Firefox and Chromium, so the leak (or excessive use of memory) is certainly in our code.

mossroy:
Fixes #161.
Each image is now converted to a Blob with a URL that is used by the img tag.
I applied the same thing for math images (Evopedia archives)


www/js/lib/util.js
```diff
@@ -194,6 +194,27 @@ define(['q'], function(q) {
     }
     
     /**
+     * Creates a Blob from the given content, then a URL from this Blob
+     * And put this URL in the attribute of the DOM node
```
peter-x: I think we should keep dom and non-dom util functions in different files (to simplify later separation into js-zim and frontend). If they are already mixed, then it is fine.

mossroy: I agree with you. I fixed that in latest commits

PR #167, opened 2016-02-08
Fix integer computation on ZIM backend to make it work on big ZIM files.

### LINKED ISSUE #156: On some ZIM files, the images and some articles read in the backend have a 0 octet size
For example I/m/Ciudad.svg.png or A/Région_française.html or A/Langues_nilo-sahariennes.html on wikipedia_fr_all_2015-11
In kiwix desktop, this image and articles display fine.

The CSS are working fine on this ZIM file (at least most of them).

mossroy:
Fixes #156
The performance can probably be optimized.
But, based on https://jsperf.com/multiplication-int-vs-bit-shifting-left , it might not be that bad (depending on the javascript implementation).

### DIFF
--- a/www/js/lib/util.js
+++ b/www/js/lib/util.js
@@ -294,6 +294,19 @@ define(['q'], function(q) {
         }
         return s;
     }
+    
+    /**
+     * Does a "left shift" on an integer.
+     * It is equivalent to int << bits (which works only on 32-bit integers),
+     * but compatible with 64-bit integers.
+     * 
+     * @param {Integer} int
+     * @param {Integer} bits
+     * @returns {Integer}
+     */
+    function leftShift(int, bits) {
+        return int * Math.pow(2, bits);
+    }
 
     /**
      * Functions and classes exposed by this module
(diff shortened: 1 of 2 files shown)


peter-x:
Unfortunately, it is not as easy as that:
```
> 0x12345678 * Math.pow(2, 32) + 0x12345678 * Math.pow(2, 0)
1311768465173141000
> 0x12345678 * Math.pow(2, 32) + 0x12345671 * Math.pow(2, 0)
1311768465173141000
```
(note the change in the last digit in the third line)

This can only be solved by using some "big integers", i.e. objects with at least two variables.

mossroy:
@peter-x , can you elaborate? I suppose you mean the algorithm does not always work?
Is it "simply" the 53-bit limitation? In this case, I suppose it's already far than enough, as 2^53 = 9007 Tera-Bytes (but I should mention that in the function description : it's not a 64-bit compatibility, but only 53-bit).

In any case, it seems to solve the issue of displaying articles and images in the wikipedia_fr_all_2015-11.zim file. After browsing many articles in it, I did not notice zero-length contents any more (but maybe I've been lucky?)

mossroy:
@peter-x , I'm not clear on the status of this PR.
If you believe it should not be merged, please give me more details

peter-x:
Sorry for the delay. Please merge it if you noticed an improvement. If
precision will turn out to be an issue, we can fix it later.

Issue #119, opened 2015-07-19
Find a way to automate unit testing


mossroy:
So that it could be ran in the background, and checked automatically.
Tools like Karma should do that https://karma-runner.github.io
It runs the tests in a browser, supports QUnit, and can be integrated in some IDEs.
The main issue is to inject the archive file(s) in the browser : not sure if it is possible

Another idea would be to run the tests outside a web browser, with NodeJS?

mossroy:
According to Michal Henretty (from Mozilla) :
> Hrm, yeah I can see this is a problem. Can you host the ZIM file somewhere and load it using ajax?
> Conversely, if you are trying to test the file input user flow, this really belongs in a webdriver/marionette type of test. But that is another conversation :)

Maybe we could read the test ZIM file from https://github.com/mossroy/evopedia-html5/raw/master/tests/wikipedia_en_ray_charles_2015-06.zim ?
It should work and would probably allow automated testing.
But we'll have to find a way to avoid downloading these 1.5MB each time we want to run the unit tests : maybe the browser cache will simply handle this ? (this might depend on how github configures the cache HTTP headers)

I don't think we need Webdriver or Marionette for now.

mossroy:
After downloading the file in ajax, I think we won't be able to have a File instance, but we probably can make a Blob instance https://developer.mozilla.org/en-US/docs/Web/API/Blob
File extends Blob, and I think we only use the .slice() function (available in Blob), and the FileReader.readAsText and FileReader.readAsArrayBuffer functions (that accept Blobs as parameter) so using a Blob instance (read from an ajax request) instead of a File instance should work.
But I'm not 100% sure : this needs to be verified.

mossroy:
This has been fixed by #180

PR #180, opened 2016-06-25
Automated testing using travis, nightwatch and sourcelabs.

peter-x:
Still testing, please do not merge yet.


peter-x:
This is ready to be merged. I wanted to add more user interface tests, but it seems that it is non-trivial to get local files (i.e. zim files) working in the test environment. We first need to add a proxy for that as discussed in #119

peter-x:
I couldn't stop until the user interface tests were also working :-)

Also added a way to "inject" a "remote archive", i.e. zimfile blob that is just kept in memory for testing. This should close #119

mossroy:
Excellent! I really love that feature

PR #182, opened 2016-06-26
Fixed performance issue.

peter-x:
The actual problem was an incorrectly computed size value resulting in gigantic reads.


mossroy:
Cool : a quick test seems to work properly without the big performance regression.
I like the fact that your code is generic, and should also work if a content is spread among more than 2 files.

@kelson42 : would it be safe to assume that all the .zimnn files have the same size? (except the last one, of course). In this case, the algorithm could be optimized a bit, and go directly to the first relevant file, instead of having to add the sizes of each file before finding the right one. But I'm not sure this optimization is really necessary.

I don't understand the TODO about float type : you don't seem to use any non-integer numbers?

Do you have a way to check the "concatenate" use-case?
Maybe a solution would be to split the "Ray Charles" ZIM file into smaller parts, find a content that needs this concatenation, and create a unit-test on it? We even could base all our unit tests on the splitted version.
(I might do that if you need, but I won't have time in the short term)

peter-x:
Removed the todo about floats (it was invalid), split up the test zim file and added a test that reads the "Ray Charles" article (I confirmed that it is indeed split among two files).

mossroy:
The file wikipedia_en_ray_charles_2015-06.zim should not be deleted, as it's used by the Travis CI build through nightwatch.
Another option would be to modify the nightwatch script, so that it uses the split zim. This would need to modify setRemoteArchive from app.js to allow multiple files to be used

mossroy:
For now, I restored the standalone zim file, and created  #185 to delete it properly (and make nightwatch use the split ones)

Issue #116, opened 2015-06-07
ZIM backend : Try to cross-compile the libZIM library in webassembly with emscripten


mossroy:
It would avoid to re-code it in javascript, and would also ease support of future evolutions of the file format.
But not sure how Emscripten would handle the file I/Os

peter-x:
Being able to use libZIM as is depends on the question whether we can use synchronous IO at the low-level javascript side.
This probably has to wait until we know how we can access the filesystem from ServiceWorkers.

mossroy:
I don't think there is a way to do synchronous I/O on files in javascript.
Regarding ServiceWorkers, I suppose we have access to the same APIs as in any javascript.

mossroy:
Dattaz, I assign you this issue, as you expressed some interest in it.
No hurry and no obligation of course

mossroy:
@thiolliere if you want to have a look, too

mossroy:
I met @bnjbvr yesterday (he was giving a conference about WebAssembly, at @mixitconf), and talked to him about this idea. He said he might put us in touch with the maintainer of emscripten if necessary.
Obviously, it's too early for now.

I found this article that tackles our issue : https://hacks.mozilla.org/2015/02/synchronous-execution-and-filesystem-access-in-emscripten/
There seems to be a notion of "virtual filesystem" that allows synchronous access to files that are preloaded in memory. It's clearly not possible for us because of the size of our ZIM files.
Refactoring the zimlib code to make asynchronous file I/O is probably complicated. It's a change that would only be interesting for us (not for the other applications using this library) and would probably make the code less readable.
Maybe the Emterpreter might be an option, but they say it would be slower. So maybe we would have to find a way to use the Emterpreter only on the code that reads the ZIM file, and compile everything else in WebAssembly, if it's technically possible.

dattaz:
I have compile libzim in webassembly ; you can check demo here : https://dattaz.github.io/libzim_wasm/ (but zimfile (ekopedia_fr_all_2017-04.zim 12M) is embedded ) ; Now we have to deal with filesystem :) 

Source code to build is here : https://github.com/dattaz/libzim_wasm

kelson42:
@dattaz  I get only - with FF 53 - a a page with black squares. Nothing interesting is displayed.

mossroy:
Yeah! @dattaz, you rock!
It works pretty well, that's promising.
[...]

PR #183, opened 2016-06-26
Directory entry and block-level cache

peter-x:
At least on my desktop firefox this pull request results in blazing fast archive performance (search is instant).

The two techniques employed are:
- cache all directory entries
- cache small reads (less than 2048) from all files into blocks of 2048 bytes

TODO:
- What is missing is the actual cache eviction for the least-recently-used cache and using the lru-cache inside the block cache.
- cleanup

The block cache brought the actual performance gain, so the directory entry cache might not be needed, but is probably good to have.

This builds upon the pull request #182 and contains commits from there.

Please test on different devices!


mossroy:
I tested a bit on my Firefox OS device, and it does make a big difference on article search!
I think it also speeds up the first searches, because of the upper/lowercase variants.
I did not notice a difference for article contents, probably because most of the time is spent on handling CSS and images in the frontend.

I'm wondering if it's worth caching all the blocks : I suppose the chances to reuse a 2KB block inside a 20GB archive are quite low... Except for indexes, and for dependencies that would be commonly used (css, javascript, icons etc).

We have to keep in mind that low-end devices don't have a lot of memory (512MB for the ZTE Open C, for example, 1GB for the Flame, but some even had 256MB), so we can't afford to waste too much. So there must be a maximum size of the cache, and a relevant use/eviction strategy.

Regarding the 2 different caches, my opinion is that we should keep only the one(s) that give a significant difference. Because it makes the code more complicated and might make debugging more difficult in the future.

peter-x:
I'm cleaning up this PR now:
 - remove direntry cache
 - add cache eviction strategy

We have to tune the actual cache size by testing on different devices.

The thing is that article data is most likely not cached already due to the simple reason that we only cache reads of size up to 2048 bytes and article data is usually more than 2k.

peter-x:
Updated the code to be ready for review and testing. Actually, I was not able to properly reproduce the speed gain. I even artificially introduced a waiting time for IO and could not see a significant speedup. Cache statistics show that we have about 50% cache hit ratio, even with tiny caches. This all needs to be tested on real devices. Please make sure to compare this to https://github.com/kiwix/kiwix-html5/pull/183/commits/7718566744d34a8b77565673c131633a5bd5b240 or later.

mossroy:
Here are a few speed comparisons on a Mozilla Flame (same CPU as ZTE Open C). Kiwix app is restarted before each scenario (in order to empty the cache). Before=current master (https://github.com/kiwix/kiwix-html5/commit/44bcaf0840a0bea7510fb03269fa91c9f2caefc9) and after=current "cache" branch (https://github.com/kiwix/kiwix-html5/commit/2ae45b9eee87c98d4a71218a897a277f24ae6645). It is measured manually so not very accurate :

- searching "a" (many results) : before=4s , after=4s
- searching "kiwix" (only one result) : before=5s60 , after=5s90
- displaying "kiwix" (after searching like above) : before=9s70 , after=9s60
- searching "a", and then searching "a" again : before=4s , after=3s
- searching "kiwix" then searching "kiwix" again : before=6s , after=6s
- displaying "kiwix" then displaying "kiwix" again : before=6s , after=4s

So I agree with you, the speedup is less than what I remember of. It only speeds up the second access, and I wrote in June that it also speeded up the first search (see previous comments).

mossroy:
@peter-x , I close this PR as the speedup was less interesting than we expected. I think it's not worth making the code more complicated in this case.

Performance still needs to be improved, so we keep the branch with your code.
There are also other ideas to test, like #219 or #116.

Jaifroid:
AS this was completed and merged in #658, I'm deleting this stale branch.

kelson42:
Yohoo... 7 years later and probably one of the last, if not the last, inherited from Evopedia.

Jaifroid:
Ha ha -- though actually we did merge the block level cache a year ago, so "only" six years!

PR #191, opened 2017-04-03
Replace peter-x travis sauce credentials

mossroy:


mossroy:
@peter-x : your sauce credentials did not seem to work any more, so I replaced them with mine.

Issue #220, opened 2017-05-02
Feedback from exploratory testing on Chrome OS


julianharty:
Here's some initial feedback on how the extension works on Chrome OS. I may update this topic from time to time, others are welcome to add their comments and feedback.

I downloaded the extension from the Chrome Store. The initial impression was it loaded a bit slowly but then rendered adequately and told me enough to be able to visit the download.kiwix.org site to download a ZIM file, etc. I duly did so (a phet_en_ ZIM) however the subsequent experience was mixed. Content rendered quite slowly (perhaps 5 seconds) for the initial screen and then there wan't an obvious home page or anything that'd indicate what to do next (this may be an issue with the ZIM file which was one reason I later downloaded additional files to test with). The Random feature could find a topic and started loading that topic (PhET exercise) but the loading doesn't seem to get beyond the splash screen (with a PhET logo) and I've yet to see or try a PhET exercise on this Chrome OS device. I tried using the search but wasn't able to find any search results (the extension says `no files found`). 

I then used wikinews in English, April 2017. Again this renders slowly - multiple seconds. However Teresa May did eventually appear in all her glory - as well as the rest of the articles on the homepage of the ZIM file. Searches seem to work (there are some data quality issues in the ZIM file outside the scope of this testing) and content generally renders, albeit slowly.

I'll test one of the small TED talks to see how well audio and video are presented.

Here are various screenshots, it's impractical to filter or check them on Chrome OS when raising this issue so apologies for any duplicates.

![screenshot 2017-05-02 at 18 21 08](https://cloud.githubusercontent.com/assets/785891/25631209/97891130-2f67-11e7-89be-7332cedac986.png)
![screenshot 2017-05-02 at 18 17 29](https://cloud.githubusercontent.com/assets/785891/25631211/978b182c-2f67-11e7-8b5c-f5930640a5ce.png)
![screenshot 2017-05-02 at 18 13 24](https://cloud.githubusercontent.com/assets/785891/25631207/9788556a-2f67-11e7-9842-2339676cbd38.png)
![screenshot 2017-05-02 at 18 13 16](https://cloud.githubusercontent.com/assets/785891/25631210/978a81fa-2f67-11e7-84d9-679e5241bebb.png)
![screenshot 2017-05-02 at 18 12 56](https://cloud.githubusercontent.com/assets/785891/25631208/97888846-2f67-11e7-92a5-75e3ff4fff16.png)
[...]

PR #222, opened 2017-05-06
Add user info on which ZIM files are compatible.

### LINKED ISSUE #221: Give more user info on what kind of ZIM files are compatible

mossroy:
To avoid users from being frustrated by incompatible ZIM files like in #220
Fixes #221

### DIFF
--- a/www/index.html
+++ b/www/index.html
@@ -103,7 +103,7 @@
                     <br />
                     <h3>Usage</h3>
                     <h4>Step 1 : download an archive</h4>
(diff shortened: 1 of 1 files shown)


kelson42:
@mossroy Probably both "wikimedia" and "wikipedia" words should be written with an uppercased first letter in that document.

julianharty:
A small positive step to helping users realise there may be problems using some ZIM files, thank you.

Issue #218, opened 2017-04-22
Simplify/document the technical details to help contributors get into it


mossroy:
As initally mentioned by @sharun-s in #196 :

> It would be great if you could add some more detail on the difference between these two modes? I think it will help future contributors to get a better idea of the code and why they exist. Right now given the two different formats zim/evo archive, two different content injection modes (probably a third coming with webrequest), the weird zim format(whose advantages aren't clear), references to firefox os/geo location etc the barrier to entry for a dev to figure out what is going on is a bit high.
> 
> I think this is a fantastic project with lot of potential for all kinds of offline use-cases and I feel you will get more devs involved and exploring those use cases, if the core pieces of this project are well documented. Thanks for all the work you guys do!

mossroy:
@sharun-s : you're probably right, even if some of this info is already available.
Here is the situation of this application : 
- It was initially targeting Firefox OS, with Evopedia archive format. Both are now discontinued (as mentioned at the end of https://github.com/kiwix/kiwix-html5/blob/master/README.md)
- Now the main target is desktop (as browser extensions), with ZIM archive format (the format used by the kiwix project). But Firefox OS and Evopedia archives are still supported
- #195 discusses about dropping Evopedia compatibility. It's true that dropping it would remove some legacy code, and help understand the code for developers. The main reason to keep it is for the geolocation feature, that might be implemented in ZIM files too, but I agree it's not a strong argument
- I agree with you that mentions of Firefox OS should not bother the user or the developer. But I'm more reluctant to drop compatibility with it. There's not a lot of specific code for it, and I still use it on my phone
- There's some more legacy code that might be removed about Phonegap : there has been an attempt to port this code on Phonegap, but it has never been finished. This could be removed to simplify, too

Regarding the archive file formats : 
- the Evopedia file format is discontinued with Evopedia itself : there will be no more new archives
- the ZIM file format is actively used, and much more generic : it can hold wikimedia content, but also many more websites, content-types etc

Regarding the different content injection modes : 
- in the first (Evopedia) versions, only jQuery mode was implemented, and was sufficient because it only worked with wikimedia content
- when switching to the ZIM file format, we needed a more generic way to inject dependencies (CSS, javascript, images etc) in a web page, to be able to support all the kinds of content it can hold. In our opinion, the right way to do it is to use a Service Worker. But we're not there yet, because it's still buggy (#145), and is not always supported (old browsers AND Firefox extensions)
[...]

PR #226, opened 2017-05-06
Remove legacy unused code for Evopedia and Cordova/Phonegap

mossroy:
@sharun-s : this should help understanding the code for new contributors. See #218 
It also simplifies a bit the UI and the understanding for the user.


sharun-s:
@mossroy I will check it out. Don't worry I haven't given up. :-) Still working on those notes. 

I am not a js dev so navigating all the anonymous functions, callbacks and async code can be quite crazy. It's not hard to understand what's going on, but it's hard to remember the flow of the code by the next day. And if I leave the code for a couple days full restart and review has to happen. Maybe this is just due to my lack of javascript experience. Anyway will keep you posted.

mossroy:
I also renamed 2 javascript files (abstractBackend and osabstraction) to try to make their purpose more clear

PR #233, opened 2017-05-19
Fixes #131 Unit Testing on Chrome.

### LINKED ISSUE #131: Unit tests do not all work on Chromium/Chrome
It's probably because there is a race condition somewhere, and we're lucky on Firefox and not on Chromium.
There's probably a place where I should wait for something to be completed before starting another one (with a callback or a Promise)

sharun-s:
Requires Chrome to be started with --allow-file-access-from-files when opening tests.html.
Chrome does not support xhr for the "file://"" protocol (which is used by the tests), unless the flag is set.
Easiest way (on Windows) is to create a Chrome shortcut with the flag in local directory that opens tests.html. Tests also work on Firefox and Edge.

**zimFile.js** 
fileArray.sort() in fn fromFileArray needs to be passed a comparision function to sort blob objects by "name". Chrome was sorting blob objects differently than Firefox and Edge. Reading the first 80 bytes of the fileArray[0] to retrieve zim header details then would fail causing tests to fail. NOTE: If array size is 10 or less the blob sort order is the same across all 3 browsers. So best to test with larger than 10 blobs.

**tests/tests.js**
* Removed xhr.status==200 check in makeblobrequest.
   It fails in Chrome. Even though the blobs are read xhr.status is set to 0.
   Probably related to "file://" handling policy.
* toSource() in the load Image 'm/RayCharles_AManAndHisSoul.jpg' test is non-standard. 
   Works only in FF. Changed to toString().
* Updated QUnit to 2.3.2 as QUnit.asyncTest is deprecated. Removed old files.

**tests/tests.html**
* Updated to use QUnit 2.3.2 files.


sharun-s:
:) for a second there I thought I broke something - "expected "contain 'Tests completed in'" but got: "20 tests completed"

mossroy:
Excellent!
It did not take you long to be able to use git, and to create a PR in GitHub :-)
In my opinion it's very to close to be mergeable.
I'll add a few comments in the code (as a review) : if you modify the code accordingly, commit, and push, it will modify the PR, so that I can merge it.

tests.html
```diff
@@ -42,5 +42,6 @@
         These unit tests should be run with Firefox.<br/>
         TODO : find out why they do not all run on Chromium
```
mossroy: I think you can remove the TODO, as this PR fixes it :-)
It also might be the right place to explain how to run the tests on Chromium/Chrome : either with the command-line argument, or through a local webserver

tests/tests.js
```diff
@@ -37,16 +37,9 @@ define(['jquery', 'zimArchive', 'zimDirEntry', 'util', 'utf8'],
...
-                } else {
-                    reject({
-                        status: this.status,
-                        statusText: xhr.statusText
-                    });
-                }
+                var blob = new Blob([xhr.response], {type: 'application/octet-stream'});
+                blob.name = name;
+                resolve(blob);
             };
```
mossroy: It's great that it fixes the issue for Chrome, but it looses the ability to fail properly if the XHR returns an error.
Maybe accepting the 0 status code in addition to status between 200 and 300 could be enough?

The idea is to have a meaningful error in QUnit if XHR fails in the future

tests/tests.js
```diff
@@ -106,237 +99,252 @@ define(['jquery', 'zimArchive', 'zimDirEntry', 'util', 'utf8'],
...
-                    equal(title.url, "-/j/local.js", "URL is correct.");
+                    assert.equal(title.url, "-/j/local.js", "URL is correct.");
                     localZimArchive.readBinaryFile(title, function(title, data) {
-                        equal(data.length, 41, "Data length is correct.");
+                        assert.equal(data.length, 41, "Data length is correct.");
                         data = utf8.parse(data);
                         var beginning = "console.log( \"mw.loader";
-                        equal(data.slice(0, beginning.length), beginning, "Content starts correctly.");
-                        start();
+                        assert.equal(data.slice(0, beginning.length), beginning, "Content dones correctly.");
```
mossroy: I suppose the starts->dones modification in the string is a mistake due to a search/replace?

tests/tests.js
```diff
@@ -81,23 +74,23 @@ define(['jquery', 'zimArchive', 'zimDirEntry', 'util', 'utf8'],
      * @param e Error
      */
     function errorHandlerAsyncTest(e) {
-        ok(false, "Error in async call", e);
-        start();
+        assert.ok(false, "Error in async call", e);
+        done();
     }
```
mossroy: Is this function still necessary?
It looks like you removed every reference to it, so I suppose it can be removed?

sharun-s:
This is really great! I didn't expect to see the testing and review tools on this page. New respect for GitHub! Thanks for providing the steps. I didn't know how long it was going to take to figure this out, and the steps gave me some sense it wasn't going to be to bad.

I will make the changes and resubmit. How does the next push get associated with this pull request though?

tests/tests.js
sharun-s: good catch

tests/tests.js
sharun-s: Yup. Have to test that condition out with Firefox. If it works will add it.

tests/tests.js
sharun-s: This is a missing piece wrt error handling that I haven't figured out yet. Every time something goes wrong in a test nothing shows up on the page. There is no indication even that the test was run. QUnit seems to swallow up all errors.

sharun-s: Okay I now see the QUnit no try-catch option which was causing all the errors to be swallowed up. Should be okay to remove this.

tests.html
sharun-s: done!

mossroy:
I suppose you forgot to add some files in your last commit, in particular tests.js ?
In any case, Travis is now happy :-)

sharun-s:
Damn. Will check...

mossroy:
Sorry, I have to leave. I will have another look later in the weekend

sharun-s:
No prob. And thanks I learnt some very interesting stuff today.

mossroy:
Cool, it now works properly on Chromium/Chrome (did not test on Edge) : I'll merge this pull request.
@sharun-s : I'll "squash" your 4 commits into only one with github (to avoid polluting the changelog with the fixes). Next time I can explain you how to do it by yourself with `git rebase -i`.

I think the way we handle the XHR in tests.js for reading the files is not perfect, but I'll open another issue for that : it's outside the scope of this PR.

Thanks a lot for your contribution.

sharun-s:
Cool will try and remember rebase -i for next time.  I was wondering about that this.status and xhr.status inconsistency but was in a bit of a hurry. Sorry!

PR #238, opened 2017-05-24
Fixes #177 History Navigation

### LINKED ISSUE #177: The back button does not always work on ZIM files
The back button of the browser (or the left arrow in the bottom bar of the screen) should allow the user to browse his previous pages.
It works fine on Evopedia archives, but not always on ZIM files

sharun-s:
"article URL" should have been used instead of "article Title" when saving browser history state. But when these terms are being read from the DirEntry "article Namespace" is being prepended to url and stored as url. The fix is to stop doing this prepending. That keeps it consistent to the reader when "DirEntry.url" is reqd and when "DirEntry.namespace + '/' + DirEntry.url" is required. Further details [here](https://github.com/kiwix/kiwix-html5/issues/177#issuecomment-296538463)
 
Changes:
www/js/zimDirEntry.js
    url should not hold namespace + url
    
 www/js/app.js
    In gotoRandom, handleTitleClick, gotoMainArticle
    now passes title.url instead of title.name() to pushBrowserHistoryState
    
www/js/app.js, tests/tests.js
    Replaced occurances of title.url with title.namespace + title.url


mossroy:
It seems to work pretty well. Thanks a lot for this fix!

But I won't be able to merge this PR as it is.
As you might have seen, it mixes your previous commits (about making unit tests working on Chrome, already merged) with your newest ones. And, it detects a conflict because I had "squashed" these previous commits into a single one.
It's partly my fault : I gave you directions to make a pull request as simply as possible, without the best practices.
So what I suggest is that we try to make this new PR properly together.

It would be possible to keep your local copy of sources and "cherry-pick" your commit, but it might be not trivial to follow.
As the source code has not changed since your last PR, I propose to start over from scratch :
- copy the files you modified (tests.js, app.js, zimDirEntry.js, and maybe .gitignore) to another location of your computer
- rename or move your local copy of source (kiwix-html5 folder), in case of
- copy this comment somewhere (to be sure you still can follow it after next step)
- on github, delete your fork of kiwix-html5 repository, fork it again from kiwix user, and clone it locally on your computer : https://help.github.com/articles/fork-a-repo/ (including adding the "upstream" remote repository)

After that, each contribution should be commited in a separate branch.
So, for this PR, you can : 
- `git checkout -b branch_name` (replace branch_name by something meaningful about the contribution)
- copy the files you modified back here and test
- `git add` and `git commit`
- then push your new branch : `git push origin branch_name` and create the PR from github

For the next contribution you would like to do, you will first have to sync with the upstream repo : 
- go back to the master branch with `git checkout master`
- re-sync with `git pull upstream master`
- then create a new branch as above, commit etc

mossroy:
As explained above, I close this PR, as it needs to be re-created properly

sharun-s:
@mossroy Hmm...I knew I had to redo this. But I wasn't totally clear about what went wrong. From your notes I _think_ I now get the reason for the conflict and the mixing of two sets of commits. 

I am not sure though...why I didn't see that conflict locally. Would like to figure it out so, I don't repeat the mistake. This is what I did - 
I did create a branch on which I made the changes. Then switched back to master. Did a sync(pull) with remote to make sure it had the latest. It didn't really pull that "squashed" commit. Instead said everything was up-to-date. I then merged myfix branch with local master and pushed to remote. 

So from your comment, if I understood it right,  you are recommending I don't merge my branch with local master before pushing to remote, but rather push from branch to remote and then update local master. Makes sense. Though it feels similar to what I did, so still wondering why I didn't catch the conflict locally. Any ideas what I might have done wrong? 

Either way will now follow this process. Thx for the steps and explanations.

mossroy:
The conflict probably comes from the squash I did, not from a mistake from you.
If I had made you squash yourself before merging (or did no squash at all), I suppose the PR would be OK.
In any case, creating PR from branches makes it much easier to separate them

sharun-s:
Ok cool.

PR #239, opened 2017-05-26
Fixes #177 History Navigation

### LINKED ISSUE #177: The back button does not always work on ZIM files

sharun-s:
"article URL" should have been used instead of "article Title" when saving browser history state. But when these terms are being read from the DirEntry "article Namespace" is being prepended to url and stored as url. The fix is to stop doing this prepending. That keeps it consistent to the reader when "DirEntry.url" is reqd and when "DirEntry.namespace + '/' + DirEntry.url" is required. Further details [here](https://github.com/kiwix/kiwix-html5/issues/177#issuecomment-296538463)

Changes:
www/js/zimDirEntry.js
url should not hold namespace + url

www/js/app.js
In gotoRandom, handleTitleClick, gotoMainArticle
now passes title.url instead of title.name() to pushBrowserHistoryState

www/js/app.js, tests/tests.js
Replaced occurances of title.url with title.namespace + title.url

### DIFF
--- a/tests/tests.js
+++ b/tests/tests.js
@@ -251,7 +251,7 @@ define(['jquery', 'zimArchive', 'zimDirEntry', 'util', 'utf8'],
             localZimArchive.getTitleByName("I/m/RayCharles_AManAndHisSoul.jpg").then(function(title) {
                 assert.ok(title !== null, "Title found");
                 if (title !== null) {
-                    assert.equal(title.url, "I/m/RayCharles_AManAndHisSoul.jpg", "URL is correct.");
+                    assert.equal(title.namespace +"/"+ title.url, "I/m/RayCharles_AManAndHisSoul.jpg", "URL is correct.");
                     localZimArchive.readBinaryFile(title, function(title, data) {
                         assert.equal(data.length, 4951, "Data length is correct.");
                         var beginning = new Uint8Array([255, 216, 255, 224, 0, 16, 74, 70,
@@ -271,7 +271,7 @@ define(['jquery', 'zimArchive', 'zimDirEntry', 'util', 'utf8'],
             localZimArchive.getTitleByName("-/s/style.css").then(function(title) {
                 assert.ok(title !== null, "Title found");
                 if (title !== null) {
-                    assert.equal(title.url, "-/s/style.css", "URL is correct.");
+                    assert.equal(title.namespace +"/"+ title.url, "-/s/style.css", "URL is correct.");
                     localZimArchive.readBinaryFile(title, function(title, data) {
                         assert.equal(data.length, 104495, "Data length is correct.");
                         data = utf8.parse(data);
@@ -290,7 +290,7 @@ define(['jquery', 'zimArchive', 'zimDirEntry', 'util', 'utf8'],
             localZimArchive.getTitleByName("-/j/local.js").then(function(title) {
                 assert.ok(title !== null, "Title found");
                 if (title !== null) {
-                    assert.equal(title.url, "-/j/local.js", "URL is correct.");
+                    assert.equal(title.namespace +"/"+ title.url, "-/j/local.js", "URL is correct.");
                     localZimArchive.readBinaryFile(title, function(title, data) {
                         assert.equal(data.length, 41, "Data length is correct.");
                         data = utf8.parse(data);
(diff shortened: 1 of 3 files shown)


.gitignore
```diff
@@ -20,6 +20,7 @@ Thumbs.db
 .tmproj
 *.sublime-project
 *.sublime-workspace
+*.lnk
```
mossroy: Is this new line related to the history navigation fix?
If it's not, it should be removed from this PR.
(and it would be a good occasion afterwards to show you how to "squash" your modifications into a single commit ;-) )

mossroy: About squashing : the goal is to have a clean git log in the end (avoiding unnecessary commits).
In our case, if you commit the removal of the line in .gitignore, you will have 2 commits in the log, that would end up in the upstream log if the PR is merged.
Squashing means merging these 2 commits, so that to hide the details of this line added and removed.
You can easily do that in your local repo with command `git rebase -i master`. It will open a text editor with 2 lines (one for each commit). For each commit, you'll be able to say what you want to do with it (by replacing the command at the beginning of the line). In our case, you'll want to put "f" (for "fixup") at the beginning of your second commit. It should result in a single commit in your log, merging the 2. You can check with `git log`
After that, you will have to push to github again with a `--force` option. It will update the PR with this single commit.

sharun-s: :) So whats the "squash" plan? 
I'll remove the line and I see in docs there is a .git/info/exclude file where I can use ignore the .gitignore changes. Should hopefully prevent this next time. I am using this to hide the shortcut to chrome with allowlocalfile access flag set  So yup totally unrelated.

mossroy: The plan is, like you said, to :
- remove the line, and commit this removal
- run `git rebase -i master` to merge these 2 commits, keeping the commit message of the first one (as it did not mention .gitignore)
- check with `git log` that you have only one commit remaining
- push to github with `--force` option, which will update the PR

sharun-s: ya sorry saw your prev comment after posting mine. And my commits are on a branch fix177. so I guess you mean git rebase -i <branch name>. Then git push --force orgin fix177 right?

mossroy: It should really be `git rebase -i master`, as master is the origin of your fix177 branch.
Then yes, after editing and checking the result, `git push --force origin fix177`.

sharun-s:
Ok that seems to have worked. Will go figure out what rebasing did when I get some time. 
Why does it give me an option to close and comment? What does close mean?

mossroy:
Cool! Thanks a lot for this fix

mossroy:
"Close and comment" might have been used before I merged, if you finally wanted to give up on this PR (luckily, you did not ;-) )

sharun-s:
phew almost clicked that. Ok thanks for the walk through! Until next time...

mossroy:
If you want to start working on another contribution later (which I hope you will :-) ), don't forget to sync your local repo with upstream before creating the branch : this should avoid most conflicts

mossroy:
Thanks again for your work. It looks like you got into the logic of the source code now, and managed to use git pretty quickly

sharun-s:
Got it. I realized the mistake I made was I hadn't setup the "upstream" remote. Now looking at the flow makes more sense. 

I am actually curious about how to setup a performance test for read and page load. I think there are some optimizations that can be done to speed things up. But I haven't yet figured out how to test if my change is making an improvement or not. Got any suggestions?

mossroy:
Regarding performance monitoring, one approach would be to use a unit test (if the ZIM file size has no influence on the optimizations you want to try), and (manually) keep track of the time it takes to run it. You can ask QUnit to run only one unit test instead of all of them.
If the ZIM file size has an influence, it could be a separate set of unit tests on a bigger ZIM file (which could not be commited with the source code because of the size).

But unit tests will only test the "backend" code, not the UI. If you want to optimize the UI, it must be another approach, like nightwatch (that is already run on each commit by Travis, to run minimal UI testing).
Regarding performance of the backend, #116 might help, but it's still a work-in-progress and we're not 100% sure it will work.

sharun-s:
@mossroy I think I will open an issue on "creating a performance test". So this discussion can carry on there instead of appending to this thread.

PR #241, opened 2017-05-28
Readme on old features and rewording of title vs direntry

mossroy:
Getting rid of this "Title" concept modifies a lot of code.
@sharun-s : is that what you expected in #218 ?


www/js/lib/zimfile.js
```diff
@@ -20,7 +20,7 @@
  * along with Kiwix (file LICENSE-GPLv3.txt).  If not, see <http://www.gnu.org/licenses/>
  */
 'use strict';
-define(['xzdec_wrapper', 'util', 'utf8', 'q'], function(xz, util, utf8, Q) {
+define(['xzdec_wrapper', 'util', 'utf8', 'q', 'zimDirEntry'], function(xz, util, utf8, Q, zimDirEntry) {
```
sharun-s: Not sure how to add a comment on the "blob" function. So am adding here.
This function took me the longest to figure out. Okay I didn't know what Promises were so that didn't help :)
But even after...I think what was missing to me was how the decompression process works. Just a brief note in the function documentation would be really useful. Like what is chunk size? Relationship to the compressed data read length. To read the 16 bytes to get blob address for example why is chunk size set to 5 KB. Basically some kind of note on how this works would be great. By changing these sizes (data length/chunk size) can reads be done faster?

sharun-s: nextCluster = readInt(clusterOffsets, 8, 8); This nextCluster doesn't seem to get used anywhere

www/js/lib/zimArchive.js
```diff
@@ -238,12 +222,12 @@ define(['zimfile', 'zimDirEntry', 'util', 'utf8'],
...
      * 
-     * @param {DirEntry} title
+     * @param {DirEntry} dirEntry
      * @param {callbackStringContent} callback
      */
-    ZIMArchive.prototype.readArticle = function(title, callback) {
-        return title.readData().then(function(data) {
-            callback(title.name(), utf8.parse(data));
+    ZIMArchive.prototype.readArticle = function(dirEntry, callback) {
+        return dirEntry.readData().then(function(data) {
```
sharun-s: Is this return here required? The callback handles the end of the read.

www/js/lib/util.js
```diff
@@ -78,16 +78,16 @@ define(['q'], function(q) {
...
+     * @returns {Array.<DirEntry>} same array of DirEntry, without duplicates
      */
-    function removeDuplicateTitlesInArray(array) {
-        array.sort(function(titleA, titleB) {
-            if (titleA.title < titleB.title) return -1;
-            if (titleA.title > titleB.title) return 1;
+    function removeDuplicateTitlesInDirEntryArray(array) {
+        array.sort(function(dirEntryA, dirEntryB) {
+            if (dirEntryA.title < dirEntryB.title) return -1;
+            if (dirEntryA.title > dirEntryB.title) return 1;
```
mossroy: I agree. It was probably meaningful with Evopedia archives, but is not any more.

www/js/lib/zimArchive.js
mossroy: True : it's unnecessary

www/js/lib/zimArchive.js
```diff
@@ -254,12 +238,12 @@ define(['zimfile', 'zimDirEntry', 'util', 'utf8'],
...
-     * @param {DirEntry} title
+     * @param {DirEntry} dirEntry
      * @param {callbackBinaryContent} callback
      */
-    ZIMArchive.prototype.readBinaryFile = function(title, callback) {
-        return title.readData().then(function(data) {
-            callback(title.name(), data);
+    ZIMArchive.prototype.readBinaryFile = function(dirEntry, callback) {
+        return dirEntry.readData().then(function(data) {
+            callback(dirEntry.name(), data);
```
mossroy: Not quite : it hides the concept of cluster

www/js/lib/zimfile.js
mossroy: I don't master this code portion (I did not write it myself).
I agree with you that it lacks some comments. And it might be a place where some optimization could take place.
I'll leave it as it is for now.

mossroy:
I pushed another commit that addresses some of your comments.
I'd like to merge this PR, test and release a version 2.1.0 that would include all the recent commits we made.
Feel free to open other issues if necessary, for a next version

sharun-s:
Ok. Sounds good to me!

PR #252, opened 2017-06-16
.zimb-series extensions without the BOM

Jaifroid:
Sorry about the BOM, I didn't realize it was there. Hopefully this gets rid of it.


Jaifroid:
PS Feel free to "Squash and merge", as I couldn't get this PR to come across in a single commit.

Jaifroid:
Hmm, as you can see I was trying to use the Ignore file feature to prevent my copy from including "Local items" in a pull request, but I think the "Local items" got committed before the Ignore file.... Now I don't know how to undo those two pull requests! As they are local items, perhaps they could be accepted and then I could just delete the resulting .vs folder, and hopefully in future it will get ignored... Sorry, I'm a bit of a noob with pull requests.

mossroy:
Thanks a lot for the fixes proposed in this PR.
But, as you saw by yourself, 3 different thinks get mixed in it : the .zimb series, ignoring VisualStudio config files, and the CSS fix for Edge Mobile.
Another issue is that, if I merge this PR as is, all the commits will get merged, which will not help understanding the changelog. I might use the "squash and merge" feature of Github to fix that, but it would make it harder for you to sync with upstream afterwards (and it would put the 3 fixes in one commit).

All in all, I think it would be better to restart from scratch, as there are not a lot of file modifications.
The goal would be to make 3 PR : one for each fix, each PR containing only the necessary commits. The right way to do that (as far as I know), is to create a new branch for each fix, and create a PR for each branch.

This is what I would suggest : 
- rename or move your local copy of the source code, containing your file modifications
- copy these instructions somewhere (because they might disappear in next step)
- delete your repository fork on github, and fork it again from kiwix-html5
- clone it again locally
- create a new branch and switch to it with `git checkout -b newbranchname`
- copy one fix in it, commit it, and push it to github with `git push origin newbranchname`
- in github, in the code tab, create a PR from this branch
- for the next fix, as they are all independent, go back to the master branch with `git checkout master`, create a new one (from master) with the same procedure etc.

If you have several commits in a branch that you would like to squash into a single one, switch to that branch, use `git rebase -i master`, and put a "f" at the start of the commit you want to squash with the previous one.
A rebase will also be advisable later in some or your branches, as soon as one of the PR is merged upstream (in order to have a clean linear changelog), but we'll see that later.

Regarding each fix : 
- .zimb series : please remove the spaces after commas, to be consistent with the .zima series
- VisualStudio config files : wouldn't it be better to add the .vs folder in .gitignore, so that to ignore the whole folder content?
- CSS fix for Edge Mobile : maybe add in the comment to check if it is still necessary if we upgrade to bootstrap 4, and create a new issue to try this upgrade later (when it will be in final release)

NB : Thanks to @mgautierfr who taught me these git practices not a long time ago ;-)

mossroy:
Another good idea is to put in your commit description a "fixes #correspondingissuenumber", so that the corresponding issue is automatically closed when the PR is fixed (see https://help.github.com/articles/closing-issues-via-commit-messages/).

Again, thanks a lot for these contributions, I'll have a look at your latest comments/issues later in the weekend

PR #260, opened 2017-06-20
Fixes #253 (EMSCRIPTEN buffer overrun)

### LINKED ISSUE #253: EMSCRIPTEN (xzdec.js) - buffer overrun on Windows 10 Mobile with very long articles
I've been investigating a suspected buffer overrun with the xzdec.js JS-compiled C++ code which happens on Windows 10 Mobile (only, not on full Windows 10) when loading a very heavy Wiki page. The page I've been using which reliably causes the buffer overrun is https://es.wikipedia.org/wiki/Bolivia (in wikipedia_es_all_2017-05.zim ). Quite a few other heavy pages cause the problem too. I suspected a limited memory environment, and it turns out to be the case. I can solve the problem by increasing the initial variable `TOTAL_MEMORY` assigned in xzdec.js, but only by modifying the module, which is a compiled one and presumably not one I should be "hacking". Is there a more elegant solution than the hack below?

**Details:**

I isolated the problem to this function in xzdec.js:
```
function enlargeMemory() {
    while (TOTAL_MEMORY <= DYNAMICTOP) { TOTAL_MEMORY = alignMemoryPage(2 * TOTAL_MEMORY) }
    assert(TOTAL_MEMORY <= Math.pow(2, 30));
    var oldHEAP8 = HEAP8; var buffer = new ArrayBuffer(TOTAL_MEMORY);
    Module["HEAP8"] = HEAP8 = new Int8Array(buffer);
    Module["HEAP16"] = HEAP16 = new Int16Array(buffer);
    Module["HEAP32"] = HEAP32 = new Int32Array(buffer);
    Module["HEAPU8"] = HEAPU8 = new Uint8Array(buffer);
    Module["HEAPU16"] = HEAPU16 = new Uint16Array(buffer);
    Module["HEAPU32"] = HEAPU32 = new Uint32Array(buffer);
    Module["HEAPF32"] = HEAPF32 = new Float32Array(buffer);
    Module["HEAPF64"] = HEAPF64 = new Float64Array(buffer);
    HEAP8.set(oldHEAP8);
    var success = _emscripten_replace_memory(buffer); assert(success)
}
```
After iterating that loop on the "Bolivia" page about three times, the line `HEAP8.set(oldHEAP8);` throws an exception and after that memory is corrupted and the app has to be exited and restarted to load anything further at all. Fortunately apps are rigorously sandboxed, so this is an inconvenience more than a security concern, I think.

After reading up on EMSCRIPTEN, which is the compiler here, I traced `TOTAL_MEMORY`, and found that it increases too rapidly because its startup value is too low. After experimenting with various values, here is what I did to solve it:
```
//var TOTAL_MEMORY = Module["TOTAL_MEMORY"] || 8e7; //- GK not enough startup memory
var TOTAL_MEMORY = Module["TOTAL_MEMORY"] || 14000000;
```
With the hex value of 14000000, the phone iterates the above loop only twice in the page load (increasing the dynamic value further) and the page finishes loading correctly.

Any thoughts on a better way to solve it? I'm wondering whether `TOTAL_MEMORY = Module["TOTAL_MEMORY"] || 8e7;` would allow me to set `TOTAL_MEMORY` in an external module. But I can't find anything referenced by `Module["TOTAL_MEMORY"]`.

Jaifroid:
Incorrectly entered HEX value for TOTAL_MEMORY in xzdec.js. See #253 for full discussion. NB this file is minified, so diffs will not look pretty, but the change is literally adding 0x in front of the hex value, plus a comment to document it.

I've tested now on UWP, Edge and Firefox. Can @mossroy check with FFOS, and someone do it with Chrome?


mossroy:
It seems to work fine on my FirefoxOS device, even with big articles with many images. I don't see any difference.
It also seems to work fine on Chromium.
On both Firefox and Chromium, it now says : 
`increasing TOTAL_MEMORY to 16777216 to be more reasonable`
(16.7MB, instead of 83886080 = 83.8MB)

Issue #272, opened 2017-06-25
Prevent unnecessary display of 404s before image load


Jaifroid:
The current code shows broken image icons before the images have finished loading, which can be a problem for heavy articles. I noticed that @sharun-s implemented a method for hiding the images until they load. I've improved it by hiding also the fixed image height and width until the images are loaded. The code is below. I've tested on Edge (mobile and desktop) and Firefox. I'd be happy to implement it here as a PR:

**EDIT: I've radically simplified the code. Please ignore this and see next comment.**

```
@@ -787,8 +787,23 @@ define(['jquery', 'zimArchiveLoader', 'util', 'uiUtil', 'cookies','abstractFiles
         $("#articleContent").contents().scrollTop(0);
 
         // Display the article inside the web page.
-        $('#articleContent').contents().find('body').html(htmlArticle);
-        
+        var $body = $(htmlArticle);
+        $body.find('img').each(function(){
+            var image = $(this);
+    	    // Prevents unnecessary 404's being produced when iframe loads images
+            $(image).attr("data-src", $(image).attr("src"));
+            $(image).removeAttr("src");
+            $(image).attr("data-height", $(image).attr("height"));
+            $(image).removeAttr("height");
+            $(image).attr("data-width", $(image).attr("width"));
+            $(image).removeAttr("width");
+            //Restore image height and size on image load
+            $(image).on("load", function (e) {
+                this.width = $(image).attr("data-width");
+                this.height = $(image).attr("data-height");
+            });
+        });
+        $('#articleContent').contents().find('body').html($body);
         
         // If the ServiceWorker is not useable, we need to fallback to parse the DOM
         // to inject math images, and replace some links with javascript calls
@@ -854,7 +869,7 @@ define(['jquery', 'zimArchiveLoader', 'util', 'uiUtil', 'cookies','abstractFiles
                 var image = $(this);
                 // It's a standard image contained in the ZIM file
                 // We try to find its name (from an absolute or relative URL)
-                var imageMatch = image.attr("src").match(regexpImageUrl);
+                var imageMatch = image.attr('data-src').match(regexpImageUrl);
                 if (imageMatch) {
                     var title = decodeURIComponent(imageMatch[1]);
                     selectedArchive.getDirEntryByTitle(title).then(function(dirEntry) {
```

Jaifroid:
[...]

PR #273, opened 2017-06-25
Implements #272 (Prevent unnecessary display of 404s)

Jaifroid:
Implements #272, elaborating on idea of @sharun-s (hiding of img.src in img.data-src).
This also hides the whitespace until the image is displayed.

### DIFF
--- a/www/js/app.js
+++ b/www/js/app.js
@@ -777,8 +777,9 @@ define(['jquery', 'zimArchiveLoader', 'util', 'uiUtil', 'cookies','abstractFiles
         $("#articleContent").contents().scrollTop(0);
 
         // Display the article inside the web page.
+        //Fast-replace img with data-img and hide image [kiwix-js #272]
+        htmlArticle.replace(/(<img\s+[^>]*\b)src(\s*=)/ig, "$1data-kiwixsrc$2");
         $('#articleContent').contents().find('body').html(htmlArticle);
-        
         
         // If the ServiceWorker is not useable, we need to fallback to parse the DOM
         // to inject math images, and replace some links with javascript calls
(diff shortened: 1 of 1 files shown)


www/js/app.js
```diff
@@ -786,9 +786,11 @@ define(['jquery', 'zimArchiveLoader', 'util', 'uiUtil', 'cookies','abstractFiles
         $("#articleContent").contents().scrollTop(0);
 
         // Display the article inside the web page.
+        //Fast-replace img with data-img and hide image [kiwix-js #272]
+        htmlArticle = htmlArticle.replace(/(<img\s+[^>]*)src(\s*=)/ig,
+            "$1style=\"display: none;\" onload=\"this.style.display='inline'\" data-src$2");
```
mossroy: In general, jQuery certainly does the parsing better (in terms of compatibility) than we might do manually, even with a regexp. I'm thinking about old browser compatibility and/or the many ways to write HTML code.
For example, I'm not 100% sure what would happen if some style, onload or data-src attributes already existed in the img tag.
It seems jQuery can also work on HTML strings with https://api.jquery.com/jQuery.parseHTML/

But I have to say you did it very well, as you handle the case-sensitivity and multiline aspects.
Regarding speed, it's certainly much faster than a complete jQuery parsing of the String, that will be done again when the string will be inserted in the DOM with .html(htmlArticle).

Another idea would be to make jQuery parse the string with .parseHTML, change the img src attributes to data-src with jQuery, and then inject the elements in the body with .append instead of .html. So that the jQuery parsing is only done once (I did not test, though). The performance must be tested, but it might be similar, and we might use all the power of jQuery if we want to do more changes in the HTML before injecting it.

Jaifroid: I'll reply to this on #272 because this PR may be closed.

Jaifroid:
Just to say I've updated the PR with just the code that eliminates 404s, and not the code that hides the display of images. You'll see it leaves gaps which may be puzzling to the user until filled with images. I have left the old code in a comment until it is decided whether to use this (depending on outcome of other speed enhancements).

mossroy:
In my opinion, this PR might be merged with a few minor changes : 
- remove the commented code
- rebase (which should allow Travis build to be successful) and squash

Removing the 404s on images is worth it, even if there might still be some 404s on other kind of contents (css, audio/video etc)

Issue #268, opened 2017-06-25
Update jQuery to 3.2.1 or 3.2.1.slim


Jaifroid:
I have tested both in Edge and Firefox, and they cause no problems I can see. Since we seem to be using only standard jQuery functionality, it may make sense to use the slim version (NB this is not minified, it's just a subset of jQuery functionality, the most used ones).

**Rationale**: keeping it up to date keeps on top of latest bugfixes and security issues.

**Disadvantage**: may not be backwardly compatible with legacy browsers or browser engines. Needs testing on FFOS, Symbian (?), Internet Explorer.

mossroy:
Officially, jQuery does not support that much browsers/versions (see http://jquery.com/browser-support/). It only supports very recent versions of Firefox and Chrome, no Edge on mobile, no IE on mobile, Symbian etc.
But that's the official support. We currently only use some basic features of jQuery, so it might work without problems, even on some older browsers. But that needs to be checked.
I think it's worth a PR.
I don't know if the slim version is enough : bootstrap uses some effects on menus (on small screens), and I don't know if it's based on jQuery or not.
I was thinking about adding some transitions (see #74), but we can stil switch to the non-slim version later, if we need it. So it's probably worth trying the slim version, to see if it works or not.
I'll do the testing on FirefoxOS

Jaifroid:
I'll include 3.2.1 and 3.2.1-slim in the repo and a commented switch between them in init.js.

(Btw, Edge mobile is supposed to be compiled from the same source as Edge desktop, and MS keeps them in sync approx. every other build.)

PR #274, opened 2017-06-26
Implements #268 Update jQuery to 3.2.1

Jaifroid:
Implements #268 .
The file selected in init.js is 3.2.1.slim.


www/js/lib/jquery-3.2.1.js
```diff
@@ -0,0 +1,10253 @@
+/*!
+ * jQuery JavaScript Library v3.2.1
```
mossroy: If we confirm the use of the slim version, we should remove this version (but it was a good idea to keep it for testing the PR before deciding)

www/js/init.js
```diff
@@ -25,9 +25,11 @@
 require.config({
     baseUrl: 'js/lib',
     paths: {
-        'jquery': 'jquery-2.1.4',
+        //'jquery': 'jquery-2.1.4', // Implements #268
+        //'jquery': 'jquery-3.2.1', // Full version
+        'jquery': 'jquery-3.2.1.slim', // Slim version
```
mossroy: Of course, we should remove the commented lines.
Please also switch to the new version of jQuery for the unit tests (it also works), in tests/init.js, and also in /nbproject/project.properties (this is for NetBeans autocompletion with requirejs : I'll do the testing)

I don't remember why we kept the version number in the file name of jQuery (and not for every other library). It might be the occasion to remove it, so that the file name would simply be jQuery.js

Jaifroid: The packaged version of jQuery comes with a version number, so I think that's why it has version. It's probably useful to future developers to be able to see at a glance which version is being used. I think especially if we use .slim future developers will need to be able to see this at a glance. I would therefore recommend keeping. (It may be the case that, with so many versions of jQuery floating around the Web, the developers of jQuery decided to use explicit version numbering in their releases.)

mossroy: ok : you can keep the full filename if you prefer

Jaifroid:
I can confirm that it works on Edge Mobile. I've now made the requested changes, including to the test files. Please check, @mossroy that it's in order, and if necessary test the tests before squash (?) and merge of this branch!

www/js/init.js
```diff
@@ -25,9 +25,9 @@
 require.config({
     baseUrl: 'js/lib',
     paths: {
-        'jquery': 'jquery-2.1.4',
+        'jquery': 'jquery-3.2.1.slim',
         'bootstrap': 'bootstrap'
-    },
+     },
```
mossroy: Very very minor : this extra space is unnecessary

mossroy:
Sorry I forgot to ask you to add something else in this PR :
I would suggest to remove the jQuery.globalEval overriding at the beginning of app.js (lines 32 to 39).
It is some FirefoxOS-specific code, to prevent calls to eval() that were forbidden in "packaged" FirefoxOS applications.

This is obsolete because jQuery 3.2.1 does not seem to call eval() any more, and this overriding might break some things.
I tested on my FirefoxOS device, and it seems to work fine without these lines.

NB : I can put that in another PR if you prefer

mossroy:
Perfect, thanks

Issue #127, opened 2015-09-13
Reopen the last article after switching between menu items


mossroy:
A cookie might keep the last article

Jaifroid:
Kiwix-js-windows now implements re-displaying the last article after switching back from "Config", without re-loading the article. (It also disables the link back to the article if the user has switched CSS display mode, as this necessitates reloading the article). I'm in the process of backporting to kiwix-js to commit just the switching between tabs to a new branch built on the main branch (so as to be able to make a PR for testing from main).

The testing branch is here: https://github.com/kiwix/kiwix-js/tree/Return-to-last-article-from-config-without-reloading , but it has no changes yet.

If anyone wants to test the full feature set, they need to do so on kiwix-js-windows until I edit this message.

mossroy:
As discussed in #291, I think switching between menus should "naturally" keep the context.
Without doing anything special, the user should see its last article after going to and back from another section. It should only be a matter of not emptying a div, in order to be able to see its content when un-hiding it.
That would be the first step.

A second step would be to store what's necessary in a cookie, so that to be able to re-open the same article after closing/re-opening the app. Which is a bit more complicated because of the different ways to open a ZIM file

mossroy:
I renamed this issue, to narrow it to what can be implemented for now : be able to return to the same article after going to another section (settings, or about)
As it's not possible to reopen the ZIM file automatically (in a portable and cross-browser way), it's not possible to do more.

Jaifroid:
OK. It is possible to remember the last page for each ZIM and load it after the user has picked the ZIM. We simply store a cookie value like `lastPageLoaded=Page_d'accueil@wikipedia_fr_all_novid_2018-10.zim` and load it if the picked ZIM name matches. In Web context, it's not so useful, but in app context it can be very useful (albeit that it has privacy implications).

mossroy:
You're right

> In Web context, it's not so useful, but in app context it can be very useful

That's why I think it's better to keep it in kiwix-js-windows, without backporting it for kiwix-js (where it would only be useful for the dead Firefox OS...).

With this renamed issue, it should be not necessary to store the info in a cookie, and there will be no privacy concern

Rbcoder1:
I am working on #523  but I need to build this functionality of returning to last article 

That's why am solving this issue first

Rbcoder1:
We just need to implement the functionality of returning to last article after unclick on config & about button 

[...]

PR #291, opened 2017-07-19
Partially implements #127

Jaifroid:
This implements the most urgent aspects of #127 (Reopen the last article after closing/reopening, or after switching between menu items). It implements a link to go back to the article, and disables the link if a major UI option is changed (so as to force a reload). It does not implement storing the article title in a cookie.

Please note that this PR also contains a channel for posting a message to the user on the "spinner" page, which is available here for future use (I use it on kiwix-js-windows). As these changes are very closely related, it seemed best to include it as an available utility in uiUtil.js. Messages posted to the user via uiUtil.poll(msg) will only be visible if the interface is taking a long time to load a file, hence only really on mobile with heavy pages.

### DIFF
--- a/www/index.html
+++ b/www/index.html
@@ -189,6 +189,7 @@
                 </div>
 
                 <div id="configuration" style="display: none;" class="category-padding">
+                <div id="returntoArticle_top" style="text-align: center;"></div>
                     <h2>Configuration</h2>
                     This application needs a ZIM archive to work. If you did not download one yet, please see the About section
                     <div id="openLocalFiles" style="display: none;">
@@ -208,13 +209,14 @@
                     <div id="contentInjectionModeDiv">
                         Do not touch unless you know what you're doing. <br />
                         Content injection mode : <br/>
-                        <input type="radio" name="contentInjectionMode" value="jquery" id="jQueryModeRadio" checked><label for="jQueryModeRadio">JQuery</label> (slow and memory hungry, but safer)
+                    <input type="radio" name="contentInjectionMode" value="jquery" id="jQueryModeRadio" checked><label for="jQueryModeRadio">&nbsp; JQuery</label> (slow and memory hungry, but safer)
                         <br>
-                        <input type="radio" name="contentInjectionMode" value="serviceworker" id="serviceworkerModeRadio"><label for="serviceworkerModeRadio">ServiceWorker</label> (faster but unstable, and not supported by all platforms. In particular, Firefox extensions do not allow ServiceWorkers for now)
+                    <input type="radio" name="contentInjectionMode" value="serviceworker" id="serviceworkerModeRadio"><label for="serviceworkerModeRadio">&nbsp; ServiceWorker</label> (faster but unstable, and not supported by all platforms. In particular, Firefox extensions do not allow ServiceWorkers for now)
                     </div>
                     <div id="serviceWorkerStatus"></div>
                     <div id="messageChannelStatus"></div>
                     <br />
+                <div id="returntoArticle_bottom" style="text-align: center;"></div>
                 </div>
 
                 <div id="welcomeText" class="container category-padding">
@@ -229,7 +231,8 @@
                     </div>
                 </div>
                 <div id="readingArticle" style="display: none;" class="container">
-                    Reading article <span id="articleName"></span> from archive... Please wait <img src="img/spinner.gif" alt="Please wait..." />
+                    <p>Reading article <span id="articleName"></span> from archive... Please wait <img src="img/spinner.gif" alt="Please wait..." /></p>
+                <div id="progressMessage"><p></p></div>
                 </div>
                 <iframe id="articleContent" src="A/dummyArticle.html" class="articleIFrame">&nbsp;</iframe>
             </article>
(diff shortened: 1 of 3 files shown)


mossroy:
One minor comment : you don't have to put the "implements #xxx" in the title of your commits or pull requests. It would be more readable to put a meaningful title, and put this string below (in the following lines of the commit comment), or in the PR description. GitHub will use them to close the corresponding issue too.

www/index.html
```diff
@@ -189,6 +189,7 @@
                 </div>
 
                 <div id="configuration" style="display: none;" class="category-padding">
+                <div id="returntoArticle_top" style="text-align: center;"></div>
```
mossroy: Instead of adding specific links, I would find more user-friendly to return to the previous article when clicking on the Home button (instead of reloading the home page).
It would also work when coming back from the About section.

www/js/app.js
```diff
@@ -771,15 +773,19 @@ define(['jquery', 'zimArchiveLoader', 'util', 'uiUtil', 'cookies','abstractFiles
      * @param {String} htmlArticle
      */
     function displayArticleInForm(dirEntry, htmlArticle) {
+        // Display the article inside the web page.
+
+        //Fast-replace img with data-img and hide image [kiwix-js #272]
+        htmlArticle = htmlArticle.replace(/(<img\s+[^>]*\b)src(\s*=)/ig, "$1data-kiwixsrc$2");
```
mossroy: This does not seem to be related to #127. I suppose it's because you did not create your branch from master?

www/js/lib/uiUtil.js
```diff
@@ -53,11 +53,33 @@ define([], function() {
...
+        var link = '<h4><a href="#" onclick="$(\'#configuration\').hide();$(\'#formArticleSearch\').show();' +
+            '$(\'#articleContent\').show();$(\'#liHomeNav\').attr(\'class\',\'active\');$(\'#liConfigureNav\').attr(\'class\', \'\');' +
+            '$(\'#liAboutNav\').attr(\'class\', \'\');">&lt;&lt; Return to ' + shortTitle + '</a></h4>';
+        document.getElementById("returntoArticle_top").innerHTML = link;
+        document.getElementById("returntoArticle_bottom").innerHTML = link;
+    }
+
+    function poll(msg) {
+        document.getElementById("progressMessage").innerHTML += '<p>' + msg + '</p>';
+    }
```
mossroy: I don't see this function ever called?

www/js/lib/uiUtil.js
```diff
@@ -53,11 +53,33 @@ define([], function() {
         }
     }
 
+    function makeReturnLink(title) {
+        //Abbreviate title if necessary
+        var shortTitle = title.substring(0, 25);
+        shortTitle = shortTitle == title ? shortTitle : shortTitle + "..."; 
+        var link = '<h4><a href="#" onclick="$(\'#configuration\').hide();$(\'#formArticleSearch\').show();' +
+            '$(\'#articleContent\').show();$(\'#liHomeNav\').attr(\'class\',\'active\');$(\'#liConfigureNav\').attr(\'class\', \'\');' +
+            '$(\'#liAboutNav\').attr(\'class\', \'\');">&lt;&lt; Return to ' + shortTitle + '</a></h4>';
```
mossroy: I think we should avoid putting some HTML inside javascript.
This HTML code could probably be put in the HTML file, with the dynamic part identified by an id and filled from the javascript. Plus using the visibility to show/hide it.

www/index.html
Jaifroid: I agree. This was more a quick solution attempt, especially given the discoverability issues on mobile (home button is hidden inside a clunky hamburger menu, taking more clicks to open and navigate around). Hence attempt (not very elegant, I agree!) to provide clear visual clues.

www/js/app.js
Jaifroid: I just moved that code, it's not an addition. I did create the branch from master, and backported code into it from kiwix-js-windows. This move was "en passant".

www/js/lib/uiUtil.js
Jaifroid: I tried, but having "<h4>" in the HTML caused an obvious gap, even when empty. I would have had to have added "element.hide()" code in several different places I was uncertain of.... It's definitely a hack...

www/js/lib/uiUtil.js
Jaifroid: Indeed, it's for future use as mentioned in the PR comment -- a message channel.

Jaifroid:
Thanks for the comments, @mossroy , I appreciate your guidance as I haven't grasped fully the intricacies of the app.

This was admittedly a quick attempt to add a feature that has been in the works for quite a while and is a real usability issue with the app,, as our Ubuntu Phone friend pointed out. There's also a discoverability issue on a phone: you have to open the hamburger menu in order to navigate between the tabs, so in-app navigation isn't intuitive on a phone.

I feel a few things need fixing urgently before I can put kiwix-js-windows in the Store, but I don't have very much time, hence the rush -- and apologies for that. It can certainly be done better than this, though I don't think I have the knowledge of the libraries to do things in a standard way. I'm happy to close the PR and leave the branch for a while in case anyone else can have a go at doing it properly.

mossroy:
Don't apologize, it's still a step towards a fix.

No special library is used to switch between sections : it's simply programmatically hiding/un-hiding some parts of the DOM when the user clicks on some menu items.
At some point, the article content is emptied : this should be done at another moment, and the home page should be displayed only if the content is empty

PR #299, opened 2017-08-08
Handle uncaught exception and display welcome text...

Jaifroid:
…if ZIM file does not have main page.

For explanation, see #300 (Display Welcome text if ZIMfile does not have main page).


Jaifroid:
No problem, @mossroy . My main editor doesn't use tabs, but because this is a very different branch from what's currently loaded in my main editor, I used a different one for the minor edit. Will correct and squash etc.

mossroy:
Great : you can squash and merge

Issue #302, opened 2017-08-27
Use Bootstrap panels to organize Configuration page


Jaifroid:
In Kiwix JS Windows I'm now using Bootstrap panels to organize the Configuration page. I'd be very happy to backport it here, if it's wanted. Obviously only the options implemented / used in Kiwix JS would be backported, so this should be decided after it is decided which, if any, of those options to adopt. So I haven't added a milestone. Screenshots below (please ignore typo in second screenshot!!).

### Top half of Configuration page:

![image](https://user-images.githubusercontent.com/4304337/29748955-a2f7ff0a-8b19-11e7-87a4-d0bc91718fba.png)
 
 
 
### Bottom half of Configuration page:

![image](https://user-images.githubusercontent.com/4304337/29748969-cb19688e-8b19-11e7-9519-a135ced3e68b.png)

mossroy:
Good idea, it makes the UI more clear and prettier. That's worth a pull request, even if it might not be a top priority.
It's probably not an issue to add/remove some options afterwards, I suppose the merge would not be very complicated

Jaifroid:
OK, I've added a PR, but with only one panel the benefit is not so obvious. But it lays down the standard grid system which is then  very easy to add to as we add more options and potentially more sections (plain Settings as well as Expert Settings, as per screenshots above).

Jaifroid:
Just to say, there is a new commit for PR #303 .

PR #303, opened 2017-08-29
Use Bootstrap panels to organize Configuration page #302

Jaifroid:
[ Kiwix-JS #302 ]
Apart from panels, I have included a skeleton Bootstrap grid layout (responsive divs with flexbox) on the page. These can be arranged as desired by adding classes described at https://getbootstrap.com/docs/3.3/css/#grid-options . More advanced options will only make sense if we introduce more options / panels and a non-expert settings section.

### DIFF
--- a/www/index.html
+++ b/www/index.html
@@ -189,37 +189,66 @@
                 </div>
 
                 <div id="configuration" style="display: none;" class="category-padding">
-                    <h2>Configuration</h2>
-                    This application needs a ZIM archive to work. If you did not download one yet, please see the About section
-                    <div id="openLocalFiles" style="display: none;">
-                        Please select the .zim file (or all the .zimaa, .zimab etc in case of a split ZIM file)<br />
-                        <input type="file" id="archiveFiles" multiple class="btn" accept=".zim,.dat,.idx,.txt,.zimaa,.zimab,.zimac,.zimad,.zimae,.zimaf,.zimag,.zimah,.zimai,.zimaj,.zimak,.zimal,.zimam,.ziman,.zimao,.zimap,.zimaq,.zimar,.zimas,.zimat,.zimau,.zimav,.zimaw,.zimax,.zimay,.zimaz,.zimba,.zimbb,.zimbc,.zimbd,.zimbe,.zimbf,.zimbg,.zimbh,.zimbi,.zimbj,.zimbk,.zimbl,.zimbm,.zimbn,.zimbo,.zimbp,.zimbq,.zimbr,.zimbs,.zimbt,.zimbu,.zimbv,.zimbw,.zimbx,.zimby,.zimbz" /><br />
-                        <strong>Only Wikimedia contents (wiki*.zim* files) have been tested</strong> for now. There are known bugs on other contents, and audio/video/dynamic contents are not supported for now.<br />
+                    <div class="container">
+                        <div class="row">
+                            <h2>Configuration</h2>
+                            <div class="column">
+                                This application needs a ZIM archive to work. If you did not download one yet, please see the About section
+                                <div id="openLocalFiles" style="display: none;">
+                                    Please select the .zim file (or all the .zimaa, .zimab etc in case of a split ZIM file)<br />
+                                    <input type="file" id="archiveFiles" multiple class="btn" accept=".zim,.dat,.idx,.txt,.zimaa,.zimab,.zimac,.zimad,.zimae,.zimaf,.zimag,.zimah,.zimai,.zimaj,.zimak,.zimal,.zimam,.ziman,.zimao,.zimap,.zimaq,.zimar,.zimas,.zimat,.zimau,.zimav,.zimaw,.zimax,.zimay,.zimaz,.zimba,.zimbb,.zimbc,.zimbd,.zimbe,.zimbf,.zimbg,.zimbh,.zimbi,.zimbj,.zimbk,.zimbl,.zimbm,.zimbn,.zimbo,.zimbp,.zimbq,.zimbr,.zimbs,.zimbt,.zimbu,.zimbv,.zimbw,.zimbx,.zimby,.zimbz" /><br />
+                                    <strong>Only Wikimedia contents (wiki*.zim* files) have been tested</strong> for now. There are known bugs on other contents, and audio/video/dynamic contents are not supported for now.<br />
+                                </div>
+                                <div id="scanningForArchives" style="display: none;">
+                                    <br /> Scanning for archives... Please wait <img src="img/spinner.gif" alt="Please wait..." />
+                                </div>
+                                <div id="chooseArchiveFromLocalStorage" style="display: none;">
+                                    <br /> Please select the archive you want to use : <select id="archiveList" class="form-control"></select>
+                                    <br /> Click <a id="btnRescanDeviceStorage">here</a> to rescan your SD Cards and internal memory
+                                </div>
+                            </div>
+                        </div>
                     </div>
-                    <div id="scanningForArchives" style="display: none;">
-                        <br /> Scanning for archives... Please wait <img src="img/spinner.gif" alt="Please wait..." />
-                    </div>
-                    <div id="chooseArchiveFromLocalStorage" style="display: none;">
-                        <br /> Please select the archive you want to use : <select id="archiveList" class="form-control"></select>
-                        <br /> Click <a id="btnRescanDeviceStorage">here</a> to rescan your SD Cards and internal memory
+
+                    <div class="container">
+                        <div class="row">
+                            <h3>Expert settings</h3>
+                            <div class="column">
+                                <div class="panel panel-danger" id="contentInjectionModeDiv">
+                                    <div class="panel-heading">Content injection mode</div>
+                                    <div class="panel-body">
+                                        Don't touch unless you know what you're doing!
+                                        <div class="radio">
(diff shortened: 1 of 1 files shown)


www/css/app.css
```diff
@@ -49,6 +49,11 @@
   padding: 0;
 }
 
+.panel-body p {
+    text-indent: -22px;
+    margin-left: 20px;
+}
```
mossroy: What is the purpose of this style overriding?
I don't see a big difference if I remove it (at least on Firefox desktop)

www/index.html
```diff
@@ -189,37 +189,57 @@
...
-                        <br /> Click <a id="btnRescanDeviceStorage">here</a> to rescan your SD Cards and internal memory
+
+                    <div class="container">
+                        <h3>Expert settings</h3>
+                        <div class="row">
+                            <div class="column">
+                                <div class="panel panel-danger" id="contentInjectionModeDiv">
+                                    <div class="panel-heading">Do not touch unless you know what you're doing!</div>
+                                    <div class="panel-body">
+                                        <p>Content injection mode:</p>
```
mossroy: I think it would be better to put "Content injection mode (do not touch unless you know what you're doing)" as the panel heading, and remove this <p> tag

www/index.html
```diff
@@ -189,37 +189,57 @@
...
+                                            <b>&nbsp;JQuery</b> (slow and memory hungry, but safer)
+                                        </p>
+                                        <p>
+                                            <input type="radio" name="contentInjectionMode" value="serviceworker" id="serviceworkerModeRadio">
+                                            <b>&nbsp;ServiceWorker</b> (faster but unstable, and not supported by all platforms)
+                                        </p>
+                                    </div>
+                                </div>
+                                <div id="serviceWorkerStatus"></div>
+                                <div id="messageChannelStatus"></div>
```
mossroy: Maybe those two divs could be put inside a panel too, with a heading like "API status"?

www/css/app.css
Jaifroid: Alignment of the text after radio buttons (or checkbox), so that, if it wraps (on narrow screens) it aligns with the start of the first word, and not with the extreme left of the radio button, which looks ugly...

www/index.html
Jaifroid: Will do.

www/index.html
Jaifroid: Good idea, I'll take a look.

www/css/app.css
mossroy: It seems to be possible to do that while keeping the <label> tag (I find it important, as it allows the user to click on the label to select it, instead of only the radio-button itself), and avoiding the fixed pixel numbers. With display:table-* : I've put an example on http://jsfiddle.net/z59bkdo0/2/
This seems to be widely supported by browsers : http://caniuse.com/#feat=css-table
NB : I'm not a CSS expert, so not sure it's the best way

Jaifroid: Great, I wasn't aware of this (and t.b.h. originally when I changed the labels to paras it was a "quick fix" I'd since forgotten about).  Will experiment.

Jaifroid:
OK, @mossroy , I've discovered Bootsrap has its own way of managing input controls, so we don't need the table-row and table-cell tweaks. See: https://www.w3schools.com/Bootstrap/bootstrap_forms_inputs.asp . However, we do still need an override on the margin-left (in app.css) because otherwise the radio buttons get pushed -20px right up against the panel wall. And we have to override the default bold font for the label.

Also, I've used bold (`<b>`) rather than strong (`<strong>`) because I feel it is purely typographical rather than a question of emphasis. If you feel strongly about using strong :-), no worries. We're overriding bootstrap's `font-weight: bold` for the label as a whole.

Finally, let me know if you prefer "Expert Settings" to be aligned with the left edge of the box, or with the left-edge of text inside the box. In the last example it was aligned with the text inside the box, but I've changed it here. Pros and cons either way. It's a simple case of where it is place, inside the overall panel div or inside the row div.

Anyway, see what you think. Screenshot is from the FFOS simulator.

image

mossroy:
You're right : it's a better idea to use the Bootstrap way to do that.
But we should not have to override the default bootstrap css, if we follow their example https://getbootstrap.com/docs/3.3/css/#forms (scroll to the radio buttons example)

I managed to have the same result with the following HTML, and without any .radio CSS overriding in app.css : 
```html
<div class="radio">
    <label>
       <input type="radio" name="contentInjectionMode" value="jquery" id="jQueryModeRadio" checked>
       <strong>JQuery</strong> (slow and memory hungry, but safer)
    </label>
</div>
<div class="radio">
    <label>
        <input type="radio" name="contentInjectionMode" value="serviceworker" id="serviceworkerModeRadio">
        <strong>ServiceWorker</strong> (faster but unstable, and not supported by all platforms)
    </label>
</div>
```

Regarding `<b>` vs `<strong>`, it's not crucial at all, and I'm no expert.
I read the explanation of Mozilla in https://developer.mozilla.org/en-US/docs/Web/HTML/Element/strong and I think we try to express the fact that this word is the one the user should read first to understand what the radio-button does. While the rest of the sentence gives more details (that can be skipped if the user already knows what he's doing). I see that as a difference of importance, that would be better expressed as a logical state in the HTML.
But, again, it's no big deal.

Regarding the alignment of "Expert settings", I like it this way (and have to admit I forgot how it was before ;-) )

mossroy:
I forgot to mention that the 2 lines that are telling which APIs are available might be isolated in a different panel, called "API Status".
But that can be done later if necessary.

Jaifroid:
Great, thanks @mossroy , will take a look. Glad you found a way round the override (or rather that I had introduced unnecessary complication!). Always good to find simplest / most efficient solution.

Jaifroid:
OK, @mossroy , hopefully this does the trick now! The problem was indeed not having an extra div with a "radio" class, and adding that means no need for a css override on label, as you discovered. All these divs look a bit excessive to me, but hey, that's the bootstrap way, and it has the advantage of being extensible. So worth getting right, to make it easier to build on when we add more options.

If you'd like to add an extra panel for the APIs at this stage, could I leave that to you to add to this PR? If you prefer PRs to be contributed by a single developer, that's OK, just tell me the HTML you prefer.

Otherwise, if you're happy with this, I'll squash / merge.

mossroy:
Sorry for the delay, I don't have much time those days.
I was about to add the API panel (OK to commit it myself, no pb). But I discovered that the icons of the bottom bar were mis-aligned in this branch (with or without the API panel) : the back and forward icons are on the left of the screen instead of the center.
There's certainly something wrong in our use of the bootstrap APIs (and it might be in the bottom bar from the beginning, revealed by your changes).

Jaifroid:
Don't worry, I'm extremely busy too. I noticed that issue, but oddly I don't think I've touched the footer... It could however have something to do with panel widths. I'll experiment when I next get a chance.

kelson42:
Would be good to go ahead on that PR :)

kelson42:
kiwix-js-windows seems to have fixed the problem in git master HEAD. Hope to see a refresh of that PR and cross the finger for an approval/merge :)

Jaifroid:
I think it's a slightly different problem here (alignment rather than button widths), but I have a better (not complete!) understanding of how the bootstrap alignment works, so will take a look soon and update PR if I can get a consistent fix.

Jaifroid:
@mossroy Going back to this bootstrap panel PR, I noticed when testing #310 that the icons are misaligned in that branch too, and that branch has none of the code from this PR. See screenshot. This implies to me that the issue with alignment here is prior to this PR and independent of it. I'm not saying it isn't a bootstrap issue more generally, but the changes proposed in this PR (and those you were going to do adding in the API panel) are probably not related to the alignment issue that has held up this PR. 

Would it be appropriate to try to fix the alignment issues in this PR, or is it better to accept the panels as one PR (after adding in the API panel), and do a new one for the alignment?

image

mossroy:
I also noticed it, and came up to the same conclusion.
As this PR is already old, we might merge it and create a new issue for the alignment of icons.
I'll try to do that this weekend too (I had to push a new commit on this branch to add a panel dedicated to the API status)

mossroy:
I pushed the commit for the API status panel : @Jaifroid please have a look if it's OK for you.
If it is, we might squash/rebase and merge this branch.
And create another issue for the alignment of icons

Jaifroid:
This looks great, @mossroy . I'll squash/rebase.

mossroy:
The squash/rebase looks wrong to me. You seem to have merged another branch into this one.
I'll try to re-do it from my local git repo

Jaifroid:
Yes, I think I did something wrong earlier, but the master branch now has the correct single commit, so I could just close this and delete this branch?

mossroy:
There was a single commit a few minutes earlier, but it just changed to 3 commits : did you push something?
I'll try to push my single commit again. If it looks good to you, you can merge the PR and delete the branch on github.
It's probable that your local branch will be out of sync, but we won't care any more

Jaifroid:
Clearly I need more practice with the squash/merge thing... Something weird happened when I did it on my local repository: it pulled in the "translate some items from French" commit from master into this branch. I managed to squash it all into one commit, but every time I pushed this branch, a "Merge branch" commit would get added back on top of the other commits...

Anyway, thanks for correcting. I'll merge this now, so we can delete the branch.

Issue #205, opened 2017-04-06
Links do not work on StackExchange archives


mossroy:
It seems to be because the articles are not all at the same directory level

kelson42:
Yes, I confirm that bug. If articles have "/" in their URL then it fails. Try for example with that file http://tmp.kiwix.org/prog.zim

kelson42:
This problem basically impacts all StackExchange ZIM files http://download.kiwix.org/zim/stack_exchange/

dattaz:
If i remember what we say in hackathon, you should inject an "base" tag  (https://developer.mozilla.org/fr/docs/Web/HTML/Element/base) when you have a redirect.

Jaifroid:
@mossroy The two edits in commit https://github.com/kiwix/kiwix-js/commit/76055a921372047560d885530a325d1a19495936 on the [Support-for-Stackexchange-links](https://github.com/kiwix/kiwix-js/tree/Support-for-Stackexchange-links) branch seem to do the trick. Question links, tags, and user profile links all seem to be working.

I tested with the FFOS Simulator (thanks for your help re-enabling it) using `mathoverflow.net_en_all_2017-10.zim` . I've also tested this code with `wikipedia_fr_tunisie_2017-12.zim
` to be sure we haven't introduced a problem with non-Stackexchange ZIMs, and so far haven't found anything that's not working.

Would you like to test that branch before I do a PR from it? It would be good also to test with other Stackexchange ZIMs, as the mathsoverflow one may have its own peculiarities or be a special case.

![image](https://user-images.githubusercontent.com/4304337/33946128-da8890e8-e018-11e7-9453-c879189121d1.png)

mossroy:
Great : I'll try to test that tomorrow. But you can create the PR if you like

Jaifroid:
OK, PR https://github.com/kiwix/kiwix-js/pull/309 created from that branch.

Jaifroid:
I've added commit https://github.com/kiwix/kiwix-js/pull/309/commits/0253fb3b7f0a2864badf083b4916f418bbf7aee7 to the PR to support "Duplicate" and "Related" question types. See comment in code.

PR #309, opened 2017-12-13
Fixing Stackexchange links #205

Jaifroid:
This seems to support links, tags, and even user profiles. Tested with `mathoverflow.net_en_all_2017-10.zim` on Firefox OS Simulator. I also just tested it with `philosophy.stackexchange.com_en_all_2017-10.zim`. My tests have been a bit cursory, but no errors found yet.

### DIFF
--- a/www/js/app.js
+++ b/www/js/app.js
@@ -850,13 +850,9 @@ define(['jquery', 'zimArchiveLoader', 'util', 'uiUtil', 'cookies','abstractFiles
                     // Add an onclick event to go to this article
                     // instead of following the link
                     
-                    if (url.substring(0, 2) === "./") {
-                        url = url.substring(2);
-                    }
-                    // Remove the initial slash if it's an absolute URL
-                    else if (url.substring(0, 1) === "/") {
-                        url = url.substring(1);
-                    }
+					//Get rid of any absolute or relative prefixes (../, ./../, /../.., etc.)
+                        url = url.replace(/^[.\/]*([\S\s]+)$/, "$1");
+
                     $(this).on('click', function(e) {
                         var decodedURL = decodeURIComponent(url);
                         pushBrowserHistoryState(decodedURL);
--- a/www/js/app.js
+++ b/www/js/app.js
@@ -854,7 +854,7 @@ define(['jquery', 'zimArchiveLoader', 'util', 'uiUtil', 'cookies','abstractFiles
                         url = url.replace(/^[.\/]*([\S\s]+)$/, "$1");
                         //Some Stackexchange links (e.g. "duplicate" and "related" questions) are missing the "question" path, so add it back
                         //Regex matches a pattern that looks like: 1234/what-is-mathematics.html and changes to: question/1234/what-is-mathematics.html
-                        url = url.replace(/^(\d+\/[\s\S]+\.html?)$/ig, "question/$1");
+                        url = url.replace(/^(\d+\/[\s\S]+\.html?)$/i, "question/$1");
 
                     $(this).on('click', function(e) {
                         var decodedURL = decodeURIComponent(url);
--- a/www/js/app.js
+++ b/www/js/app.js
@@ -852,6 +852,9 @@ define(['jquery', 'zimArchiveLoader', 'util', 'uiUtil', 'cookies','abstractFiles
                     
 					//Get rid of any absolute or relative prefixes (../, ./../, /../.., etc.)
                         url = url.replace(/^[.\/]*([\S\s]+)$/, "$1");
+                        //Some Stackexchange links (e.g. "duplicate" and "related" questions) are missing the "question" path, so add it back
+                        //Regex matches a pattern that looks like: 1234/what-is-mathematics.html and changes to: question/1234/what-is-mathematics.html
+                        url = url.replace(/^(\d+\/[\s\S]+\.html?)$/ig, "question/$1");
 
                     $(this).on('click', function(e) {
                         var decodedURL = decodeURIComponent(url);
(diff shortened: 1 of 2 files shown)


mossroy:
These changes make most of StackOverflow work, which is cool.
But there are still a few bugs. For example, if I open the article "Iterate through a HashMap" (from stackoverflow.com_eng_all_2017-05.zim), the "possible duplicate" link points directly to "46898.html" and does not work (because not matched by your regexp to add "question/" before it).

More generally, I think we should not introduce stackoverflow-specific code, or make any assumption on the ZIM structure.
Historically, the code makes a (wrong) assumption that all articles are at the same level : A/articlename.html. It's because we seem to always have this structure in mediawiki-based ZIM files and we initially only tested with this kind of ZIM files.
But it was wrong to assume that, as it is not in OpenZIM spec (http://www.openzim.org/wiki/Article_Format#Article_Text). At least stackoverflow-based ZIM files use some subdirectories for articles.

I think we need to find a generic way to support any directory structure. Because there might be other structures in other ZIM files, and even sotoki (that generates ZIM files from stackoverflow) might change its structure in the future.

But it's probably not trivial : we might need to resolve (in javascript) the target URL based on the current directory and the relative URL.
After a quick search, there seems to exist some already-made implementations that might do that : 

- https://gist.github.com/tkw1536/6980004 (based on jQuery) or even https://jsfiddle.net/ecmanaut/RHdnZ/ (without the jQuery dependency) : these implementations make the browser resolve the URL itself. This might be adequate in our case
- https://gist.github.com/Yaffle/1088850 (which implements the logic in pure javascript)

NB : this issue is specific to our "jQuery" mode. In ServiceWorker mode (and probably WebRequest mode), this would be handled by the browser itself.
NB2 : for my culture, why do you use "[\s\S]" syntax in your regexp expressions? How does it differ from a simple dot, that would match anything?

Jaifroid:
Hmm, OK, thanks @mossroy , I'll have a think about the above and get back to you. It would be nice to find a simple, generic way to prepend the full path. Note that the Stackexchange ZIMs still use "A/" as the namespace for articles. I don't need to add "A/" in my regex because it is added by the regex in [this line]( https://github.com/kiwix/kiwix-js/pull/309/files#diff-1916c6001d87418e72ffda9cfa7b2d82R262) of zimArchive.js.

If none of this is relevant to ServiceWorker, I'm a bit confused by the fact that we prepend "A/" in zimArchive.js -- doesn't ServiceWorker use zimArchive to load the ZIM for decompression? ServiceWorker is a black box for me...

Regarding `[\s\S]` this is an old habit that derives from the fact that JavaScript doesn't have the regex `/s` modifier that makes the dot match newline and end-of-line characters. You never quite know, when dealing with HTML, where there might be an end-of-line that breaks the regex, so I always just instinctively use this pattern when I want to match "anything" in JavaScript. However, it is true that in these specific cases we're dealing with variables pulled from the DOM, and information won't be split across multiple lines, so you're right that a dot will do the trick. Old habits die hard!

mossroy:
You're right: this "A/" prefix should not be in zimArchive.js, it was certainly a workaround for the relative URLs between articles in mediawiki ZIM files. It does not create issues for the ServiceWorker mode as the URL is always complete. But, yes, if we refactor the logic of resolving relative URLs, we should remove this from zimArchive.js

After re-thinking about this, the solution of https://jsfiddle.net/ecmanaut/RHdnZ/ would be un-optimal for us because it would create & delete a <base> tag for each link it parses.
In our case, the base URL is always the same. So a better solution would be to put it in the DOM once for all :
- in jQuery mode, we could always add this <base> tag in the DOM (with the basepath of the current article), as suggested by @dattaz in #205 
- then we could simply retrieve the URL resolved by the browser

It has to be tested bu, if it works, it would probably be a simple and elegant way to solve this issue (and it would allow to remove the "A"/ prepending in zimArchive.js too)

mossroy:
If it's not clear or if you don't have time, I'll try to do it when I have some time

Jaifroid:
One issue might be that the browser is seeing the path of the current document (as given by `document.location.pathname`) as always "/www/index.html". Is the idea that we modify this by dynamically injecting <base href="something"> (e.g. "/www/A/question/") so that the relative links can be recognized? 

There is another complication: it seems questions in Stackexchange ZIMs can be retrieved using either a "full syntax":

`question/33074/can-humans-maintain-a-free-relationship-with-technology.html`

Or with a "short syntax":

`question/33074.html` 

You can modify the URL on-the-fly by breaking on [this line](https://github.com/kiwix/kiwix-js/pull/309/files#diff-bca463ed8ccf5e2058dfe307a95e8501R860) and changing the value of url before continuing. With the above question from the Philosophy Stackexchange ZIM, either URL will retrieve the same page.

What you might have seen could have been merely a short-form reference rather than a relative URL? I can't test the particular example you give because the Stackexchange ZIM you refer to is 52Gb!!

In the meantime, I've modified the regex in the PR / branch to take into account short-form URLs -- as a "quick-n-dirty" approach just for now and for testing.

Jaifroid:
PS I tried changing the `[\s\S]+` notation to `.+` and something broke. I think it's to do with the "greediness" of the dot.... So keeping `[\s\S]+` for now.

dattaz:
Full syntax and short syntax are just redirection in the zim.
And for the zim you can download smaller here : https://download.kiwix.org/zim/stack_exchange/  (i'm quite sure you will reproduce your issue)

Jaifroid:
Thanks for the info @dattaz . I was hoping the Stackexchange ZIMs weren't storing each page twice!

Jaifroid:
I'm closing this PR for now because the more generic solution in https://github.com/kiwix/kiwix-js/pull/310 is better.

PR #310, opened 2017-12-15
Support subdirectories in articles

### LINKED ISSUE #205: Links do not work on StackExchange archives

mossroy:
Please do not merge for now.

I create the PR to have some space to discuss.
I've implemented the idea of the "base" tag, that should fix #205 
But it's still quick and dirty

### DIFF
--- a/www/js/app.js
+++ b/www/js/app.js
@@ -699,7 +699,7 @@ define(['jquery', 'zimArchiveLoader', 'util', 'uiUtil', 'cookies','abstractFiles
         $("#prefix").val("");
         findDirEntryFromDirEntryIdAndLaunchArticleRead(dirEntryId);
         var dirEntry = selectedArchive.parseDirEntryId(dirEntryId);
-        pushBrowserHistoryState(dirEntry.url);
+        pushBrowserHistoryState(dirEntry.namespace + "/" + dirEntry.url);
         return false;
     }
     
@@ -784,6 +784,8 @@ define(['jquery', 'zimArchiveLoader', 'util', 'uiUtil', 'cookies','abstractFiles
     // Compile some regular expressions needed to modify links
     var regexpImageLink = /^.?\/?[^:]+:(.*)/;
     var regexpPath = /^(.*\/)[^\/]+$/;
+    // Pattern for ZIM file namespace - see http://www.openzim.org/wiki/ZIM_file_format#Namespaces
+    var regexpZIMUrlWithNamespace = /(?:^|\/)([-ABIJMUVWX]\/.+)/;
     // These regular expressions match both relative and absolute URLs
     // Since late 2014, all ZIM files should use relative URLs
     var regexpImageUrl = /^(?:\.\.\/|\/)+(I\/.*)$/;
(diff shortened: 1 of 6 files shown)


Jaifroid:
Thank you for this, @mossroy , it looks like a more complete and generic solution. I'll do some testing.

www/js/app.js
```diff
@@ -813,8 +825,16 @@ define(['jquery', 'zimArchiveLoader', 'util', 'uiUtil', 'cookies','abstractFiles
...
+                // Compute current link's url (with its namespace), if applicable
+                var fullUrl = this.href;
+                var urlMatch = fullUrl.match(regexpUrl);
+                var url;
+                if (urlMatch) {
+                    url = urlMatch[0];
+                }
+                else {
+                    url = $(this).attr("href");
+                }
```
Jaifroid: @mossroy I couldn't resist trying all this as a one-liner....!
`var url = regexpUrl.test(this.href) ? this.href.match(regexpUrl)[0] : $(this).attr("href");`
I'm half joking, but perhaps there's no need to define `fullUrl`, when you could just match against `this.href`.

www/js/lib/zimArchive.js
```diff
@@ -256,10 +254,6 @@ define(['zimfile', 'zimDirEntry', 'util', 'utf8'],
      */
     ZIMArchive.prototype.getDirEntryByTitle = function(title) {
         var that = this;
-        // If no namespace is mentioned, it's an article, and we have to add it
-        if (regexpTitleWithoutNameSpace.test(title)) {
-            title= "A/" + title;
-        }
```
Jaifroid: Nice that you could clean this up!

Jaifroid:
@Mossroy PS I forgot to say I've tested this branch/PR against a couple of Stackexchange ZIMs, and a couple of Wikipedia ZIMs, and it all seems to function as intended. Nice work!

(I haven't tested ServiceWorker mode.)

Jaifroid:
@mossroy This just occurred to me:
The regex
`var regexpUrl = /(?:[-ABIJMUVWX]\/.*)$/; `
matches this string:
`Synthesis of RNA/12345.html`
And the matched part is:
`A/12345.html`
So we need a way to avoid that kind of rare but possible match of a capital letter at the end of part of the string. Your regex also matches `A/` (the match is `A/`), which might not be a problem but could be avoided by using `.+` instead of `.*`, to ensure at least one character after the forward slash.

Finally, I think the end-of-string anchor is redundant.

Consider rewriting as:

`var regexpUrl = /(?:^|\/)([-ABIJMUVWX]\/.+)/;`

The match, if any, will then be stored in match[1] rather than match[0]. By putting `(?:^|\/)` at the beginning, the regex will match either the start of the string followed by the pattern we want OR a / anywhere in the string followed immediately by the pattern we want. That way we don't match `Synthesis of RNA/12345.html` either, but we would match `Synthesis of RN/A/12345.html`. I tested this quickly and it seems to work (if you try, remember to change `urlMatch[0]` to `urlMatch[1]`).

mossroy:
Thanks for all this valuable feedback.
I agree with you and will update this PR (probably this weekend)

mossroy:
I pushed a commit with the changes you suggested. But it's not ready to merge yet.
The ServiceWorker mode seems to be broken, with no big surprise.
I'll try to have a look tomorrow.

mossroy:
I've pushed another commit that fixes some issues in ServiceWorker mode, and does a bit more code cleanup.
There are still bugs in ServiceWorker mode when subdirectories are used for articles. But it's not worse than before. I've created #312 to solve that later

mossroy:
I won't have more time to work on that today.
But it's probably worth merging it as it solves the issue in jQuery mode.
@Jaifroid : if it's good for you, I might rebase/squash.

I'll probably have a bit of time to work on kiwix-js by the end of the first week of January : maybe it would be worth releasing a version 2.2 at that time, that would include all the latest improvements we commited? That would in particular allow to use stackexchange ZIM files in browser extensions and Ubuntu touch.

Jaifroid:
@mossroy Would be good to work towards a 2.2 at end Jan. 👍

Issue #251, opened 2017-06-16
"Timed out while waiting for element <#formArticleSearch>" in Travis


mossroy:
Happened on build 287. Resolved by restarting the build.
See logs : 
[travis_build_287.zip](https://github.com/kiwix/kiwix-html5/files/1080654/travis_build_287.zip)

mossroy:
It does not seem to be the exact same symptom as in #228 
Maybe Travis is simply not very reliable...

mossroy:
Same issue (4 or 5 times) with build 289

mossroy:
Happened on build 294 too (nightly build) : maybe Sauce is getting slower sometimes.
In this case, we should raise the timeout

mossroy:
It seems to happen on every build now.
Let's wait a few days to see if Sauce fixes something

mossroy:
Raising the timeout and handling the onreadystatechange does not seem to solve the issue.
Maybe it's because the browser inside Sauce can not access github.io any more? Because one of them has blocked the traffic?

mossroy:
Reading the ZIM file from localhost instead of github.io does not seem to help either...
And, for some unknown reason, build 311 passed : https://travis-ci.org/kiwix/kiwix-html5/builds/244899133
Strange...

mossroy:
This currently makes all nightly builds fail.

mossroy:
The behavior now seems random, and I really think it comes from the CI platform itself (Sauce or Travis) : 
For the same PR https://github.com/kiwix/kiwix-html5/pull/261 , 2 builds are run at almost the same time, with the same code : https://travis-ci.org/kiwix/kiwix-html5/builds/245382199?utm_source=github_status&utm_medium=notification for the push fails, but  https://travis-ci.org/kiwix/kiwix-html5/builds/245385433?utm_source=github_status&utm_medium=notification for the pr succeeds

mossroy:
It might come from the version of Firefox (53 by default)
I tested with : 
- Firefox 52 : fails
- Firefox 54 : fails
- Chrome 59 : works
- Firefox 45 : works (even after restarting the job many times)
I don't understand the reason behind that, as all these versions of Firefox are working fine on my desktop...

So we might simply switch to version 45 for now (or to the last working version : I'll try to find which one). This version would also have the ServiceWorker support (if we want to add UI tests on it, which is not the case for now), but would not be enough for WebAssembly support (which needs Firefox >=52).

mossroy:
[...]

PR #317, opened 2017-12-18
Let UI tests run on the latest version of Firefox.

### LINKED ISSUE #316: Let the UI tests run (with Sauce/Travis) on the latest version of Firefox
Because of #251 , we had to force a version 47 of Firefox for the UI tests (else the tests were not reliable).
This was certainly a platform bug, not related to kiwix-js itself.

It would now be better to use the latest version of Firefox, in order to have the recent technologies (WebAssembly, full WebRequest API, Quantum technologies, headless mode etc)

mossroy:
Instead of using only version 47 (See #251)
Fixes #316

### DIFF
--- a/nightwatch.js
+++ b/nightwatch.js
@@ -1,5 +1,5 @@
-'use strict'
-var TRAVIS_JOB_NUMBER = process.env.TRAVIS_JOB_NUMBER
+'use strict';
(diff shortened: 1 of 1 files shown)


mossroy:
I close this PR, because #322 does the same, while adding more browsers

Issue #313, opened 2017-12-17
Make vCard geo: and tel: URIs work


Jaifroid:
Wikivoyage ZIMs use URIs of type:

`geo:37.04741,22.11429`

and

`tel:+302721023999`

These are part of the vCard schema, see in particular: https://en.wikipedia.org/wiki/Geo_URI_scheme . They are currently not handled in jQuery mode at least. It's quite easy to make them into links that will be handled by whatever the underlying Operating System (or browser) has registered to handle such links. For example, on MS systems or in Edge the "tel:" URI opens a dialogue box asking the user if they want to launch Skype or whatever other apps are registered to handle "tel:" URIs.

I can do a PR for this.

mossroy:
I agree.
It's probably only a matter of letting these URIs as they are : the jQuery mode certainly wrongly thinks it's a URL it should retrieve from the ZIM.
If the URI is untouched, the browser/OS will try to handle the link appropriately.

I don't know it there is a more generic way to recognize this kind of link than looking for "tel:" or "geo:" prefix.
Ideally, we would have a generic way to recognize the URI scheme, and ignore all but the links in the ZIM content.
For example, the "mailto", "ftp", "irc" schemes should be ignored too.

If necessary, we can use the complete url (given by "this.href") instead of the raw href string (given by $(this).attr("href"))

Jaifroid:
The basic and easy way to support geo: and tel: would be like in commit https://github.com/kiwix/kiwix-js/commit/a16be0e1bc93ac33813d002184045eb6a2bbc13c .

Adding the href target means that the user doesn't navigate away from the Kiwix app (which is not a great thing to happen when reloading it involves picking the ZIM file again).

A generic approach I guess would involve recognizing *only* links that should be opened from the ZIM and leaving all others alone. Is there enough to identify a ZIM link other than by elimination of those that are obviously not in the ZIM?

PR #319, opened 2017-12-20
Fix anchor links, that were broken because of the base tag

### LINKED ISSUE #318: Links in article to footnotes do not function correctly
When I click / tap on a footnote reference marker, I get the message shown in the screenshot below (this is Kiwix JS in the FFOS Simulator with a recent Wikipedia-style ZIM). On Kiwix JS Windows nothing at all happens. Could this error have been introduced with the code to support subdirectories in ZIMs, or has it always existed and we've just not noticed it till now?

image

mossroy:
Regression introduced by #310
Fixes #318

### DIFF
--- a/www/js/app.js
+++ b/www/js/app.js
@@ -782,10 +782,12 @@ define(['jquery', 'zimArchiveLoader', 'util', 'uiUtil', 'cookies','abstractFiles
     };
     
     // Compile some regular expressions needed to modify links
-    var regexpImageLink = /^.?\/?[^:]+:(.*)/;
+    // Pattern to find the path in a url
     var regexpPath = /^(.*\/)[^\/]+$/;
-    // Pattern for ZIM file namespace - see http://www.openzim.org/wiki/ZIM_file_format#Namespaces
+    // Pattern to find a ZIM URL (with its namespace) - see http://www.openzim.org/wiki/ZIM_file_format#Namespaces
     var regexpZIMUrlWithNamespace = /(?:^|\/)([-ABIJMUVWX]\/.+)/;
+    // Pattern to match a local anchor in a href
+    var regexpLocalAnchorHref = /^#/;
     // These regular expressions match both relative and absolute URLs
     // Since late 2014, all ZIM files should use relative URLs
     var regexpImageUrl = /^(?:\.\.\/|\/)+(I\/.*)$/;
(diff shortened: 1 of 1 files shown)


www/js/app.js
```diff
@@ -837,8 +837,14 @@ define(['jquery', 'zimArchiveLoader', 'util', 'uiUtil', 'cookies','abstractFiles
                         return false;
                     });
                 }
-                else if (url.slice(0, 1) === "#") {
-                    // It's an anchor link : do nothing
+                else if (util.startsWith($(this).attr("href"), "#")) {
```
Jaifroid: Hi @mossroy , I think this could be written much more simply as:
`else if (this.hash) {`
It matches if the URL is an anchor, and doesn't need jQuery. And doesn't need a utility written for it.
Yours works fine, but this could be easier?

www/js/app.js
```diff
@@ -837,8 +837,14 @@ define(['jquery', 'zimArchiveLoader', 'util', 'uiUtil', 'cookies','abstractFiles
...
+                    // It's an anchor link : we need to make it work with javascript
+                    // because of the base tag
+                    var anchor = $(this).attr("href");
+                    $(this).on('click', function(e) {
+                        $('#articleContent').first()[0].contentWindow.location.hash = anchor;
+                        return false;
+                    });
```
Jaifroid: @mossroy So, I've tested a slight simplification of your approach, and I think substituting this for the above else if clause would do the same job, without requiring the utility function:
```
                else if (this.hash && ~this.href.indexOf("/#")) {
                    // It's an anchor link : we need to make it work with javascript
                    // because of the base tag
                    var anchor = this.hash;
                    $(this).on('click', function (e) {
                        $('#articleContent').first()[0].contentWindow.location.hash = anchor;
                        return false;
                    });
                }
```
What do you think? If you prefer your approach for reasons of clarity, that's completely fine by me.

www/js/app.js
mossroy: The problem is that we want to match only local anchors, but don't want to match anchors that would be in another URL.
I suppose (this.hash) is true for any anchor

www/js/app.js
mossroy: It's not clear to me of what you're trying to test with `~this.href.indexOf("/#")`
I'm not used to this tilde operator, but I understood that, in this context, it is true if "/#" is found in this.href, and false if it is not found.
Why are you looking for "/#"?

Jaifroid: The ``~this.href.indexOf("/#")`` ensures that the anchor is a local one, anticipating your previous comment that ``this.hash`` on its own could match an anchor on an external site.

A perhaps less elegant alternative could be:

``else if (/\/#[^/]+$/.test(this.href)) {`` 

(no need for ``this.hash`` in this case).

Jaifroid:
@mossroy I've discovered another case we have to deal with: links back to the same page. As an example, load the latest French "Wiktionnaire" ZIM and load up the article on "gratuit". There you will see a table of agreement (see screenshot below), and the word "gratuit" linked back to self. Looking at the href it is simply "", i.e. blank. However, clicking on it throws an error message that there is no such page "A/" in the archive.

To exempt such links we would need to amend the above test like so:

`                else if (!url || (this.hash && ~this.href.indexOf("/#"))) {`

The `!url` test catches an empty URL in this rule, and sets the link to an empty anchor, so when clicked, nothing happens (which is what we want, rather than an error message).

image

www/js/app.js
mossroy: I don't understand why the string "/#" would indicate a local anchor. If I there is in the article a hyperlink to http://www.geeksphone.com/#our-passion for example (this is a real URL), it will match even if it is not a local anchor.

mossroy:
You're right : we need to handle this case of an empty href more properly

www/js/app.js
Jaifroid: Ah, OK, I hadn't thought of such URLs, as none of the pages I tested had that form -- silly me for trying to get around using jQuery...

So, if all you want to do is test if the URL begins with a hash, then ` else if (/^#/.test($(this).attr("href")))` will do the job. Or store it in a variable earlier and test for both "" and # in the same condition.

mossroy: I agree with you, and will push an updated fix during the weekend.
I'll probably refactor a bit this part of code, which should hopefully fix #313 too

mossroy:
I've updated the PR : could you please test and review it?
It was the occasion to refactor and remove some code that was obsolete (for the Evopedia content)

www/js/app.js
```diff
@@ -819,47 +821,45 @@ define(['jquery', 'zimArchiveLoader', 'util', 'uiUtil', 'cookies','abstractFiles
...
-                else if (url.substring(0, 4) === "http") {
-                    // It's an external link : open in a new tab
-                    $(this).attr("target", "_blank");
+                else if (regexpLocalAnchorHref.test(href)) {
+                    // It's an anchor link : we need to make it work with javascript
+                    // because of the base tag
+                    var anchor = href;
+                    $(this).on('click', function(e) {
+                        $('#articleContent').first()[0].contentWindow.location.hash = anchor;
+                        return false;
```
Jaifroid: In the above, is it necessary to define anchor, when its value is equal to href, so you could just use href? I tested commenting out `var anchor = href;` and just using `= href;` instead of `= anchor;` in the function, and it works fine (I tested it just in case there was a conflicting assignment of href at the time that the .onclick function runs). However I can see that it might be desirable to keep this for reasons of clarity in understanding the code, and if so that's fine.

www/js/app.js
```diff
@@ -819,47 +821,45 @@ define(['jquery', 'zimArchiveLoader', 'util', 'uiUtil', 'cookies','abstractFiles
...
-                else if (url.match(regexpImageLink)
-                    && (util.endsWith(lowerCaseUrl, ".png")
-                        || util.endsWith(lowerCaseUrl, ".svg")
-                        || util.endsWith(lowerCaseUrl, ".jpg")
-                        || util.endsWith(lowerCaseUrl, ".jpeg"))) {
-                    // It's a link to a file of Wikipedia : change the URL to the online version and open in a new tab
-                    var onlineWikipediaUrl = url.replace(regexpImageLink, "https://" + selectedArchive._language + ".wikipedia.org/wiki/File:$1");
-                    $(this).attr("href", onlineWikipediaUrl);
+                else if (this.protocol !== currentProtocol
+                    || this.host !== currentHost) {
```
Jaifroid: Nice!

www/js/app.js
```diff
@@ -819,47 +821,45 @@ define(['jquery', 'zimArchiveLoader', 'util', 'uiUtil', 'cookies','abstractFiles
...
+            var currentHost = location.host;
 
             // Convert links into javascript calls
             $('#articleContent').contents().find('body').find('a').each(function() {
+                var href = $(this).attr("href");
                 // Compute current link's url (with its namespace), if applicable
-                var url = regexpZIMUrlWithNamespace.test(this.href) ? this.href.match(regexpZIMUrlWithNamespace)[1] : $(this).attr("href");
-                if (url === null || url === undefined) {
-                    return;
+                var zimUrl = regexpZIMUrlWithNamespace.test(this.href) ? this.href.match(regexpZIMUrlWithNamespace)[1] : "";
```
Jaifroid: Good that you renamed this to `zimUrl`, it makes the subsequent code much clearer.

Jaifroid:
Sorry, I meant "ready to merge".

www/js/app.js
mossroy: I agree, this variable is not necessary

mossroy:
I've squashed and rebased the branch (after removing the unnecessary variable)
Please merge the PR if it's OK for you.

Jaifroid:
Great, many thanks for this, @mossroy !

Issue #320, opened 2017-12-20
Add automated UI tests on older browsers


mossroy:
After #316, I think it might be worth adding another (older) browser on which we run the same UI tests as on latest Firefox.

The goal would be to check we are still compatible with some older but officially supported browsers (the ones listed in https://github.com/kiwix/kiwix-js/blob/master/README.md).

Testing on Firefox 45 would probably be the easiest, but it would be interesting to test on another engine

mossroy:
Fixed by #322

PR #322, opened 2017-12-23
Use several browsers for automated unit and UI tests

mossroy:
Instead of only one version of Firefox, this PR makes the automated unit and UI tests run on : 
- Firefox 45
- latest version of Firefox
- latest version of Chrome
- latest version of Edge
- Internet Explorer 11 (skipping the unit tests because they don't work, but the UI tests do work)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,10 @@ before_script:
     - npm install nightwatch http-server
     - "./node_modules/http-server/bin/http-server . &"
 script:
-    - "./node_modules/nightwatch/bin/nightwatch -c nightwatch.js --env firefox"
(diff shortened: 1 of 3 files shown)


mossroy:
This fixes #320 and #316

mossroy:
I also added Chrome 58 (in addition to the latest version of Chrome).
But, for now, some UI tests seem to be unstable on some browsers : at least Edge and IE11. They work most of the time, but sometimes fail on a timeout waiting for element <#formArticleSearch> to be visible, just like in #251

mossroy:
I think the UI tests are stable now.
I changed the nightwatch script to wait for the file input to be displayed before injecting the zim files. And I also limited the number of concurrent tests by separating recent and old browsers.

PR #330, opened 2018-01-05
Fix a regression on some images in ServiceWorker mode.

### LINKED ISSUE #329: Some images are not displayed in ServiceWorker mode
It's a regression due to #272  
I see how to fix it

mossroy:
Fixes #329
It was affecting all  tags of the first displayed article, only
in ServiceWorker mode.


mossroy:
Replacing  by  should only be done in jQuery mode. And it was also done in ServiceWorker mode.

PR #331, opened 2018-01-05
Fix the jQuery version and the visibility of last lines in About section

mossroy:


www/index.html
```diff
@@ -178,6 +178,11 @@ <h3>License information</h3>
                     </ul>
                     <h3>Other platforms/versions</h3>
                     Other Kiwix clients exist on various platforms. See the official site : <a href="http://www.kiwix.org/" target="_blank">http://www.kiwix.org/</a>
+                    <!-- We need to add a few lines else the bottom bar hides the last actual lines of text -->
+                    <br/>
+                    <br/>
+                    <br/>
+                    <br/>
```
Jaifroid: Ah yes, I did the same or similar in Kiwix JS Windows!

mossroy:
I'm not proud of the workaround to see the last lines of the About section. In the long term, we should find a better way.
This bottom bar also caused issues in the automated UI tests

Issue #335, opened 2018-01-15
Cache the CSS contents (in jQuery mode)


mossroy:
Following the discussions in #130, this is about putting the CSS contents of a ZIM file in a cache.
@Jaifroid suggests to port/adapt the work of @sharun-s in kiwix-js, as it seems to be a very efficient optimization.

Jaifroid:
Thanks, @Mossroy. The suggested code is in https://github.com/kiwix/kiwix-js/tree/Backport-CSSCache .

In testing, it would be a good idea to use one of the "modern" ZIMs, I suggest `wikipedia_en_medicine_novid_2018-01.zim`. The reason for using this or similar to test is that the new mobile-oriented ZIMs require *a lot* of stylesheets. Other than the front page, many pages require between four and seven stylesheets to be decompressed. Some of them are dummy stylesheets containing nothing, but they're still costly to decompress. Hence you see a dramatic speed improvement in Kiwix JS once the app has had a chance to cache most or all of the required CSS. Watch the Console to see what it's doing.

mossroy:
I had a look at the code and tested on PC with Ray Charles archive (I could not test on my FirefoxOS device, and with the ZIM file you suggests, but will do later).
The place to remove the cache content is certainly right for the FirefoxOS DeviceStorage way to read ZIM files.

I did not measure the performance, but it gives a real visual improvement : as the CSS is instantly loaded, the content "moves" much less after displaying the text.
This optimization is (unfortunately) specific to the jQuery mode. But it does not seem to add a lot of complexity.

As there should not be a lot of CSS content in a ZIM file, I think it's not really necessary to put a size limit for this cache.

Regarding the code itself, some variable/function naming look questionable, like newCC. But we can discuss that in the PR.

I'll try to test more later.

Jaifroid:
OK, thanks @mossroy. Just to be clear, I *haven't* put in any code to clear the cache anywhere except in `setLocalArchiveFromFileList`. And that is definitely not hit on changing archives using the FFOS Device Storage menu. So I leave it to you to insert that code at the best spot for FFOS users. I don't know if there's a "universal" spot we could find to insert the cache-clearing code only once?

Feel free also to change any of the variable names to anything more consistent. Then I can do a PR.

mossroy:
I've refactored the code a bit and pushed on your branch. It's open to comments.
I've added the code to clear the cache on FFOS (you were right, I spoke too soon) : it's working fine on my devices.
[...]

PR #336, opened 2018-01-16
Backport css cache

Jaifroid:
This is the PR of the code already discussed. We should do a bit more testing on a variety of ZIMs, and when we're satisfied no further changes are required, I'll do a squash.

### DIFF
--- a/www/js/app.js
+++ b/www/js/app.js
@@ -529,12 +529,23 @@ define(['jquery', 'zimArchiveLoader', 'util', 'uiUtil', 'cookies','abstractFiles
                         + " storages found with getDeviceStorages instead of 1");
                 }
             }
+            resetCssCache();
             selectedArchive = zimArchiveLoader.loadArchiveFromDeviceStorage(selectedStorage, archiveDirectory, function (archive) {
                 cookies.setItem("lastSelectedArchive", archiveDirectory, Infinity);
                 // The archive is set : go back to home page to start searching
                 $("#btnHome").click();
             });
             
+        }
+    }
+    
+    /**
+     * Resets the CSS Cache (used only in jQuery mode)
+     */
+    function resetCssCache() {
+        // Reset the cssCache. Must be done when archive changes.
+        if (cssCache) {
+            cssCache = new Map();
         }
     }
 
@@ -547,6 +558,7 @@ define(['jquery', 'zimArchiveLoader', 'util', 'uiUtil', 'cookies','abstractFiles
     }
 
     function setLocalArchiveFromFileList(files) {
+        resetCssCache();
         selectedArchive = zimArchiveLoader.loadArchiveFromFiles(files, function (archive) {
             // The archive is set : go back to home page to start searching
             $("#btnHome").click();
@@ -792,6 +804,10 @@ define(['jquery', 'zimArchiveLoader', 'util', 'uiUtil', 'cookies','abstractFiles
     // Since late 2014, all ZIM files should use relative URLs
     var regexpImageUrl = /^(?:\.\.\/|\/)+(I\/.*)$/;
     var regexpMetadataUrl = /^(?:\.\.\/|\/)+(-\/.*)$/;
+    
+    // Cache for CSS styles contained in ZIM.
+    // It significantly speeds up subsequent page display. See kiwix-js issue #335
+    var cssCache = new Map();
 
     /**
      * Display the the given HTML article in the web page,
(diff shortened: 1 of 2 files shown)


Jaifroid:
Add a reminder to self to remove the console logging from this code before squashing.

www/js/lib/uiUtil.js
```diff
@@ -42,6 +42,36 @@ define([], function() {
         });
         jQueryNode.attr(nodeAttribute, url);
     }
+
+    /**
+     * Replace the given CSS link (from the DOM) with an inline CSS of the given content
+     * 
+     * For some reason, Firefox OS does not accept the syntax <link rel="stylesheet" href="data:text/css,...">
```
Jaifroid: @mossroy I think the reason is CSP (Content Security Policy). See  https://developer.mozilla.org/en-US/docs/Archive/B2G_OS/Firefox_OS_apps/Building_apps_for_Firefox_OS/CSP . Should we change this comment to:
"Due to CSP, Firefox OS...."?

mossroy: You're right : adding a link to this page gives more explanation on the reason.
It's not because the browser engine is old, but because of some CSP restrictions that have been put intentionally on it.
And some other platforms (like browser extensions) can apply some CSP restrictions : https://developer.mozilla.org/en-US/Add-ons/WebExtensions/Content_Security_Policy

mossroy: All right, no problem

www/js/lib/uiUtil.js
```diff
@@ -42,6 +42,36 @@ define([], function() {
         });
         jQueryNode.attr(nodeAttribute, url);
     }
+
+    /**
+     * Replace the given CSS link (from the DOM) with an inline CSS of the given content
+     * 
+     * Due to CSP, Firefox OS does not accept <link> syntax with href="data:text/css..." or href="blob:..."
```
Jaifroid: @mossroy I updated comment in line with discussion.

Jaifroid:
I've been testing this PR with the philosophy Stackexchange ZIM:

image

Good to see that subdirectories seem to be working in the cache. Not sure what is causing the cross domain error, though -- clearly something in a CSS file. I assume it makes no difference, but thought I should note it here. Page looks fine.

mossroy:
I could not reproduce the cross-origin issue with the philosophy stackexchange ZIM : could give more details on how you had this error?

Jaifroid:
> I could not reproduce the cross-origin issue with the philosophy stackexchange ZIM

That was in Edge, but I also just reproduced it in Firefox:
1. Open index.html (of this branch) in Firefox ESR;
2. Enable developer mode / console;
3. Pick Philosophy Stackexchange ZIM (I used `philosophy.stackexchange.com_en_all_2017-10.zim`);
4. This appears in my Console:
```
downloadable font: download failed (font-family: "Glyphicons Halflings"
style:normal weight:normal stretch:normal src index:1): bad URI or 
cross-site access not allowed source: file:///C:/Users/XXXXX/Source/
I/static/bootstrap/fonts/glyphicons-halflings-regular.woff2
```
However, I'm not worried by this -- I think the CSS has an import font rule that is not allowed. I don't think it has anything to do with our branch. More of a curiosity than anything else.

mossroy:
OK I reproduced the CSS error.
It's typically the kind of problems that will be very hard to solve in jQuery mode. In this case, it's the content of a CSS (bootstrap, here), that references another file (a font, here). In jQuery mode, we parse the DOM and replace some hyperlinks with their content. But we don't recursively parse the CSS contents we inject; and we would need a way to inject content in CSS (I don't know if it's possible).
It's a real limitation of the jQuery mode : we can not exhaustively find the links and replace them by the corresponding content.

mossroy:
@Jaifroid : the CSS console error is not related to this PR. I agree with you that it should be discussed in a new issue.

www/js/app.js
```diff
@@ -792,6 +804,9 @@ define(['jquery', 'zimArchiveLoader', 'util', 'uiUtil', 'cookies','abstractFiles
     // Since late 2014, all ZIM files should use relative URLs
     var regexpImageUrl = /^(?:\.\.\/|\/)+(I\/.*)$/;
     var regexpMetadataUrl = /^(?:\.\.\/|\/)+(-\/.*)$/;
+    
+    // Cache of the CSS as blobs 
+    var cssBlobCache = new Map();
```
Jaifroid: Just one more question before I squash/merge, @Mossroy:
This is *not* technically a blob cache any more. It started off as holding blob URLs, but due to CSP, it's become just a CSS cache (the cache just stores the CSS filename and the raw CSS extracted from the ZIM). Would you prefer the function names and variable names to be changed to ...CssCache... instead of ...CssBlobCache... ? It's easy to do in my editor.

mossroy: Yes. Please do.

www/js/app.js
```diff
@@ -792,6 +804,9 @@ define(['jquery', 'zimArchiveLoader', 'util', 'uiUtil', 'cookies','abstractFiles
     // Since late 2014, all ZIM files should use relative URLs
     var regexpImageUrl = /^(?:\.\.\/|\/)+(I\/.*)$/;
     var regexpMetadataUrl = /^(?:\.\.\/|\/)+(-\/.*)$/;
+    
+    // Cache for CSS styles contained in ZIM [kiwix-js #335]
```
Jaifroid: @mossroy Are you in favour of this kind of reference in comments to a discussion in the repository for future devs? Alternatively, I could put:

`// Cache for CSS styles contained in ZIM (significantly speeds up subsequent page display)`

mossroy: I don't have a strong opinion on that.
It's probably a good idea, as it might help the reader to have more information. But I think it should not be put *instead* of a relevant comment.
In this case, it might be a mix of both like : 

```javascript
// Cache for CSS styles contained in ZIM.
// It significantly speeds up subsequent page display. See kiwix-js issue #335
```

By the way, is the bracket syntax `[kiwix-js #335]` a way to make it automatically recognized by some tools?

Jaifroid: The bracket syntax was just a form of abbreviation, not a tool thing.
Will use your suggestion.

Issue #341, opened 2018-02-19
Void the iframe between page loads


Jaifroid:
After much experimentation with MathJax script loading on kiwix-js-windows (https://github.com/kiwix/kiwix-js-windows/issues/27), I've come to the conclusion that it's good practice to void the iframe between page loads. This can be done with a single line of code added to `function readArticle(dirEntry)`:

`document.getElementById("articleContent").src = "dummyArticle.html";`

The advantages of this are:

- It clears out the iframe DOM, preventing any memory leaks
- It clears any scripts loaded in the iframe (including MathJax) and any variables attached, clearing the memory they use
- It doesn't appear to delay page load at all

I was getting a very strange bug with MathJax whereby all equations on some mathoverflow pages would randomly get shifted down by one line, making the page illegible. But I noticed this never happened if I exited the app and went straight back to that page. The problem turned out to be the persistence of the MathJax scripts and configuration between page loads and even between different ZIM loads. As MathJax loads very fast and asynchronously anyway (much faster than decompressing a single SVG), the cleanest solution was to void the iframe. This solved the issue.

The main reasons to consider this for kiwix-js are a) it's good practice not to have bits of the DOM of an unrelated document hanging around (even between ZIM loads); and b) it will be necessary if we ever reinstate JavaScript loading or decide to backport MathJax from kiwix-js-windows.

mossroy:
It's probably a good idea.
I would advise to try to use "about:blank" for the iframe src attribute, instead of the dummyArticle (which I'd like to get rid of). And it's probably a bit faster.

Jaifroid:
Thanks @mossroy , I can confirm that about:blank works perfectly as an alternative. I've incorporated this now into kiwix-js-windows 0.9.9.1 beta. Voiding the iframe also helps with some CSS issues I was noticing when switching from Wikipedia ZIMs to Stackexchange-family ZIMs (very small fonts -- with this voiding code, the fonts show at the correct size).

Jaifroid:
Was about to post the note below when I've discovered links aren't working in Firefox! So I'll try with dummyArticle again instead....

I've made a PR and have tested quickly in Edge, but not yet in FFOS (works fine in Edge).
[...]

Issue #312, opened 2017-12-17
ServiceWorker mode fails to inject CSS/images when there are subdirectories in article paths


mossroy:
For example on stackexchange ZIM files.

It was already the case before #310

mossroy:
I think this comes from ZIM redirects, when the original and target URLs are not in the same directory.

For example, when you have a link to A/question/123456/description.html , and this URL redirects (in the openZIM way) to A/question/123456.html , the ServiceWorker currently directly sends the final content.
This is an issue because the browser believes it is in A/question/123456/ directory, even if the content of A/question/123456.html has links with URLs relative to A/question/ directory.
As a consequence, the browser resolves the links in a wrong directory : that's why they don't work.

I'm working on a fix for that : when there is a redirect in a ZIM file (in ServiceWorker mode), we need to send an HTTP redirect to the browser, so that it knows that the URL has changed, asks for it by itself, and know in which directory it is.

PR #342, opened 2018-02-20
Void the iframe in readArticle function #341

Jaifroid:
24th Feb: The latest commit ([Inject baseUrl](https://github.com/kiwix/kiwix-js/pull/342/commits/3b5f7b28ae323e1c4f2e0002433c89846279a770)) incorporates #343 (Add the <base> tag before the document is injected into the DOM). It has been tested on Microsoft Edge, Internet Explorer, Firefox and FFOS simulator.

### DIFF
--- a/www/dummyArticle.html
+++ b/www/dummyArticle.html
@@ -7,7 +7,6 @@ and open the template in the editor.
 <html>
     <head>
         <meta charset="utf-8">
-        <base href="A/">
         <title>dummy Article</title>
     </head>
     <body>
--- a/www/index.html
+++ b/www/index.html
@@ -257,7 +257,7 @@
                 <div id="readingArticle" style="display: none;" class="container">
                     Reading article <span id="articleName"></span> from archive... Please wait <img src="img/spinner.gif" alt="Please wait..." />
                 </div>
-                <iframe id="articleContent" src="dummyArticle.html" class="articleIFrame">&nbsp;</iframe>
+                <iframe id="articleContent" class="articleIFrame"></iframe>
             </article>
             <footer>
                 <div id="navigationButtons" class="btn-toolbar">
--- a/www/js/app.js
+++ b/www/js/app.js
@@ -748,6 +748,8 @@ define(['jquery', 'zimArchiveLoader', 'util', 'uiUtil', 'cookies','abstractFiles
             selectedArchive.resolveRedirect(dirEntry, readArticle);
         }
         else {
+            //Void the iframe
+            document.getElementById("articleContent").src = "dummyArticle.html";
             selectedArchive.readArticle(dirEntry, displayArticleInForm);
         }
     }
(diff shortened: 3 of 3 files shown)


mossroy:
@Jaifroid : your tests are very interesting. I'll have a look as soon as possible

mossroy:
Mmmh, this is not simple.
As it is, this PR seems to break the ServiceWorker mode on stackexchange ZIMs.

I think these tests reveal some weaknesses in our existing code.
I did not realize that we were injecting the html content in the "body" part of the frame. It's not right to do so, as the head tags end up in the body. Using the innerHTML as you did is probably a better approach (even if we should try to select the frame by its name/id instead of its position).
Unfortunately, I think we were using some side-effect of this badly-injected content when there are subdirectories in the URL (like the stackexchange ones).

I'm not very satisfied by the way I handled the base tag for ServiceWorkers. I had to put it in the dummyArticle.html (instead of removing this file).

I have no simple solution in mind, but we might find better solutions to these underlying problems. I hope I'll have some time to work on that in the middle of March

Jaifroid:
I agree -- I suspected these changes would have an impact on ServiceWorker. It would indeed be good to arrive at a universal solution.

Remember that ServiceWorker is already broken with #310, and we have an issue for that #312 . I'm pretty sure these things are related to the fact that the baseUrl tag is injected too late, after the browser has already got to work launching http: or file: requests, so probably ServiceWorker never saw the baseUrl, or saw it only on click.

If code has been introduced in ServiceWorker to work around that issue, it would then break once we inject the baseUrl along with the document into the DOM. But this is the correct way to do it, and I've shown that it is perfectly possibly to do away with dummyArticle too: we then need to work around the link issue in Firefox, which is one we ourselves introduced by checking for the document protocol.

So I agree that this exposes issues which should make us redo several aspects of the code, but it will put us on a firmer footing.

Regarding selecting the iframe, if you prefer, I can easily use the iframe's ID. It was just a bit more concise this way. Will update.

No rush of course. It has to be right. I don't know the ServiceWorker code, I'm afraid...

mossroy:
I finally could take some time to work on that.

I was wrong when I said that this PR causes a regression in ServiceWorker mode. The issue I was seeing was #312 (on which I made some progress, by the way).

After some more testing, I come to the same conclusions that you made : 

- Injecting the HTML string in the body tag is wrong. The head tag must come from the ZIM file too. And I did not find any better way than the one you used : 
`document.getElementById("articleContent").contentDocument.documentElement.innerHTML = htmlArticle;`
- Injecting the base tag in javascript in ServiceWorker mode (as it was already done in jQuery mode) is certainly a good idea. It probably allows us to remove the base tag from the dummyArticle
- I trust your tests when you say we need to put the base tag before putting the HTML in the DOM (#343). It makes sense. I dislike modifying the HTML with a regexp here, but do not see a better way for the moment. We might have an issue if there is already a base tag inside the html content (from the ZIM file), because your regexp does not remove an existing base tag before adding it. But it's probably better to let it this way for now : it might be unsafe to remove a tag from the header with a regexp, and I don't see a good reason for a base tag inside a ZIM file.

Regarding the use of dummyArticle vs about:blank, I'll test a bit more. Then I'll add a few comments in the review of this PR

Jaifroid:
OK, great, I'll make that change (which I had already made in kiwix-js-windows) and will squash, etc.

PR #345, opened 2018-03-12
Send an HTTP Redirect in SW mode when there is a ZIM redirect.

### LINKED ISSUE #312: ServiceWorker mode fails to inject CSS/images when there are subdirectories in article paths

mossroy:
It is necessary to let the browser know in which directory it is.
Else the relative links can be broken if the original URL is in a
different directory than the final URL.

Fixes #312


Jaifroid:
I can't see any problems with this code, @mossroy , but with the big caveat that I've never delved into the serviceworker code, nor can I test it with my current setup. The changes look simple enough. I'm sure you can go ahead and merge!

PR #346, opened 2018-03-12
Accept links to articles that have an anchor.

### LINKED ISSUE #344: Links with anchors do not work (at least in StackExchange ZIM files)
For example, in askubuntu.com_en_all_2017-06.zim, go to article A/question/58172/How-to-revert-to-GNOME-Classic-Desktop.html (by searching for "How to revert to GNOME Classic Desktop" for example)
All the first links in this page have an anchor in the URL, like A/answer/15811.html#a15811

These links fail (both in jQuery and ServiceWorker modes), because we try to look for an article "15811.html#a15811", instead of "15811.html"

mossroy:
The anchor has to be removed before passing to the ZIM backend

Fixes #344

### DIFF
--- a/www/js/lib/uiUtil.js
+++ b/www/js/lib/uiUtil.js
@@ -73,14 +73,15 @@ define([], function() {
         link.replaceWith(cssElement);
     }
         
-    var regexpRemoveUrlParameters = new RegExp(/([^\?]+)\?.*$/);
+    var regexpRemoveUrlParameters = new RegExp(/([^?#]+)[?#].*$/);
     
+    /**
+     * Removes parameters and anchors from a URL
+     * @param {type} url
+     * @returns {String} same URL without its parameters and anchors
+     */
     function removeUrlParameters(url) {
-        if (regexpRemoveUrlParameters.test(url)) {
-            return regexpRemoveUrlParameters.exec(url)[1];
-        } else {
-            return url;
-        }
+        return url.replace(regexpRemoveUrlParameters, "$1");
     }
 
     /**
--- a/service-worker.js
+++ b/service-worker.js
@@ -36,7 +36,7 @@ self.addEventListener('activate', function(event) {
     console.log("ServiceWorker activated");
 });
 
-var regexpRemoveUrlParameters = new RegExp(/([^\?]+)\?.*$/);
+var regexpRemoveUrlParameters = new RegExp(/([^?#]+)[?#].*$/);
 
 // This function is duplicated from uiUtil.js
 // because using requirejs would force to add the 'fetch' event listener
@@ -44,12 +44,14 @@ var regexpRemoveUrlParameters = new RegExp(/([^\?]+)\?.*$/);
 // in recent versions of the browsers.
 // Cf https://bugzilla.mozilla.org/show_bug.cgi?id=1181127
 // TODO : find a way to avoid this duplication
+
+/**
+ * Removes parameters and anchors from a URL
+ * @param {type} url
+ * @returns {String} same URL without its parameters and anchors
+ */
 function removeUrlParameters(url) {
-    if (regexpRemoveUrlParameters.test(url)) {
-        return regexpRemoveUrlParameters.exec(url)[1];
-    } else {
-        return url;
-    }
+    return url.replace(regexpRemoveUrlParameters, "$1");
 }
     
 console.log("ServiceWorker startup");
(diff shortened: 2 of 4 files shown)


mossroy:
In ServiceWorker mode, the anchor works naturally (in the sense that the built-in behavior of the browser makes it scroll to the anchor when the target page is displayed).
In jQuery mode, it does not work for now, we would have to implement it ourselves. It's probably possible to make it work, but would involve to pass the anchor parameter in many functions, including in the ZIM backend : I would like to avoid "polluting" the backend with workarounds for the UI

www/js/lib/uiUtil.js
```diff
@@ -74,13 +74,16 @@ define([], function() {
...
+     * @returns {String} same URL without its parameters and anchors
+     */
     function removeUrlParameters(url) {
-        if (regexpRemoveUrlParameters.test(url)) {
-            return regexpRemoveUrlParameters.exec(url)[1];
-        } else {
-            return url;
-        }
+        var urlWithoutParameters = regexpRemoveUrlParameters.test(url) ? url.match(regexpRemoveUrlParameters)[1] : url;
+        return regexpRemoveAnchor.test(urlWithoutParameters) ? urlWithoutParameters.match(regexpRemoveAnchor)[1] : urlWithoutParameters;
```
Jaifroid: @mossroy This all works fine, but it's unnecessarily verbose. You can just do:
```
function removeUrlParameters(url) {
        return url.replace(regexpRemoveUrlParameters, "$1");
}
```
Alternatively, there's no real need to precompile such a simple regex. So you could get rid of the two regexp definitions above this function and here you can just do:
```
function removeUrlParameters(url) {
        return url.replace(/([^?#]+)[?#].*$/, "$1");
}
```
I tested this on Firefox, Edge and FFOS, and it has exactly the same effect as yours. I understand that for readability you may prefer the version with the precompiled regex.

Jaifroid:
@mossroy This is the issue I've found, which is a bit confusing, and may be to do with how Stackexchange links to answers.

On the Philosophy Stackexchange ZIM, in the question "Why does the universe obey scientific laws?" there is a self-anchor, supposedly to an answer on the same page, that looks like this:
```
<p>
    I like <a href="../answer/14355.html#a14355">immortal squish's answer</a>, but I'm going to take it a step further.
</p>
```
On both master and on this branch/PR, that link causes the page to navigate to 14355.html. That's all well and good, but 14355.html is the same page, with the same answer and URL "title=A/answer/14355.html#a14355". Now clicking again on the link to Immortal squish's answer, the page loads again. It seems we haven't (in jQuery mode) recognized this anchor as being on the same page. I'm not sure there's an easy solution to this, as it seems Stackexchange has various redirects to different forms of the URL, which are also redirects in the ZIM, I believe. Anyway, thought I should mention it in case it's at all relevant to this PR, given that you investigated it in relation to Stackexchange.

www/js/lib/uiUtil.js
```diff
@@ -74,13 +74,16 @@ define([], function() {
     }
         
     var regexpRemoveUrlParameters = new RegExp(/([^\?]+)\?.*$/);
+    var regexpRemoveAnchor = new RegExp(/([^#]+)#.*$/);
```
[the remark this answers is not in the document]
mossroy: Great! You're right.

www/js/lib/uiUtil.js
mossroy: I agree. I'll push a commit to use your much better method.
Regarding precompiling the regexp, I did not do a benchmark but I suppose it's worth it as you avoid to recompile it on each function call.

mossroy:
Indeed, in jQuery mode, we currently recognize local anchors only if the src attribute starts with a #. 
In ServiceWorker mode, I thought the problem would not occur (as the browser is able to know by itself if the current URL is the same as the link). But A/answer/14355.html redirects (in the ZIM fashion) to A/question/14355.html, which generates an HTTP redirect, so that the browser is in fact, in the end, on A/question/14355.html page. So the link to A/answer/14355.html is considered as another page, so the page also reloads in ServiceWorker mode. The only difference with jQuery mode is that the page is scrolled to the anchor.

I agree that it's not optimal. On the other hand, it seems to be the same behavior as in the original content : https://philosophy.stackexchange.com/questions/14352/why-does-the-universe-obey-scientific-laws/14355#14355

Except if @dattaz had another opinion on that, I would leave that as it is.

dattaz:
> But A/answer/14355.html redirects (in the ZIM fashion) to A/question/14355.html, which generates an HTTP redirect, so that the browser is in fact, in the end, on A/question/14355.html page.

 A/answer/14355.html redirects (in the ZIM fashion) to A/question/[ID of the question that have answers 14355].html

Anyway if it's works, it's totally ok from me

Issue #348, opened 2018-03-14
Unhandled exception swallowed by Q in Javascript code


Jaifroid:
I was investigating some odd behaviour I couldn't account for in Firefox, and was stepping through the code, when suddenly I ended up in Q catching an exception which I managed to trace back to our unused JavaScript parsing code. I have tested this on master, and it is there too.

The error is:
```
TypeError: Unable to get property 'match' of undefined or null reference
at Anonymous function (file:///C:/....../kiwix-js/www/js/app.js:948:17)
```
And line 948 is the last of the lines below:
```
// Load Javascript content
            $('#articleContent').contents().find('script').each(function() {
                var script = $(this);
                // We try to find its name (from an absolute or relative URL)
                var srcMatch = script.attr("src").match(regexpMetadataUrl);
```
To reproduce, put a break on that line and load up any wiki page that might have a javascript in its head. I was opening the landing page of the latest French Wiktionary. Then step over the line.

The exception gets swallowed, but leaves something in a strange state. Whether this is the actual cause of the iframe weirdness I'm investigating, I'm not sure, but I landed straight in this exception, very oddly, from the onreadystatechange code when I attempted to step over it. In any case, it's not good to have
unhandled exceptions hiding in our code. I suggest we comment out the entire Load JavaScript content block for now. Do you agree @mossroy ?

Jaifroid:
Added a commit in new branch to deal with this: https://github.com/kiwix/kiwix-js/commit/4821952f3413d2b6fe757c5a98a9d8741bfb3f17.
Obviously it would be better to fix, but this code is currently unused, I believe. Hence I have added a note for future development.

mossroy:
It looks like a bug in the case when a script tag has no src attribute (which is possible when you use inline javascript).
In this case, we don't have anything to inject, so I think we might simply add a test before that line, like :  
```
if (script.attr("src)) {
    // We try to find its name (from an absolute or relative URL)
    var srcMatch = script.attr("src").match(regexpMetadataUrl);
    ...
}
```

Jaifroid:
[...]

Issue #286, opened 2017-07-16
Make kiwix-js work from file:/// on Chrome in jQuery mode


Jaifroid:
Excuse my naivety, but I don't use Chrome on my main machine for various reasons (principally the number of "phone home" background services it installs on Windows). However I've installed on a secondary machine for testing Version 59.0.3071.115 (Official Build) (64-bit) for Windows 10. So I'm shocked to find that kiwix-js (main) refuses to show the selected ZIM file article when index.html is run from the file system (i.e., not served). Since Edge, Firefox and IE all run kiwix-js fine from the FS, I wonder if I'm doing something wrong or failing to set some flag. The ZIM file loads, because the search box works, but no page (article) is ever shown, and I get this exception when trying to move between tabs in the app:

![image](https://user-images.githubusercontent.com/4304337/28245383-3c1787d4-69fd-11e7-9105-f739c722fa02.png)

Line 135 of app.js is `$("#articleContent").contents().empty();`. Sorry if this is "obvious", but I'd have also thought it's a major use case to be able to download the files and run in a browser without installing node.js or a web server, etc.... ? NB this is failing against kiwix-js main branch, nothing to do with my Preload stylesheets branch. Standard Windows 10 Pro 64bit. (PS I realize this is caused by https://developer.mozilla.org/en-US/docs/Web/Security/Same-origin_policy , but why only this browser?)

mossroy:
Different browsers have different security policies and implementation.
I don't think it has been a major use-case, at least so far.
But maybe it might be fixed. I noticed that, if I remove the "src" attribute of the iframe, it works well (in jQuery mode).

But this src attribute is necessary for the ServiceWorker mode. Or at least it's the only way we had found to make it work at that moment.

Jaifroid:
Ah, OK. I assumed one obvious use for the app would be to distribute along with a ZIM file so that it could be run on any end-user's computer just by launching index.html, without installing apps or going to stores. As Chrome is a popular browser, it would be a shame for it not to work this way. Maybe we can programmatically change the "src" attribute if the user is running Chrome and has not set/enabled ServiceWorker mode.

sharun-s:
Start chrome with the --allow-file-access-from-files flag. I have two shortcuts on desktop. One I call localChrome with the flag set and the other without. This has been discussed before in multiple places just search for "--allow-file-access-from-files" from issues search.

Jaifroid:
[...]

PR #349, opened 2018-03-20
Make jQuery routines into functions and eliminate unhandled exception #348

Jaifroid:
I've organized all the jQuery routines into functions. I thought it best to leave these functions within the scope of displayArticleInForm, not because it's necessary, but because it's neater, I think. Let me know if you would prefer them to be in global scope. I've tested on FFOS and the functions work fine in this scope.

I've disabled the loadJavaScriptJQuery function to prevent overhead until it is worked on. However, I've fixed the line that was causing an unhandled exception as per #348.

Tested and working in: Firefox ERS, FFOS, Edge and Internet Explorer. Please test in Chrome if you can.


mossroy:
In fact, the issue with Chrome on file:// was already open : #286

Issue #347, opened 2018-03-12
Sometimes the article is briefly displayed then replaced by a blank page


mossroy:
I noticed this issue with the current version of Firefox (58.0.2), not with Chromium. Both in jQuery and ServiceWorker mode.
If I look at the source of the white page, I see that it's our dummyArticle.html.
It's probably a regression of #342, as I cannot reproduce this issue with the commit before the merge of this PR.

mossroy:
It seems to be a race condition between loading the dummyArticle (in order to void the frame) and displaying the actual article.
I managed to fix that with the following code in readArticle function (app.js) : 
            ```
            //Void the iframe
            document.getElementById("articleContent").src = "dummyArticle.html";
            $("#articleContent").off("load");
            $("#articleContent").on("load", function(){
                selectedArchive.readArticle(dirEntry, displayArticleInForm);
            });
```
But it might not be the best way to do it.

mossroy:
I'm surprised that the browser can take so much time to load a very simple HTML file.
It might be one of the recent optimizations of Firefox (since v57), where they give a lower priority for loading some content considered less important (ads, trackers etc).

Instead of waiting for the iframe to be ready to start reading the article, it would be better to start reading it right away (it can take some time). At the end of the backend job, instead of calling displayArticleInForm, we might first check that the iframe is loaded. I did not investigate on how to do that yet.

Jaifroid:
Hmm. When I was initially experimenting, I couldn't put the voiding code in displayArticleInForm, which is where I thought it should naturally go, because it emerged that whatever method I used to load the iframe.src (even just blanking it) was actually happening async. That's why I had to put it in readArticle, since the callback there gives time for the article to load.

Maybe we need to wrap it in a promise or something. Either a simple setTimeOut promise or using q?

Jaifroid:
I haven't seen this issue in Edge or Firefox ESR (52.6), after many hours of testing in Kiwix-JS-Windows, so it must indeed be something in the latest FF.

mossroy:
We must avoid a setTimeout, because it would just hide the race-condition issue, that could still occur on slow devices.

I would suggest to move the code that voids the frame to the moment we really need it. It has the advantage to be more readable, and more user-friendly in case of an unexpected error (for example, if something goes wrong in the backend, the current code leaves the user with an empty page. If we move the code in displayArticleInForm, it would leave the existing content).
[...]

Issue #152, opened 2016-01-08
Enable injection of javascript content coming from ZIM files in jQuery mode


mossroy:
The current Ray Charles ZIM file does not have any testable javascript content.
And it was causing issues in Firefox and Firefox OS, so I disabled the injection of javascript (both in jQuery and ServiceWorker modes)

So we have to find a test ZIM file that has some, and re-enable injection of javascript

julianharty:
The PhET Simulations ZIM file is both small and relies on Javascript to render and use the exercises. It might be a useful testbed to test the support of JavaScript in kiwix-html5...

Jaifroid:
@mossroy Were the JavaScript issues due to the decompression engine hanging, or were they to do with the interaction of JavaScript with the loaded page or app operation? If hanging, then it was probably a similar issue to overloading the engine with SVG decompression or even CSS decompression. As plain-text files are highly compressed, the engine deals poorly with decompressing several things thrown at it at once.

I'm wondering if it will be feasible to re-enable JavaScript injection with a bit of pipeline management.

mossroy:
It would probably be worth testing again : I don't remember exactly what were the symptoms, and they might have disappeared (if we're very lucky), with the improvements of recent browser engines.

In any case, it will be more difficult in jQuery mode, because, unlike images and CSS, the injection timing of javascript code can sometimes be important. For example, some javascript code is sometimes launched on the "load" event of the page, because the javascript code of the "head" section should be ready at this time. If we don't inject the javascript at the right time, these javascript calls might fail.
In ServiceWorker mode, it should work naturally, as the browser engine will handle this by itself.

mossroy:
After a quick test, I managed to enable javascript in ServiceWorker mode, and it does seem to solve #355.
I'll try to test more later, and try to enable it in jQuery mode (which might be harder)

mossroy:
In jQuery mode, I manage to inject the javascript content in the src tag, but it does not seem to be executed by the browser engine. I don't know if there's a way to ask for that.
I tried to use eval() to execute it instead : it works, but seems to be executed in the global context of the document. As a consequence, the jQuery script (coming from the ZIM file) tries to load and sees that it is already loaded (because kiwix-js also uses jQuery). This eval() should be run in the context of the iframe : I don't know if it's possible.

[...]

PR #353, opened 2018-04-06
Remove dummy article

Jaifroid:
@mossroy I have now applied the changes from Kiwix JS Windows that remove dependency on dummyArticle and void the iframe using document.open(), write() and close(). Changes are in the Remove-dummyArticle branch on which this PR is based.

PR is for testing, not yet for merging.

As explained yesterday, to make this code work, it is necessary (since we are not extracting scripts from the ZIM) to neutralize active JavaScript that otherwise gets executed on page load, and causes an unhandled exception.

The culprit here is the extra JavaScript that has been injected into Wikimedia Mobile-style ZIMs to enable the opening and closing of section headers with a click. Unnecessarily IMHO, this JavaScript relies on jQuery (it could easily be rewritten not to require jQuery). So we have to neutralize it unless we want to attach jQuery to every html page pulled from the ZIM. I'm not convinced that these JavaScript routines should be hard-coded in the ZIM -- they should be injected by the client IMHO. Maybe @kelson42  can comment on why this functionality is hard-coded in the mobile-style ZIMs rather than provided by each client.

Anyway, once we neutralize those JavaScript blocks and the open-close toggles, and remove the injected class ('client-js') that has been added to the html tag of every page, we can now use this standard open-write-close method to inject the html into the iframe. I believe it solves a lot of the problems we were seeing by injecting the html using innerHTML (which doesn't execute the JavaScript, but also prevents some other things from rendering properly). Plus we squash #347 by not loading dummyArticle into the iframe.

It is quite likely that this code won't work initially in Service Worker mode, because the code for neutralizing the JavaScript is currently applied only inside a `if (contentInjectionMode === 'jquery')` block. Some of those should, I believe, be applied to both injection modes unless Service Worker mode is correctly pulling scripts from the ZIM. I leave that to you to test.

This code is tested on: IE11, Edge, Firefox ESR, Firefox Quantum and FFOS emulator. I have not tested on Chrome yet.

Please test thoroughly!


kelson42:
@Jaifroid It looks like you are disturbed by the feature which allows to open/close section with a click on Wikipedia based ZIM files (mobile version). This is indeed a feature of the content and I do not see why this should be handle outside the content (by the reader) as this is a specificity of the content/ZIM. We will have more and more ZIM file with sophisticated javascript in them (also based on jquery). If that is a problem with Kiwix-js, then we should really find a way to improve Kiwix-js to avoid that.

Jaifroid:
Sorry, @kelson42 , I'm not disturbed by it, I just have to work around it, because Kiwix JS doesn't currently support extracting and executing JavaScript from the ZIM. The script in question has this comment:

`// this function is to be able to open/close section in mobile version since the code from wiki meant to do that doesn't work`

So I thought it must have been added during scraping, and therefore not in the original HTML. It seemed like a functionality that the client should provide (and could easily provide). Clicking on sections that have the openToggle onclick event currently causes Kiwix JS to crash, but so does loading any page from such a ZIM if we do it in a way that executes JavaScript, because of the use of jQuery in that script. This is clearly a limitation of our client, but we have to work around it for now.

Good to know that the roadmap for the ZIM includes dynamic functionality -- a good reason for getting JavaScript working in Kiwix JS. Personally, I'm not a fan of jQuery, but that's down to personal coding style and history.

Jaifroid:
I've been doing a bit of reading on the merits of inserting HTML using .innerHTML vs document.write(). Main difference is that with the former, as I noted above, inline scripts do not execute, whereas with the latter they do. See https://developer.mozilla.org/en-US/docs/Web/API/Element/innerHTML . Document.write() doesn't work in XHTML documents that are served as XML (rarely the case, and not relevant to Kiwix JS, I think). Note that I found this warning on the above page, with regard to the use of innerHTML in browser extensions, and thought it might be relevant:

image

Jaifroid:
Following above discussion, and since we cannot use the supplied routine, I've implemented functionality for opening and closing section headers. Like the supplied routine, this includes the ability to open the section (if it is closed) that contains the target of any footnote, endnote or other form of note that has been clicked on (the section is derived from the target). These are the reasons why we cannot use the supplied routine in Kiwix JS:

* Due to the Content Security Policy, Firefox OS does not support use of JavaScript in inline HTML, e.g. in the `onclick = toggleOpenSection();` which is hard-coded in the mobile-style ZIM's supplied routine;

* Even if we decide not to support FFOS at some point in the future, we would have to extract or inject jQuery into every page taken from the ZIM, which seems wasteful for a few lines of code;

* The supplied routine is buggy, and does not work on ZIMs that contain reference sections with different ID's than the hard-coded id="References". E.g. several WikiMed articles have reference sections with id="Notes" and even id="Footnotes" (sometimes in addition to id="References"), and potentially others. A more generic routine is needed, and ideally one that doesn't use inline JavaScript but attaches the onclick event using addEventListener.

I've written a generic version that doesn't need jQuery here: https://github.com/kiwix/kiwix-js/pull/353/files#diff-bca463ed8ccf5e2058dfe307a95e8501R860 . No doubt it can be improved, but it's suitable for use in our client and doesn't rely on the inline onclick event in the HTML. It is tested on FFOS, IE11, Edge and Firefox with a range of ZIMs.

Note that a page is started with all headings open, rather than closed, as I believe it could be disorientating to users not used to having all headers closed. This could easily be changed, say, just for ZIMs that have a mobile style. A good ZIM to test this routine on is the latest WikiMed. @Mossroy, I completely understand if you don't want to use such a routine in Kiwix JS, but I'd be grateful if you could try it out and let me know of any problems. I'll probably use it in Kiwix JS Windows, but I wrote it out of interest, and as a challenge.

Jaifroid:
That's a great summary, @Mossroy. I'm very happy to separate out the issues. One led to another as logical consequences of each other, which is why they got pushed into this PR.

I think I'll keep this branch as a repository of ideas and port the issues into different branches.

Yes, we can start with `document.write()`, though I think it very quickly requires neutralization of JavaScript on the page at least until we have implemented #152.

Jaifroid:
Closing. See #341 for continuation.

Issue #355, opened 2018-04-11
On small screens and recent mediawiki ZIM files, the sections can not be "opened"


mossroy:
The CSS detects that the screen is small, and collapses the sections.
Normally, a click on the section name should open it, through the javascript "toggleOpenSection" call.
But it is based on jQuery, which is not loaded in the iframe (because no javascript is injected at all for now. See #152).
As a consequence, the section content can not be read at all.

On Firefox OS, the CSP seems to block the inline javascript call too (but this old platform is not a priority).

The best way to solve this would be to enable javascript, by implementing #152. I think it's the first approach we should try.

mossroy:
The CSP of browser extensions also blocks inline javascript by default, both on Firefox and Chrome.
But the console error of Chrome gives some interesting hints, that should be tested : 

> Refused to execute inline script because it violates the following Content Security Policy directive: "script-src 'self' blob: filesystem: chrome-extension-resource:". Either the 'unsafe-inline' keyword, a hash ('sha256-kuvNbvNowm0sZo4Jsuyt1dmRAsyQrNP0d1OW1SQe1Sk='), or a nonce ('nonce-...') is required to enable inline execution.

Jaifroid:
Does it block the whole script in `<script>...</script>` tags on the page, or just the .onclick events? And does it block `windows.onload` inline as well? If it blocks everything that is not an external script or a DOM attached event, we'll have to take the script block and transform it into a script blob to attach to the DOM I guess.

Alternatively, we could conclude that the header open-close routine is generic enough to our main use-case (Wikimedia) that we support it internally. But I do understand the reasons for not doing so, and for keeping Kiwix JS as generic and universal as possible. And if we're ever to support PhET, we have to have to have generic mechanisms as far as possible.

mossroy:
It confirm it blocks at least the code inside `<script>...</script>` tags, and most probably all the javascript that is embedded inside the HTML (including the ones in onclick attributes).

The "unsafe-inline" CSP directive can be added in the manifest.json, but is ignored by Firefox and Chrome : both browsers consider it's too dangerous and don't allow this directive in extensions. See https://developer.mozilla.org/Add-ons/WebExtensions/manifest.json/content_security_policy and https://developer.chrome.com/extensions/contentSecurityPolicy

The only exception they seem to allow is to put in the manifest.json a hash of each allowed content.
[...]

PR #358, opened 2018-04-23
Open headers and blocks #355

Jaifroid:
See comment in #355.


www/js/app.js
```diff
@@ -932,6 +932,12 @@ define(['jquery', 'zimArchiveLoader', 'util', 'uiUtil', 'cookies','abstractFiles
         }
 
         function loadCSSJQuery() {
+            // Ensure all headers are open for clients that lack JavaScript support [kiwix-js #355]
+            var iframe = document.getElementById('articleContent').contentDocument;
+            var collapsedBlocks = iframe.querySelectorAll('.collapsible-block:not(.open-block)');
+            for (var i = collapsedBlocks.length; i--;) {
```
Jaifroid: @mossroy As there are potentially a large number of blocks in a long document,  I've optimized this loop for speed using a decrement iterator. See https://stackoverflow.com/questions/3520688/javascript-loop-performance-why-is-to-decrement-the-iterator-toward-0-faster-t for explanation of syntax (it's not a mistake!).

Jaifroid:
OK, but the problem with modifying the CSS is that the open-close feature will then stop working even in browsers that support inline javascript.

The `querySelectorAll` method is pretty efficient, and by using only native methods (no slow jQuery) I doubt you'd notice any slowdown on even the longest documents on slow devices. I can't see the risk: we're only selecting elements that have the class  ".collapsible-block". Therefore we are only selecting elements that have been explicitly marked for use with this specific CSS. And all we're doing with my code is adding a CSS class ".open-block" to those (selected) elements that don't already have it. If a future ZIM uses some completely different method for opening-closing blocks, then it's unlikely that ".open-block" will mean anything, so in that case it won't do anything, or else the ".collapsible-block" classes will be removed, so this code won't then touch anything. Note that non-Wikimedia ZIMs that don't have ".collapsible-block" are not touched by this code (the querySelector will simply not match). Try it out with an old Wikipedia ZIM or with Stackexchange ZIMs.

If you want a minimally invasive method, simply remove client-js from the html tag with a one-liner, then everything is open. But again it prevents the open-close code from working in any environment.

mossroy:
Mmmh, you're probably right. I did not check the that it did not prevent the open/close feature to work.

www/js/app.js
mossroy: Interesting!
I've created https://jsperf.com/loop-with-increment-or-decrement-operator to test that on recent browsers. The result really depends on the browser engine.
From what I tested : 
- On Firefox 52.x, the result is the same
- On Firefox OS 2.2 (the simulator, using the Firefox 37 engine), the decrementing loop is 20% faster
- On IE11, the decrementing loop is 30% faster
- But on Chrome 65, the decrementing loop is 20% slower

In any case, I think this is not crucial, as the number of collapsed blocks should not be very high (around 10, I would say).
I suppose that modifying the DOM could be much more expensive than the loop itself.

Feel free to keep the decrementing loop : in this case, please add a comment above with the stackoverflow link.

Regarding the cost of DOM modification, it might be worth optimizing it (in general in the jQuery mode). See section 5 of http://desalasworks.com/article/javascript-performance-techniques/ or section 7 of https://www.sitepoint.com/10-ways-minimize-reflows-improve-performance/ : the browser re-computes the layout in each loop iteration.
Regarding this PR, as the CSS class does not exist yet when we run this loop, maybe it's fast enough. Else jQuery allows to apply this kind of changes in a batch.

I did not run benchmarks on this, and it might not be crucial for this PR. But it might be an idea to optimize our jQuery mode in general

Jaifroid: Thanks for the ideas. 
I checked here: https://jsperf.com/jquery-addclass-vs-element-classlist-add . 
Test is: jqueryObject.addClass("test"); vs domElement.classList.add("test");
On Edge jQuery is 88% slower.
On Firefox 52, jQuery is 80% slower.
On IE 11, jQuery is 87% slower.
I'll have an explore to see whether jQuery delays the screen repaint but as the for loop is extremely fast, I imagine there would be no time for a full repaint before all the nodes are updated. Anyway, I'll check out what method jQuery uses to mass-update a collection of nodes and will get back.
PS The decrement loop works on all the browsers we support, I believe, but I agree with you that it's probably not a performance bottleneck! I just discovered it and thought I'd try it out.

mossroy: What should be compared is a batch-update with jQuery like (untested) : 
`jQuery('.collapsible-block:not(.open-block)').addClass('open-block');`
From what I understood, this code does all the DOM changes at the same time, triggering only one repaint.

Jaifroid: I'm not noticing any issue on even large pages such as the Paris page in English Wikivoyage run in FFOS simulator. There are no multiple repaints that are noticeable to the user.

Doing some research on this, I've come across several times the idea that browsers queue reflows and repaints "so if you do four dom modifications in one sequential piece of Javascript, the browser will wait for that piece of Javascript to finish running and then do one reflow and one repaint". So long as your DOM modifications are done in one synchronous piece of JavaScript, no reflow or repaint is triggered until the script stops running. Since we are using synchronous code here and no callback, the update to all the nodes causes a single repaint when the script stops running (or when it makes an asynchronous call). See https://stackoverflow.com/questions/37039667/executing-multiple-dom-updates-with-javascript-efficiently 

Still, if you would prefer to use jQuery here, it's fine by me. All it does is abstract an underlying for or while loop more slowly than doing it ourselves. Underneath jQuery are just standard JavaScript calls and a bunch of syntax parsers that provide the abstraction.

Jaifroid: Hey @mossroy , I set up a more realistic performance test on real-world data using the two routines we want to compare, here: https://jsperf.com/queryselectorall-classlist-add-vs-jquery/1 

My results are:
Edge 16: jQuery is 16% slower
IE 11: jQuery is 15% slower
Firefox 52: jQuery is 31% slower

Jaifroid: Actually, use this one:
https://jsperf.com/queryselectorall-classlist-add-vs-jquery-addclass/1 
It has more blocks that match so that the loop runs a few times on each iteration.

Jaifroid: Alternative to all of this is that I can do a replacement regex that runs on the raw html before we inject it into the DOM. Would be interesting to compare that to the pure JavaScript test...

Jaifroid:
I've just realized that we need to invert the V in the h2 tag too, so that when it's open it looks like a caret, as in the image below. Fortunately it's a simple amendment to the querySelector to include headers that do not have the .open-block class. I'll push it to the PR now for testing.

image

mossroy:
I think you can squash/rebase/merge this branch. (I'll do it if you don't have the time)
Tomorrow I'll try to release a version 2.3 based on the master branch

Jaifroid:
Please go ahead, @mossroy . And feel free to make any changes to comments or coding style to fit with the app overall.

Issue #366, opened 2018-05-04
History back function does not reload article from ZIM in Firefox (regression)


Jaifroid:
This is a regression from #354. In Firefox ESR and Firefox Quantum, pressing the "Back" button to go back in the browsing history -- which calls the `history.back()` function -- replaces only the iframe's state back to the point at which we last called `document.write()`. In the released version 2.3 of Kiwix JS, Firefox correctly causes the required page to be re-read from the ZIM, resetting the state of the outer document as well. Since we now have an `iframe.onload` event, in the current version of Kiwix JS the `parseAnchors()` and other functions are called without voiding the iframe, which attempts to attach onclick events to all links again, even though onclick events are still somehow attached to the document referencing the previous page. This results in a mess (links to internal pages do not function, etc.).

Curiously this only happens with Firefox, not with Edge. Firefox appears unique in maintaining a history state for the iframe window, whereas Edge and Chromium do not, so they correctly reset the state of the outer document as well on history back. See https://stackoverflow.com/questions/3254985/back-and-forward-buttons-in-an-iframe and in particular this:

> in Firefox around v42, sometimes the browser's own "back" button results in per-frame history.back(), rather than a top-frame history.back(), for no apparent reason (= inconsistent behavior of the native "back" button)

It seems we need to intercept `history.back()` and ensure it is applied to the top frame, if that is possible. If that isn't possible, then we will need to intercept the iframe.onload event, derive the articleContent's title and path from the HTML, as it doesn't seem to be available in the URL (from initial testing) and manually reload the article.

It is not enough merely to void the iframe.onload() event with a blank function (I have tested this): the article still remains in an inconsistent state, because the iframe hasn't been voided correctly. The inconsistent state, whereby event listeners for the previous article are still attached to elements in the DOM of the new article looks like a Firefox bug to me, related to its per-frame history. Or it could be a jQuery bug related to #361 (jQuery keeps hooks on the old article even after the new one is loaded).

Jaifroid:
Possibly a better solution than manipulating window.history is suggested here:

https://stackoverflow.com/questions/19624452/workaround-to-firefox-creating-new-history-after-each-document-write 

It does seem to have an effect on first trials.

Jaifroid:
[...]

Issue #361, opened 2018-04-30
"TypeError: can't access dead object" exceptions within Firefox extension


mossroy:
It looks like a regression since 2.2
When running kiwix-js as a Firefox extension, I randomly get this exception when clicking on a link, leading to a "Error reading article with title xxx" instead of displaying the article.

mossroy:
See https://developer.mozilla.org/docs/Web/JavaScript/Reference/Erreurs/Dead_object

After trying various versions of nightly builds, I see that the issue appeared after merging #354.
I suppose we're somewhere referencing a DOM object from the iframe, that has been destroyed by our  articleContent.write().
I need to find which one and fix this because it looks like a blocker for a release.

Jaifroid:
When you say "randomly" is it not every link? Can you isolate which links are producing the error? It may have something to do with the event listeners? Might we have to revoke the listener in the click function?

Jaifroid:
I mean, put removeEventListener (or jQuery off) as part of the redirection here:
https://github.com/kiwix/kiwix-js/blob/master/www/js/app.js#L905 
(or anywhere that a new document is loaded)? Just a guess.

mossroy:
The bug happens on the Ray Charles archive for example.
Steps to reproduce :
- on Firefox, install the latest nightly extension by clicking on the .xpi file from https://download.kiwix.org/nightly/2018-04-29/
- open the Ray Charles archive
- it displays the main article : click on the first link "A fool for you" (it usually works)
- then click on the first link "Ray Charles"

Jaifroid:
Hmm, I've not been able to reproduce (yet) in Firefox Quantum on Ray Charles, or on the latest WikiMed. But I am getting a lot of CSP errors in console:

![image](https://user-images.githubusercontent.com/4304337/39419030-a0d1c802-4c55-11e8-96c0-e83e046ccf59.png)

I wonder if it's a race condition with the old DOM elements not being fully garbage collected before starting to access the new DOM. Strange that it's random.

mossroy:
The stacktrace of the exception is :
`Sizzle@moz-extension://e3199cf6-77e7-4308-8133-66400fd6f608/www/js/lib/jquery-3.2.1.slim.js:819:6
find@moz-extension://e3199cf6-77e7-4308-8133-66400fd6f608/www/js/lib/jquery-3.2.1.slim.js:2922:4
[...]

PR #368, opened 2018-05-12
Fix for dead object #361 and history regression #366

Jaifroid:
PR of combined changes from #366 (incorporating fix for #361 already in #366). See comment there.

### DIFF
--- a/www/js/app.js
+++ b/www/js/app.js
@@ -865,6 +865,7 @@ define(['jquery', 'zimArchiveLoader', 'util', 'uiUtil', 'cookies','abstractFiles
         // Completely void iframe [kiwix-js #341] and inject new article (NB iframe.onload runs *after* this)
         var articleContentDocument = iframeArticleContent.contentDocument;
         articleContentDocument.open('text/html', 'replace'); // Prevents incorrect history state in Firefox [kiwix-js #366] 
+        articleContentDocument.write("<!DOCTYPE html>"); // Ensures browsers parse iframe in Standards mode
         articleContentDocument.write(htmlArticle);
         articleContentDocument.close();
         
--- a/www/js/app.js
+++ b/www/js/app.js
@@ -711,7 +711,6 @@ define(['jquery', 'zimArchiveLoader', 'util', 'uiUtil', 'cookies','abstractFiles
         $("#prefix").val("");
         findDirEntryFromDirEntryIdAndLaunchArticleRead(dirEntryId);
         var dirEntry = selectedArchive.parseDirEntryId(dirEntryId);
-        pushBrowserHistoryState(dirEntry.namespace + "/" + dirEntry.url);
         return false;
     }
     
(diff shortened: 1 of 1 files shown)


Jaifroid:
Whoops, I see there's a typo in the commit title (objexts instead of objects). I'll correct before merge. More haste, less speed. Also, @mossroy , looks like this PR has failed one of the integration tests...? Did I do something wrong?

Jaifroid:
@mossroy , I've pushed a commit to this PR that ensures the iframe is run in Standards mode, not in Quirks Mode. The problem was that we were not writing a DOCTYPE declaration to the iframe. I'm not sure if this is because the ZIM articles don't include a doctype or because we accidentally exclude it. The added code is simply this:

`articleContentDocument.write("<!DOCTYPE html>"); // Ensures browsers parse iframe in Standards mode` 

And I see that this fixes the integration tests too.

Jaifroid:
Could it be that pesky onload event for when the user changes mode? It's just a wild guess.
Or maybe pushing the doctype before the rest of the document is an issue? I kind of doubt it, but you never know with JavaScript and the DOM...

Jaifroid:
Another quick thought: might the exception produced in the iframe by the lack of jQuery be causing problems? Does the regression only happen with recent Wikimedia ZIMs? On every such article I get the message in screenshot below in Console. Until we can extract the JavaScript (see #152 ) we should probably neutralize this.

image

mossroy:
The regression is certainly not related to the `'$' is not defined` console errors : this appears after the page is loaded (onload event), after the CSS and JS fail to load.
It appears on old ZIM files too (the Ray Charles one, for example). It does not seem to be related to the onload event we add (because the issue is still there if I comment this code).

I tested with the .innerHTML method (commit aa2c65d1785b54e1a3c987adf9e060a14cce0e89 of branch Fix-history-back-regression) : it works fine (the CSS HTTP requests are trapped by the ServiceWorker in all cases).

I suspect that the document.write method does not set a source that is inside the scope of the ServiceWorker (or something similar). I need to check if there is a way to do so.
Else we might switch (again) the method ;-)

Jaifroid:
Hmm, OK I've pushed a branch with the latest version of the innerHTML fix for testing here:

mossroy:
Yes, it would be better to add `<meta charset="utf-8">` in the head section. I would also advise to add a title there, like `placeholder where articles will be injected`.

Regarding the document.write method, I managed to workaround the problem on Chrome, by setting a src to the iframe tag (in index.html), pointing to an empty html file like your article.html. But it only seems to work on Chrome, not on Firefox.

Regarding the innerHTML method, I'd be fine to finally choose it, but there are 2 issues : 
- The way to copy the CSS classes of the HTML tag is not generic. For example, some CSS styles might be applied with the `style` attribute (without referring to a CSS class), and there might be other attributes too. It would be better to find a way to copy all attributes of this tag, in order to be more generic (NB : why are you applying the CSS styles to the body tag instead of the htm tag?)
- It might not be accepted by the extension reviews. As you already pointed, Chrome might accept (see https://wiki.mozilla.org/Add-ons/Reviewers/Guide/Reviewing#Step_2:_Automatic_validation), but Mozilla will probably not (see the warning in the middle of https://developer.mozilla.org/en-US/docs/Web/API/Element/innerHTML). But we might submit it to find out, we don't have much to loose...

I tested this innerHTML method only quickly with Firefox 60ESR and Chrome 66 (both in ServiceWorker mode, with a recent Wikivoyage ZIM), and did not find any issue.

Jaifroid:
Both of those links point to the Mozilla documentation, with the first one giving the detail about the review process, and it seems to imply that the rejection would apply if it's clear that the HTML is from an external unsanitized source, which isn't the case with our code. We might need to explain to the reviewer...

Regarding the method of copying the html tag's attributes, the issue here is that I can't find any way of accessing or writing to the html element of the iframe's contentDocument (or the contentWindow). Neither outerHTML on the documentElement nor attempts to use createElement and insert it seem to work (see discussion of this in #366). If you have a way of doing it (apart from srcdoc) that I've overlooked, it would be great to try it out.

I'm a bit worried about copying anything in the html element and applying it to the body, so all I've done is copy classes. We could end up adding something that shouldn't be in the body element.

Note that to date we have got by with inserting the full html document into the *body* element of the iframe (which is clearly wrong), so whatever we do here is an improvement. But of course it would be great to optimize this as much as possible.

mossroy:
You're right about the links about extensions. In any case, it's worth proposing the extension and see what they say.
OK I understand for the html/body CSS classes.
It would be worth creating a PR for the https://github.com/kiwix/kiwix-js/tree/InnerHTML-dead-objects-and-history-fix branch, with the UTF-8 and title fix in article.html.
I started testing it and did not find any issue so far.

Jaifroid:
PR superseded by #369

Issue #370, opened 2018-05-21
Read CSS as UTF8 strings instead of binary


Jaifroid:
This has puzzled me for some time and I believe I've now discovered the cause. At least in jQuery mode, we currently use `zimArchive.readBinaryFile()` to decode CSS files, whereas for articles (i.e. html files) we use `zimArchive.readArticle()`. The only difference between these two is that the latter decodes the binary array using `utf8.parse()` whereas the former transforms the binary array using a slightly hacky `util.uintToString()` function that also claims to convert a binary array to UTF8 (we call that in jQuery mode in app.js to turn the binary into a string).

If we examine the footer of a typical article from full English Wikipedia, e.g. Piracetam.html (NB this needs to be a desktop-style ZIM, not mobile, to see the difference), using Firefox Quantum, this is what we see in our current master branch:

![image](https://user-images.githubusercontent.com/4304337/40304369-6eceadb2-5cee-11e8-9360-41153bd7d4f1.png)

If we instead use the same routine to decode the CSS file that we use to decode the HTML file, this is what we get:

![image](https://user-images.githubusercontent.com/4304337/40304439-a7efd684-5cee-11e8-990e-69a699db822c.png)

The second is clearly correct. Commit https://github.com/kiwix/kiwix-js/commit/abadc19543447abd6f4e081cc5c81626c836324c shows suggested changes (so far only in jQuery mode) that could solve this issue. I note that all files (other than the article) are sent to the Service Worker in binary format: it may be that the browser knows how to decode the CSS files, but it would be worth testing whether the first image above is what is shown in Service Worker (on master), or the second. If it is the first, then a similar fix may be needed in Service Worker.

Jaifroid:
PS Ray Charles articles show the same effect as the article above on master vs the [Decode-CSS-as-UTF8-byte-array](https://github.com/kiwix/kiwix-js/tree/Decode-CSS-as-UTF8-byte-array-instead-of-uint8-binary) branch.

mossroy:
Excellent discovery!
I reproduce the issue, and can confirm it does not happen in ServiceWorker mode (with Firefox or Chrome). Your branch seems to solve the issue in jQuery mode.
Cool! That's worth a PR

Jaifroid:
I changed title of this issue because I think technically we "encode" binary as UTF8 rather than "decode". "Read" should avoid (my) confusion.

Jaifroid:
Merged #371

PR #371, opened 2018-05-21
Read CSS as utf8 strings instead of binary

Jaifroid:
See discussion in #370 . It has yet to be determined whether JavaScript files should also be decoded as UTF8, or, since we will store them as Blobs, whether a binary format is appropriate for the browser to decode. Since we do not yet use the JavaScript function, that should be left to future investigation.

### DIFF
--- a/tests/tests.js
+++ b/tests/tests.js
@@ -148,7 +148,7 @@ define(['jquery', 'zimArchive', 'zimDirEntry', 'util', 'uiUtil', 'utf8'],
                 assert.ok(htmlArticle.match("^.*<h1[^>]*>A Fool for You</h1>"), "'A Fool for You' title somewhere in the article");
                 done();
             };
-            localZimArchive.readArticle(aFoolForYouDirEntry, callbackFunction);
+            localZimArchive.readUtf8File(aFoolForYouDirEntry, callbackFunction);
         });
         QUnit.test("check findDirEntriesWithPrefix 'A'", function(assert) {
             var done = assert.async();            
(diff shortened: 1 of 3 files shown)


mossroy:
According to http://www.openzim.org/wiki/ZIM_file_format#Character_Encoding , everything should be encoded in UTF-8 inside ZIM files. @kelson42 : could you confirm it's also true for CSS and Javascript files?

mossroy:
@ernesst : the day after this PR will be merged, could you please test a nightly build on ubuntu touch? We're preparing a release of a version 2.3.

Jaifroid:
@mossroy , I've updated the tests.js, and tests are now passing.

Regarding JavaScript, as the most likely scenario is that we will be storing the scripts as Blobs, which are by definition binary format, and we'll then ask the browser to load them via the src attribute of the script tags, we should leave it to the browser to read/encode the JavaScript contents even in jQuery mode. I've tested this approach in the javaScript-support branch, and the scripts work fine that way.

So what I've done here is to approximate minimally the method used there. However, it won't actually work here, because with the innerHTML mode of injection, scripts attached by changing the src of a script tag do not read the files (by design). Therefore, the scripts will need to be attached by deleting the old script and creating a new script element. This is working in javaScript-support, but I feel it is a different issue and a different PR.

So let me know if you're happy with the skeleton I've put here. It's disabled in any case in the iframe onload section. I've added a note to point to #152 .

ernesst:
@mossroy sure no prob.

mossroy:
@kelson42 : yes, that's the wiki page I had read.
But when it says "both article data and URLs should be handled accordingly", it was not clear to me if "article data" was only the HTML page or also the javascript, CSS files that come with it.
Maybe this sentence could be more precise about this kind of content.

In any case, we now decode every textual content in UTF-8.

kelson42:
@mossroy yes, the sentence is not so clear - but I have nothing more clear in mind. It just means that if the URL in the ZIM url table is encoded in UTF8 (it should be) then you better have an HTML/CSS page in UTF8 too if it points to it... because nothing will make a re-encoding on the fly if a link points from the HTML/CSS document to that resource.

Issue #376, opened 2018-05-25
Multiple 404 errors in Chromium and Edge when running from localhost in jQuery mode


Jaifroid:
This is effectively the same error as #272. It occurs also with stylesheets when the browser is running kiwix-js from localhost -- screenshot below is from Chromium. I see this in the console log of Chromium and Edge (but not Firefox). I imagine it's also occurring under the hood when running from the file:// protocol, but the console log doesn't bind the network stack in file:// protocol.

The solution is the same as for images. We can add this line below the current regex for images, but for the sake of completeness, we should do the same for scripts. Full suggested revised code is:
```
    if (contentInjectionMode === 'jquery') {
        // Fast-replace img and script src with data-kiwixsrc [kiwix-js #272]
        htmlArticle = htmlArticle.replace(/(<(?:img|script)\s+[^>]*\b)src(\s*=)/ig, '$1data-kiwixsrc$2');
        // Replace link hrefs to prevent 404 errors in http:// mode
        htmlArticle = htmlArticle.replace(/(<link\s+[^>]*\b)href(\s*=)/ig, '$1data-kiwixhref$2');            
    }

```
Then, in `loacCSSJQuery()`, just as in `loadImagesJQuery()` we reference `link.attr("data-kiwixhref")` instead of `link.attr("href")`.

![image](https://user-images.githubusercontent.com/4304337/40539454-b7a9c96a-600c-11e8-91c0-4bf2466746e7.png)

mossroy:
I agree (I saw that too), but I'll change the label to "enhancement", as it has no impact for the user.

Jaifroid:
Commit https://github.com/kiwix/kiwix-js/commit/64061cc8d9ca88957e60c80dd57cc7e33e7a3e4c implements this fix. Tested on Edge, IE11, FFOS and Chromium (Edge and Chromium in http: mode as well as file:// mode). I'm happy to wait to do a PR after v2.3 or, as it's small, if you decide to change the milestone, I can do a PR now. No rush. -- I don't want to delay 2.3.

Jaifroid:
Forgot to update `loadJavaScriptJQuery()` too. Done (branch [Fix-404-errors-in-http-mode](https://github.com/kiwix/kiwix-js/compare/Fix-404-errors-in-http-mode)).

mossroy:
Please create the PR : we'll see if it can be in 2.3 or not.

PR #378, opened 2018-05-26
Fix 404 errors in http mode

Jaifroid:
See #376 .

### DIFF
--- a/www/js/app.js
+++ b/www/js/app.js
@@ -807,10 +807,12 @@ define(['jquery', 'zimArchiveLoader', 'util', 'uiUtil', 'cookies','abstractFiles
     var regexpZIMUrlWithNamespace = /(?:^|\/)([-ABIJMUVWX]\/.+)/;
     // Pattern to match a local anchor in a href
     var regexpLocalAnchorHref = /^#/;
-    // These regular expressions match both relative and absolute URLs
-    // Since late 2014, all ZIM files should use relative URLs
-    var regexpImageUrl = /^(?:\.\.\/|\/)+(I\/.*)$/;
-    var regexpMetadataUrl = /^(?:\.\.\/|\/)+(-\/.*)$/;
+    // Regex below finds images, scripts and stylesheets with ZIM-type metadata and image namespaces [kiwix-js #378]
+    // It first searches for <img, <script, or <link, then scans forward to find, on a word boundary, either src=["'] 
+    // OR href=["'] (ignoring any extra whitespace), and it then tests everything up to the next ["'] against a pattern that
+    // matches ZIM URLs with namespaces [-I] ("-" = metadata or "I" = image). Finally it removes the relative or absolute path. 
+    // DEV: If you want to support more namespaces, add them to the END of the character set [-I] (not to the beginning) 
+    var regexpTagsWithZimUrl = /(<(?:img|script|link)\s+[^>]*?\b)(?:src|href)(\s*=\s*["']\s*)(?:\.\.\/|\/)+([-I]\/[^"']*)/ig;
     
     // Cache for CSS styles contained in ZIM.
     // It significantly speeds up subsequent page display. See kiwix-js issue #335
@@ -828,8 +830,9 @@ define(['jquery', 'zimArchiveLoader', 'util', 'uiUtil', 'cookies','abstractFiles
         $("#articleContent").contents().scrollTop(0);
 
         if (contentInjectionMode === 'jquery') {
-            // Fast-replace img src with data-kiwixsrc [kiwix-js #272]
-            htmlArticle = htmlArticle.replace(/(<img\s+[^>]*\b)src(\s*=)/ig, '$1data-kiwixsrc$2');
+            // Replaces ZIM-style URLs of img, script and link tags with a data-url to prevent 404 errors [kiwix-js #272 #376]
+            // This replacement also processes the URL to remove the path so that the URL is ready for subsequent jQuery functions
+            htmlArticle = htmlArticle.replace(regexpTagsWithZimUrl, "$1data-kiwixurl$2$3");            
         }
 
         // Compute base URL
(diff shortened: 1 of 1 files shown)


Jaifroid:
OK, I hadn't given much through to this. On reflection, it's very important that we match the same elements that we will match later in the functions that insert images, css and scripts, so if we make a search more specific here, we need to make it more specific later too, otherwise we'll end up with mismatched tags. Here is a summary of what we currently select (in italics) plus discussion. If it's too long, skip to the conclusion at the end in bold:

**Images**: *all `img` tags in query selector, then filter those that have an `src` that matches an internal ZIM file*. We could tighten up our image selection here, but it will only be useful if it is possible that there are other images in the ZIM that link to external files. As, by definition, all ZIMs must work entirely offline, it doesn't seem necessary to do a more specific regex here for images.

**Links**: *all link tags that have a relationship (`rel`) attribute of "stylesheet" in the query selector, and then we filter for those whose `href` matches `regexpMetadataUrl`*. We can certainly make the regex here more specific to match only those `link` tags that have `rel="stylesheet"`. However, it begs the question as to why we *only* select for "stylesheet" relationship types later. Surely it would be valid, for example, for the ZIM to contain stylesheet relationship types of "alternate",  "author", "help", "license", "next", "pingback", "prefetch", "preload", "prev" or "shortlink" (these are the ones that seem to be legal for links and not deprecated or browser-specific). The real question is: **Can there be a `link` tag with a ZIM-style href that we do *not* want to process?** If the answer is no, then our later query selector should look for all links with an href matching a ZIM link, and not limit it to stylesheet types, as we currently do. And in that case, we should tighten up this regex to select only `link` tags where the href matches a ZIM-style link. The relationship type is then irrelevant. 

**Scripts**: *all `script` tags, and then filter for those whose `src` matches a ZIM link*. Like images, we could make the regex more specific here, though it doesn't seem useful, as by definition an `src` that points to an external source that is not a ZIM reference can have no meaning in an offline ZIM.

**Conclusion: We should tighten up the `link` tag regex to select only those that have an `href` that matches `regexpMetadataUrl`. If that matches, then it is safe to ignore the relationship type, as that is the only thing we're interested in. We should further simplify the later query selection `'link[rel=stylesheet]'` to `'link[data-kiwixhref]'`, since the `rel` should not be limited to stylesheet types.**

Jaifroid:
I'll push a proposed simplification for testing later this morning.

Jaifroid:
Commit https://github.com/kiwix/kiwix-js/pull/378/commits/f2638f05f5169048b8370dd607220621491a3b86 goes to the core of the matter by only selecting link, img and script tags where the href or src matches a ZIM-style URL. I think this solves the whole problem, because that is really all we are interested in. We don't need to do complex HTML parsing, because zimwriter / mw-offliner should already have done that for us, and should have decided when the link should be in the ZIM or not. That's all we need to know, right? So that's all we need to test for.

Since we're doing this test here in jQeury mode, we can take the opportunity to simplify the testing later in the jQeury functions that deal with links, images and scripts. Effectively, we're also only interested there in elements that have matched this earlier regex.

Note that I believe this code now makes redundant the two regexes `regexpImageUrl` and `regexpMetadataUrl`. Assuming they are only used in app.js, and not elsewhere (I couldn't find them elsewhere on a quick search), then if this code is acceptable, they can be removed. Or they can be kept for instructional purposes for future devs, to show them how to do this kind of matching.

I've tested this commit on Edge, IE11, Firefox ESR, FFOS, and Chromium, on Wikimedia and one Stackexchange ZIM. I'm going to add a regex101 in a moment to test for regex optimization (I already did some optimization to minimize regex search steps).

Jaifroid:
Here's the regex101 link: https://regex101.com/r/kjldDq/2 
This is for the JavaScript version. In Firefox ESR this makes 24 matches in 2 milliseconds, which is pretty fast! To see more clearly what has matched, use this one:
https://regex101.com/r/kjldDq/1 
This one runs in PHP and therefore highlights all sub-matches more clearly.

mossroy:
You're right, I like this approach. And it makes the code a bit more simple.
In theory, there is a little difference between the regexp and what you match with jQuery after : the regexp matches for example a script with a matching href (instead of src) attribute, or a link with a matching src (instead of href) attribute, or a img link with a matching href (instead of src) attribute. But theses combinations are apparently not valid in the HTML spec, so I think it's not a problem.

www/js/app.js
```diff
@@ -811,6 +811,9 @@ define(['jquery', 'zimArchiveLoader', 'util', 'uiUtil', 'cookies','abstractFiles
     // Since late 2014, all ZIM files should use relative URLs
     var regexpImageUrl = /^(?:\.\.\/|\/)+(I\/.*)$/;
     var regexpMetadataUrl = /^(?:\.\.\/|\/)+(-\/.*)$/;
```
mossroy: I think you can safely remove these 2 unused lines (and their comment above)

www/js/app.js
```diff
@@ -811,6 +811,9 @@ define(['jquery', 'zimArchiveLoader', 'util', 'uiUtil', 'cookies','abstractFiles
...
+    // Regex below finds images, scripts and stylesheets with ZIM-type metadata and image namespaces [kiwix-js #378]
+    // DEV: If you want to support more namsespaces, add them to the END of the character set [-I] (not to the beginning)
```
mossroy: Little typo on "namsespaces" : "namespaces"
As the regexp is not trivial, I think it's necessary to describe it a bit more in the comment.

Jaifroid: OK, will see if I can come up with a clearer comment.

Jaifroid:
> the regexp matches for example a script with a matching href (instead of src) attribute, or a link with a matching src (instead of href) attribute, or a img link with a matching href (instead of src) attribute. But theses combinations are apparently not valid in the HTML spec, so I think it's not a problem.

Yes, I was aware of that, but we have a wider problem if the ZIM has valid ZIM URLs assigned to the wrong attribute for the tag. We could do two separate regexex, but I felt there was a minor performance gain to sweeping through the HTML only once and picking everything up in that sweep.

mossroy:
I agree that it's probably better to keep this version

Jaifroid:
I added this description to the new regex to make it clear how it works. Is this sufficient? Too much?
```
// Regex below finds images, scripts and stylesheets with ZIM-type metadata and image namespaces [kiwix-js #378] 
// It first searches for <img, <script, or <link, then scans forward to find, on a word boundary, either src=["']  
// OR href=["'] (ignoring any extra whitespace), and it then tests everything up to the next ["'] against a pattern that 
// matches ZIM URLs with namespaces [-I] ("-"" = metadata or "I" = image). Finally it removes the relative or absolute path.  
// DEV: If you want to support more namespaces, add them to the END of the character set [-I] (not to the beginning)
var regexpTagsWithZimUrl = /(<(?:img|script|link)\s+[^>]*?\b)(?:src|href)(\s*=\s*["']\s*)(?:\.\.\/|\/)+([-I]\/[^"']*)/ig; 
```

mossroy:
This description is perfect.
If you tested enough, I think you can rebase/squash/merge

PR #383, opened 2018-05-30
In ServiceWorker mode, set the iframe src to display articles.

### LINKED ISSUE #382: In ServiceWorker mode, always set the src of the iframe instead of using the displayArticleInForm function
The displayArticleInForm is used, even in ServiceWorker mode, when an article is opened programmatically (welcome page, after a search, random etc).
But it would be simpler (and more consistent) to simply set the src of the iframe in these cases, and let the ServiceWorker do the job (as it already does when the user clicks on a link inside an article).
It also seems to significantly improve performance on complex pages (see #381)

mossroy:
Instead of calling displayArticleInForm like in jQuery mode.
I also removed some tests on jQuery/ServiceWorker modes in displayArticleInForm, as this function is now only used in jQuery mode.
And I renamed it to displayArticleContentInIframe

Fixes #382

### DIFF
--- a/www/js/app.js
+++ b/www/js/app.js
@@ -743,11 +743,26 @@ define(['jquery', 'zimArchiveLoader', 'util', 'uiUtil', 'cookies','abstractFiles
      * @param {DirEntry} dirEntry
      */
     function readArticle(dirEntry) {
-        if (dirEntry.isRedirect()) {
-            selectedArchive.resolveRedirect(dirEntry, readArticle);
+        if (contentInjectionMode === 'serviceworker') {
+            // In ServiceWorker mode, we simply set the iframe src and show it when it's ready.
+            // (reading the backend is handled by the ServiceWorker itself)
+            var iframeArticleContent = document.getElementById('articleContent');
+            iframeArticleContent.onload = function () {
+                iframeArticleContent.onload = function () {};
+                // Actually display the iframe content
+                $("#readingArticle").hide();
+                $("#articleContent").show();
+            };
+            iframeArticleContent.src = dirEntry.namespace + "/" + dirEntry.url;
         }
         else {
-            selectedArchive.readUtf8File(dirEntry, displayArticleInForm);
+            // In jQuery mode, we read the article content in the backend and manually insert it in the iframe
+            if (dirEntry.isRedirect()) {
+                selectedArchive.resolveRedirect(dirEntry, readArticle);
+            }
+            else {
+                selectedArchive.readUtf8File(dirEntry, displayArticleContentInIframe);
+            }
         }
     }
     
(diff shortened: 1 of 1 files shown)


mossroy:
I've tested on Firefox 60 and Chromium 66, in ServiceWorker and jQuery modes, with various ZIM files (including stackoverflow.com_eng_all_2017-05.zim).

I also removed some tests on jQuery/ServiceWorker modes in displayArticleInForm, as this function is now only used in jQuery mode. And I renamed it to displayArticleContentInIframe.

www/js/app.js
```diff
@@ -743,11 +743,26 @@ define(['jquery', 'zimArchiveLoader', 'util', 'uiUtil', 'cookies','abstractFiles
      * @param {DirEntry} dirEntry
      */
     function readArticle(dirEntry) {
-        if (dirEntry.isRedirect()) {
-            selectedArchive.resolveRedirect(dirEntry, readArticle);
```
Jaifroid: I realize the code that provides the Service Worker with the content does its own resolveRedirect, but wouldn't it save a hop if we allowed this redirect code to run for both modes (jQuery and ServiceWorker) here, then set the iframe's src to the dirEntry.namespace + url of the new dirEntry? Maybe it's unlikely we'd actually encounter a redirect here? Anyway, just a thought.

Jaifroid:
Here's an example of the spinner still showing while images are being loaded, in Service Worker. I'm not familiar enough with Service Worker to remember if this is the standard behaviour in that mode. (Sorry for the milkiness of this image, I put a breakpoint to be sure I could screenshot the area.)

image

www/js/app.js
mossroy: Yes, it would certainly work and speed up a little bit.
It's completely possible to encounter a redirect here.
On the other hand, we would "only" save the binary search done by getDirEntryByTitle.
And we had here the opportunity to do things exactly the same between an article opened after a search (or a random search, or the main article), and an article opened while browsing (after a click on a hyperlink). If we handle the redirection differently, we re-introduce a little difference.
I'm torn between doing it or not (or doing it later). Maybe we should benchmark it?

mossroy: OK I created #384 for that, let's leave this portion as it is, for now.

mossroy:
Regarding the spinner issue, I don't reproduce it on Firefox or Chrome.
We call `$("#readingArticle").hide();` in the onload event, so it's normal that it disappears after the page is completely ready. But it's also at this moment that we call `$("#articleContent").show();`, so the article should not be visible before we hide the spinner...
Maybe Edge makes the iframe visible as soon as we set a new src to it?
In any case, both behaviors look acceptable to me.

Regarding the ServiceWorker mode in general, there are many things that should be easier and faster to achieve with it, using the browser capabilities instead of reimplementing things ourself. That's why I'm willing to focus on it for the next hackathon, in order to (hopefully) make it the default mode (when available on the device/browser). It does not mean we would drop jQuery mode support, because it's still necessary in many contexts.

mossroy:
Can I merge this PR?

Jaifroid:
Yes, looks good to me!

Issue #367, opened 2018-05-09
Remove jQuery and use native DOM functions instead


mossroy:
Following the suggestions of @Jaifroid in #361, it would be interesting to see if we can remove the use of jQuery library.
The reasons why we introduced it (notably cross-browser compatibility) might not be relevant now. We make a basic use of jQuery.

If removing this library works on our target browsers, does not make the code much more complicated, and gives a few performance enhancement, it might be worth it.

Jaifroid:
Here is a great resource for viewing the jQuery source code of any particular jQuery function:

https://j11y.io/jquery/ 

I think it would be difficult to complete a transition from jQuery to native DOM methods all in one go. I think it would have to be in stages. Some of the work has been done in the [fix-TypeError-tests](https://github.com/kiwix/kiwix-js/tree/fix-TypeError-tests) branch, which I have just rebased off the latest version of `Fix-history-back-regression` branch.

Related thought: It might be worth having some blue-sky thinking about experimenting with a contemporary app framework like React (with or without [React-JSX](https://reactjs.org/docs/introducing-jsx.html)). It is possible to use React to compile to a web app with node, but there is also [React-native](https://facebook.github.io/react-native/) that will compile to Android, iOS and [UWP native apps](https://github.com/Microsoft/react-native-windows) off the same JavaScript codebase.

mossroy:
Thanks for this useful link.
What you already commited in the branch looks promising.
Why couldn't the transition be done in one shot? Because I think it's necessaey

mossroy:
Regarding React, it's probably another github issue to create and discuss on. I don't really see what would be its benefits for us, for now

Jaifroid:
I just thought it might be difficult to maintain two separate codebases in sync with each other for the amount of time needed to go through the hundreds of lines that use jQuery in minor form - things like `$('btnTop').on('click', function(){…..});` which would need to be changed to `document.getElementById('btnTop').addEventListener('click', function(){…});`. There's a *lot* of that kind of code. Anyway, GitHub has some good tools for keeping things in sync.

As for React, the main benefit would be the possibility of having a single code-base in JavaScript that can compile to JavaScript, PWA, UWP, iOS and Android native apps. Kind of like a contemporary, UI-centred version of Cordoba. Anyway, it was just something that occurred to me as a long-term possibility to bear in mind. It's not worth creating an issue for it yet.

kelson42:
Is this ticket not deprecated?

[...]

PR #385, opened 2018-05-30
Prevent render till css fulfilled from cache or zim

### LINKED ISSUE #381: Serious performance regression on rendering the index page of stackoverflow.com_eng_all_2017-05.zim
The index page (that is automatically loaded when you choose this ZIM file) is very heavy.
In version 2.2, it already took some time to load it.
But now, applying the extra CSS (that were not loaded in 2.2) takes almost 10 minutes on my computer, with the browser eating a whole CPU all this time.

It's not a bug, but the fact that we are now able to apply all CSS styles has this consequence. And it makes this ZIM file unusable.

I did not have the time to investigate more on that. It would be worth testing with the desktop version

Jaifroid:
Please try out this PR, @mossroy, which fixes #381 for jQuery mode. I've tested load of the Stackoverflow home page in Edge, Chromium and Firefox in jQuery mode (I haven't tried to load that page in IE, because IE is very slow on normal ZIMs, but other ZIM files work fine in IE with this code). I've also tested the code on FFOS, but not with that ZIM.

It works like this: if (some of) the CSS files are cached, then preload them into the html string before injecting the string into the iframe. If they are not cached (or if some of them aren't), then inject the HTML but keep the iframe hidden until all the CSS files have been fulfilled from the ZIM. Then render the iframe. Since browsers don't render content where `style.display == "none"`, we can take advantage of this behaviour to prevent rendering until all CSS is available and injected through one method or other.

The code is based on a more complete caching solution I've been developing as a replacement for the clunky (but functional) caching-in-the-filesystem that I was doing this time last year for Kiwix JS Windows. The more complete solution uses indexedDB or localStorage to persist the cached assets between ZIM loads (in jQuery and ServiceWorker), and includes the ZIM filename in the cacheKey. It is too big a change to contemplate just for this issue, however, so I've adapted relevant bits.

But you'll notice some embryonic features that will make it easier to develop the following:

1. Addition of JavaScript to the cache (hence I've renamed it assetsCache) -- this will be absolutely vital if we ever hope to support JavaScript; 
2. Asynchronous cache calls (necessary for indexedDB).

You'll also notice that the assetsCache is declared app-wide. This is so that we don't need to pass it around between app.js and uiUtil.js (and a future cache.js) in function calls. Here we're still erasing it between ZIM loads. If you don't like this and have a better way to avoid passing the cache around, please suggest!

The per-ZIM persistent cache *really* makes the app fly, b.t.w.!

### DIFF
--- a/www/js/app.js
+++ b/www/js/app.js
@@ -544,8 +544,8 @@ define(['jquery', 'zimArchiveLoader', 'util', 'uiUtil', 'cookies','abstractFiles
      */
     function resetCssCache() {
         // Reset the cssCache. Must be done when archive changes.
-        if (cssCache) {
-            cssCache = new Map();
+        if (assetsCache) {
+            assetsCache = new Map();
         }
     }
 
@@ -829,10 +829,6 @@ define(['jquery', 'zimArchiveLoader', 'util', 'uiUtil', 'cookies','abstractFiles
     // DEV: If you want to support more namespaces, add them to the END of the character set [-I] (not to the beginning) 
     var regexpTagsWithZimUrl = /(<(?:img|script|link)\s+[^>]*?\b)(?:src|href)(\s*=\s*["']\s*)(?:\.\.\/|\/)+([-I]\/[^"']*)/ig;
     
-    // Cache for CSS styles contained in ZIM.
-    // It significantly speeds up subsequent page display. See kiwix-js issue #335
-    var cssCache = new Map();
-
     /**
      * Display the the given HTML article in the web page,
      * and convert links to javascript calls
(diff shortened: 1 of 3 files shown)


Jaifroid:
@mossroy With that Stackoverflow home page, I've noticed that [this line](https://github.com/kiwix/kiwix-js/pull/385/files#diff-bca463ed8ccf5e2058dfe307a95e8501L849) makes it extremely burdensome on the CPU to load a new page. Presumably jQuery is going through each and every one of the tag links and removing event listeners. If I comment out that line, then clicking on one of the tags opens the tag reasonably fast. With it in, we get a serious lock-up before the new page eventually loads.

I realize we're going round in circles here (we put this in to avoid the dead objects error). This page is not typical, so maybe we don't need to do anything about it. The longer-term solution is #367 (Remove jQuery), so that we don't have the overhead of jQuery's legacy caching of objects.

Jaifroid:
Just quickly, I thought to compare/benchmark a simplified version which doesn't preload anything into the html string, but merely prevents rendering until the CSS is ready (using `loadCSSJQuery`). It's in this branch:

https://github.com/kiwix/kiwix-js/tree/Prevent-render-without-preload

This code is less disruptive to our current way of doing things. It may not be quite so performant once CSS is in the cache. But maybe for a V2.3 release, it is a more palatable change. Let me know if you'd like me to PR that branch instead of this.

mossroy:
Please create a PR for this second branch.
Its code is much simpler, which seems less risky as a last step before a release.
A few minor remarks on it : 
- please use jQuery to display the content, to be consistent with the other places where we change this visibility
- the cssCount and cssFullfilled variables are not technically related to the CSS cache in this version. Maybe they could be simple variables inside loadCSSJQuery function and renderIfCSSFulfilled function might a function declared inside loadCSSJQuery (in order to have access to these variables). This way, the CSS cache would be independent from this optimization.

Jaifroid:
Agreed, closing.

PR #387, opened 2018-06-01
Prevent render until all CSS is fulfilled

Jaifroid:
@mossroy , I agree that this is a less risky way of solving (or rather, ameliorating) #381 for a v2.3 release.

I've implemented your helpful suggestions in commit https://github.com/kiwix/kiwix-js/commit/127ff299de66c16398e250dc56fbd696a0388ee8. Tested on IE, Edge, FFOS, Firefox Quantum.

I've also, optionally, added a commit https://github.com/kiwix/kiwix-js/commit/41f1b7042ac87ba83e171e0a42416a836f67bd42 which displays an additional message to the user when the app is retrieving CSS from the ZIM instead of the cache. Because that is slow, and because first impressions often count strongly with apps, giving this additional small piece of feedback could be a user-friendly gesture. The message won't display again on subsequent page loads unless a further piece of CSS needs to be cached.

It's easy to drop that commit if you feel it's unnecessary. On FFOS, because of the very small screen size, it does look a bit cluttered the first time. But I think nowadays most mobiles have more width. However, the message could be abbreviated to just "Caching styles...". See what you think.

### DIFF
--- a/www/index.html
+++ b/www/index.html
@@ -255,7 +255,10 @@
                     </div>
                 </div>
                 <div id="readingArticle" style="display: none;" class="container">
-                    Reading article <span id="articleName"></span> from archive... Please wait <img src="img/spinner.gif" alt="Please wait..." />
+                    <p>Reading article <span id="articleName"></span> from archive... Please wait <img src="img/spinner.gif" alt="Please wait..." /></p>
+                </div>
+                <div id="cachingCSS" style="display: none;" class="container">
+                    <p>Caching styles to speed up future page display...</p>
                 </div>
                 <iframe id="articleContent" class="articleIFrame" src="article.html"></iframe>
             </article>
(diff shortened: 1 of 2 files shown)


Jaifroid:
Maybe commit https://github.com/kiwix/kiwix-js/pull/387/commits/b985966d72625fe3440371594f8d17eb21cadf9d is a better way to display the progress message (it replaces the previous message rather than adding it on an extra line -- better for small screens?).

mossroy:
It looks like great job!
I've quickly tested on Firefox and a Firefox OS device, and it worked fine.
It's a good idea to add information about what's going on : it's certainly a relevant feedback for the user. I tend to prefer your first version (with the css message added to the original one) than the second (where it replaces it). It looks fine on my Firefox OS device too.

Jaifroid:
Merged.

mossroy:
@ernesst : could you please test the ubuntu touch nightly build when it will be available in https://download.kiwix.org/nightly/2018-06-01/ ?
We're about to release version 2.3

Issue #390, opened 2018-06-05
Correct the mimetype for SVG images in jQuery and ServiceWorker


Jaifroid:
As noted in #219 and #278, SVG images (the usual format for equation fallback images) have no hope of being displayed correctly currently in ServiceWorker or in jQuery modes, because no correct mimetype is applied to them. This would also address, but not fully fix #308 (#389 would be needed fully to enable display of equations).

Jaifroid:
I should add that there is already an issue #137 for a possible ZIM-based solution. I shall propose a pragmatic solution, but we should also see if a more generic one is possible either now or later (I don't feel I have the knowledge to do the latter).

mossroy:
Fixed by #391 
A more generic way to handle mimetypes is tracked on #137

Jaifroid:
@kelson42 I think the issue is that we don't have a function to read the ZIM's mimetype table, so we can't (currently) derive the mimetype from the dirEntry's mimetype byte, as it's meaningless without the table. If the mimetype table is always the same for every ZIM, then we could do it. But we don't know that, I think, looking at the specs. It's quite low-level stuff...

kelson42:
@Jaifroid OK, no the mime-type is not always the same. This should be retrieved at each file opening.

kelson42:
@Jaifroid @mossroy This is a quite dangerous approach on the longer term. I have for example in project to implement https://github.com/openzim/mwoffliner/issues/278. That mean that web pages would not end anymore with an ".html" extension.

Jaifroid:
I completely agree. It shouldn't be *too* difficult to read the mimetype table on ZIM opening. We don't even need xzdec to do that. It's just like reading a dirEntry, I believe, but it's low-level stuff I don't feel very confident about atm. Hopefully we'll get that implemented before the change you mention, though. We have issue #137 to work on it.

Issue #137, opened 2016-01-06
Ability to read and handle the different MIME-types included in the ZIM file


mossroy:
For now, the MIME types are often hard-coded, despite the fact that they are available in the ZIM file with each content.
We should be able to retrieve this information and use it

kelson42:
This ticket might start to be critical if https://github.com/openzim/mwoffliner/issues/278 is implement, which probably will happen in the next 2/3 quarters.

mossroy:
I agree that it's something that needs to be done in the mid-term.
But, to be precise : with the current implementation, removing the .html extension in URLs should not hurt kiwix-js. We're using the suffix to guess the MIME-type of images, CSS, javascript etc, but not for the article HTML content. Currently, it's the use of namespace "A" that tells us it's an article.
So https://github.com/openzim/libzim/issues/15 should be a bigger problem than https://github.com/openzim/mwoffliner/issues/278.

All these guesses were a quick-and-dirty way to move on, when developing kiwix-js. The right way to do it is to read and use the MIME-types from the ZIM file.
So far, nobody took the time to implement that. I think it's because we never faced any bug on that (so far), and because it's low-level file reading that not many people like to work on.

As the ZIM file format changes and improves, it will be more a more difficult to maintain the ZIM backend in pure javascript. That's why I think we need to work on #116 first.

Jaifroid:
@mossroy Looking at the spec, I don't think it's difficult to add a function to `zimArchive.js` to get the MIME type table, and it would be a more robust solution for some code I have that is currently having to "guess" the type from the namespace.

I envisage that we would want to store the MIME Type List in the selectedArchive (prototype) to avoid a large number of binary lookups at the head of the ZIM file (where the Mime Type List is stored) for every article loaded.

Could you let me know if the following be a good way to do this?

1. Create a prototype `mimeTypeList` (array or map) in `zimArchive.js` and populate it once on archive load (or, alternatively, the first time that a MIME type lookup is requested); and
2. Create a prototpye lookup function in `zimArchive.js` that returns a MIME type string looked up from the `mimeTypeList` (using `dirEntry.mimeType` value).

Jaifroid:
And we'll need code in `zimfile.js` equivalent to `ZIMFile.prototype.dirEntryByUrlIndex` that reads the zero-terminated values stored at the position indicated in `zf.mimeListPos` (this value is already defined in `zimfile`.

Jaifroid:
[...]

PR #391, opened 2018-06-05
Add more specific mimetype support for images #390

Jaifroid:
This proposes a pragmatic solution to #390. For ServiceWorker, I have merely added SVGs to the pragmatic code that is already there (that type was missing).

For jQuery, I've used a more compact solution. These regexes are extremely small and quick for the system to match, and I'm not sure anything much would be saved by compiling them in advance, given that the bottleneck is extraction from the ZIM. Speedup here is probably immaterial and the operations can easily be completed in a minute fraction of the time between images arriving. EDIT: ah, of course I forget, unless we're using a Worker, there is no "time between", as there is only one thread. But it's still probably minimal.

In any case, this is probably "meanwhile" code, unless someone would like to bite the bullet with #137 now?

### DIFF
--- a/service-worker.js
+++ b/service-worker.js
@@ -84,6 +84,7 @@ var regexpJPEG = new RegExp(/\.jpe?g$/i);
 var regexpPNG = new RegExp(/\.png$/i);
 var regexpJS = new RegExp(/\.js/i);
 var regexpCSS = new RegExp(/\.css$/i);
+var regexpSVG = new RegExp(/\.svg$/i);
 
 // Pattern for ZIM file namespace - see http://www.openzim.org/wiki/ZIM_file_format#Namespaces
 var regexpZIMUrlWithNamespace = new RegExp(/(?:^|\/)([-ABIJMUVWX])\/(.+)/);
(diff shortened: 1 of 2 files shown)


Jaifroid:
I don't know how smart the browsers are, but it's no problem to precompile them.
The idea of the generic catch-all regex is that it already does image/tiff and image/jpeg if a url ends .tiff or .jpeg. That's why I later only need to test for .jpg and .tif. Maybe I should add a note to explain, for maintainability. Will ask you to check code again when I've edited it.

mossroy:
Based on https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions , a regular expression literal (like you did) is compiled only once, when the script is loaded. Which is what we want.
So you don't need to modify your code : it's more readable to keep your code like  `/\.jpg$/i.test` than to externalize this regexp in a variable.
So ignore my remark on that.

mossroy:
I created #393 to use literals like you did when it's more readable, in the rest of the code.

Regarding the different file extensions, I now understand. Sorry, I read your code too fast.

NB : don't hurry to work on the current open PRs if you are busy these weeks. We have no rush

Issue #389, opened 2018-06-05
Batch decompression jobs so as to manage memory use and avoid crashing xzdec


Jaifroid:
Recent Wikimedia ZIMs which often have up to six or more stylesheets per page, can crash xzdec when all of them are sent to the decompressor at the same time, especially on devices with less than 1GB RAM. This also occurs on any machine with pages that have a lot of SVGs (e.g. maths pages with equations), which are stored also in compressed format. The issue doesn't affect JPEGs and other images that are merely extracted from the ZIM without xzdec (because their format is already highly compressed).

Batching arises out of experiments in #219 , where it is proving necessary for faster compiles of xzdec on lower-end devices. However, a successful PR addressing the issue could also, potentially, fix #278, #288 and #308.

Jaifroid:
@mossory : PR #392 is set up to experiment with this issue.

Some notes on that PR:

* The PR works best if you cherry-pick (in your own repository) the commit in #390 on top of it. You can then (and only then) test for SVG-heavy pages such as Wilson-Cowan model in WikiMed or Sine.html. NB, FFOS doesn't seem able to display the SVGs at least in my simulator. They are decompressed just fine, and inserted as blob: urls, but they're blank only in FFOS (there is no crash).

* The option in init.js is for convenience of experimenting with different batch sizes. It may not be needed in final code. My experiments have tended to suggest that a batch size of one is best. It seems xzdec is not set up to manage well with more than one file at a time -- it doesn't seem to provide any saving of, say, overhead in setting up memory, etc. On the contrary, it forces xzdec to expand its memory tables (which is why we get the crash when sending several files for decompression).

* The reversal of the initial order of sending css vs images in app.js is there because in my tests, it has made a difference to the speed at which the page is displayed when processing a page with lots of images. The queue lets through any images that won't require decompression, so it's useful to have sent at least one CSS through first.

* As I've put in a comment, we have to guess which dirEntries need to be send to the compressor. This is because the testing to find out if the blob is in a compressed cluster only happens (presumably for performance reasons) *after* the read has begun. I can't find any "generic" way of finding out whether a specific dirEntry needs to be decompressed. If we forced a file read in this queue, we would damage performance (badly, I presume).

* I experimented with placing the queue at the blob level in zimfile.js, rather than at dirEntry level. Unfortunately, at the point where compressionType is tested, and then data received back, xzdec has not released its file handle. Not good.

[...]

PR #392, opened 2018-06-05
Send dirEntries to decompressor in batches

Jaifroid:
As per #389, here is some slightly reworked code to batch the decompressing of dirEntries. This PR is for testing, to see if this is a good way to do things. A more explicit queue could be devised that forces certain file types to be processed before others.

No rush at all on this, and it's probably best to deal with #390 first. I won't be able to work on this now for approx. 2 weeks. I'll add some notes in #389.

### DIFF
--- a/www/js/app.js
+++ b/www/js/app.js
@@ -874,8 +874,8 @@ define(['jquery', 'zimArchiveLoader', 'util', 'uiUtil', 'cookies','abstractFiles
             pushBrowserHistoryState(dirEntry.namespace + "/" + dirEntry.url);
             
             parseAnchorsJQuery();
-            loadImagesJQuery();
             loadCSSJQuery();
+            loadImagesJQuery();
             //JavaScript loading currently disabled
             //loadJavaScriptJQuery();            
         };
--- a/www/js/init.js
+++ b/www/js/init.js
@@ -22,6 +22,12 @@
  */
 'use strict';
 
+// Parameters that define overall operation of app
+var params = {};
+params['xzMaxJobs'] = 1; // Controls the maximum number of jobs that can be sent simultaneously to the decompressor
+                         // WARNING: setting this value too high can crash the app: see note at head of zimArchive.js 
+
+// Do not touch unless you know what you are doing
 require.config({
     baseUrl: 'js/lib',
     paths: {
--- a/www/js/lib/zimArchive.js
+++ b/www/js/lib/zimArchive.js
@@ -23,6 +23,13 @@
 define(['zimfile', 'zimDirEntry', 'util', 'utf8'],
     function(zimfile, zimDirEntry, util, utf8) {
     
+    // DEV: This controls the number of jobs sent to the decompressor at any one time. The value should be set in
+    // init.js and will need fine-turning according to your build of xzdec, and the memory constraints of your target
+    // environment(s). Some low-end mobiles can only support a value of 1.
+    var MAX_DECOMPRESSOR_JOBS = params.xzMaxJobs || 4;
+    // Counter to track the number of decompression-type jobs sent to xz
+    var xzJobs = 0;
+
     /**
      * ZIM Archive
      * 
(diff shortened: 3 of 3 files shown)


mossroy:
I quickly tested branch New-compile-of-xzdec (which includes almost the same implementation) : it does fix the crash on recent wikimedia ZIM files (with many CSS files), on my 512MB Firefox OS device.
It's great news!
Anyway, I'm wondering if it might have a negative impact on performance. Are you able to measure the difference (without the recompiled version of xzdec)?

I'm not clear on why the memory consumption is increased without your batch limit.
In theory, as we didn't implement a WebWorker for that, the decompression should not be multi-threaded (i.e. parallelized on multiple CPUs), so the memory should be released between each decompression. But maybe the code generated by emscripten uses callbacks that let several decompressions run in parallel (i.e. the event loop gives CPU to one then another one, even if the first one is not finished). This might be the reason?

I did not understand what did not work when you tried to put the queue at a lower level (closer to the decompressor itself), which I found a good idea. Ideally, this might be implemented in xzdec_wrapper.js?

Jaifroid:
@mossroy , I'm travelling to a conference, so can't explain fully, but I tried to put the queue round about [here](https://github.com/kiwix/kiwix-js/blob/New-compile-of-xzdec/www/js/lib/zimfile.js#L202), returning the queue instead of the promise for the decompressed blob (didn't intercept non-decompression jobs), and only returning the resolved promise when done (I kept track of the compressionType for the blob/cluster, so I was sure it was an xz job, and not a plain blob reader returning its data).

It worked, in that I could get the decompressor not to start up a new job until the data were extracted, but starting a new job at that point (after data extracted) still caused xzdec to expand its memory arrays (and hence crash an implementation limited to one job). I believe it's because the dirEntry has not at that point been returned even though the data are available? Then again, I wasn't always sure that blob == file (do some files potentially require more than one blob read?). It's very object-orientated and hard to get my head around... Maybe you can see a better way? I did think it was the logical place to delay decompression.

Regarding performance, you'd need to test the PR after rebasing it on the SVG commit, and set the maxJobs variable to 1, 2 or more to test. I didn't see obvious performance problems, but ran out of time...

I'm actually not so happy with the queuing mechanism now when there are lots of SVGs, even though it seem to work... I feel there must be a better way than setTimeout.

mossroy:
I tested a bit this File-level-batch-control branch on the home page of wikivoyage_en_all_novid_2018-03.zim (which uses 4 CSS styles), comparing the performance with master.
On Chromium, there's a slight difference : 1.65s on master, 1.81s on the branch
On Firefox, the difference is more important : 0.84s on master, 1.39s on the branch. I saw that a lot of time seemed to be "lost" because of the setTimeout waits (where the CPU is idle). If I reduce the timeout to 50ms on the same branch, it takes around 1s.
If I change the batch size to 4 (with a timeout of 50ms), it takes 0.88s

So there is a slight performance regression, that could be reduced by adjusting the timeout.
I would suggest to keep a batch size of 1 for compatibility with low-memory devices. I tend to think it's not worth using so much more memory to gain only 10% of performance on this use-case (especially because the CSS styles are cached for next pages).
I did not test the SVG use-case.

If we stick with a batch size of 1, maybe the code might be simplified? I'll try to look at the implementation : I agree with you that it's not easy.

mossroy:
I've created another branch https://github.com/kiwix/kiwix-js/tree/No-parallel-decompression where I'm experimenting another implementation, based on xzdec_wrapper.js
It can only use a batch size of 1, and is currently set with a delay of 50ms.
In terms of performance, it seems to be the same as this branch.
On my Firefox OS device with 512MB, it allows to read the recent wikipedia ZIM files with many CSS, and the memory usage seems to stay low.

Could you test this branch, and let me know what you think of it?

Jaifroid:
Thanks, @mossroy , will test the new approach for no parallel decompression shortly.

mossroy:
I've created a PR (#398) on my branch, to help keep track of the discussions about it

mossroy:
I close this PR, as it has been replaced by #398

PR #397, opened 2018-07-03
Read the language of a ZIM file right after it is loaded.

### LINKED ISSUE #395: Add support of ZIM meta information reading
We need to easily achieve to get the value of these metadata:
http://www.openzim.org/wiki/Metadata

This is necessary to apply a proper voice filtering (based on ZIM content language) after @kasemmarifet has implemented the TTS system.

mossroy:
So that to populate ZIMArchive.language
Fixes #395


mossroy:
As the _language attribute of a ZIMArchive is finally not used in #400, I might modify this #397 PR (or create another one) to only expose getMetadata function, without reading the _language attribute automatically (it's no use doing it on every ZIM file as it is currently unused).
@Jaifroid : would you prefer a new commit on this PR, or another PR?

mossroy:
I close this PR, as it is replaced by #431

PR #398, opened 2018-07-05
Prevent xzdec_wrapper from decompressing several chunks at a time.

### LINKED ISSUE #389: Batch decompression jobs so as to manage memory use and avoid crashing xzdec

mossroy:
Fixes #389

### DIFF
--- a/www/js/lib/xzdec_wrapper.js
+++ b/www/js/lib/xzdec_wrapper.js
@@ -25,6 +25,18 @@ define(['q'], function(q) {
     xzdec._init();
     
     /**
+     * Number of milliseconds to wait for the decompressor to be available for another chunk
+     * @type Integer
+     */
+    var DELAY_WAITING_IDLE_DECOMPRESSOR = 50;
+    
+    /**
+     * Is the decompressor already working?
+     * @type Boolean
+     */
+    var busy = false;
+    
+    /**
      * @typedef Decompressor
      * @property {Integer} _chunkSize
      * @property {FileReader} _reader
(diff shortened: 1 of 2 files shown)


mossroy:
The delay of 50ms can be discussed : I found it efficient on my desktop machine, but it does not seem optimal on my Firefox OS devices, where I can see the decompressor often takes much longer to be released.

I did not test on the SVG use-case : could you test how it behaves with this branch?

Jaifroid:
@mossroy I've ported this change to a dev branch of Kiwix JS Windows for testing in UWP, and so far they are working fine, and seem fast and efficient. I am using them with the version of the new asm-only fixed-memory compile of xzdec.js which only supports decompressing one file at a time (and will crash if attempting to decompress more than one). I have not yet encountered a crash on several different pages on different ZIMs.

I can also confirm that this version decompresses Sine.html and Wilson-Cowan model fast and efficiently, extracting the many SVGs on those pages with no crash. This is a good test of non-parallel decompression.

So, so far this is looking very good. I'll do more testing in different scenarios.

mossroy:
@Jaifroid : you were mentioning some more testing you wanted to do on this PR.
Could you keep me informed when you are done?
If we merge this PR, it would supersede #392

PR #402, opened 2018-07-12
Add favicon links in HTML header, to the existing PNG icons.

### LINKED ISSUE #401: feature request: add favicon
For those of us using kiwix-js on desktop browsers, it would be helpful to have the favicon showing on the Kiwix tab. I know this was mentioned in passing at https://github.com/kiwix/kiwix-js/issues/280#issuecomment-314118966, but since that issue is really about the window title, I figured the favicon should have its own issue.

Kiwix 2.3.0 on Firefox 61.0.1 on macOS

mossroy:
And remove obsolete favicon.ico (that was still using an evopedia icon).
Fixes #401


Jaifroid:
Sorry to have been slow on this! The favicon appears to work fine (based on my testing) in Firefox and in Chromium with both localhost and file access. In Edge and IE it works with localhost, but in neither does it work with file access (a generic icon is shown). This is probably an issue with the browsers (Microsoft's implementation) rather than something we need to do anything about, but I shall check to see if there's a simple fix worth implementing.

PR #406, opened 2018-08-05
Workaround serviceworker state loss after a while

mossroy:
By keeping it alive periodically.


Jaifroid:
Sorry to be slow @mossroy -- I'm on holiday abroad. Will get to this soon.

Issue #219, opened 2017-04-22
Recompile xzdec with Emscripten in WebAssembly (instead of asm.js) to improve performance


mossroy:
Now that the main browsers support WebAssembly (see http://caniuse.com/#feat=wasm), it would certainly speed up the application to use it for xz decompression.
Emscripten also proposes a polyfill to allow running the code on non-WebAssembly-compliant browsers, using asm.js as a fallback.

The C source code is https://tukaani.org/xz/embedded.html , but I don't know how @peter-x compiled it with Emscripten

mossroy:
Implementing this might be a first attempt to introduce the WebAssembly technology, before working on #116

mossroy:
I investigated a bit on that topic.
I managed to compile xz-embedded with Emscripten, but it does not correspond to the APIs of our current xzdec.js. So it seems useless for now.

Here are the steps I followed to compile xz-embedded :
- Install emscripten, following https://kripken.github.io/emscripten-site/docs/getting_started/downloads.html#sdk-download-and-
- git clone https://git.tukaani.org/xz-embedded.git (cf https://tukaani.org/xz/embedded.html)
- cd xs-embedded/userspace
- Modify the Makefile : 
    - replace cc by the full path to emcc in the CC vairable
    - remove the flag -ggdb3 in CFLAGS variable
    - add ".html" after the "-o $@" of every executable, so that it appears as "-o $@.html"
- Run emmake make
It generates an asm.js version. To generate a wasm version, we can add "-s WASM=1" in CFLAGS variable

Jaifroid:
Is xz-embedded superior to xzdec? A quick look through the site doesn't clearly highlight the differences, except that the former was developed for the Linux kernel, and the latter for POSIX-like systems.

If you recompiled xzdec do you think the API might still be the same?

mossroy:
I suppose xz-embedded is designed for low-spec devices (slow CPU, not a lot of RAM), but it's only a guess. 
I tried to compile xzdec with emscripten but it seems more complicated (and I could not spend a lot of time on it)

mossroy:
@peter-x managed to find the directory from where he generated the file xzdec.js : many thanks to him.
It is based on xz-embedded, but with some added code (that's why I did not find the same APIs).
There is a compile.sh script inside the directory, that should let us recompile it with a current version of emscripten (I did not have the time to try that yet) :
[...]

PR #408, opened 2018-09-12
Add xzdec source files, as provided by peter-x

mossroy:
This is the C source files of xzdec, and the original emscripten script provided by @peter-x a long time ago, that he used to generate xzdec.js.

It's a first step for #219 and lets us comply with the rule of the Firefox store, that asks to provide the source files of any generated file (like xzdec.js in our case).

This is the first commit of branch emscripten-script-xzdec (where I did some tests on recompiling xzdec)


Jaifroid:
Thanks for doing this, @mossroy . Would you like to merge this PR before we add compiles of xzdec as separate PR(s) , or do you want to add a new working compile of xzdec to this PR?

mossroy:
I had your first option in mind

PR #410, opened 2018-09-15
Asm-enabled compile with memory fixed at 5 x 16MiB

Jaifroid:
This PR isn't (yet) for merging. It is for the purposes of comparison. We should also do a PR which uses logic to select between this or similar asm-compile of xzdec and a wasm compile. I don't think clients, browsers and OS's are yet sufficiently capable for us to have a wasm-only build.

Compile string:
emcc --memory-init-file 0 -O3 -s WASM=0 -s MALLOC="emmalloc"
-s TOTAL_MEMORY=83886080 -s NO_FILESYSTEM=1
-s AGGRESSIVE_VARIABLE_ELIMINATION=1 -s DOUBLE_MODE=0 -s PRECISE_I64_MATH=0
-s NO_DYNAMIC_EXECUTION=1 -s LEGACY_VM_SUPPORT=1
-s EXPORTED_FUNCTIONS="['_init', '_init_decompression', '_input_empty', '_get_in_buffer',
'_set_new_input', '_decompress', '_get_out_pos', '_get_out_buffer', '_out_buffer_cleared',
'_release']" -DXZ_USE_CRC64=1 -DXZ_INTERNAL_CRC64=1 *.c -o xzdec.js


mossroy:
Cool, thanks!
Could you also commit the compile string in emscripten/xzdec/compile.js ?

There are a few things that are not clear to me : 
- When you're writing "asm", I suppose you mean "asm.js" (https://en.wikipedia.org/wiki/Asm.js)? "asm" alone is a bit confusing, as it means (to me) https://en.wikipedia.org/wiki/Assembly_language
- I don't understand how asm.js could be "not used", and replaced by "standard interpretation" : asm.js is standard javascript. In our case, it is inside xzdec.js file, that can be run on any javascript engine. As there is no WebAssembly (wasm) in this branch, it's the only way the browser has to do the decompression. Or did I miss something?

Jaifroid:
Ah, sorry about the shorthand. Yes, it's asm.js, as compiled by Emscripten.

Regarding the performance hit with the console open (which is obvious when debugging, at least in older Firefox and in Edge, not sure about Quantum), see https://stackoverflow.com/questions/23453712/typeerror-asm-js-type-error-disabled-by-debugger . 

Maybe it depends on browser architecture, but I think browsers have optimized asm.js interpreters that are very fast but don't provide debug hooks (?). But asm.js can also be interpreted by standard JavaScript Just-In-Time (JIT) interpretation, which is much slower, and at least some browsers fall back to JIT if a debugger is attached to the browser engine. It's really, really obvious in Edge. Asm-js code executes about 3 X slower the moment you open the console. It's a pain for doing any kind of timing. But the Edge Chakra engine has really outstanding performance with asm.js enabled -- it flies through 'Trigonometric Functions' for example.

One thing I didn't mention is that the filesize of this compile of xzdec.js is 57,485 bytes, compared to the previous (old) compile of 201,588 bytes.

Note that compiling without `ALLOW_MEMORY_GROWTH=1` ( https://github.com/kripken/emscripten/blob/master/src/settings.js#L81 )  is what allows full asm optimization - `'use asm'` - as opposed to the `'almost asm'` mode used by the previous xzdec.js. The latter prevents browsers using asm-specific optimizations, since memory expansion is not compatible with those (see https://github.com/kripken/emscripten/issues/6541 ).

Jaifroid:
Note, if testing this PR, I'm getting very good results with the version of it that has a queue in jQuery mode: https://github.com/kiwix/kiwix-js/tree/Add-simple-jQuery-queue . It is built on top of this PR, for comparison of the user experience. Please test that version if possible.

mossroy:
Thanks a lot @Jaifroid for these explanations and this PR.
I've done a simple non-academic performance test, by simply running the unit tests and see how long it takes to run all of them. I've tested on Firefox 62.0.
With master version, it's around 1500ms on my laptop
With this PR (and without debugging console open), it's around 1050 or 1100ms. If I open the console afterward, Firefox says : 
With this PR (and with debugging console open), it's around 1500ms, and says "TypeError: asm.js type error: Disabled by debugger"

So it looks like it's the browser asm.js optimizations that make the difference (and the fact that your compile options allow them with the "use asm").
It's around 30% faster, which is great!
It's the same performance gain I was getting by using wasm version, as you already said in #219.

mossroy:
When asm.js optimizations are on, Firefox still says : 
`Successfully compiled asm.js code (total compilation time 25ms; unable to cache asm.js in synchronous scripts; try loading asm.js via <script async> or createElement('script'))`
So I thought there might be room for more improvements.
I've added the async attribute, and Firefox manages to use the cache.
But, instead of a performance increase, I get a slight decrease : around 1100-1150ms the complete the unit tests.
Firefox says : "Successfully compiled asm.js code (loaded from cache in 96ms)"
So, surprisingly, it seems to take more time to load from the cache than to compile!

I tested on Chromium 69 and did not see a performance difference between the async and non-async versions.

So I won't add this async attribute for now. Plus it would force us to add some code to wait for the script to be loaded, to be sure it's ready before we use it (even if I did not run into any problem on my machine, it might happen on some other)

Jaifroid:
Interesting! I hadn't noticed that Firefox message before. Thanks for the timing info -- I wasn't sure how to do that without console.

mossroy:
I've tested this PR on Firefox OS : it works fine too.
I compared the performance with version 2.3.1 on the following use-case : click on the app icon, and wait for the first display of the main page of wikipedia_fr_all_2016-06.zim.
It takes 12s52 on version 2.3.1, and 10s19 with this PR, which is around 20% faster.
But Firefox OS 2.5 is based on Gecko 44, which might not include all the asm.js optimizations of recent Firefox engines.

I also tested running the "Split article 'A/Ray_Charles.html'" reading unit test alone (on Firefox 62) : on master, it takes around 600ms. With this PR, it's around 240ms, which is 60% faster.

mossroy:
You can squash and merge this PR.
I'll probably test a few other emscripten attributes, to see if we can gain a bit more. Like removing the LEGACY_VM_SUPPORT or switching the wasm (which would force us to handle the fallback for old browsers)
And I'll create another PR if they're worth it.

Jaifroid:
OK, great! To make a good WASM, note that Kripken says you *should* enable memory expansion, as it's beneficial for WASM (but detrimental for ASM) -- see https://github.com/kripken/emscripten/blob/master/src/settings.js#L81. This is one reason why he actually recommends that if we want a fallback from WASM we should handle the fallback ourselves, because there is no single set of build parameters that is optimized for both builds. See discussion with him and Brion from Wikipedia starting here: https://github.com/kripken/emscripten/pull/5837#issuecomment-394200262 .

Issue #412, opened 2018-09-17
Add a queue for extraction of assets


Jaifroid:
As explained in #219 ([here](https://github.com/kiwix/kiwix-js/issues/219#issuecomment-421743277) and [here](https://github.com/kiwix/kiwix-js/issues/219#issuecomment-421786359)), current master delays rendering the page until all the CSS is extracted from the ZIM, but CSS is heavily compressed and is often therefore extracted after all other assets on the page have been extracted. This can lead to long waits for the page to render if a page has many images on it. This is especially the case when there are lots of SVGs (almost any page on a mathematics ZIM). This problem affects jQuery and Service Worker modes. It is ameliorated by caching, but not eliminated. We therefore need a queueing mechanism to control the order of extraction of various assets.

Another argument in favour of queuing would be to allow us to prioritize subcategories of assets, for a better user experience. For example, we could ensure the following extraction order simply by queuing functions to do the following things:

 1. Extract html of page;
 2. Extract all CSS required by page;
 3. Render page; 
 4. Extract all JPEG or PNG images above the fold; 
 5. Extract any SVGs above the fold; 
 6. Extract JavaScript.; 
 7. Extract remaining JPEGs etc; 
 8. Extract remaining SVGs.

I propose one possibility for an extensible queuing framework in https://github.com/kiwix/kiwix-js/tree/Add-simple-jQuery-queue . Currently it only applies to jQuery mode. We could in principle hook into Service Worker requests as they come into app.js and queue responses to create a similar effect in Service Worker mode. It may well be possible to implement a "universal" queuing mechanism in a lower-level process such as zimDirEntry.js.

Before finalizing any framework, and before creating a PR, we need to decide whether the framework should include ServiceWorker mode or only jQuery mode. Your thoughts, @mossroy ?

mossroy:
In ServiceWorker mode, the browser seems to prioritize the requests as we would expect : first the HTML, then all the CSS, then the images that are immediately visible, then the other images. I tested on both Firefox and Chromium.
It makes sense because it's what browsers do when you open an online page. In ServiceWorker mode, we benefit from this behavior for free.
So I don't think we should try to implement this queuing for ServiceWorker mode. And we would have the issue of not knowing if all the requests have been sent to the ServiceWorker (where we can count them in jQuery mode).

[...]

Issue #196, opened 2017-04-05
Switch to ServiceWorker mode by default


mossroy:
The ServiceWorker mode is faster and cleaner.
At startup, we should first try to enable it by default (except if the user has chosen jQuery mode before).
If it fails, we fallback to jQuery mode.

sharun-s:
It would be great if you could add some more detail on the difference between these two modes? I think it will help future contributors to get a better idea of the code and why they exist. Right now given the two different formats zim/evo archive, two different content injection modes (probably a third coming with webrequest), the weird zim format(whose advantages aren't clear), references to firefox os/geo location etc the barrier to entry for a dev to figure out what is going on is a bit high. 

I think this is a fantastic project with lot of potential for all kinds of offline use-cases and I feel you will get more devs involved and exploring those use cases, if the core pieces of this project are well documented. Thanks for all the work you guys do!

mossroy:
Thanks for your feedback @sharun-s 
I think what you mentioned is not only related to jQuery & ServiceWorker modes, so I created #218 to discuss all this. I'll answer in this new issue

Jaifroid:
In #383 @mossroy wrote:

> Regarding the ServiceWorker mode in general, there are many things that should be easier and faster to achieve with it, using the browser capabilities instead of reimplementing things ourself. That's why I'm willing to focus on it for the next hackathon, in order to (hopefully) make it the default mode (when available on the device/browser). It does not mean we would drop jQuery mode support, because it's still necessary in many contexts.

I agree that the Service Worker approach is cleaner, and the logical way to do things. Effectively, we've written an emulation of ServiceWorker in jQuery, which is quite high maintenance. However, I perceive a number of issues with Service Worker still:

* Firefox (even Quantum) still has not (seemingly) enable Service Worker mode as standard. Running Firefox from localhost, on Windows, I am unable to use Service Worker. Perhaps https: is required? Edge and Chromium now work out of the box, running from localhost in http: ;

* No browser allows Service Workers under the file:// protocol. This does not affect browser extensions, of course. But I've always felt that a major advantage of Kiwix JS is that you can simply run it by clicking on index.html. In fact, I've always thought of this as *the* major flexibility and use-case for the app. The simplicity of being able to distribute an offline ZIM reader, with a ZIM, on, say, a thumb drive, is a major advantage of Kiwix JS. We cannot expect users to have servers installed on their machines. I realize this advantage fails with Chrome.

[...]

PR #417, opened 2018-09-18
Display the content immediately after a search in ServiceWorker mode.

### LINKED ISSUE #413: In ServiceWorker mode, render the page before extracting all resources
In ServiceWorker mode, the browser prioritize the requests by itself, just like #412 tries to do in jQuery mode : HTML, then CSS, then visible images, then other images.
But, for some reason, the page does not render before all images are read.
On big pages, it would be a much better user feeling to be able to read the beginning of it, while the rest is read from the ZIM file. Just like browsers do on any online page.

mossroy:
Instead of waiting for the "load" event.
On big articles, it allows some browsers to display the content even if it is not fully loaded (missing images etc).
It can make a big difference in the user perception of performance.
Fixes #413

### DIFF
--- a/www/js/app.js
+++ b/www/js/app.js
@@ -773,16 +773,23 @@ define(['jquery', 'zimArchiveLoader', 'util', 'uiUtil', 'cookies','abstractFiles
      */
     function readArticle(dirEntry) {
         if (contentInjectionMode === 'serviceworker') {
-            // In ServiceWorker mode, we simply set the iframe src and show it when it's ready.
+            // In ServiceWorker mode, we simply set the iframe src.
             // (reading the backend is handled by the ServiceWorker itself)
+            // But we still need to empty the article content first.
+            $('#articleContent').contents().remove();
             var iframeArticleContent = document.getElementById('articleContent');
-            iframeArticleContent.onload = function () {
-                iframeArticleContent.onload = function () {};
-                // Actually display the iframe content
-                $("#readingArticle").hide();
+            iframeArticleContent.onload = function() {
+                // The iframe is empty
+                iframeArticleContent.onload = function () {
+                    // The content is fully loaded by the browser : we can hide the spinner
+                    iframeArticleContent.onload = function () {};
+                    $("#readingArticle").hide();
+                };
+                iframeArticleContent.src = dirEntry.namespace + "/" + dirEntry.url;
+                // Display the iframe content
                 $("#articleContent").show();
             };
-            iframeArticleContent.src = dirEntry.namespace + "/" + dirEntry.url;
+            iframeArticleContent.src = "article.html";
         }
         else {
             // In jQuery mode, we read the article content in the backend and manually insert it in the iframe


Jaifroid:
Great, I'll test this PR shortly on Edge etc.!

Jaifroid:
This PR will cause a regression to #381 (serious performance regression on rendering the index page of Stackoverflow.com) in the Edge browser in Service Worker mode only. I believe it won't cause that regression in Firefox or Chromium, because they appear to wait for the stylesheets to be ready before rendering the page. Edge, however, does not wait, and starts rendering whatever the Service Worker receives as soon as it receives it. With this PR I see (visually) multiple page redraws as the CSS is applied in Edge, and CSS is applied after some images are already downloaded. This means that at least for Edge we would need to prioritize sending CSS to the Service Worker, and ideally not render until CSS has been sent.

On the other hand, this PR provides a clear user optimization for Firefox and Chrome, but a regression for Edge in Service Worker mode only with a minor use case. (It does slightly spoil the experience in Edge, because the page is not really readable while CSS is being applied -- it jumps around.)

Thoughts? Swings and roundabouts!

mossroy:
Interesting. Thanks a lot for catching up so quickly with what I'm doing at the hackathon. It's very valuable to have your feedback and your help.

Regarding Edge in SW mode, does this PR give a significant performance regression? Or "only" a visual annoyance?

Jaifroid:
> Regarding Edge in SW mode, does this PR give a significant performance regression? Or "only" a visual annoyance?
It's the same regression we had with #381, i.e. the little boxes on the Stackoverflow ZIM are redrawn each time a new CSS file is sent to SW, so the browser spends all its time trying to redraw the gazillions of little boxes and the page takes forever to load. It's the reason we prevented render until all CSS was loaded in the first place, to solve #381. The solution via PR #387 (Prevent render until CSS fulfilled) was to count the CSS and render when all of them were ready, but that fix was only in jQuery mode, because SW wasn't affected (until this PR).

It seems we have four possibilities:

1. Not apply this PR;
2. Deem Edge on Service Worker to be an "edge" case (pun intended!) and apply this PR;
3. Apply a version of #387 to Service Worker, i.e. the [Service Worker queue](https://github.com/kiwix/kiwix-js/tree/Service-Worker-queue) branch, or
4. Apply the more comprehensive solution with the [Universal queue](https://github.com/kiwix/kiwix-js/commits/Universal-queue).

Your call! They're not necessarily mutually exclusive options (except 1, perhaps). Others could be seen as interim stages towards a wider solution.

mossroy:
As discussed in #412, I believe adding queuing in SW mode breaks the separation of responsibilities of the layers, and might prevent us from doing refactorings like using the libzim as the backend.
So that would leave us with options 1 or 2.

Jaifroid:
Probably better to go for 2, as it provides the greater good to the greater number of users, and users of Edge are much more likely to be using the app in jQuery mode. However the PR doesn't display the spinner and "Reading article xxxxxx from archive. Please wait...." when clicking on a link in an article. It only does it when searching for a file. Can you add that? It seems inconsistent as it currently is.

mossroy:
I tend to agree.
But some time we'll certainly implement #196 to switch to SW mode by default. hopefully the Edge engine will be improved in between?

Regarding the spinner, I agree that it's not consistent and should be improved.
But, instead of trying to add the spinner everywhere, I would suggest to not display it after a search (only in SW mode). Let me explain : 
In SW mode, the browser makes use of its own way to inform the user that a page is loading. The way it is done depends on the browser brand/version. See https://www.stevesouders.com/blog/2013/06/16/browser-busy-indicators/
I think we should simply rely on that, without trying to add our own way to inform the user.

Plus it would be technically complicated. I only see a few bad solutions : 
- modify the DOM from handleMessageChannelMessage, which looks wrong to me
- parse and modify the HTML string in the ServiceWorker, to inject some javascript code that handles this
- parse and modify the DOM in the load event of the iframe to inject some javascript code that handles this

Jaifroid:
I think this bug (?) in Edge would be a blocker for moving the entire app to SW mode, as the experience is then very unprofessional. Given, however, that there are pragmatic solutions we would be forced to adopt in that case, and given that Service Worker is a marginal use case that is currently warned against in the app, I've got no objection to your proceeding with option 2.

Personally, I would go with a pragmatic solution to #155 (Adding a spinner), because it's just two lines of code. But I completely respect (and learn from, and also greatly benefit from) your more thorough and pure approach! So I leave that decision to you! :-)

mossroy:
Thanks @Jaifroid : I've merged this PR, and created #419 for the spinner inconsistency in SW mode.

PR #422, opened 2018-09-22
Update README, CHANGELOG, and reword warnings on ServiceWorker mode.

### LINKED ISSUE #421: Reword the warnings about ServiceWorker mode
So far, it was unstable so we stated serious warnings to discourage its use, except for testing.
Now it's a bit different : the ServiceWorker mode is more stable.
But there are still a few issues. At least : 
- recent wikimedia ZIM files, when viewed on small screens, have the sections collapsed, with no possibility to open them
- when switching between jQuery and ServiceWorker modes, I still manage to break the UI. Workaround is to reload the page or app. I've created #423 for that

mossroy:
Fixes #421

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@ TODO
 ## Kiwix-JS v2.4.0
 Released on 2018-09-22
 
-Performance improvement on decompression of content (now twice faster)
+Performance improvement on decompression of content (now twice as fast)
 
 Stability improvement for the ServiceWorker mode
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,10 +4,17 @@ Please note that this application has changed its name over time.
 It was first called Evopedia (and was using the file format of Evopedia)
 Then it was renamed Kiwix-html5 (and uses ZIM files), then was renamed Kiwix-JS.
 
-## Kiwix-JS v2.4.0
+## Kiwix-JS v2.5.0
 Released on TODO
 
 TODO
+
+## Kiwix-JS v2.4.0
+Released on 2018-09-22
+
+Performance improvement on decompression of content (now twice faster)
+
+Stability improvement for the ServiceWorker mode
 
 ## Kiwix-JS v2.3.1
 Released on 2018-09-08
--- a/www/index.html
+++ b/www/index.html
@@ -11,10 +11,11 @@
         Home page : http://www.kiwix.org
 
         Main authors of this application :
-        Mossroy - mossroy@mossroy.fr
+        Mossroy - https://github.com/mossroy - mossroy@mossroy.fr
         Peter-x - https://github.com/peter-x
+        Jaifroid - https://github.com/Jaifroid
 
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -4,11 +4,18 @@ Please note that this application has changed its name over time.
...
-## Kiwix-JS v2.4.0
+## Kiwix-JS v2.5.0
 Released on TODO
 
 TODO
 
+## Kiwix-JS v2.4.0
+Released on 2018-09-22
+
+Performance improvement on decompression of content (now twice faster)
```
Jaifroid: > (now twice faster) 

"(now twice as fast)" is better English

www/index.html
```diff
@@ -217,17 +218,16 @@ <h3>Expert settings</h3>
...
                                                 <input type="radio" name="contentInjectionMode" value="jquery" id="jQueryModeRadio" checked>
-                                                <strong>JQuery</strong> (slow and memory hungry, but safer)
+                                                <strong>JQuery</strong> (stable and safer)
                                             </label>
                                         </div>
                                         <div class="radio">
                                             <label>
                                                 <input type="radio" name="contentInjectionMode" value="serviceworker" id="serviceworkerModeRadio">
-                                                <strong>ServiceWorker</strong> (faster but unstable, and not supported by all platforms)
+                                                <strong>ServiceWorker</strong> (still a bit experimental, and not supported by all platforms)
```
Jaifroid: > still a bit experimental, and not supported by all platforms 

English is fine, but maybe could be more concise: "experimental, not supported by all platforms".

www/index.html
```diff
@@ -11,10 +11,11 @@
         Home page : http://www.kiwix.org
 
         Main authors of this application :
-        Mossroy - mossroy@mossroy.fr
+        Mossroy - https://github.com/mossroy - mossroy@mossroy.fr
         Peter-x - https://github.com/peter-x
+        Jaifroid - https://github.com/Jaifroid
```
Jaifroid: Thank you! Very kind of you to add me!

CHANGELOG.md
mossroy: OK thanks, updated

www/index.html
mossroy: To me, the SW mode is now usable in most cases (when available). It's not really "experimental" any more, even if we know a few edge cases that cause problems.
So I'd keep the "still a bit", unless you have a better way to say that?

www/index.html
mossroy: That's normal : I should have done that earlier.
Thanks again for all the work you do on this project.

www/index.html
Jaifroid: @mossroy: OK, so maybe "Somewhat experimental …."? I think "Still a bit" implies that the user knows how it used to be, which is not necessarily the case. But I'm fine with "Still a bit" if you prefer it.

mossroy: OK, updated with "somewhat experimental" and pushed

mossroy:
I've merged this (but forgot to squash before : oops)

PR #423, opened 2018-09-22
Fix broken ServiceWorker when switching between modes.

### LINKED ISSUE #145: Sometimes the ServiceWorker ends up in a state where it does not handle anything more, even if it's registered and enabled
This stucks the application.
I saw that both on Chromium and Firefox dev edition.

Workaround : remove manually the ServiceWorker and reload

mossroy:
This should finally fix #145

### DIFF
--- a/www/js/app.js
+++ b/www/js/app.js
@@ -311,6 +311,8 @@ define(['jquery', 'zimArchiveLoader', 'util', 'uiUtil', 'cookies','abstractFiles
                     refreshAPIStatus();
                 });
             } else {
+                // We need to set this variable earlier else the ServiceWorker does not get reactivated
+                contentInjectionMode = value;
                 initOrKeepAliveServiceWorker();
             }
         }


Jaifroid:
I've tested this on Edge, but don't see any change in behaviour, but then after the stability fix I wasn't getting a broken Service Worker when switching between SW and jQuery when running from localhost.

When running with the **file:// protocol** (which doesn't of course support Service Worker), I still get the following issue at least in Edge: if I click on Service Worker mode, the system allows me to select the mode (giving the usual warning that will be eliminated with #422 ), and **doesn't** then give a warning that the mode is unusable, even though the first API status is orange, and Service Worker remains selected. Now, if I try to load a file (or if a file was already loaded, navigate to a page) it of course doesn't work. If I now go to "Configure" and switch back to jQuery mode, the app gets into an unusable state. I cannot search (no results), the Random button appears to find a page but never displays it, and clicking on Home produces a blank page.

N.B. The above is **not** specific to this PR, so I should report it elsewhere now that we're removing the warning about Service Worker mode, but thought it might be easily fixable if you recognize these symptoms? Will check other browsers shortly.

Jaifroid:
Just tested on Firefox Quantum and Chromium:

* On Chromium, I get exactly the same behaviour as in Edge, i.e. a broken app after switching back from Service Worker mode when using the file:// protocol.

* On Firefox Quantum, it also allows me to turn on Service Worker mode even though the first API status is orange. I cannot load a page of course. However, switching back to jQuery mode allows the app to continue functioning as normal. Is this what you fixed in this PR? If so, the fix only works in Firefox.

Jaifroid:
OK, I didn't notice that instability since the previous SW fix. I'll add my report to #424.

PR #424, opened 2018-09-22
Automatically fallback to jQuery mode if the SW fails to register.

### LINKED ISSUE #405: If the ServiceWorker fails to register, we should automatically fallback to jQuery mode
We already do that if the ServiceWorker API is unavailable.
But when it's available and can not be registered (for example when running from file://), we currently do not do the fallback automatically.

mossroy:
Fixes #405

### DIFF
--- a/www/js/app.js
+++ b/www/js/app.js
@@ -309,6 +309,9 @@ define(['jquery', 'zimArchiveLoader', 'util', 'uiUtil', 'cookies','abstractFiles
                 }, function (err) {
                     console.error('error while registering serviceWorker', err);
                     refreshAPIStatus();
+                    alert("The ServiceWorker could not be properly registered. Switching back to jQuery mode. Error message : " + err);
+                    setContentInjectionMode("jquery");
+                    return;
                 });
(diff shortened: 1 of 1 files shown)


Jaifroid:
As reported in #423, here are the results of my testing this issue:

Just tested on Firefox Quantum and Chromium:

* On Chromium and Edge, I get a broken and unusable app after switching to Service Worker mode when using the file:// protocol. I am allowed to switch to this mode, even thought the first API status is orange, and the mode remains selected, even thought the API remains orange. Switching back to jQuery mode does not correct the situation -- the app remains unusable until I reload.

* On Firefox Quantum, it also allows me to turn on Service Worker mode in file:// protocol, even though the first API status is orange, and it remains selected despite not working. However, switching back to jQuery mode allows the app to continue functioning as normal.

Jaifroid:
Apologies, sorry, I'm confuse about where I'm reporting. I hadn't realized this was a **new** PR. Will test properly and get back to you!! Please ignore above for now.

Issue #426, opened 2018-09-27
Prevent redundant extraction of assets after navigation


Jaifroid:
[ Issue previously reported in #425 ]

Kiwix JS continues to extract assets for a given page (CSS, images, JavaScript) that have been queued up for extraction even after the user has navigated away from that page. This issue affects both jQuery and Service Worker modes. On short and simple pages, extraction is now so quick that it is not noticeable. But on longer pages with hundreds of images, and especially on pages with SVG images that are heavily compressed, it can significantly delay either the display of the next page or the extraction of assets from it. The problem is noticeable particularly when using Kiwix JS with, e.g., `wikipedia_en_maths_novid_2018-06.zim` where almost every article has many SVG images of equations.

Jaifroid:
[ Comment moved from #425 ]
I've just done a quick test on Edge and Firefox, with a couple of console log entries in zimarchive.js to log the files extracted. Basically, if we've sent images to be extracted or decompressed in `loadImagesJQuery()` or have sent them to the backend in SW, the engine will continue to work on them, even after the next page is loaded. We don't have a "stop" mechanism (and maybe it would be good to put one in).

The test was done on master (not on the spinner branch) by loading full English Wikipedia, letting the landing page settle, then searching for "Paris". After a few seconds of letting "Paris" load, I quickly press the Home button to take me back to the landing page. You can see the console log entries below (in Edge, but exactly the same in FF). I've highlighted where the html of each page was extracted. Visually, the landing page was visible in the browser, while in the console log the images from the Paris page continued to be extracted for several seconds. (Having Console open usefully slows things down in this case!) I need to do more tests in SW mode, because there are so many console messages it's a little hard to pin things down, but I've just reproduced the issue in FF a few times (I should make a separate issue to report this problem).

In Kiwix JS Windows, I use "lazy loading" of images: the visible area plus the next x images are loaded, then no more, so the issue is minimized (JQuery only, since mobile doesn't support SW). But it would be good to work out a "stop" signal to send to the backend? 

![image](https://user-images.githubusercontent.com/4304337/46085372-ce4d5180-c19d-11e8-8229-b99ed64e2c14.png)

Jaifroid:
[ Comment moved from #425 ]

I pushed a branch to test the issue of redundant image calls after navigation. I'll move these posts to a new issue if you think it *is* an issue. The branch is:

https://github.com/kiwix/kiwix-js/tree/Test-redundant-image-loads-after-navigation 

[...]

Issue #425, opened 2018-09-26
Use subtler CSS spinner and make article transitions smoother


Jaifroid:
I've been thinking about the article loading experience **in jQuery mode**, as it is currently a little "clunky", with flashes to white screen, and an old-fashioned gif spinner. I've had a go at implementing a new concept, using a subtler CSS spinner in overlay over the existing article (if there is one), and keeping the article displayed until the new HTML is ready to be injected. There's a screenshot of the spinner over a page from FFOS below, and the branch is:

https://github.com/kiwix/kiwix-js/tree/Use-a-CSS-spinner-overlay 

This is just a prototype, as I didn't want to do any more work on it until I had your opinion. In particular, I haven't done anything specific to Service Worker mode (though the new spinner does display where we currently use it in SW mode). And I haven't yet gone rigorously through the different scenarios and transitions. I haven't yet done anything with the "Caching CSS" message (it still pops up occasionally in its old spot).

One feature I've tried to work out here is keeping the page display in place until it needs to be replaced, and using just the new spinner as the main visual queue that something is happening. To me, it feels a lot smoother like this, even if there remain some inconsistencies that need to be worked through. When you've had a chance to test it in jQuery mode, could I have your opinion, @mossroy ?

![image](https://user-images.githubusercontent.com/4304337/46070643-38ea9700-c176-11e8-9f63-0ffb06cf244e.png)

Jaifroid:
Great!

On the point about images continuing to load, I'm afraid they do on our current master branch in both jQuery mode and SW. I've just done a quick test on Edge and Firefox, with a couple of console log entries in zimarchive.js to log the files extracted. Basically, if we've sent images to be extracted or decompressed in `loadImagesJQuery()` or have sent them to the backend in SW, the engine will continue to work on them, even after the next page is loaded. We don't have a "stop" mechanism (and maybe it would be good to put one in).

[...]

Issue #78, opened 2014-04-01
Simplify the source code for hiding/showing of HTML elements


mossroy:
It's becoming a bit complicated to maintain all these .hide() and .show()

Jaifroid:
There is a lot of repeated code, whether it's jQuery hide and show statements or native JS equivalents. These hide or show UI elements when the Home button is clicked, when the Random button is clicked, when a search is launched, when the user goes to Configuration or About. A lot of the code is the same and could be tidied away into non-jQuery functions.

PR #427, opened 2018-10-08
Add a modern CSS spinner

Jaifroid:
@mossroy I've taken the CSS spinner code as far as I can without hooking into the Service Worker messaging channel. So it works fully in jQuery mode, and in a limited way in Service Worker mode: when loading a new page by searching for it, with the Random button, or when loading the landing page, but not when clicking on a link within a page. The latter would require hooking into the messaging channel, and so would require a policy decision if we want uniformity across the app. It can be left for future decision.

Regarding your comment in #425 that keeping the old page on-screen until we have the new html to replace it might cause the app to keep requesting images unnecessarily, I have done extensive testing in #426 : there is a broader issue affecting master, not this PR specifically, so I have opened a separate issue for it. 

Could you please check this code and test? I'd be grateful if you could test in particular that #361  ("TypeError: can't access dead objects") is not reintroduced. It shouldn't be, as we still have the specific fix for that in place. There is also a related query embedded in a comment in app.js line 748 addressed to you, which I'll remove when the query is resolved. No rush at all!

### DIFF
--- a/www/css/app.css
+++ b/www/css/app.css
@@ -60,11 +60,17 @@
     padding-left: 1px !important;
 }
 
-#searchingForArticles .floating {
-    position: relative;
-    z-index: 10;
-    top: 20px;
-    height: 0px; 
+#articleList {
+    margin-bottom: 12px !important;
+}
+
+#articleListWithHeader {
+    position: absolute; 
+    top: 0; 
+    left: 5%;
+    padding-left: 12px;
+    padding-right: 12px;
+    background: lightblue;
 }
 
 .articleIFrame {
@@ -82,6 +88,45 @@
 
 .apiUnavailable {
     color : orange;
+}
+
+.loader {
+    position: absolute;
+    margin: 0 auto;
+    left: 0;
+    right: 0;
+    border: 5px solid lightgrey;
+    -webkit-animation: spin 1s linear infinite;
+    animation: spin 1s linear infinite;
+    border-top: 5px solid steelblue;
+    border-radius: 50%;
+    width: 50px;
+    height: 50px;
+}
+
+.status {
+    position: absolute;
+    top: 50%;
+    width: 100%;
+    z-index: 1;
+}
+
+.message {
+    position: absolute;
+    margin: 70px auto 0;
+    left: 0;
+    right: 0;
+    font-size: smaller;
+    background: whitesmoke;
+    color: dimgrey;
+    border: 1px solid gainsboro;
+    padding: 1px 4px;
+    width: 200px;
+}
+
+@keyframes spin {
+    0% { transform: rotate(0deg); }
+    100% { transform: rotate(360deg); }
 }
 
 /* Reduce the size of some elements for small screens */
--- a/www/index.html
+++ b/www/index.html
@@ -44,6 +44,13 @@
         <link href="css/bootstrap-theme.css" rel="stylesheet" media="screen"/>
     </head>
     <body role="application">
+        <!-- Status indicators -->
+        <div id="searchingArticles" style="display: none;" class="status">
+            <div class="loader"></div>
+            <div id="cachingCSS" style="display: none;" class="message">
+                Caching styles...
+            </div>
+        </div>
         <section id="search-article" role="region">
             <header id="top">
                 <nav class="navbar navbar-default" role="navigation">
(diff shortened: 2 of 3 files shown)


Jaifroid:
Thanks @mossroy : 

- I'll experiment with a darker colour -- I agree it's sometimes hard to notice;
- I was having trouble finding a good place for "Caching styles..."; it's not so necessary now that xzdec is much faster, though could still be useful on low-end devices. I'll experiment with what it looks like placing it over the spinner, otherwise use more noticeable style;
- Good catch about the scrolling -- I think it's not necessary, will check;
- I think the only purpose of `$('#articleContent').contents().remove();` would be to tell jQuery to remove its hooks, so I think in SW mode it only needs to be performed once when the user switches to SW mode after having used jQuery mode -- I'll look for a place to relocate it in the switching code.

Jaifroid:
Hi @mossroy , if you get a moment, could you give me an opinion on the (visual) implementation now?

- More visible foreground and background colours for spinner;
- In jQuery mode, Caching styles... message now centred beneath spinner;
- Caching styles... message is updated with abbreviated name of each css file as it is extracted.

Note that the caching message(s) will only show in jQuery mode, as the cache isn't implemented for SW mode. Also the caching message will only be displayed on first load of the landing page, and very rarely on a page that needs a new CSS file that has not yet been cached.

Specifically, does the caching message strike a good balance between being discreet but still being noticeable/informative? Is it too obtrusive? Would it be better relocated to bottom-right or top-right corners? (We can't use bottom left because many browsers use that area to show the URL of a link that has been clicked.) 

An example over blank background shown below.

image

Jaifroid:
Great, thanks! I can certainly test adding a line to the onclick code to activate the spinner sooner for these edge cases. I think it can be part of this PR.

I've still got code clearing-up to do and one small thing to fix from earlier discussion. I'll ask you to look over final code when it's ready (probably at weekend) before merging.

Jaifroid:
@mossroy I've completed the pending issues (see commit titles). Could you test your slow I/O files with this final code? The spinner should display on click of UI elements in both modes, and inline article links in jQuery mode.

If you get a moment to look through the code, it would be great. Quite a lot of redundant hiding and showing has been deleted.

There is one thing that might need decision: I got rid of the "Reading article xxxxx ...." message before I implemented the css caching info box beneath the spinner, on the basis that users know what they've clicked on, and the spinner is enough feedback. However, it would now be possible to use the css caching info box to present that message. I tend to think it's still not necessary, and displaying the message on each page load would mean that we would always see the box. Currently it is only shown when there is new CSS to cache, so only for the first couple of page loads usually. I feel it's neater just to use the spinner as feedback. The "loss" of feedback is mitigated by the fact that the article list stays in place now until the page is ready, and when a user clicks on an inline link, browsers usually display the link in their status message bottom left.

However, I thought I should point this out, since it is removing a feature that exists in master.

mossroy:
I tested with the same slow I/Os (on Firefox 60ESR and IE11, both in jQuery mode) : the user feedback is much better now, with the spinner appearing right after the click.
A small regression : the spinner does not seem to appear any more when clicking on "Random" button.

Regarding the "loss" of "Reading article xxx", I don't have a strong opinion. We can probably leave it this way : it's still possible to add it back later if we feel it's necessary. In a sense, it's also more consistent with the SW mode.

I had a quick look at the code : I like to see some code removal in this hiding/showing code :-), see #78.

I did not test in SW mode, but if you did, you might squash/rebase/merge this PR.

Jaifroid:
Thanks for spotting the regression. It was a bit "random", and a puzzle, because I can't see the rogue `hide()` statement that turns off the spinner. It's clearly shown at the very beginning of the `goToRandomArticle()` function, and nothing in the chain should turn it off until the html is received and all CSS is loaded. I wonder if there is a manifestation of #426 here: a delayed CSS file from a previous page comes through async and triggers turning off the spinner (because we have a tendency when testing to mash the Random button (at least I do), and because I moved the `show()` statement to before the dirEntry is sourced, it gets turned off before we read the file. In any case, I've fixed it with an extra `show()` statement. But it feels like a brute force kludge... Hmmm.

Jaifroid:
Found another regression in this PR with FFOS, when clicking on a search result, we get a severely truncated display area once the new page is shown, at least in the simulator. It's to do with display of the on-screen keyboard, I guess. I need to remove the article list before that code is called...

It's never easy!!

Jaifroid:
OK, [this commit](https://github.com/kiwix/kiwix-js/pull/427/commits/a363a93ea9644c068901c8368ae3fdf9968f8bef) solves that one. Again, feels a bit like brute force, but it does make senes to call `resizeIFrame()` after we render CSS.

I'm going to leave this till tomorrow before squashing etc. @mossroy , I don't know if you want to test on a real FFOS device first? If not, it does seem to work in the simulator. Anyway, I aim to merge early tomorrow morning unless I hear to the contrary.

Issue #428, opened 2018-10-13
Clicking on footnote/endnote references reloads the page in latest Wikipedia English novid ZIM


Jaifroid:
I'm not sure if this is a bug with our code, or whether it is a problem with the format of the latest English Wikipedia file `wikipedia_en_all_novid_2018-06.zim`. In this file (I haven't tested the non-novid version), links to footnote references appear like this:

`href="Ottoman_Empire.html#cite_note-:0-13"`

In other words, the link contains the url of the page. This causes our anchor parsing routine to miss the fact that this is an anchor reference, and instead the page is re-loaded. The link is not followed, and the footnote is not shown.

The "problem" (if it isn't a bad encode) is in this regex:

`var regexpLocalAnchorHref = /^#/;`

It clearly does not match the above href. We can work around with relocating this to the anchor parsing routine like this (roughly):
```
var regexpLocalAnchorHref = new RegExp('^(?:#|' + dirEntry.url + '#)([^#]+$)');
var anchorRef = href.replace(regexpLocalAnchorHref, '$1');
document.getElementById('articleContent').contentWindow.location.hash = anchorRef;
```
Should we work around? I guess this ZIM file is a popular one that will be around for a long time.

mossroy:
This probably affects jQuery mode, but not SW mode, I suppose?
In any case, it's probably worth working on it, as it does not seem very complicated to solve.
Be careful on the RegExp that dirEntry.url might contain special characters that would be interpreted by the RegExp (like #?*+ etc)

Jaifroid:
I confirm that the issue only affects jQuery mode. In SW mode, the page is not reloaded and the browser jumps to the selected footnote. A regex like the one above should be fine with query strings, percent encodes, and other special characters, but I believe `#` is a reserved delimiter, so can only appear once (legally) in a URL. In any case this regex will only match either a URL starting with `#` or a URL starting with the current page's own URL immediately followed by `#`, so it should be OK.

Jaifroid:
Contrary to the above comment, I've discovered that SW mode is affected by this issue, but in a lesser way. To reproduce, load `wikipedia_en_all_novid_2018-06.zim` in master (SW mode), and search for "Presidential $1 Coin Program" (or any other page with a $ or other percent-encoded character in the URL). Click on a footnote, and the page reloads. However, in SW mode, the browser does jump to the footnote after page reload. It causes all images to reload, etc.

Jaifroid:
[...]

PR #429, opened 2018-10-14
Support local anchors with URL

Jaifroid:
This PR addresses #428 . I think it will need quite wide testing. I've tested on pages such as `$1,000_Reward.html` which has a footnote href encoded as `%241%2C000_Reward.html#cite_note-TH-1`, as well as pages with parentheses in the URL, in Edge, but it needs wider testing before merging, on both the full English Wikipedia and other ZIMs (especially ZIMs with subdirectories, which I haven't tested yet).

### DIFF
--- a/www/js/app.js
+++ b/www/js/app.js
@@ -741,7 +741,7 @@ define(['jquery', 'zimArchiveLoader', 'util', 'uiUtil', 'cookies','abstractFiles
                     iframeArticleContent.onload = function () {};
                     $("#searchingArticles").hide();
                 };
-                iframeArticleContent.src = dirEntry.namespace + "/" + dirEntry.url;
+                iframeArticleContent.src = dirEntry.namespace + "/" + encodeURIComponent(dirEntry.url);
                 // Display the iframe content
                 $("#articleContent").show();
             };
@@ -814,8 +814,6 @@ define(['jquery', 'zimArchiveLoader', 'util', 'uiUtil', 'cookies','abstractFiles
     var regexpPath = /^(.*\/)[^\/]+$/;
     // Pattern to find a ZIM URL (with its namespace) - see http://www.openzim.org/wiki/ZIM_file_format#Namespaces
     var regexpZIMUrlWithNamespace = /(?:^|\/)([-ABIJMUVWX]\/.+)/;
-    // Pattern to match a local anchor in a href
-    var regexpLocalAnchorHref = /^#/;
     // Regex below finds images, scripts and stylesheets with ZIM-type metadata and image namespaces [kiwix-js #378]
     // It first searches for <img, <script, or <link, then scans forward to find, on a word boundary, either src=["'] 
     // OR href=["'] (ignoring any extra whitespace), and it then tests everything up to the next ["'] against a pattern that
@@ -881,10 +879,15 @@ define(['jquery', 'zimArchiveLoader', 'util', 'uiUtil', 'cookies','abstractFiles
         function parseAnchorsJQuery() {
             var currentProtocol = location.protocol;
             var currentHost = location.host;
+            // Percent-encode dirEntry.url and add regex escape character \ to the RegExp special characters - see https://www.regular-expressions.info/characters.html;
+            // NB dirEntry.url can also contain path separator / in some ZIMs (Stackexchange). } and ] do not need to be escaped as they have no meaning on their own. 
+            var escapedUrl = encodeURIComponent(dirEntry.url).replace(/([\\$^.|?*+\/()[{])/g, '\\$1');
+            // Pattern to match a local anchor in an href even if prefixed by escaped url
+            var regexpLocalAnchorHref = new RegExp('^(?:#|' + escapedUrl + '#)([^#]+$)');
             $('#articleContent').contents().find('body').find('a').each(function() {
-                var href = $(this).attr("href");
+                var href = $(this).attr('href');
                 // Compute current link's url (with its namespace), if applicable
-                var zimUrl = regexpZIMUrlWithNamespace.test(this.href) ? this.href.match(regexpZIMUrlWithNamespace)[1] : "";
+                var zimUrl = regexpZIMUrlWithNamespace.test(this.href) ? this.href.match(regexpZIMUrlWithNamespace)[1] : '';
                 if (href === null || href === undefined) {
                     // No href attribute
                 }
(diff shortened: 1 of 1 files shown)


mossroy:
This seems to work fine, both in jQuery and SW mode.
In jQuery mode, I did not test every possible character combination but the principle looks good to me. Is it normal that you espace the character "{", but not "}"? Same thing for "[" and "]".
Maybe you could simply use the code from https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions#Using_special_characters (at the end of the paragraph)?

These special characters seem to sometimes create bugs when ZIM files are generated by mwoffliner.
For example, in wikipedia_fr_all_novid_2018-03.zim, if you search for "? (album de Nena)", the cite notes point to namespace I instead of A (and are not working, even on Kiwix desktop)

Jaifroid:
According to one of the Regex Gurus (maker of Regex Buddy, Jan Goeyvaerts) there are 12 special characters in major Regex engines (including JavaScript flavours) -- see: https://www.regular-expressions.info/characters.html .

The reason I, and more importantly he, doesn't include `]` and `}` is because if the Regex engine encounters these characters without a corresponding opening one first, then they are treated as literal. If we had a string with one of these as an opening character, it would be escaped by my regex, so would not in fact open the pair. That makes sense, since Regex engines are tuned to the minimal interpretable syntax. This doesn't work for `)` because the possibility of embedding parentheses would make that ambiguous. (Incidentally, no backslash is needed for special characters inside `[ … ]` character class list except for the backslash itself and of course `^` cannot be the first character or it negates the character class.)

What this code does (ignoring the percent-encoding) is to turn a (fictional) string like

`$1_Coin_Program[me]_(2005).html` 

into 

`\$1_Coin_Program\[me]_\(2005\)\.html` 

… which is what is needed to construct a valid Regex with this string in it. I am pretty sure the other characters in the MDN list are all combinations with characters that are already escaped by my regex, so the combination will never be parsed as such, but I will double check.

Regarding the mwoffliner bug, I reported it, and it was marked as fixed in March: https://github.com/openzim/mwoffliner/issues/205 . If you can reproduce with a more recent ZIM than March, it would be well worth re-opening (I think I've seen it in a recent WikiMed, but need to find that again). I guess the issue is caused by a regex in mwoffliner not using a complete list of special characters.

Jaifroid:
@mossroy Please note that I've rebased this branch/PR, and so you may need to reset your local copy of the branch (if you have one).

As a point of information, is it *ever* possible that `dirEntry.url` could contain a path, e.g. `/my/path/An_article.html`? I think the answer is no, but if there is a remote possibility, I need to include `\/` in the character list of the regex that parses the `dirEntry.url`. Or should I include it for the sake of completeness, in case anyone needs to generalize and use the pattern elsewhere in the future?

Jaifroid:
I confirm the problem with question marks in footnote references, seemingly an issue with mwoffliner. In the article `Boy_or_Girl?.html` in `wikipedia_en_all_novid_2018-06.zim`, the footnote anchor href appears as:

`../I/Boy_or_Girl%3F.html#cite_note-1`

Whereas the target is:

`<li id="cite_note-1">`

@mossroy , I could work around this issue by matching the href starting from the first character after the last `/` (effectively ignoring the path). Should I do this? And will you report the issue to mwoffliner? I think it's probably a new issue there, but it would be good to reference https://github.com/openzim/mwoffliner/issues/205 .

Jaifroid:
PS For jQuery mode only, the workaround would simply be to have `(?:^|.*/)` instead of `^` at the beginning of the programmatically constructed regex. So, instead of

`var regexpLocalAnchorHref = new RegExp('^(?:#|' + escapedUrl + '#)([^#]+$)');`

we just need:

`var regexpLocalAnchorHref = new RegExp('(?:^|.*/)(?:#|' + escapedUrl + '#)([^#]+$)');`

(Tested with `Boy_or_Girl?.html`.) For SW mode, we would need to parse/alter the html to fix this, though it could be done inside `serviceworker.js`.

mossroy:
All right, thanks for the explanations about your regexp! Maybe it would be worth putting a comment in the code to explain that.

Regarding the wrong hyperlinks in cite notes, it is indeed probably a bug in mwoffliner.
In this case, I'm not in favor of trying to put a workaround in kiwix-js, either in jQuery or SW mode. It's a very minor issue that should be fixed in mwoffliner.
If you have time to create the issue on mwoffliner, please do. Else I'll try to do it later.

Regarding a dirEntry.url containing a path, I suppose it is the case for stackoverflow ZIM files? But maybe they don't have cite notes...

Jaifroid:
Thanks, I've added the comments. Let me know if you're happy with them (I included the URL that gives a complete explanation).

You're right about Stackexchange ZIMs, I just checked. I'd misremembered and thought there was a separate "path" property of dirEntry, but in fact the path is in `dirEntry.url`. Therefore I have included escaping of `/`, otherwise we could construct invalid regexes that might cause a crash.

Fine about not working around the mwoffliner issue. I've reverted it. I'd be grateful if you could report the issue to mwoffliner, as it adds a bit of weight to have more voices. It is in fact a completely different issue from https://github.com/openzim/mwoffliner/issues/205. This issue seems like a problem with an unescaped `?` somewhere in an `if...else if...else if...` clause. Probably mwoffliner recognizes file types by the extension `.html`, but the `?` in the url causes a match to fail, triggering a default to set filetype as image. Just a guess.

Jaifroid:
@mossroy , I'm not sure if this is a separate issue, or we can pop in a correction under this PR (as it deals with the same subroutine). When attempting to read the article `Immunotherapy` in the latest WikiMed, the app crashed with a malformed URI in an anchor href:

`<a class="external text" href="http://DOI:%2010.3923/pjbs.2018.135.150">DOI: 10.3923/pjbs.2018.135.150</a>`  

The problem is `http://DOI:%2010.3923/pjbs.2018.135.150`, which is wrong. According to https://en.wikipedia.org/wiki/Digital_object_identifier it should either be `DOI:10.3923/pjbs.2018.135.150` or `https://doi.org/10.3923/pjbs.2018.135.150`, but not the hybrid we have here. Quite rightly the browser (at least Edge, didn't test in others) complains and leaves host and protocol undefined, causing a crash for us.

There is a simple solution, which is to use jQuery, which prevents undefined and allows the page to load correctly. Since we're already extensively using jQuery in this page, it's a tiny adjustment, but I've added a comment. Can we include this here? It's commit https://github.com/kiwix/kiwix-js/pull/429/commits/233138358233c2a68a0508c28fcb832fefac25f8 so you can check it out. It's almost a style correction. But I can make a new issue if you prefer.

mossroy:
I've created https://github.com/openzim/mwoffliner/issues/412 for the first issue.
I did not manage to reproduce your second issue (with DOI links) : in wikipedia_en_medicine_novid_2018-10.zim, article Immunotherapy, the DOI links look good to me.

Jaifroid:
> I did not manage to reproduce your second issue (with DOI links) : in wikipedia_en_medicine_novid_2018-10.zim
Apologies, the article was "Cancer Immunotherapy", and it was from the September version of WikiMed. You can see the invalid DOI in the first footnote of this old version of the article from September: https://en.wikipedia.org/w/index.php?title=Cancer_immunotherapy&direction=prev&oldid=859467602 . In the latest version of the page, the reference has been corrected, so you won't see it in the October ZIM. I was doubting my sanity trying to find this again, but I'm glad Wikipedia keeps its history!!! In any case, even though it was transient, it serves to discover a problem in our code with malformed hrefs. I'll make an issue for it.

PR #431, opened 2018-10-21
Add API for reading ZIM Metadata.

### LINKED ISSUE #395: Add support of ZIM meta information reading

mossroy:
Fixes #395
It is especially useful to read the language of the ZIM file


Jaifroid:
I've tested this with the following test code in Edge:

`selectedArchive.getMetadata("Language", function(code) { console.log("Language = " + code); });`

It returns the code successfully to the console, and doesn't cause any errors when placing it inside `displayArticleContentInIframe` (for example). So I think this all looks good and can be merged.

PR #432, opened 2018-10-24
Prevent app crash with malformed hrefs

Jaifroid:
This problem is a bit tricky. In essence, the malformed href causes a crash on any attempt to access any property of the href. This means we have two choices: use jQuery, with its in-built error handling, or else we have to use try … catch. So I'm posting two commits here, one with jQuery and the second with try … catch. Both solve the issue. @mossroy , if you prefer the jQuery solution, or a variant of it, I'll revert the second commit. Although generally I prefer not to use jQuery when it's unnecessary, here I think it provides a tangible benefit, so I'm inclined towards the jQuery solution.

### DIFF
--- a/www/js/app.js
+++ b/www/js/app.js
@@ -884,39 +884,42 @@ define(['jquery', 'zimArchiveLoader', 'util', 'uiUtil', 'cookies','abstractFiles
             var escapedUrl = encodeURIComponent(dirEntry.url).replace(/([\\$^.|?*+\/()[{])/g, '\\$1');
             // Pattern to match a local anchor in an href even if prefixed by escaped url
             var regexpLocalAnchorHref = new RegExp('^(?:#|' + escapedUrl + '#)([^#]+$)');
-            $('#articleContent').contents().find('body').find('a').each(function() {
-                var href = $(this).attr('href');
-                // Compute current link's url (with its namespace), if applicable
-                var zimUrl = regexpZIMUrlWithNamespace.test(this.href) ? this.href.match(regexpZIMUrlWithNamespace)[1] : '';
-                if (href === null || href === undefined) {
-                    // No href attribute
+            $('#articleContent').contents().find('body').find('a').each(function () {
+                // Attempts to access any properties of 'this' with malformed URLs causes app crash in Edge/UWP [kiwix-js #430]
+                try {
+                    var testHref = this.href;
+                } catch (err) {
+                    console.error("Malformed href caused error:" + err.message);
+                    return;
                 }
-                else if (href.length === 0) {
+                var href = this.getAttribute('href');
+                if (href === null || href === undefined) return;
+                // Compute current link's url (with its namespace), if applicable
+                // NB We need to access 'this.href' here because, unlike 'this.getAttribute("href")', it contains the fully qualified URL [kiwix-js #432]
+                var zimUrl = regexpZIMUrlWithNamespace.test(this.href) ? this.href.match(regexpZIMUrlWithNamespace)[1] : '';
+                if (href.length === 0) {
                     // It's a link with an empty href, pointing to the current page.
                     // Because of the base tag, we need to modify it
-                    $(this).on('click', function(e) {
-                       return false; 
+                    $(this).on('click', function (e) {
+                        return false;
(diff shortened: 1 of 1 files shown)


Jaifroid:
Don't bother downloading the September WikiMed, because I've now discovered the crash only occurs in Edge and in UWP (and apologies for not having tested more widely before -- I was more preoccupied with fixing it). Firefox and Chromium don't have a problem with the malformed URL, even though it doesn't yield a valid URL. I think it's still worth doing the jQuery solution, because it's generic. However, I need to find out at what point we changed to `this.href`, and why -- it seems a bit odd that we have a variable, `href`, which has just been initialized with the URL using jQuery, and then not to use it. We can't use `this.href` in the jQuery solution, because any attempt to access the properties of `this` causes a crash in Edge and in UWP (which uses the Edge engine).

My suspicion is that we started to rewrite some of these functions without jQuery when we got the problem with #361 . For a short while, we started substituting pure JS in these routines, then we reverted that when we found another fix, but maybe not fully? However, I'll locate the spot in the commit history where it changed (if it's not too far in the remote past!), before proceeding.

Jaifroid:
The use of `this.href` instead of `$(this).attr("href")` was introduced in this commit https://github.com/kiwix/kiwix-js/commit/4a933eb554913c01bda825334b3fa0152dbd1cf9#diff-bca463ed8ccf5e2058dfe307a95e8501R826 . It seems to be the case that you used that syntax simply because you'd deleted `var url = $(this).attr("href");` (as it was then). It doesn't look particularly significant, as my proposed simplification/revert to the already defined variable `href` https://github.com/kiwix/kiwix-js/pull/432/commits/b09959cb3a9b1772a2d0ade8f0906bd48c0b6ce6#diff-bca463ed8ccf5e2058dfe307a95e8501R890 would not make a substantial difference except that jQuery works around the browser quirk here (probably simply by catching exceptions). Do you agree, @mossroy , or does it prompt you to remember a reason for writing the code this way? In principle, the jQuery code merely means: get the attribute "href" of the currently selected anchor. If you agree, I'll use the jQuery solution, as `try...catch` is ugly.

Jaifroid:
PS Version of the historical commit without yellow highlighting: https://github.com/kiwix/kiwix-js/commit/4a933eb554913c01bda825334b3fa0152dbd1cf9 (line 826).

Jaifroid:
Will do :-)

Jaifroid:
@mossroy : Just to let you know your memory was correct, and there is a difference between what is returned by `this.href` and what is returned by jQuery. The former returns, e.g. `file:///C:/Users/xxxx/Source/Repos/kiwix-js/www/A/question/84/how-important-are-accents-in-written-spanish.html`, while the latter returns `question/84/how-important-are-accents-in-written-spanish.html` for the same href. This accounts for your having made that change in a commit entitled "Handle subdirectories in article namespace", where you need the href to show the protocol.

I'm looking at the best options in the light of this. Thanks for your caution!

mossroy:
Nice work! I just regret that I did not add a comment in the code to explain that.
Could you please add it in your PR (if this line of code is finally kept)?

Jaifroid:
I'll definitely add a comment to explain.

Just FYI, in case you're curious, here's what I found out:

* jQuery `$(this).attr('href')` = JS `this.getAttribute('href')` -- both of these merely return the content of the href attribute of the anchor link;
* jQuery `$(this).prop('href')` = JS `this.href` -- both of these get the fully qualified URL that the browser has worked out by combining base href with the protocol and the anchor link's href.

Unfortunately, jQuery's `$(this).prop` syntax, which is the one we would need here, doesn't do any error checking, and the crash in Edge/UWP happens the same as with `this.href`, `this.protocol`, `this.host`, etc.

Therefore we're back to Option 2, using `try...catch`.

I'll revert the revert and test more.

Jaifroid:
Thanks @mossroy , I ran out of time this week, and I need to do a bit more testing of this PR in different scenarios. I aim to do that at the weekend, after which I'll squash/merge if I find no more issues with the code. It's interesting how what I thought at first would be a "little fix" turned out to have far more complications than I'd realized!

Issue #339, opened 2018-01-31
PhET zim not read on Kiwix-JS


Popolechien:
I basically get the splash screen and that's it - no way to access content (either Kiwix JS windows or browser extension). See image:
![phet](https://user-images.githubusercontent.com/21675885/35639344-2cc641fa-0677-11e8-8fe2-4a03e0915a20.png)

mossroy:
This is a known issue. Kiwix-js does not currently inject the javascript files at all.
It was causing issues in the browser engine (but it was a long time ago, so maybe this could be tested again).
See #152 .
Anyway, I don't know if enabling javascript would be enough to make PhET files work. There might be other contents that are not well supported by kiwix-js.

Popolechien:
Ok thanks -  I'd seen #152 but wasn't entirely sure it was the same problem.

mossroy:
To be honest, I'm not 100% sure either. But I don't see how these PhET experiments would work without a bunch of javascript code, so I did not investigate more.

mossroy:
Javascript is now enabled in ServiceWorker mode. PhET seems to work properly in this mode.

kelson42:
@mossroy I have tried with last nightly and https://download.kiwix.org/zim/phet/phet_fr_2018-11.zim. The welcome page is empty = not displayed properly to me. I'm in Jquery mode as service worker does not work out-of-the-box on FF.

Jaifroid:
@kelson42 In master, only SW mode currently supports PhET. I have an experimental branch, [JavaScript Support](https://github.com/kiwix/kiwix-js/tree/javaScript-support) that include JS support in jQuery mode. However, the front page of PhET is not currently working in that branch in jQuery mode. If you press the random button in that branch, you'll see that a random experiment will load and is usable. But I need to debug the landing page. It's probably because it uses ReactJS (I think), and that involves a lot of inline JavaScript, which my branch tries to put into attached scripts so that a page will work with as restrictive CSP. I may have to ditch that workaround.

mossroy:
In ServiceWorker mode, the latest gutenberg seems to work well too, and also various ZIMs that have videos (tedx, oer4schools, dirtybiology etc, even if moving to a specific time in the video does not work on all browsers)

Jaifroid:
[...]

PR #434, opened 2018-11-11
Enable javascript injection in ServiceWorker mode

mossroy:
See #152 : it implements this issue, but only in ServiceWorker mode.

### DIFF
--- a/service-worker.js
+++ b/service-worker.js
@@ -127,19 +127,6 @@ function fetchEventListener(event) {
                     console.log("It's a layout dependency : " + title);
                     if (regexpJS.test(title)) {
                         contentType = 'text/javascript';
-                        var responseInit = {
-                            status: 200,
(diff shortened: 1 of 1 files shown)


mossroy:
It also seems to fix #339 in ServiceWorker mode

Jaifroid:
I've tested this branch in Edge. It works fine in Wikimedia ZIMs I tested. And, the good news is that it now enables use of PhET ZIMs. I was able to load one of the simulations in a recent PhET ZIM! See screenshot below... I think this amply proves the robustness of the code! (Note that there is an issue accessing the Load button for the PhET simulations, because the buttons are hidden under our bottom bar, but that is another issue: I'll make an issue for it.)

image

Jaifroid:
PS I'm not sure why one of the travis checks is failing.

Issue #172, opened 2016-06-21
Display video content of ZIM files


mossroy:
They are currently not handled.
It might cause difficulties because of their size, either in jQuery or ServiceWorker mode.

Jaifroid:
As mentioned by @mossroy in #339, videos are now working in SW mode. I believe adding support for them in jQuery mode should be a matter of:

* Parsing and transforming links of the following type in `parseAnchorsJQuery`: "../../I/Free-software-free-society-Richard-Stallman-at-TEDxGeneva-2014/video.webm";

* Extracting the video file to a BLOB, probably in advance of rendering the page (?) so that it can be inserted into the above link, or else finding a way to extract it when the user clicks play -- the latter would be difficult, as we don't have an equivalent of the Fetch intercept in jQuery mode;

* Enabling JavaScript support in jQuery mode (#152 ) so that the video player will work - this part is more-or-less solved.

One issue is whether the video data are currently steamed to SW, or whether the engine is extracting the data and posting the entire byte array to the SW via postmessage. I think it's the latter, as I don't see anything that handles streaming in the backend, but I may be wrong. I assume we can't easily emulate streaming in jQuery mode, but we can extract the video BLOB.

@mossroy , can you think of anything else?

Jaifroid:
It looks like we might be able to play "directly" the data steam in the Uint8Array using the [MediaSource interface](https://developer.mozilla.org/en-US/docs/Web/API/MediaSource) of the Media Source Extensions API. It allows you to replace the media source of an html5 video player with a JavaScript function, and in particular it accepts the output of the function as an array buffer, which should be perfect for us. It might save us the overhead of having to convert the entire data stream of a very large video file (potentially) into a blob.

I'm a bit out of my depth here, however... There is a code example on the msdn page, in which it looks like we might be able to substitute the `fetchAB` function with our own `readBinaryFile` function's output for the video source. I guess we just have to try it... There's a working demo of the code on that page [here](http://nickdesaulniers.github.io/netfix/demo/bufferAll.html).

Jaifroid:
[...]

Issue #440, opened 2018-11-20
Read dirEntry titles correctly in PhET and TED ZIMs


Jaifroid:
Experiments in PhET ZIMs contain dirEntries in the A article namespace, but our search algorithm is, for some reason, failing to retrieve the titles when searching for articles. This can be seen in the screenshot below. I have checked that the dirEntry for `Etats de la matière`, for example, contains the correct title in the dirEntry title field, but typing "E" or "e" in our search bar finds nothing. I believe the problem is with our search, not with the ZIM format.

Since Kiwix JS in jQuery mode can run experiments, but cannot currently run the proprietary User Interface that these ZIMs have (coded in React JSX), supporting search in these ZIMs would make the ZIMs useable by Kiwix JS clients that can only use jQuery mode (this includes Kiwix JS Windows currently), as well as for clients that cannot use inline JavaScript due to restrictive CSPs (basically clients running in extensions as currently configured at least).

![image](https://user-images.githubusercontent.com/4304337/48761035-892a3500-ec9f-11e8-96cf-4908992ec06d.png)

Jaifroid:
I've changed the title, because I've discovered this issue also in `ted_en_global_issues_2018-07.zim` (see bottom screenshot below). The title of this talk is "How we can bring mental health support to refugees" but nothing can be found in our search index under H or h. This is in current master in Service Worker mode. In fact the proprietary interface is the only way to access these videos in the ZIM. Again, the dirEntry appears to be fine, but we don't appear to be able to read it:

![image](https://user-images.githubusercontent.com/4304337/48971361-515d1d80-f010-11e8-9d25-5a829dbaa4bc.png)

![image](https://user-images.githubusercontent.com/4304337/48971284-7309d500-f00f-11e8-9f75-35c8aa6be656.png)

Jaifroid:
@mossroy I've debugged this, and the problem is a false assumption made in [this line](https://github.com/kiwix/kiwix-js/blob/master/www/js/lib/zimArchive.js#L185) in `zimArchive.js` (the first `if` test below):

![image](https://user-images.githubusercontent.com/4304337/48971636-41473d00-f014-11e8-80ed-5a216a50431a.png)

It seems that in these ZIMs, the assumption that the ZIM places empty titles at the end, is false. If I delete this test, and leave only the other two, then the result of searching for the letter "H" is:

[...]

PR #441, opened 2018-11-23
Enable media playback in JQuery mode

Jaifroid:
The evolution of this code is tracked in #172 . I believe the issues I identified there have been solved.

The code is generalized to work with `<video>` and `<audio>` blocks, together with any `<source>` tags contained within them. Hence I call it "media playback" rather than just "video". I have tested with video and audio provided in different ZIMs, but note that I did not find any ZIM using pure `audio` blocks -- the scrapers seem to put any audio in video blocks (and it plays fine like that). Details of some pages to test with in #172.

Playback works in browsers without needing JavaScript-in-the-ZIM to be enabled, although sometimes getting to the content is a little challenging without JS (but basically our standard search usually works). I have tested the code with FFOS (works in simulator, actual device will depend on available codecs), Edge, Chromium, Firefox ESR, IE11 (no webm support), and UWP (works in Windows 10, no webm codec in Windows 10 Moible). For the browsers which don't support webm, it seems to degrade gracefully, indicating that there is no support.

One issue with TEDx ZIMs is that they contain a proprietary UI that uses extensive JS, including having links of the type `<a href="javascript:doSomeProprietaryFunction();">Go to top</a>`. Or else `<body onload="raw javascript">`. These cause an exception when clicked, or in the worst case, cause an app crash on loading the page, and they won't work even in SW mode in most extensions. But this is a different issue. I just thought it should be signalled with regard to testing this PR.

### DIFF
--- a/www/js/app.js
+++ b/www/js/app.js
@@ -809,12 +809,12 @@ define(['jquery', 'zimArchiveLoader', 'util', 'uiUtil', 'cookies','abstractFiles
     var regexpPath = /^(.*\/)[^\/]+$/;
     // Pattern to find a ZIM URL (with its namespace) - see http://www.openzim.org/wiki/ZIM_file_format#Namespaces
     var regexpZIMUrlWithNamespace = /(?:^|\/)([-ABIJMUVWX]\/.+)/;
-    // Regex below finds images, scripts and stylesheets with ZIM-type metadata and image namespaces [kiwix-js #378]
-    // It first searches for <img, <script, or <link, then scans forward to find, on a word boundary, either src=["'] 
+    // Regex below finds images, scripts, stylesheets and media sources with ZIM-type metadata and image namespaces [kiwix-js #378]
+    // It first searches for <img, <script, <link, etc., then scans forward to find, on a word boundary, either src=["'] 
     // OR href=["'] (ignoring any extra whitespace), and it then tests everything up to the next ["'] against a pattern that
     // matches ZIM URLs with namespaces [-I] ("-" = metadata or "I" = image). Finally it removes the relative or absolute path. 
     // DEV: If you want to support more namespaces, add them to the END of the character set [-I] (not to the beginning) 
-    var regexpTagsWithZimUrl = /(<(?:img|script|link)\s+[^>]*?\b)(?:src|href)(\s*=\s*["']\s*)(?:\.\.\/|\/)+([-I]\/[^"']*)/ig;
+    var regexpTagsWithZimUrl = /(<(?:img|script|link|video|audio|source|track)\s+[^>]*?\b)(?:src|href)(\s*=\s*["']\s*)(?:\.\.\/|\/)+([-I]\/[^"']*)/ig;
     
     // Cache for CSS styles contained in ZIM.
     // It significantly speeds up subsequent page display. See kiwix-js issue #335
@@ -828,7 +828,7 @@ define(['jquery', 'zimArchiveLoader', 'util', 'uiUtil', 'cookies','abstractFiles
      * @param {String} htmlArticle
      */
     function displayArticleContentInIframe(dirEntry, htmlArticle) {
-        // Replaces ZIM-style URLs of img, script and link tags with a data-url to prevent 404 errors [kiwix-js #272 #376]
+        // Replaces ZIM-style URLs of img, script, link and media tags with a data-url to prevent 404 errors [kiwix-js #272 #376]
         // This replacement also processes the URL to remove the path so that the URL is ready for subsequent jQuery functions
         htmlArticle = htmlArticle.replace(regexpTagsWithZimUrl, "$1data-kiwixurl$2$3");            
 
@@ -866,6 +866,7 @@ define(['jquery', 'zimArchiveLoader', 'util', 'uiUtil', 'cookies','abstractFiles
             loadCSSJQuery();
             //JavaScript loading currently disabled
             //loadJavaScriptJQuery();            
+            insertMediaBlobsJQuery();
         };
      
         // Load the blank article to clear the iframe (NB iframe onload event runs *after* this)
@@ -1027,6 +1028,38 @@ define(['jquery', 'zimArchiveLoader', 'util', 'uiUtil', 'cookies','abstractFiles
                     }
                 }).fail(function (e) {
                     console.error("could not find DirEntry for javascript : " + title, e);
+                });
+            });
+        }
+
+        function insertMediaBlobsJQuery() {
+            var iframe = iframeArticleContent.contentDocument;
+            Array.prototype.slice.call(iframe.querySelectorAll('video[data-kiwixurl], audio[data-kiwixurl], source[data-kiwixurl], track[data-kiwixurl]'))
+            .forEach(function(mediaSource) {
+                var source = mediaSource.dataset.kiwixurl;
+                if (!source || !regexpZIMUrlWithNamespace.test(source)) {
+                    console.error('No usable media source was found!');
+                    return;
+                }
+                var mediaElement = /audio|video/i.test(mediaSource.tagName) ? mediaSource : mediaSource.parentElement;
+                var mimeType = mediaSource.type;
+                // Check mimeType
+                if (!mimeType) {
+                    // Try to guess type from file extension
+                    var mediaType = mediaElement.tagName.toLowerCase();
+                    if (!/audio|video/i.test(mediaType)) mediaType = 'video';
+                    if (/track/i.test(mediaSource.tagName)) mediaType = 'text';
+                    mimeType = source.replace(/^.*\.([^.]+)$/, mediaType + '/$1');
+                }
+                selectedArchive.getDirEntryByTitle(decodeURIComponent(source)).then(function(dirEntry) {
+                    return selectedArchive.readBinaryFile(dirEntry, function (fileDirEntry, mediaArray) {
+                        var blob = new Blob([mediaArray], { type: mimeType });
+                        mediaSource.src = URL.createObjectURL(blob);
+                        // In Firefox and Chromium it is necessary to re-register the inserted media source
+                        // but do not reload for text tracks (closed captions / subtitles)
+                        if (/track/i.test(mediaSource.tagName)) return;
+                        mediaElement.load();
+                    });
                 });
             });
         }


www/js/app.js
```diff
@@ -814,12 +814,12 @@ define(['jquery', 'zimArchiveLoader', 'util', 'uiUtil', 'cookies','abstractFiles
...
     var regexpZIMUrlWithNamespace = /(?:^|\/)([-ABIJMUVWX]\/.+)/;
-    // Regex below finds images, scripts and stylesheets with ZIM-type metadata and image namespaces [kiwix-js #378]
+    // Regex below finds images, scripts, stylesheets and media sources with ZIM-type metadata and image namespaces [kiwix-js #378]
     // It first searches for <img, <script, or <link, then scans forward to find, on a word boundary, either src=["'] 
     // OR href=["'] (ignoring any extra whitespace), and it then tests everything up to the next ["'] against a pattern that
     // matches ZIM URLs with namespaces [-I] ("-" = metadata or "I" = image). Finally it removes the relative or absolute path. 
     // DEV: If you want to support more namespaces, add them to the END of the character set [-I] (not to the beginning) 
-    var regexpTagsWithZimUrl = /(<(?:img|script|link)\s+[^>]*?\b)(?:src|href)(\s*=\s*["']\s*)(?:\.\.\/|\/)+([-I]\/[^"']*)/ig;
+    var regexpTagsWithZimUrl = /(<(?:img|script|link|video|audio|source)\s+[^>]*?\b)(?:src|href)(\s*=\s*["']\s*)(?:\.\.\/|\/)+([-I]\/[^"']*)/ig;
```
Jaifroid: Just a quick explanation here: all video and audio I've encountered have been in the `I` namespace. The spec doesn't mention video or other media, and no specific namespace seems to exist for it. There are two image namespaces, with the second one being `J`, but the detailed specs are missing for image handling, so it's not clear when if ever `J` is used. I have never encountered it, and it seems to be reserved for image text (?).

mossroy:
Thanks a lot, it looks great!
I only did a quick test on a Firefox OS device, with dirtybiology ZIM file, and it worked perfectly!
At first sight, the code looks clean too.

I'll try to have a closer look, but I really lack some time those days.

Jaifroid:
Hmm, I've just finished downloading Dirtybiology, the latest ZIM dated 2018-06, and I'm getting in both Edge and Firefox, "Video format or MIME type is not supported". The type of videos in this ZIM is "video/mp4" rather than webm. Now mp4 is definitely supported in Windows, though I guess the specific encoding could be a problem with my system, though I doubt it. I'll investigate...

mossroy:
I was using an old ZIM of dirtybiology (2015-11) : I'll try later with 2018-06

Jaifroid:
OK, the videos on the sub-pages of that ZIM work, it's the video on the front page that doesn't. Looking at it in detail, it doesn't appear to have a source. It looks like it might be loaded dynamically in JavaScript.

But mp4 videos on other pages in the ZIM seem to work OK on my system.

To summarize, I think this is a "false alarm" based on the fact that the front page of the ZIM probably populates the video block with dynamic info. Maybe a random video, or "video du jour" type of thing...

Jaifroid:
I'm experimenting with not using the DataView method of converting the uint8array to blob, even though it's theoretically faster. The reason is that a) it blocks Internet Explorer needlessly (I'm not sure why, because DataView is supposed to be fully supported from IE10); and b) I'm not actually convinced it offers a more efficient access to the underlying data in our case. In fact, I can't see any of the theoretical advantages from my quick tests -- the blob is loaded just as fast with either direct use of the uint8array or a DataView of its buffer.

IE11 is able to play the videos in the recent TEDx ZIM `ted_en_global_issues_2018-07.zim`, because they are in `video/mp4` format, so it's worth continuing support, I think, if it doesn't offer any disadvantages to the other browsers.

Jaifroid:
There is a problem when there are too many subtitle tracks for a video (e.g. in TEDx ZIMs). The screen grab below is from latest global issues TEDx ZIM in Firefox, after pressing the captions button in the player.

Only the subtitles from Danish to Russian can be selected, and the others are impossible to click. In Edge, only the top five subtitles are shown, and the others are also impossible to reach. Furthermore, subtitle files are highly compressed and it takes a very long time to decompress all of these (I've monitored them -- they carry on decompressing for at least a minute after loading the page). I've been thinking that all this decompression is a bit useless, when at most the viewer will want one of the subtitle languages. 

@mossroy @kelson42 , what would you advise in this situation? I could add a little box beneath the video,, or in a bootstrap modal at the bottom fo the screen, allowing the user to select which subtitle file to decompress, and insert just that one as the top track only when the user asks for it. It would make the requested language accessible and would save a lot of unnecessary decompression. But is this intervening in a problem that should be fixed at ZIM level? Should I make an issue somewhere for this?

image

Jaifroid:
I've rebased this PR on master for ease of testing media-based ZIMs which couldn't be searched without #440 #443.

Jaifroid:
There is what I believe is a specific problem with accessing subtitles in `tedxgeneva-2014_fr_all_2015-03.zim`. The subtitles are definitely in the ZIM (you can see them if you press the space bar in our search box), but do not seem to be properly referenced, which means that they can't be found, either in this branch in jQuery mode, or in master in SW mode. The problem can be seen in the page's source code below. It looks like the subtitles are referenced as if the page referencing them were inside the subdirectory of the video.

Should we try to work around this, or put it down to a bad encode of this ZIM?

Subsequent TEDx ZIMs I've tested don't have this problem, but I've only tested the global issues ZIM.

```
<video class="video-js vjs-default-skin" controls preload="auto" width="480" height="270" 
    data-setup='{"autoplay": true, "preload": "true"}'>
    <source src="../../I/Free-software-free-society-Richard-Stallman-at-TEDxGeneva-2014/video.webm" type="video/webm" />
    <track kind="subtitles" src="en.vtt" srclang="en" label="English" />
    <track kind="subtitles" src="es-ES.vtt" srclang="es-ES" label="" />
    <track kind="subtitles" src="hu.vtt" srclang="hu" label="Hungarian" />
</video>
```

kelson42:
@Jaifroid @mossroy I would not recommend to workaround the problem. If there is a bug please report it. If it is done in a way that it ask for too much of decompression, please report it too (but to me it sounds really strange).

Jaifroid:
Thanks @kelson42 . The issue with decompression of lots of vtt files is just the usual one: decompressing text-based assets (html, js, css, svg, vtt) is "costly" and slow for our app, whereas extracting assets that are stored in the ZIM with no additional compression (jpeg, png, webm, mp4, etc.) is fast. So, when we load a video page from TEDx that has 30+ subtitle assets, it takes a lot of CPU to decompress them all. You can really feel a mobile device heating up, for example, and even on a high-spec laptop it can cause the fan to kick in. Since the user only needs one of the subtitle files, I was trying to think of ways to prevent unnecessary decompression.

I think the display bug shown in screenshot above is a Firefox issue.

Jaifroid:
Some more info on events associated with text tracks here: 

https://stackoverflow.com/questions/50778257/texttracklist-onchange-event-not-working-in-ie-and-edge 

The suggested workaround doesn't work (I tested) because we don't have a valid subtitle URL (until we extract the file from the ZIM).

Jaifroid:
So I've pushed a custom text track selection dropdown to this PR for testing. This is what it looks like visually:

image

Jaifroid:
@mossroy I assumed that would only fire when a track is added programmatically to the video block. The tracks are already in the video DOM on page load/render, so they're not being added. However, I'll test to see, in case my assumption was wrong.

mossroy:
I would make the same assumption, but am not sure

Jaifroid:
Unfortunately, the `addtrack` event of the video element's `textTracks` object, like the `onerror` event, fires on DOM load in Firefox for each of the subtitle tracks. In both Edge and FF it doesn't fire on attempting to select a different subtitle, at least one with an invalid src (I tested both with a ZIM src and with no src / data-kiwixurl).

image

Jaifroid:
OK, we might be able to work around with a combination of an `onchange` event on the `textTracks` object in Firefox and Chromium, and an `onerror` event for Edge. I had previously tested FF for a change event of the video element itself, but trying again now with the change event of the textTracks *object*, it fires in both FF and Chromium on attempting to load a subtitle file, even though the src is invalid. In Edge, it doesn't fire (because the src is invalid), but the `onerror` event fires instead. In IE11, no event fires, though the textTracks object does exist.

Clearly, Edge doesn't attempt to load the asset until the user selects it, and then reports an error instead of a change. Firefox and Chromium both report the error on DOM load, which suggests they might be trying to load the assets up-front. Difference in loading strategy seems to account for the difference in events here.

The difficulty in working out a combination of onerror and onchange is that we'd need some browser detection, to prevent FF and Chromium from reacting to the onerror event, since it fires for each track on DOM load. And IE would be left with no support for subtitles.

mossroy:
If I sum up the situation (please correct me if I'm wrong) : 
- You managed to make videos work in jQuery mode (on all browsers, even if some video formats are not supported by IE)
- You managed to make subtitles work in jQuery mode too (on all browsers), by loading all of them on page load. But it can be slow when there are many, because subtitles are compressed, and our current ZIM backend is not very fast
- To optimize performance, you are testing 2 different ways : 
    - modify the page to add a drop-down for the user to select the subtitle. This way, you can read only the selected subtitle
    - use a combination of onchange and onerror events to read only the selected subtitle

Maybe this could be split into 2 different PRs : a first one which enables videos and subtitles (without optimization), and another one to test some optimizations.
The first one might be merged, and give some time to work on optimizations.

Regarding optimizations, my first reflex would be to work on #116 instead (compile libzim with emscripten). I know it will take more time to implement that, but it should improve performance and features of our ZIM files reading. I feel that our time would be better invested to work on that.

Anyway, if you prefer to work on this optimization (which I would respect), I would prefer the second way (trying to use the events sent by the browser), because it looks more generic, and can not break the layout of the page.
I don't think leaving IE11 behind, on that topic, is a problem. It does not support webm (which is AFAIK the most common video codec used in ZIM files), and is not a priority IMHO.

But I think we must not make some code that depends on the browser brand. Because we can not know how these browsers will evolve in the future. And it's not unusual that this kind of detection fails to do what we want : for example how should Firefox for iOS be detected? (It's a browser made by Mozilla, but that does not use the Gecko engine). See https://css-tricks.com/browser-detection-is-bad/
Anyway, I think we can make an exception for Internet Explorer : we know for sure that this browser will not evolve, so we might detect it in order to use a fallback mode for this browser (like injecting all the subtitles).

Jaifroid:
That's a very clear and correct summary of the situation @mossroy . I would quite like to be able to release a version of Kiwix JS (and Kiwix JS Windows) that can claim to support TEDx ZIMs, so it might be best to revert this PR to the version that extracts all the subtitles, and then (after any further tests) merge it. That will then provide base support on all browsers.

I would also like to continue working on a solution (under a new issue / PR) for preventing extraction of all subtitles, because I think it's very close, and if I stop now, I'll probably forget what has already been tested. Besides, working on #116 just throws more power at the problem, and doesn't prevent the problematic behaviour of unnecessarily extracting 30 subtitle files when the user only needs one (which is actually a "bug"). From what I can tell, in SW mode the `video.js` plugin is used to manage video playback rather than the native widget, although that has its own problems (I can't seem to select subtitles at all in SW mode in Chromium in the latest TEDx global issues ZIM).

I completely agree that browser detection is bad. I'll think of how we could use capability detection instead. That actually could be quite easy with IE (as it seems to lack certain methods of the textTracks object).

Thank you, this exchange has been very helpful and I can see a way forward.

Jaifroid:
@mossroy This PR is now ready for your review (no rush, when you have time). I've reverted to the version that extracts all subtitle files after video load, and have cleaned up the code a little. It's tested and working in Edge, IE11, FF ESR and Chromium with `ted_en_global_issues_2018-07.zim`. In FFOS I tested video playback, which is fine, but not subtitle display (because I haven't loaded one of the newer properly subtitled ZIMs onto the fake SD card). I can't see any reason why it wouldn't work on FFOS, so long as the HTML5 widget in FFOS has subtitle / CC support.

Jaifroid:
> But there does not seem to be any GUI to choose between subtitles (or disable them). In dirtybiology_fr_all_2018-06.zim, it does not display any subtitle (maybe the platform only displays the default subtitle).

It would be very easy to have an option on the Configuration page: Set default subtitle language: xxxx. Then we could mark one of the languages as default and it might play on clients that don't have an interface for choosing subtitles.

> L’élément n’a pas d’attribut « src ». Le chargement de la ressource média a échoué. article.html
Le chargement de toutes les ressources possibles a échoué. Le chargement du média a été arrêté.

I've seen the equivalent message in English. It's because we've changed `src` to `kiwix-dataurl`. We do this to prevent 404s, but clearly something checks here that the src is missing. We'd probably get a failure message if we left the src in place also. The only way to avoid it would be to pre-extract the video before rendering the DOM, but I feel that would cause too long a delay.

> On dirtybiology_fr_all_2018-06.zim, the home page displays a black video zone with no content, but I suppose this comes from the lack of javascript support?

Yes, these are dynamic videos that are supposed to be filled with some (possibly random, or possibly "vidéo du jour") content. We could, in principle, put a random video in there, but the code would not be generic and would have to be checked on each ZIM update. I guess we'd be better off trying to get #152 [JavaScript in the ZIM] working acceptably.

mossroy:
Thanks for these answers. So I would have left this PR as it is : don't do more work to comply with Firefox OS, or to avoid these error logs, or anything specific to some ZIM files.

PR #444, opened 2018-12-02
Update CI to deal with new download.kiwix.org server

kelson42:
We have move `download.kiwix.org` server to a new server. A few things have changes:
* the CI user is now `ci` instead of `nightlybot`
* the path is now `/data/download/` instead of `/var/www/download.kiwix.org/`

Here is a patch to get Kiwix JS nightlies uploaded again (this has stopped to work during this week).


kelson42:
@mossroy Yes, should be OK. Thx.

Issue #446, opened 2018-12-06
Support relative asset URLs in the Article namespace


Jaifroid:
It appears that it may be valid for a text track to have an `src` URL that is relative to the directory containing the media source. While I can't find precise info on this, mdn has this description of the `src` attribute of the `<text>` element:

> **src** Address of the track (.vtt file). Must be a valid URL. This attribute must be specified and its URL value must have the same origin as the document — unless the <audio> or <video> parent element of the track element has a crossorigin attribute.

The subtitle `src` attributes in the TEDx Geneva 2014 ZIM are all relative `src`'s of type `src="en.vtt"`. I have seen one or two other examples of code on the Web that uses this format. But what has really convinced me that this is valid is that both Firefox and Edge "expand" the src correctly. If I programmatically read the src, I get a correct URL that I can then transform into a URL that can be read from the ZIM. So it seems that the video DOM establishes the location of the media source, and then allows other URLs to be relative to that.

This is easy to fix in our current code (precisely because the browsers expand the `src` correctly).

EDIT: Chromium also expands the `src` correctly.

Jaifroid:
I've realized this is a larger issue, so I've changed the title to "Support relative asset URLs". It is also pertinent to the way we recognize which image URLs need to be extracted from the ZIM. It's not difficult to fix in a completely generic way, and I will shortly add a PR.

Jaifroid:
I've changed this issue title to "Support relative asset URLs in the Article namespace", because I think the only assets that could be relative to an article's location are those that are in the Article namespace. And the only assets I'm aware of that are in the A namespace are subtitles / closed captions.

PR #447, opened 2018-12-07
Recognize relative asset URLs

Jaifroid:
This PR addresses #446 . It works by adding any relative asset URLs to the regex we have that turns ZIM-style URLs (src or href) into data-kiwixurl. This then allows our existing functions for both images and videos to find these URLs without any further modification of those functions.

### DIFF
--- a/www/js/app.js
+++ b/www/js/app.js
@@ -812,9 +812,9 @@ define(['jquery', 'zimArchiveLoader', 'util', 'uiUtil', 'cookies','abstractFiles
     // Regex below finds images, scripts, stylesheets and media sources with ZIM-type metadata and image namespaces [kiwix-js #378]
     // or with URLs that are relative to the current article.
     // It first searches for <img, <script, <link, etc., then scans forward to find, on a word boundary, either src=["'] 
-    // OR href=["'] (ignoring any extra whitespace), and it then tests everything up to the next ["'] against either a pattern 
-    // that matches ZIM URLs with namespaces [-IJ] ("-" = metadata or "I" / "J" = image), or (if that fails) a pattern that matches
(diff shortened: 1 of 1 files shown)


Jaifroid:
A bit more info:

This PR works out the full URL from a relative URL in the regex function, using our `baseUrl`. It effectively transforms all relative URLs into full ZIM URLs ready for processing by the jQuery functions and places those URLs in a `data-kiwixurl` attribute. This is just one way of doing it. 

The other way of doing it, would be to ignore relative URLs in this regex (which is what we've done up till now), and instead, in the image and media insertion functions, re-examine every asset to see if it contains a relative url/src, and if so, use the URL that the browser has worked out for itself from the `baseUrl`. The advantage of this method is that it lets the browser do its job and the regex is a bit less complicated. The disadvantage is that we could still end up with some relative assets showing 404s or registering failures, and we have to process every image/media asset again in the respective function.

Swings and roundabouts. Let me know what you think would be best.

Jaifroid:
Thinking about this PR, I think the only assets that could be relative to an article's URL would be assets in the A namespace, so I've changed the issue #446 title to reflect this. The only assets in the A namespace I'm aware of are some ZIM subtitiles. In some ZIMs, these are in the A namespace, whereas in other ZIMs they are in the - (metadata) namespace.

So now I'm not sure whether the solution to this is best handled in this regex (which would make it generic, in case other asset types in the future could ever be in the A namespace) or in the insert media function.

Jaifroid:
I'm closing this PR because I'm now 99% convinced that #448 is the better way to do this, and that the issue only occurs with text tracks in some ZIMs.

PR #451, opened 2018-12-11
Interpret noscript tags in jQuery mode

### LINKED ISSUE #433: In jQuery mode, inject noscript tags
We currently do not support javascript in jQuery mode.
ZIM files generated by mwoffliner in early 2018 rely on javascript to open the sections. That's why we had to implement a dirty workaround : https://github.com/kiwix/kiwix-js/blob/master/www/js/app.js#L960 (see #355).
This has been very recently fixed in mwoffliner : https://github.com/openzim/mwoffliner/issues/324
So future ZIM files generated by mwoffliner should be able to work on kiwix-js in jQuery mode with a more generic and reliable fix, as there is now a `<noscript>` tag in the HTML that opens the sections.

So we should add some code that inject the content of `<noscript>` tags (if any).
A simple way to do that would be to remove these tags in the HTML before injecting it in the iframe. But we need to check if there's a risk to remove actual content (like in a https://developer.mozilla.org/en-US/docs/Web/API/CDATASection content).
Another way to do it is to inject the content while parsing the DOM. It would allow a safe detection of these tags. In theory, it would slow down the content display (and imply some redraw), but we already wait for the CSS to be injected before actually displaying, so it should not add some redraw.

Should we keep the dirty workaround? I unfortunately think so, in order to be still compatible with the ZIM files generated in 2018.
But we should disable it when testing with noscript, and add some comments to explain why we still keep it (it will look redundant)

mossroy:
This should fix #433, but it's not ready to be merged yet.

In this commit, I commented the original workaround for section toggling, so that you can test the injection of noscript tags on recent wikimedia ZIM files (I tested with the last one one of https://github.com/openzim/mwoffliner/issues/324).
It's a generic way to handle noscript tags, that might be useful for any kind of ZIM file.

However, I think we will still need to keep the original workaround too, to keep compatibility with ZIM files generated in early 2018.
So, before merging this PR, I'll uncomment this original workaround (and add a comment above to explain why we keep it). Both pieces of code seem to play well together.


www/js/app.js
```diff
@@ -949,18 +952,25 @@ define(['jquery', 'zimArchiveLoader', 'util', 'uiUtil', 'cookies','abstractFiles
                 });
             });
         }
+        
+        function loadNoScriptTags() {
+            // For each noscript tag, we replace it with its content, so that the browser interprets it
+            $('#articleContent').contents().find('noscript').replaceWith(function () {
+                return this.textContent;
```
Jaifroid: Is there a reason to use `textContent` rather than `innerHTML`? Do we want to interpret the HTML or filter it out in some way? It's a genuine question, because I don't know what normally goes inside noscript tags, or quite what the difference is between `textContent`, `innerText`, jQuery's `text`, etc.

Jaifroid:
I saw https://developer.mozilla.org/en-US/docs/Web/API/Node/textContent#Differences_from_innerHTML which explains the difference between `innerHTML` and `textContent`. From the example given on that page, it seems `textContent` removes HTML, and returns just the text, getting rid of things like `<span>` and `<div>`. We probably want to keep those, right?

mossroy:
Yes, this can be confusing.
From what I understood : 
- if javascript is enabled, the noscript tag is understood by the browser as containing only text. See https://html.spec.whatwg.org/multipage/scripting.html#the-noscript-element
- as a consequence, textContent and innerHTML should both give the right text content. textContent is the standard one (supported by all modern browsers, including IE11. See https://caniuse.com/#feat=textcontent), innerHTML was a proprietary implementation of Microsoft, that you still need to use if you need to support IE<=8 (which is not our case)

Maybe I should add some comments in the code?

Jaifroid:
Let's take the example that is given on the whatwg.org page you linked:
```
 <noscript>
  <input type=submit value="Calculate Square">
 </noscript>
```
Are you saying that `this.textContent` and `this.innerHTML` will both return an empty string instead of `<input type=submit value="Calculate Square">` when JS is turned on in the browser (which it must be of course)? If so, that's not going to be useful to us, is it? As we need the html contained within the noscript tags to be exposed so that the browser will interpret it, even though JS is turned on. Sorry if I've misunderstood -- I've never really used noscript!

mossroy:
In your example, `this.textContent` should return the String `<input type=submit value="Calculate Square">`, which is what we need.
I've created https://jsfiddle.net/25bx6smy/1/ as a test-case : could you test on Edge?

Jaifroid:
Great! That works in Edge. Thanks for the text case!
Given that I was confused by this, maybe it would be a good idea to add a comment to explain that the browser interprets any content inside the noscript tags, even html, as if it were text when JS is turned on in the browser. Just something brief like that.

mossroy:
I just pushed a comment about the use of .textContent, and uncommented the original workaround.

mossroy:
@Jaifroid : if it's OK for you, I'll merge this PR on wednesday.

Jaifroid:
Apologies, yes the comment is fine, please merge.

PR #455, opened 2018-12-19
Fix video seeking issue in SW mode on Chrome and Edge.

### LINKED ISSUE #452: Chrome can not seek inside videos in ServiceWorker mode
At least on TED ZIM files.
When you click on a position in the video, it restarts the video from the beginning, instead of jumping to that position.

It works fine on Firefox (both in jQuery and SW modes). It also works fine on Chrome in jQuery mode.

mossroy:
This is a quick and dirty attempt.
It seems to fix #452 but is not ready to be merged yet.
It looks like the `'Accept-Ranges': 'bytes'` made it work.

Could you test it on Edge?

NB : it also seems to fix the subtitles issue

### DIFF
--- a/service-worker.js
+++ b/service-worker.js
@@ -76,6 +76,10 @@ var regexpPNG = new RegExp(/\.png$/i);
 var regexpJS = new RegExp(/\.js/i);
 var regexpCSS = new RegExp(/\.css$/i);
 var regexpSVG = new RegExp(/\.svg$/i);
+var regexpWEBM = new RegExp(/\.webm$/i);
+var regexpMP4 = new RegExp(/\.mp4$/i);
+var regexpOGG = new RegExp(/\.og[mvg]$/i);
+var regexpVTT = new RegExp(/\.vtt$/i);
 
 // Pattern for ZIM file namespace - see http://www.openzim.org/wiki/ZIM_file_format#Namespaces
 var regexpZIMUrlWithNamespace = new RegExp(/(?:^|\/)([-ABIJMUVWX])\/(.+)/);
@@ -97,11 +101,17 @@ function fetchEventListener(event) {
                 // The namespace defines the type of content. See http://www.openzim.org/wiki/ZIM_file_format#Namespaces
                 // TODO : read the contentType from the ZIM file instead of hard-coding it here
                 if (nameSpace === 'A') {
-                    // It's an article
-                    contentType = 'text/html';
+                    if (regexpVTT.test(title)) {
+                        // It's a subtitle
+                        contentType = 'text/vtt';
+                    }
+                    else {
+                        // It's an article
+                        contentType = 'text/html';
+                    }
                 }
                 else if (nameSpace === 'I' || nameSpace === 'J') {
-                    // It's an image
+                    // It's an image or another kind of media
                     if (regexpJPEG.test(title)) {
                         contentType = 'image/jpeg';
                     }
@@ -111,6 +121,15 @@ function fetchEventListener(event) {
                     else if (regexpSVG.test(title)) {
                         contentType = 'image/svg+xml';
                     }
+                    else if (regexpWEBM.test(title)) {
+                        contentType = 'video/webm';
+                    }
+                    else if (regexpMP4.test(title)) {
+                        contentType = 'video/mp4';
+                    }
+                    else if (regexpOGG.test(title)) {
+                        contentType = 'video/ogg';
+                    }
                 }
                 else if (nameSpace === '-') {
                     // It's a layout dependency
@@ -119,6 +138,10 @@ function fetchEventListener(event) {
                     }
                     else if (regexpCSS.test(title)) {
                         contentType = 'text/css';
+                    }
+                    else if (regexpVTT.test(title)) {
+                        // It's a subtitle
+                        contentType = 'text/vtt';
                     }
                 }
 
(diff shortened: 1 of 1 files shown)


Jaifroid:
Yes, that fixes the issue with webm seeking in Edge and in Chromium! Well done!
For some reason it takes longer for the video to load in Edge (but not in Chromium), than it does without this PR, but it's not really a big difference. That's probably an Edge peculiarity.
Regarding subtitles, subtitles still can't be loaded in Edge or in Chromium in either `tedxgeneva-2014_fr_all_2015-03.zim`, or in `ted_en_global_issues_2018-07.zim`. It's clearly a different issue, I think, so if you corroborate, I can open a new issue for it.
One other thing: `ted_en_global_issues_2018-07.zim` contains mp4 video, not webm, so you ned to add a text for it, and set a mimetype of `video/mp4`.

mossroy:
I've pushed a few fixes on this PR.
With these fixes, subtitles work on my Chromium, with the ZIM files you mentioned.
Could you please test again (on Chrome and Edge)?

Jaifroid:
Hmm, testing this commit in SW mode with Chromium 66.0.3359 and with Edge, I do not see any interface for subtitles either in dirtybiology or in `ted_en_global_issues_2018-07.zim`. This is what the video inteface looks like for me in Chromium on the lecture "A Call to Men". Below that is what it looks like in Edge. And below that is what the same video looks like in Chromium in JQuery mode.

image

image

image

Jaifroid:
PS I've double-checked that Chromium hasn't cached the previous version of the code. I can see the 'video/ogg' regex in the developer tools when I open the Service Worker code.

mossroy:
Strange.
I'm using Chromium 71.0.3578.80 (on Ubuntu 18.04), here is what I see on the same ZIM file (in SW mode), with the same video, after selecting the English subtitles : 
image
Maybe the behavior of Chromium has changed between versions 66 and 71 regarding webvtt?

Jaifroid:
Yes, I haven't updated my Chromium install for a while. I could update it, but then we might lose a tool for diagnosing the issue with earlier versions, although we still have the issue with Edge (which is up-to-date except that I haven't yet installed the problematic October 2018 Windows update). Should I update Chromium for Windows, or should I keep this version to try to diagnose further?

Jaifroid:
PS I find it strange that both Chromium 66 and Edge work in JQuery mode, and have no problem with the subtitles in that mode. The only difference is `video.js` as far as I can tell. Remember that the CC menu showed up when I disabled video.js in SW mode, but there was still a problem with finding the underlying files as I reported [here](https://github.com/kiwix/kiwix-js/issues/452#issuecomment-448563382).

Jaifroid:
Which leads me to think that the problem is with finding the subtitle files, not actually with the playback system. Video.js probably looks for them but can't find them, and then disables showing the CC menu. Whereas without video.js it thinks there are files, but can't find them when selecting them. Is there any clue as to why the files can't be found/extracted in these browser versions?

mossroy:
Keep your Chromium version for now, and try to see if there is any info in the console and/or in the network tab of dev tools regarding .vtt files.
I'll try to download a version 66 of Chromium and test with it

mossroy:
I tested with Chromium 66 (downloaded from the instructions at the bottom of https://www.chromium.org/getting-involved/download-chromium, leading to https://www.googleapis.com/download/storage/v1/b/chromium-browser-snapshots/o/Linux_x64%2F540276%2Fchrome-linux.zip?generation=1519938236072713&alt=media).
The subtitles are working fine but, for some reason, I have the audio but not the video. So I'm not 100% sure my test is valid.

Jaifroid:
OK, two possibilities have occurred to me. One is that it may be to do with Linux vs Windows. I have the ESR release of Firefox installed in Windows, which doesn't support SW, so I haven't tested FF on Windows with this subtitle issue (my Firefox Quantum is installed in WSL [Ubuntu subsystem on Windows]). The second speculation is that the problem could have something to do with the configuration of the Apache server I'm using to provide a web server on localhost. It's also running in WSL. It works for everything else, so I kind of doubt it's the issue. I'll try to troubleshoot the issue in both Chromium and Edge a bit this afternoon to see if I can find a common problem.

Jaifroid:
Just a quick note that Firefox Quantum running in WSL/Ubuntu can display the subtitles using the same Apache server instance, so I think that rules out a misconfiguration of the server.

service-worker.js
```diff
@@ -111,6 +121,15 @@ function fetchEventListener(event) {
...
+                    }
+                    else if (regexpMP4.test(title)) {
+                        contentType = 'video/mp4';
+                    }
+                    else if (regexpOGG.test(title)) {
+                        contentType = 'video/ogg';
+                    }
                 }
                 else if (nameSpace === '-') {
                     // It's a layout dependency
```
Jaifroid: @mossroy Because of this test, I believe lines 142-145 below won't catch subtitles that are in the A namespace. I assume you're aware of this, but have decided that subtitles in the A namespace are an anomaly that won't be repeated in future ZIMs. I just thought I'd check.

service-worker.js
```diff
@@ -97,11 +101,17 @@ function fetchEventListener(event) {
                 // The namespace defines the type of content. See http://www.openzim.org/wiki/ZIM_file_format#Namespaces
                 // TODO : read the contentType from the ZIM file instead of hard-coding it here
                 if (nameSpace === 'A') {
-                    // It's an article
-                    contentType = 'text/html';
+                    if (regexpVTT.test(title)) {
+                        // It's a subtitle
+                        contentType = 'text/vtt';
+                    }
```
Jaifroid: @mossroy Apologies, ignore previous message, I see you repeat the test, once in A namespace and once in - namespace.

mossroy: Yes.
I'm wondering if I should test the file extensions first, and ignore the namespace to set the content-type. I think I still need to fallback to text/html in the A namespace (because not 100% sure every HTML page inside this namespace ends with .html or .htm)

Jaifroid:
In my Chromium 66 Service Worker is not even *asking* for the subtitle files, same in Edge. I know `video.js` removes the subtitle file references from the video DOM (there's no trace of them when inspecting), so it must be failing to recognize something about the file URLs, doesn't see them as valid.

Jaifroid:
I'm afraid I've come up blank, and I don't have a lot of time at the moment to try to debug the internals of video.js (it's minified, and although I can pretty print it in the Chromium dev tools, it's still pretty hard to understand due to incomprehensible variable names). Maybe we should just merge this and open another issue to investigate the subtitle problem on Edge / Chromium on Windows. In the meantime, it's somewhat more likely that users wanting to view ZIMs with subtitled videos will be doing so in jQuery mode (where there seem to be no problems with subtitles), due to the difficulty, for normal users, of setting up a localhost Web server. Merging this will also help us to see if the issues persist in Chrome extensions (if they support SW mode).

service-worker.js
Jaifroid: I think it's safest to test the namespace, even if it means repeating the test. So this is fine.

service-worker.js
```diff
@@ -76,6 +76,10 @@ var regexpPNG = new RegExp(/\.png$/i);
 var regexpJS = new RegExp(/\.js/i);
 var regexpCSS = new RegExp(/\.css$/i);
 var regexpSVG = new RegExp(/\.svg$/i);
+var regexpWEBM = new RegExp(/\.webm$/i);
+var regexpMP4 = new RegExp(/\.mp4$/i);
+var regexpOGG = new RegExp(/\.og[mvg]$/i);
```
Jaifroid: @mossroy I should probably add ogg support in the equivalent code in jQuery mode. I'll create an issue for that.

PR #459, opened 2019-01-05
Switch Kiwix and OpenZIM URLs to https, and fix invalid OpenZIM ones

### LINKED ISSUE #458: Some OpenZIM wiki URLs are not working any more, and we should switch all Kiwix/OpenZIM URLs to HTTPS
For example http://www.openzim.org/wiki/OpenZIM is mentioned several times, and should be replaced by https://wiki.openzim.org/wiki/OpenZIM

mossroy:
Fixes #458

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,18 +1,18 @@
 # Kiwix JS
 
 ## Summary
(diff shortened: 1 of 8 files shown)


kelson42:
@mossroy Looks good to me even if all the old links should still work perfectly.

Issue #457, opened 2019-01-02
Release a version 2.5


mossroy:
I think we have implemented a few significant improvements, at least :
- video playback capability (including subtitles) in both SW and jQuery mode
- javascript support (in SW mode), and noscript tag support (in jQuery mode)
- and some other fixes and improvements

It might be worth a new release?

kelson42:
@mossroy That said, it looks like that the milestone if far from being over... But I support to release often, in particular if a few important features are newly available.

Jaifroid:
I agree that this release is important enough to do it ahead of schedule (yay!). It makes a whole class of ZIMs usable (those with video).

mossroy:
I also think we have enough improvements to make a release.
I'll update the github milestones, to rename the existing milestone 2.6 to 2.7 (and do the same for the following ones), and move the remaining issues targeted on milestone 2.5 to a new 2.6 milestone.

It's probably necessary to update the README and the about section to reflect the new ZIM compatibilities of this version, before doing the 2.5 release.

mossroy:
Second thought, I'm not sure we should update README and about section about new ZIM compatibilities, at least for now.
Because the lack of javascript support in jQuery mode (#152) prevents us from saying we really support ted, dirtybiology etc : the videos can be read, but the home pages don't. So it's more difficult to access them. @Jaifroid is working on that but it's complicated and/or ZIM-specific.
And jQuery mode is still the default one. Until I implement #196, we do not switch to SW mode automatically. And it's still not supported in Firefox extensions.
In SW mode, ted, PheT and dirtybology are truly supported, but I'm not 100% sure they will work in browser extensions (when distributed through the stores). Because some of them have inline javascript. In theory, it should be blocked (see https://developer.chrome.com/extensions/contentSecurityPolicy#JSExecution), but, for some reason, it works when I manually install the extension on Chromium 71. Will it still work when distributed through the Google store? I don't know.

To sum up, I would not claim about new ZIM files compatibility for now.

Jaifroid:
Could we call it "Preliminary support for x" and add a brief explanation that users need to use the built-in search functionality?

The way I've worked around lack of JS in Kiwix JS Windows is documented in [https://github.com/kiwix/kiwix-js-windows/issues/64](Provide a UI for accessing content of inaccessible ZIMs).

[...]

Issue #439, opened 2018-11-18
Handle epub file "downloads" from the ZIM in jQuery mode


Jaifroid:
This would support "downloading" (extracting) the epub version of book files stored in recent Project Gutenberg ZIMs.

I believe this can be achieved with the [download attirbute](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#Attributes) of the anchor element in html5.

The epubs are linked from the text version of book pages. They are also linked from the proprietary UI that these ZIMs employ, but books can only be accessed from our own search box in jQuery mode currently, due to the fact that the JavaScript that renders the UI contains URLs that are loaded dynamically, and to support this, the loading function inside the script would need to be patched (!), and patching would be pretty non-generic.

Still, the ZIMs are searchable and usable from our own interface, but it is only possible to search by title, currently.

It should be noted that it is quite likely that in Extensions and other contexts, the JS which currently works in SW mode would be blocked. For example, it uses XMLHttpRequests in the script, has some script in `<script>…</script>` tags, etc.

I wonder if we shouldn't ask for an alternative interface to be provided for non-JS-capable clients.

Jaifroid:
I've added a branch with preliminary support for download links in jQuery mode: [Support-epbub-downloads-from-Gutenberg-ZIMs](https://github.com/kiwix/kiwix-js/tree/Support-epbub-downloads-from-Gutenberg-ZIMs) . It does it via BLOBs, which I think is the only way in html5 (see [this stackexchange article](https://stackoverflow.com/questions/4184944/javascript-download-data-to-file-from-content-within-the-page/)). I've so far only tested in Edge, but it opens a nice epub reader if you ask to open the file when offered the choice (Edge has built-in support for epubs).

I've tried to make it generic, so that it can be extended for any filetypes.

@mossroy If you have a moment to look at the code, and have any preliminary advice about possible gotchas, or about making it more generic, let me know (or else wait for a PR and do it then). I have to do more testing before PR.

Jaifroid:
OK, this code works in Edge and Chromium, but not currently in any edition of Firefox or IE11. Strange. Maybe the latter browsers don't recognize mimetype `application/epub+zim`?

Jaifroid:
I have rebased the branch [Support-epbub-downloads-from-Gutenberg-ZIMs](https://github.com/kiwix/kiwix-js/tree/Support-epbub-downloads-from-Gutenberg-ZIMs) on top of current master.

[...]

PR #462, opened 2019-01-07
Support epub and other downloads from ZIM in jQuery mode

Jaifroid:
This code addresses issue #439 . Currently the epub links in Guttenberg ZIMs are useless in jQuery mode. This PR handles the "download" of the epub from the ZIM to the OS's filesystem, or to open the file using the browser's open method. The code is tested and working with `gutenberg_es_all_2018-10.zim` in Edge, Chromium 66, Firefox ESR, FFOS simulator, and IE11.

Since JS doesn't work, to choose a book to test, type a space in the search field, and choose any title with the word "(cover)" in the name. Then click on the big red download icon (not the HTML5 icon, which is a simple html/text-based version of the book). In some browsers the file will automatically be saved in the download location set in the browser. In others, the user is offered a choice to save the file or open it.

The user will need to have an epub app/reader installed unless the OS provides one. In Windows 10, Edge acts as the system epub reader, with a nice resizable interface and page flipping (unless the user has installed a different epub app).

In FFOS Simulator, a message pops up at the top of the app screen to say that the file has been successfully downloaded, but when I tap on the message (the one at the top of the screen, not the alert box at the bottom), I then get a message that the file does not exist in the file system. I guess this is a quirk of the simulator not having a physical filesystem, or else that there is no installed app to handle epub files.

There is no rush to merge this before #457 (2.5 Release), unless it's felt desirable to add to the feature set of 2.5. I've tested the code fairly extensively, but it could do with testing on other systems (e.g. Linux), and I don't know how downloads are likely to be handled in extensions.

### DIFF
--- a/www/js/app.js
+++ b/www/js/app.js
@@ -78,6 +78,7 @@ define(['jquery', 'zimArchiveLoader', 'util', 'uiUtil', 'cookies','abstractFiles
     // Define behavior of HTML elements
     $('#searchArticles').on('click', function(e) {
         $("#welcomeText").hide();
+        $('.alert').hide();
         $("#searchingArticles").show();
         pushBrowserHistoryState(null, $('#prefix').val());
         searchDirEntriesFromPrefix($('#prefix').val());
@@ -857,6 +858,11 @@ define(['jquery', 'zimArchiveLoader', 'util', 'uiUtil', 'cookies','abstractFiles
     // common in unsupported packaged UIs, e.g. PhET ZIMs.
     var regexpActiveContent = /<script\b(?:(?![^>]+src\b)|(?=[^>]+src\b=["'][^"']+?app\.js))(?!>[^<]+(?:importScript\(\)|toggleOpenSection))(?![^>]+type\s*=\s*["'](?:math\/|[^"']*?math))/i;
     
+    // DEV: The regex below matches ZIM links (anchor hrefs) that should have the html5 "donwnload" attribute added to 
+    // the link. This is currently the case for epub and pdf files in Project Gutenberg ZIMs -- add any further types you need
+    // to support to this regex. The "zip" has been added here as an example of how to support further filetypes
+    var regexpDownloadLinks = /^.*?\.epub($|\?)|^.*?\.pdf($|\?)|^.*?\.zip($|\?)/i;
+    
     // Cache for CSS styles contained in ZIM.
     // It significantly speeds up subsequent page display. See kiwix-js issue #335
     var cssCache = new Map();
@@ -890,6 +896,9 @@ define(['jquery', 'zimArchiveLoader', 'util', 'uiUtil', 'cookies','abstractFiles
         
         // Tell jQuery we're removing the iframe document: clears jQuery cache and prevents memory leaks [kiwix-js #361]
         $('#articleContent').contents().remove();
+
+        // Remove from DOM any download alert box that was activated in uiUtil.displayFileDownloadAlert function
+        $('#downloadAlert').alert('close');
         
         var iframeArticleContent = document.getElementById('articleContent');
         
@@ -961,12 +970,26 @@ define(['jquery', 'zimArchiveLoader', 'util', 'uiUtil', 'cookies','abstractFiles
                     // It's an external URL : we should open it in a new tab
                     this.target = "_blank";
                 } else {
-                    // It's a link to another article
-                    // Add an onclick event to go to this article
+                    // It's a link to an article or file in the ZIM
+                    var decodedURL = decodeURIComponent(zimUrl);
+                    var contentType;
+                    var downloadAttrValue;
+                    // Some file types need to be downloaded rather than displayed (e.g. *.epub)
+                    // The HTML download attribute can be Boolean or a string representing the specified filename for saving the file
+                    // For Boolean values, getAttribute can return any of the following: download="" download="download" download="true"
+                    // So we need to test hasAttribute first: see https://developer.mozilla.org/en-US/docs/Web/API/Element/getAttribute
+                    // However, we cannot rely on the download attribute having been set, so we also need to test for known download file types
+                    var isDownloadableLink = this.hasAttribute('download') || regexpDownloadLinks.test(href);
+                    if (isDownloadableLink) {
+                        downloadAttrValue = this.getAttribute('download');
+                        // Normalize the value to a true Boolean or a filename string or true if there is no download attribute
+                        downloadAttrValue = /^(download|true|\s*)$/i.test(downloadAttrValue) || downloadAttrValue || true;
+                        contentType = this.getAttribute('type');
+                    }    
+                    // Add an onclick event to extract this article or file from the ZIM
                     // instead of following the link
                     $(this).on('click', function (e) {
-                        var decodedURL = decodeURIComponent(zimUrl);
-                        goToArticle(decodedURL);
+                        goToArticle(decodedURL, downloadAttrValue, contentType);
                         return false;
                     });
                 }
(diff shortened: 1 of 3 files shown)


mossroy:
I quickly tested on Firefox 64 on Ubuntu : it seems to work well.
I agree it would be better to release version 2.5 first.
It will give me more time to review this PR, and we might fix #463 too.

mossroy:
I suppose this branch needs to be rebased on master. In particular to re-use the bootstrap alerts code that has already been merged?

Jaifroid:
I agree, but I'm a bit short of time to do that right now. Can we leave this to 2.6? I should have more time in a couple of weeks.

mossroy:
Of course, no problem!

Jaifroid:
I have mechanically rebased this branch / PR on master, but it is *not* yet ready for review, as I need to ensure it is re-using the existing bootstrap alert code first.

Jaifroid:
I've squashed this branch to make review easier. I think it's ready for review, though I've run out of time to test on all browsers today, can test more at w/e. I've tested the latest code on Edge and IE11. I tested the first few iterations of the code on all browsers, and the code has only been re-organized, not changed in any substantial way. I'm keen to get your feedback, @mossroy . We may still need to think further about the issue of creating bootstrap alerts in code, but it's all tucked into uiUtil.js.

Jaifroid:
Tested and working on Firefox ESR 52.5.2.

mossroy:
It seems to work well on Firefox 65 too (on Linux).
Some gutenberg books can also be downloaded as PDF : it's probably easy to add support for that too. For example "Le roman de la rose - Tome I" in gutenberg_fr_all_2018-10.zim

I'll have a look at the code

Jaifroid:
> Some gutenberg books can also be downloaded as PDF : it's probably easy to add support for that too. For example "Le roman de la rose - Tome I" in gutenberg_fr_all_2018-10.zim

Thank you! I hadn't seen examples of PDFs before, so it's good to know that file type is also supported. It's just a small edit in two places to enable PDF download. I can push a commit with that now.

mossroy:
It now seems to work well for PDF too, thanks!

NB : It's not related to this PR, but I opened https://github.com/openzim/gutenberg/issues/79 for some non-working hyperlinks in this book (I suppose it's a problem inside the ZIM file itself)

mossroy:
Downloading epub files works perfectly well on my Firefox OS device too, and can be read with an epub reader.
But I noticed another small issue in jQuery mode : there is a blue arrow button at the bottom left of each HTML book, that normally allows to go to the beginning of the book.
It works in SW mode (even if it is half-hidden by our bottom bar), but it fails in jQuery mode with this error message : "Article with title A/ not found in the archive"

www/js/app.js
```diff
@@ -961,12 +970,30 @@ define(['jquery', 'zimArchiveLoader', 'util', 'uiUtil', 'cookies','abstractFiles
...
-                    // Add an onclick event to go to this article
+                    // It's a link to an article or file in the ZIM
+                    var decodedURL = decodeURIComponent(zimUrl);
+                    var contentType;
+                    // Some file types need to be downloaded rather than displayed (e.g. *.epub)
+                    // The HTML download attribute can be Boolean or a string representing the specified filename for saving the file
+                    // For Boolean values, getAttribute can return any of the following: download="" download="download" download="true"
+                    // So we need to test hasAttribute first: see https://developer.mozilla.org/en-US/docs/Web/API/Element/getAttribute
+                    // However, we cannot rely on the download attribute having been set, so we also need to test for known download file types
+                    var downloadAttribute = this.hasAttribute('download') || regexpDownloadLinks.test(href);
```
mossroy: I find this variable name misleading : it should indicate that it's a boolean meaning that this link should be downloaded.
Something like "isDownloadableLink" or "shouldBeDownloaded"?

Jaifroid: OK, thanks, I'll rename.

www/js/app.js
```diff
@@ -961,12 +970,30 @@ define(['jquery', 'zimArchiveLoader', 'util', 'uiUtil', 'cookies','abstractFiles
...
+                    var contentType;
+                    // Some file types need to be downloaded rather than displayed (e.g. *.epub)
+                    // The HTML download attribute can be Boolean or a string representing the specified filename for saving the file
+                    // For Boolean values, getAttribute can return any of the following: download="" download="download" download="true"
+                    // So we need to test hasAttribute first: see https://developer.mozilla.org/en-US/docs/Web/API/Element/getAttribute
+                    // However, we cannot rely on the download attribute having been set, so we also need to test for known download file types
+                    var downloadValue = this.hasAttribute('download') || regexpDownloadLinks.test(href);
+                    if (downloadValue) {
+                        downloadValue = this.getAttribute('download') || true;
+                        downloadValue = /download|true/i.test(downloadValue) || downloadValue;
```
Jaifroid: @mossroy Because the download attribute can be Boolean, or it can contain a string value representing the filename to use when saving the file in the filesystem, I've unified the two separate variables I had (downloadAttribute and downloadValue) into a single variable (downloadValue). I've tested this code with different values of the download attribute, and it seems to function as intended, i.e., if the attribute contains "", "download" or "true", it will be set to Boolean true (also true if there is no download attribute, but regexpDownloadLinks matches). But if the download attribute contains any other string value, it will be set to the string. However, I could only test it indirectly, as I can't manipulate the actual value of the attribute at debug time, and can only manipulate the variable itself (downloadValue). I'm 99% sure it works, but could you have a look and see if the code looks logically correct to you?

Jaifroid:
> But I noticed another small issue in jQuery mode : there is a blue arrow button at the bottom left of each HTML book, that normally allows to go to the beginning of the book.

I've checked this and the html for this blue arrow is:

`<a href="#"><i class="fa fa-2x fa-chevron-circle-up"></i></a>`

I can't see any JavaScript in the html, or any script tag, so I guess that `#` alone is what causes the jump to the top of the page.

We probably need to add code to `parseAnchorsJQuery` to recognize a `#` on it own and not link it to an article. Should I add that in this PR, or should I make a new issue?

Jaifroid:
> But we really need to investigate on #470 because the code of these display* functions in uiUtil.js is partly redundant and mixes presentation (which should be in HTML files) and code (that should stay in this js file)

I agree. We will have to change the function of clicking on the bootstrap alert's close button (the x top right) so that it hides the alert rather than removing it from the DOM. Anyway, I'll tackle this in #470.

mossroy:
Yes, handling # hyperlinks is a separate issue, it would be better to fix it in another PR.

www/js/app.js
mossroy: I still find it complicated to read/understand.
You're right about the way to interpret the download attribute, and about the way to generate a file name. But maybe this code (that generates a filename) could be moved completely in parseAnchorsJQuery function, so that it then passes a filename (as a String) to goToArticle and displayFileDownloadAlert. It would simplify the understanding of the parameters of these functions (even if they were well documented).

Then, I would suggest to avoid reusing a variable for different purposes/types of content (except if it's really necessary, maybe for optimization).
For example, there could be a "isDownloadableLink" (boolean), then a "downloadAttributeValue" (string), and finally a "downloadFileName" (string). The first two ones are just temporary variables, used to generate the third one. As only the third one is used afterwards, this logic might even be externalized in a separate function, like "getDownloadFileName(this, decodedURL)", but maybe it's not necessary as there is not that much code to put in it.

Jaifroid: OK, I'll have another go!

Jaifroid: Thinking about this, I'm not keen on only passing a string in the `download` parameter of `goToArticle` because I think that function should remain generic in line with the HTMNL5 spec. In the spec, a download can be triggered either with a Boolean or with a filename. So it should be possible to trigger a download with just `download=true` passed to `goToArticle` . There might be other cases in the future where we want to trigger a download of content in line with the HTML5 spec. Given that, I think the only logical place to construct a filename if one is missing is in the appropriate function in `uiUtil.js`.

What I've done is to simplify the code in `parseAnchorsJQuery` as best I can following your advice above, except that I haven't included a `donwloadFileName` string here for the reasons stated. I have also moved the generation of a `contentType`, if it is missing from the HTML, into `uiUtil.js` so that we have consistent behaviour: if the download attribute is missing, we just pass a Boolean or null; if a `contentType` is missing, we just pass null. Then the download function in `uiUtil.js` takes care of constructing any missing data. I believe this is now better, and has decluttered `parseAnchorsJQuery`.

Let me know if you agree.

mossroy: All right, it can make sense.

Regarding the test `/download|true/i.test(downloadAttrValue)`, maybe it should be fixed for the case where downloadAttrValue contains one of those words without being equal to it.
I suppose you only want to match equality.
For example, if the download attribute is "truelle.zip", I suppose you want downloadAttrValue to be "truelle.zip" and not true.

Appart from that, I think you can squash/rebase/merge.

Thanks for this PR

Jaifroid:
> Regarding the test `/download|true/i.test(downloadAttrValue)`, maybe it should be fixed for the case where downloadAttrValue contains one of those words without being equal to it.
> I suppose you only want to match equality.
> For example, if the download attribute is "truelle.zip", I suppose you want downloadAttrValue to be "truelle.zip" and not true.

Thank you, yes, I hadn't thought of that. I'll fix with:

`/^(download|true)$/i.test(downloadAttrValue)`

mossroy:
All right, that should be good.

Issue #466, opened 2019-01-10
Show a Bootstrap alert in jQuery mode when active content is detected


Jaifroid:
As discussed in #457, it would be ideal to display a Bootstrap alert when active content that cannot be displayed in jQuery mode is detected. This would be similar to that used in Kiwix JS Windows (see screenshot below), but without the Archive Index feature. @mossroy has suggested the following wording for the alert:

> Unable to display active content: this ZIM file is not fully supported in jQuery mode. Try to switch to ServiceWorker mode in the settings (if supported by your platform). Else some content might be available through a search (type some letters in the search zone). [Permanently hide]

It would be ideal if this can be backported in time for release 2.5. The implementation isn't difficult.

![image](https://user-images.githubusercontent.com/4304337/50907955-84246f00-1420-11e9-8bcf-0570262dccf9.png)

Jaifroid:
PR #467 addresses this issue, and is explained there. I merely add some screenshots here.

![image](https://user-images.githubusercontent.com/4304337/51009132-d1026580-1546-11e9-9727-2b396f868d08.png)

![image](https://user-images.githubusercontent.com/4304337/51009177-06a74e80-1547-11e9-84fa-f4b51d273bf6.png)

![image](https://user-images.githubusercontent.com/4304337/51009219-2dfe1b80-1547-11e9-824a-83105f606392.png)

![image](https://user-images.githubusercontent.com/4304337/51009228-3fdfbe80-1547-11e9-918b-56e79b52b9b1.png)

Issue #438, opened 2018-11-17
Cookies.js uses deprecated escape and unescape functions


Jaifroid:
You can see the use of these in `cookies.js` here: https://github.com/kiwix/kiwix-js/blob/master/www/js/lib/cookies.js#L26 

See the big red warning sign regarding `escape-unescape` on https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/escape .

This is causing me some problems with UTF-8 characters in document titles in an experimental branch where I encode the last-visited page title in a cookie. It seems that `escape-unescape` are pre-UTF8 era, and we should use `encodeURIComponent` and `decodeURIComponent` instead where we are not using simple ASCII.

We don't need a framework for our limited handling of cookies. because it's very easy to get and set them without a framework, since a) we set the values ourselves, so we control them; b) most (all?) of the key-value pairs we use are ASCII; and c) if we extend them in the future we can simply `encodeURIComponent` and `decodeURIComponent` anything else ourselves.

These are the only patterns we need in our current code, I think, and they're one-liners:

* Getting the value of `lastContentInjectionMode`:

    `value = document.cookie.replace(/^.*lastContentInjectionMode=([^;]*).*$/, '$1');
`
* Setting the value of `lastContentInjectionMode`:

    `document.cookie = 'lastContentInjectionMode=' + lastContentInjectionMode + ';expires=Fri, 31 Dec 9999 23:59:59 GMT';`

And for good measure, though I don't think we use these patterns:

* Checking that value of `myKey` is `true`:

    `if (/myKey=true\b/i.test(document.cookie) { do something }`

* Deleting `myKey`:

    `document.cookie = myKey + '=;expires=Thu, 21 Sep 1979 00:00:01 UTC';`

mossroy:
I took this `cookies.js` file from Mozilla, but it was a long time ago.
It has been removed from https://developer.mozilla.org/en-US/docs/DOM/document.cookie, but the last available version in the history of this page has replaced escape/unescape with encodeURIComponent/decodeURIComponent : https://developer.mozilla.org/en-US/docs/Web/API/Document/cookie$revision/948801#A_little_framework_a_complete_cookies_readerwriter_with_full_unicode_support

[...]

PR #467, opened 2019-01-11
Add active content warning in jQuery mode

Jaifroid:
This is a first go at addressing #466. I have backported the UI features and detection of the active content warning from Kiwix JS Windows. It intelligently detects an unsupported UI and puts a yellow Bootstrap `alert-warning` just below the search box. The alert is only shown on the landing page (otherwise it would become very annoying), and is removed when the user presses Config or searches for articles. By default, it will appear again whenever the user goes back to a landing page with active content.

The alert can be dismissed temporarily (with the x) or permanently. Hyperlinks to the appropriate settings on the Config page are provided (to activate SW mode, and to hide the warning permanently), and these hyperlinks jump to (and highlight) the setting the user needs to activate. The highlighting feature is backported from Kiwix JS Windows where it is maybe more useful because there are a lot more options that can be confusing. However, it still seems useful here, and educates the user about the Config settings. It also will become more useful as display settings are added.

I've taken the liberty of including a small bit of related code that is part of the same UI feature set in Kiwix JS Windows. It responsively changes the colour of the API status panel from yellow `panel-warning` (when the SW API is unavailable or unregistered) to green `panel-success` when the SW API is available and registered. I think this makes things clearer for the user and is aesthetically pleasing.

### DIFF
--- a/www/js/app.js
+++ b/www/js/app.js
@@ -50,6 +50,18 @@ define(['jquery', 'zimArchiveLoader', 'util', 'uiUtil', 'cookies','abstractFiles
     var selectedArchive = null;
     
     /**
+     * A global parameter object for storing variables that need to be remembered between page loads
+     * or across different functions
+     * 
+     * @type Object
+     */
+    var params = {};
+
+    // Set parameters and associated UI elements from cookie
+    params['hideActiveContentWarning'] = cookies.getItem('hideActiveContentWarning') === 'true';
+    document.getElementById('hideActiveContentWarningCheck').checked = params.hideActiveContentWarning;
+    
+    /**
      * Resize the IFrame height, so that it fills the whole available height in the window
      */
     function resizeIFrame() {
@@ -159,6 +171,7 @@ define(['jquery', 'zimArchiveLoader', 'util', 'uiUtil', 'cookies','abstractFiles
         $('#articleListWithHeader').hide();
         $("#searchingArticles").hide();
         $('#articleContent').hide();
+        $('.alert').hide();
         refreshAPIStatus();
         return false;
     });
@@ -178,22 +191,31 @@ define(['jquery', 'zimArchiveLoader', 'util', 'uiUtil', 'cookies','abstractFiles
         $('#articleListWithHeader').hide();
         $("#searchingArticles").hide();
         $('#articleContent').hide();
+        $('.alert').hide();
         return false;
     });
     $('input:radio[name=contentInjectionMode]').on('change', function(e) {
         // Do the necessary to enable or disable the Service Worker
         setContentInjectionMode(this.value);
     });
-    
+    $('input:checkbox[name=hideActiveContentWarning]').on('change', function (e) {
+        params.hideActiveContentWarning = this.checked ? true : false;
+        cookies.setItem('hideActiveContentWarning', params.hideActiveContentWarning, Infinity);
+    });
+
     /**
      * Displays of refreshes the API status shown to the user
      */
     function refreshAPIStatus() {
+        var apiStatusPanel = document.getElementById('apiStatusDiv');
+        apiStatusPanel.classList.remove('panel-success', 'panel-warning');
+        var apiPanelClass = 'panel-success';
         if (isMessageChannelAvailable()) {
             $('#messageChannelStatus').html("MessageChannel API available");
             $('#messageChannelStatus').removeClass("apiAvailable apiUnavailable")
                     .addClass("apiAvailable");
         } else {
+            apiPanelClass = 'panel-warning';
             $('#messageChannelStatus').html("MessageChannel API unavailable");
             $('#messageChannelStatus').removeClass("apiAvailable apiUnavailable")
                     .addClass("apiUnavailable");
@@ -204,15 +226,19 @@ define(['jquery', 'zimArchiveLoader', 'util', 'uiUtil', 'cookies','abstractFiles
                 $('#serviceWorkerStatus').removeClass("apiAvailable apiUnavailable")
                         .addClass("apiAvailable");
             } else {
+                apiPanelClass = 'panel-warning';
                 $('#serviceWorkerStatus').html("ServiceWorker API available, but not registered");
                 $('#serviceWorkerStatus').removeClass("apiAvailable apiUnavailable")
                         .addClass("apiUnavailable");
             }
         } else {
+            apiPanelClass = 'panel-warning';
             $('#serviceWorkerStatus').html("ServiceWorker API unavailable");
             $('#serviceWorkerStatus').removeClass("apiAvailable apiUnavailable")
                     .addClass("apiUnavailable");
         }
+        apiStatusPanel.classList.add(apiPanelClass);
+
     }
     
     var contentInjectionMode;
@@ -640,6 +666,7 @@ define(['jquery', 'zimArchiveLoader', 'util', 'uiUtil', 'cookies','abstractFiles
      */
     function searchDirEntriesFromPrefix(prefix) {
         if (selectedArchive !== null && selectedArchive.isReady()) {
+            $('#activeContent').alert('close');
             selectedArchive.findDirEntriesWithPrefix(prefix.trim(), MAX_SEARCH_RESULT_SIZE, populateListOfArticles);
         } else {
             $('#searchingArticles').hide();
@@ -713,12 +740,11 @@ define(['jquery', 'zimArchiveLoader', 'util', 'uiUtil', 'cookies','abstractFiles
             $("#searchingArticles").show();
             if (dirEntry.isRedirect()) {
                 selectedArchive.resolveRedirect(dirEntry, readArticle);
-            }
-            else {
+            } else {
+                params.isLandingPage = false;
                 readArticle(dirEntry);
             }
-        }
-        else {
+        } else {
             alert("Data files not set");
         }
     }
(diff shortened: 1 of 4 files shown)


Jaifroid:
@mossroy I've now tested and debugged this PR in Edge, Firefox ESR, FFOS Simulator, Chromium 66, and IE11 on Windows. It's ready for review.

mossroy:
I quickly tested on Firefox 60 ESR and Chrome 71 : it worked very well !
I really like the highlighting feature, and also the changing colors of the API panel.
I'll have a look at the code

www/js/app.js
```diff
@@ -49,6 +49,18 @@ define(['jquery', 'zimArchiveLoader', 'util', 'uiUtil', 'cookies','abstractFiles
      */
     var selectedArchive = null;
     
+    /**
+     * A global parameter object for storing variables that need to be remembered between page loads
+     * or across different functions
+     * 
+     * @type Object
+     */
+    var params = {};
```
mossroy: I can't make my mind about the use of this `params` object.
On which criteria would you put some variables in it, instead of declaring them individually? How different are they from some other variable like contentInjectionMode, lastContentInjectionMode, serviceWorkerRegistration etc?

www/js/app.js
```diff
@@ -49,6 +49,18 @@ define(['jquery', 'zimArchiveLoader', 'util', 'uiUtil', 'cookies','abstractFiles
...
+
+    // Set parameters and associated UI elements from cookie
+    params['hideActiveContentWarning'] = /hideActiveContentWarning=true/i.test(document.cookie);
```
mossroy: Why don't you use the functions of cookies.js to read the cookie like you did line 203 to set it? Is it related to #438?

www/js/app.js
```diff
@@ -1132,6 +1171,8 @@ define(['jquery', 'zimArchiveLoader', 'util', 'uiUtil', 'cookies','abstractFiles
                 $("#searchingArticles").hide();
                 alert("Article with title " + title + " not found in the archive");
             } else {
+                params.isLandingPage = false;
+                $('#activeContent').alert('close');
```
mossroy: Just for my understanding : what is this .alert function?

www/js/app.js
```diff
@@ -836,6 +869,12 @@ define(['jquery', 'zimArchiveLoader', 'util', 'uiUtil', 'cookies','abstractFiles
      * @param {String} htmlArticle
      */
     function displayArticleContentInIframe(dirEntry, htmlArticle) {
+        // Display Bootstrap warning alert if the landing page contains active content
+        var containsActiveContent = regexpActiveContent.test(htmlArticle);
```
mossroy: It would be better to move this test, so that it is executed only if `!params.hideActiveContentWarning && params.isLandingPage`
It would save some useless CPU cycles

www/js/lib/uiUtil.js
```diff
@@ -84,12 +84,45 @@ define([], function() {
         return url.replace(regexpRemoveUrlParameters, "$1");
     }
 
+    /**
+     * Displays a Bootstrap warning alert with information about how to access content in a ZIM with unsupported active UI
+     */
+    function displayActiveContentWarning() {
+        // We have to add the alert box in code, because Bootstrap removes it completely from the DOM when the user dismisses it
```
mossroy: Could you give me some more explanation?
What was the issue?

www/js/app.js
Jaifroid: @mossroy They are not really different. It just simplified the back-porting of the code from Kiwix JS Windows. The architecture I adopted there, first suggested by sharun-s I believe, was to declare a global params object in `init.js` and populate it from cookie at startup, set UI checkboxes, etc., or set default parameters which the user can then override in the UI. I have about 20 parameters that control different things about the state of the app (not all stored in the cookie by any means), and it simplifies things greatly to have them all kept together in the one object, and in one spot which the developer can easily manipulate (a bit like an old .ini file). There's an argument for tidying up and centralizing the use of such parameters in Kiwix JS, and things like `params['lastContentInjectionMode']` would be prime candidates for such tidying. However, I recognize that this is importing a bit of "foreign" code, and if you'd rather not go down that route, at least for now, it's fine, I can easily rename the two params I use in this PR. Just let me know if you'd prefer me to rename to standard variables.

www/js/app.js
Jaifroid: @mossroy Actually, not directly related to #438 in this instance. It's just a very concise and fast way of setting the state of the paramater to true or false. Saves some CPU!! With cookies plugin it returns a text value from the cookie which I then have to test with a conditional to transform it to true/false. This was the most concise way of getting a Boolean. Happy to change if you don't mind the code expanding a bit.

www/js/app.js
Jaifroid: @mossroy This is the standard way of closing a Bootstrap alert. See https://getbootstrap.com/docs/4.0/components/alerts/#methods . However, in one instance it caused an exception in both FF and in IE11, and I think it was because Bootstrap hadn't finished loading by the time it was called. So, you will see one alternative in the code which is simply a jQuery .hide() command. That doesn't actually dismiss the alert, just hides it, whereas this .alert function does.

www/js/app.js
Jaifroid: @mossroy OK, thanks, will do!

www/js/lib/uiUtil.js
Jaifroid: @mossroy PS from the Bootstrap API linked above:

> Note that closing an alert will remove it from the DOM.

www/js/app.js
Jaifroid: Commit https://github.com/kiwix/kiwix-js/pull/467/commits/778bc0a6a565775f98a441621a5a499ca1d740b1 addresses this comment.

www/js/app.js
mossroy: It's true that your implementation is probably faster.
On the other hand, I don't think we have performance issues regarding reading/setting cookies.
And this implementation might lead to some nasty bugs : if I set a cookie called "zzhideActiveContentWarning", it probably matches the regexp. Same thing if the value is "truezz".
In general, when there is no known performance issue on a feature, I prefer to centralize in a common implementation (even if it's a bit slower) : this lets us fix this kind of bugs in a single place, and also lets us easily refactor the code when necessary (like for #438).
So I'd prefer that you use the functions of cookies.js here

www/js/app.js
mossroy: OK, thanks for the pointer

www/js/app.js
mossroy: Looks good

www/js/app.js
Jaifroid: `/\bhideActiveContentWarning=true\b/.test(document.cookie)` would mitigate against both of the attacks on this value that you mention. However, I think your point is more philosophical ;-) and about consistency, so I'm happy to change it to use cookie.js.

www/js/lib/uiUtil.js
mossroy: I think we should investigate to find a better way to do that.
It should be avoided to have big chunks of HTML in a javascript file.

A few ideas : 
- the alert DOM element could be hidden somewhere else, and copied when we need to display it. So that bootstrap only removes the copy when the user dismisses it
- or, as you were mentioning, we could override the behavior of the X, like in https://stackoverflow.com/questions/20494887/my-bootstrap-alert-will-not-display-after-being-closed or in https://stackoverflow.com/questions/13550477/twitter-bootstrap-alert-message-close-and-open-again

But there might be better ideas.

Anyway, if it takes too much time, this might be left as is for now, provided that it will be fixed in next version (through a new issue)

www/js/lib/uiUtil.js
```diff
@@ -84,12 +84,45 @@ define([], function() {
...
+        var alertHTML =
+            '<div id="activeContent" class="alert alert-warning alert-dismissible fade in">' +
+                '<a href="#" class="close" data-dismiss="alert" aria-label="close">&times;</a>' +
+                '<strong>Unable to display active content:</strong> This ZIM is not fully supported in jQuery mode.<br />' +
+                'Content may be available by searching above (type a space or a letter of the alphabet), or else ' +
+                '<a id="swModeLink" href="#contentInjectionModeDiv" class="alert-link">switch to Service Worker mode</a> ' +
+                'if your platform supports it. &nbsp;[<a id="stop" href="#displaySettingsDiv" class="alert-link">Permanently hide</a>]' +
+            '</div>';
+        document.getElementById('alertBoxHeader').innerHTML = alertHTML;
+        ['swModeLink', 'stop'].forEach(function(id) {
```
mossroy: It would be good to add a comment here, explaining what the following code does (moving to the configuration section, and highlighting the relevant option)

www/js/app.js
Jaifroid: I've added issue #469 .

www/js/lib/uiUtil.js
Jaifroid: I have created #470 to deal with this issue. I'm wary about overriding Bootstrap standard behaviour, so would prefer the copy option, but we shall see.

www/js/lib/uiUtil.js
Jaifroid: Will do.

Jaifroid: Commit https://github.com/kiwix/kiwix-js/pull/467/commits/e179c98d3f58553b26b80085fce0f020493decea adds the requested comments.

Jaifroid:
@mossroy I believe remaining issues with this PR are either resolved or pushed to the two new issues I have created. Let me know, by approving the PR, whether it's ready to be squashed/merged.

PR #324, opened 2017-12-27
More complete scenario for UI tests

mossroy:
Should fix #323 
Not ready to be merged yet

PR #474, opened 2019-01-29
More complete scenario for ui tests

### LINKED ISSUE #323: Run a more complete scenario for UI tests
Currently the scenario is quite basic : we choose the Ray Charles archive, make a search, display an article, and check that a particular string is displayed in this article.
We might test more things, like :
- check that some images are displayed
- check that the CSS style is applied
- click on some links to browse a few articles
- use the back/forward buttons
- etc

mossroy:
Following #324, new attempt to fix #323.
This is not ready to be merged yet.

The error message is now more meaningful than it was : 
After checking that element <#mwBA> has css property "float: right", it fails with : 
```
Error while running .moveTo() protocol action: unknown error: Element <iframe id="articleContent" class="articleIFrame" src="article.html" style="height: 551px;" cd_frame_id_="ec9c22917e519d9c1fac47d6df4e53e3"></iframe> is not clickable at point (920, 671). Other element would receive the click: <a class="btn btn-lg" id="btnTop" title="Top" href="#top">...</a> 
An error occurred while running .moveToElement() command on <#mwBTI>: unknown error: Element <iframe id="articleContent" class="articleIFrame" src="article.html" style="height: 551px;" cd_frame_id_="ec9c22917e519d9c1fac47d6df4e53e3"></iframe> is not clickable at point (920, 671). Other element would receive the click: <a class="btn btn-lg" id="btnTop" title="Top" href="#top">...</a> 
Error while running .clickElement() protocol action: unknown error: Element <iframe id="articleContent" class="articleIFrame" src="article.html" style="height: 551px;" cd_frame_id_="ec9c22917e519d9c1fac47d6df4e53e3"></iframe> is not clickable at point (920, 671). Other element would receive the click: <a class="btn btn-lg" id="btnTop" title="Top" href="#top">...</a> 
An error occurred while running .click() command on <#mwBTI>: unknown error: Element <iframe id="articleContent" class="articleIFrame" src="article.html" style="height: 551px;" cd_frame_id_="ec9c22917e519d9c1fac47d6df4e53e3"></iframe> is not clickable at point (920, 671). Other element would receive the click: <a class="btn btn-lg" id="btnTop" title="Top" href="#top">...</a> 
```

So the issue comes from the bottom bar, that hides the relevant elements

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,6 +13,7 @@ script:
     # The free account on Sauce does not allow more than 5 concurrent sessions (including the main one)
     # So we separate the recent and old browsers in order to respect this limit.
     - "./node_modules/nightwatch/bin/nightwatch -c nightwatch.js --env firefox,chrome,edge"
+    - "./node_modules/nightwatch/bin/nightwatch -c nightwatch.js --env edge40,edge42"
     - "./node_modules/nightwatch/bin/nightwatch -c nightwatch.js --env firefox45,chrome58,ie11"
     - pkill node
 deploy:
--- a/README.md
+++ b/README.md
@@ -1,4 +1,5 @@
 # Kiwix JS
+[![Build Status](https://travis-ci.org/kiwix/kiwix-js.svg?branch=master)](https://travis-ci.org/kiwix/kiwix-js)
 
 ## Summary
 Kiwix is an offline Wikipedia viewer. See the official site: https://www.kiwix.org/
--- a/browser-tests/nightwatch_runner.js
+++ b/browser-tests/nightwatch_runner.js
@@ -39,6 +39,7 @@ module.exports = {
         browser
             .url('http://localhost:8080/')
             .waitForElementVisible('#archiveFiles', 20000)
+            // Select the ZIM archive of Ray Charles
             .execute(function() {
                 window.setRemoteArchives('http://localhost:8080/tests/wikipedia_en_ray_charles_2015-06.zimaa',
                     'http://localhost:8080/tests/wikipedia_en_ray_charles_2015-06.zimab',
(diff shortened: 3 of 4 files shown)


mossroy:
I managed to make these UI tests work on almost all browsers. I had to hide the bottom bar to make things work : it's not great but it's better than nothing.

In any case, I still have issues with Microsoft Edge, where it seems to depend on the version :
- Edge 40 fails to find the content of the first article
- Edge 42 works fine
- latest Edge (currently 44) also fails, but with an unknown error when trying to read some text content in the second article

It might be edge cases (pun intended) related to the way the tests are run.

It might be worth adding the version that is included in Windows Mobile? @Jaifroid : which one is it? (I need one of the version numbers listed in https://en.wikipedia.org/wiki/Microsoft_Edge#Release_history)

Jaifroid:
Apologies, @mossroy, I only just saw you request to know the version of Edge in Windows Mobile. It is 40 (40.15254.552.0 on a fully up-to-date phone). The EdgeHTML engine is 15.15254. However, don't spend time trying to make the UI tests work on Edge 40, as I'm sure you know that Windows Mobile now has less than a year left of official patch/security update support from Microsoft, and there is only one make of Windows Mobile still on sale brand new (Wiley Fox).

The other thing I'm sure you're aware of is that MS has officially announced that it will incorporate the Open Source Chromium engine into Edge going forwards. See https://github.com/MicrosoftEdge/MSEdge/blob/master/README.md

mossroy:
Thanks @Jaifroid.
I agree with you, so I disabled the failing UI tests on Edge 40 and 44.
Yes I knew about the switch to Chromium engine. When they release this version, I'll re-enable all the tests on latest Edge, to test it completely.

mossroy:
I just squashed and rebased to ease your review

mossroy:
Thanks, that's what I expected from you

Issue #478, opened 2019-03-02
Handle hyperlinks containing a single # in parseAnchorsJQuery function


Jaifroid:
Hyperlinks of the type `<a href="#">…..</a>` are not handled correctly in jQuery mode, as they attempt to launch an article. We need to add a test to `parseAnchorsJQuery` to prevent this behaviour.

Jaifroid:
A test case for this PR is documented [here](https://github.com/kiwix/kiwix-js/pull/462#issuecomment-468939961).

PR #487, opened 2019-04-22
Handle single # in hyperlink

Jaifroid:
This PR attempts to address #478. A single # in a hyperlink is defined as scroll to top of page. In jQuery mode, we need to recognize the # due to the base tag, and then we have to implement a scroll to top of the iframe window. 

The main use case for this that I'm aware of is the HTML5 version of Gutenberg books. They have a blue caret at the bottom of the page which should scroll to the top of the book. Therefore, this PR needs to be tested against, for example, `gutenberg_fr_all_2018-10.zim`. In jQuery mode, press a space in the search bar and open any book cover. Then go to the HTML5 version of the book, scroll down a little, and click the caret. The iframe window should scroll to the top.

@mossroy Can you think of any gotchas? I know `href="#"` is sometimes used with JavaScript, to allow programming of a click. But as we currently don't support JavaScript in the ZIM in jQuery mode, I don't think this code should interfere with any scripted behaviours.

### DIFF
--- a/www/js/app.js
+++ b/www/js/app.js
@@ -939,8 +939,8 @@ define(['jquery', 'zimArchiveLoader', 'util', 'uiUtil', 'cookies','abstractFiles
             // Percent-encode dirEntry.url and add regex escape character \ to the RegExp special characters - see https://www.regular-expressions.info/characters.html;
             // NB dirEntry.url can also contain path separator / in some ZIMs (Stackexchange). } and ] do not need to be escaped as they have no meaning on their own. 
             var escapedUrl = encodeURIComponent(dirEntry.url).replace(/([\\$^.|?*+\/()[{])/g, '\\$1');
-            // Pattern to match a local anchor in an href even if prefixed by escaped url
-            var regexpLocalAnchorHref = new RegExp('^(?:#|' + escapedUrl + '#)([^#]+$)');
+            // Pattern to match a local anchor in an href even if prefixed by escaped url; will also match # on its own
+            var regexpLocalAnchorHref = new RegExp('^(?:#|' + escapedUrl + '#)([^#]*$)');
             $('#articleContent').contents().find('body').find('a').each(function () {
                 // Attempts to access any properties of 'this' with malformed URLs causes app crash in Edge/UWP [kiwix-js #430]
                 try {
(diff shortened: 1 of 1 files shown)


Jaifroid:
Tested and working (with `gutenberg_fr_all_2018-10.zim`) on Firefox ESR, IE11, Edge 44 (EdgeHTML), and Edge 75 (Chromium).

Jaifroid:
Thanks, @mossroy . I can certainly add a test for an onclick attribute (even though I think any code in it won't run in most environments due to the CSP). We would have to revisit this in the future if I ever manage to get JS-in-the-ZIM working in an acceptable-enough way.

I guess an alternative is not to `return false` in the jQuery function, i.e., allow propagation? I'd need to test that to be sure it doesn't lead to a navigation, though.

mossroy:
The purpose of the improvement I suggest is not to run any extra javascript code (we know we can't do that properly, at least for now) or to allow propagation. The only goal is to handle a bit better the following case :
```
<a href="#" onclick="doSomething();return false;">
```
It used to be a common (and dirty) way to have a hyperlink that runs some javascript. In this case, the expected behavior is that a click on it will run `doSomething()`, but will not scroll to top. The `href` is only used to make the content look like a normal hyperlink.

The current PR will scroll to top, which could be surprising for the user. I think it would be better if it does nothing at all if there is a `href="#"` AND a non-empty `onclick` attribute.

But it's just an idea : it might be difficult to test it, as I don't know if we have such case in the existing ZIM files.
I'd be perfectly fine if you prefer to merge this PR without that improvement.

Jaifroid:
I think it's a good idea. I'll try!

Jaifroid:
I've pushed some code to test for the `onclick` attribute. Could you have a quick look over it @mossroy ? 

It's not quite the logic you describe: what I've put here is that we attach the click event only if the `onclick` attribute is empty or null, regardless of what's in the `href` attribute. The reasoning here is that if the click has been programmed, then it doesn't matter what's there, it will do whatever the onclick event instructs, and we shouldn't interfere.

It's hard to test this positively, as I'm not sure we have examples of inline onclick events. However, I've tested negatively, and the getter returns `null` if there is no `onclick` attribute. I'm assuming, but don't know for sure, that it will return `""` if there is an empty onclick attribute. Either way, my test should ignore it if it isn't a function or text.

I've tested with our standard hyperlinks, and the code doesn't interfere with navigation.

mossroy:
The code looks good.
I never saw something like `<a href="#something" onclick="doSomething()">`, so I don't think you take a big risk with this implementation.

NB : having an onclick does not always mean the href will never be used : if the javascript returns true, the href will be used.

Jaifroid:
> NB : having an onclick does not always mean the href will never be used : if the javascript returns true, the href will be used.

Hmm. In that case, maybe it's safer to put the condition inside the function. I was thinking of saving some CPU cycles! I realize this is all very marginal.

Jaifroid:
@mossroy I've now tested this code on Firefox OS simulator with Spanish Gutenberg. The scroll to top function works fine (it even does a fairly smooth scroll), and the code does not interfere with normal hyperlinks on other ZIM files.

Are you happy for me to squash/merge?

mossroy:
It looks good! You can squash/rebase/merge

Issue #485, opened 2019-04-18
Titles apparently missing in wikivoyage_en_all_novid_2019-04.zim


Jaifroid:
As per https://github.com/openzim/mwoffliner/issues/605 we appear to have an issue with missing titles when `wikivoyage_en_all_novid_2019-04.zim` is viewed in Kiwix JS. I initially thought this was an mwoffliner issue, but it appears it may not be, and that the issue may be with code in Kiwix JS. I open this issue in order to investigate whether the problem is indeed with Kiwix JS.

Jaifroid:
Here is another screenshot to show the issue:

![image](https://user-images.githubusercontent.com/4304337/56377055-5e69eb80-6201-11e9-9cee-cd60584b2097.png)

Jaifroid:
For comparison, here is the dirEntry for "[Lausanne]":
![image](https://user-images.githubusercontent.com/4304337/56441267-71a1b780-62e4-11e9-825d-1cd1919f7f58.png)

And here is the dirEntry for "A Taste of Coastal Texas":
![image](https://user-images.githubusercontent.com/4304337/56441212-3dc69200-62e4-11e9-8e3e-14e5828b944c.png)

As can clearly be seen, the title property of the dirEntry is blank for "[Lausanne]", but not for "A Taste of Coastal Texas".

@ISNIT0 noticed, and I confirm this, that all the one-word titles are missing in Kiwix JS, but all titles with two or more words are present in the corresponding dirEntry.

As @ISNIT0 showed Lausanne appears with a title field when extracted with `zimdump`. However,  `zimdump` uses `libzim` to read the Zim contents, and `libzim` has [this line](https://github.com/openzim/libzim/blob/684c355a6c56208c3d9e0be7990c4923b7ed8c50/src/_dirent.h#L84) which fills in the missing title from the URL when reading the dirEntry.

Therefore, it seems to me, that the most likely scenario is that this ZIM is in fact missing titles for one-word dirEntries. @mossroy , would you concur? Is there any other test we can do to show this?

mossroy:
That's what seems the most probable to me too.
If we want to be 100% sure, we should check that the title is not emptied by our javascript code.
You could put a breakpoint in a more low-level javascript function, like line 142 of zimfile.js, which constructs a dirEntry from the ZIM file.
[...]

PR #488, opened 2019-04-25
Provide getTitleOrUrl function to return url if title is empty

Jaifroid:
This addresses #485. Is this the right way to define the function? I've tested in Chromium and Edge in jQuery mode so far, and it seems to work.

### DIFF
--- a/www/js/lib/zimDirEntry.js
+++ b/www/js/lib/zimDirEntry.js
@@ -108,6 +108,16 @@ define([], function() {
     };
 
     /**
+     * Defines a function that returns the URL if the title is empty, as per the specification
+     * See https://wiki.openzim.org/wiki/ZIM_file_format#Directory_Entries
+     * 
+     * @returns {String} The dirEntry's title or, if empty, the dirEntry's (unescaped) URL   
+     */
+    DirEntry.prototype.getTitleOrUrl = function() {
+        return this.title ? this.title : this.url;
+    };
+
+    /**
      * Functions and classes exposed by this module
      */
     return {
--- a/tests/tests.js
+++ b/tests/tests.js
@@ -176,7 +176,7 @@ define(['jquery', 'zimArchive', 'zimDirEntry', 'util', 'uiUtil', 'utf8'],
             var callbackFunction = function(dirEntryList) {
                 assert.ok(dirEntryList && dirEntryList.length === 5, "Article list with 5 results");
                 var firstDirEntry = dirEntryList[0];
-                assert.equal(firstDirEntry.title , 'A Fool for You', 'First result should be "A Fool for You"');
+                assert.equal(firstDirEntry.getTitleOrUrl() , 'A Fool for You', 'First result should be "A Fool for You"');
                 done();
             };
             localZimArchive.findDirEntriesWithPrefix('A', 5, callbackFunction);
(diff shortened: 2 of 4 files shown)


Jaifroid:
Further info: this PR needs testing against a ZIM with new-style URLs, e.g. `wikivoyage_en_all_novid_2019-04.zim`. It's what I used in my quick tests so far (also now tested in SW mode on Chromium and Edge).

As an aside, note that this ZIM is buggy. For example, the Lausanne article, in SW Mode, has all the section titles closed and there is no way to open them. Also, the home page is missing, so instead we geta a huge list of all the titles in the ZIM which takes ages to load.

www/js/lib/zimDirEntry.js
```diff
@@ -107,6 +107,15 @@ define([], function() {
         return new DirEntry(zimfile, data);
     };
 
+    /**
+     * Defines a getter function that returns the URL if the title is empty
```
mossroy: Here, I would not say it is "getter function" (might be confusing).
And I would refer to the spec to explain why we implement this fallback

Jaifroid: OK, will do.

Jaifroid:
I'll take a look at these extra places. Unless the code is very obvious, I might need to leave the unit tests to you, as I've never worked with them before.

mossroy:
OK no problem : I can handle the unit tests if you're not comfortable with them

Jaifroid:
> OK no problem : I can handle the unit tests if you're not comfortable with them

That would be great, thanks!

mossroy:
I commited the changes in Unit tests, and Travis seems happy with them

Jaifroid:
Hmm, checks have failed in this latest. Maybe we pushed too many changes too quickly?

www/js/lib/util.js
```diff
@@ -86,12 +86,12 @@ define(['q'], function(q) {
...
         array.sort(function(dirEntryA, dirEntryB) {
-            if (dirEntryA.title < dirEntryB.title) return -1;
-            if (dirEntryA.title > dirEntryB.title) return 1;
+            if (dirEntryA.getTitleOrUrl() < dirEntryB.getTitleOrUrl()) return -1;
+            if (dirEntryA.getTitleOrUrl() > dirEntryB.getTitleOrUrl()) return 1;
             return 0;
         });
         for(var i = 1; i < array.length; ){
-            if(array[i-1].title === array[i].title){
+            if(array[i-1].getTitleOrUrl() === array[i].getTitleOrUrl()){
```
Jaifroid: @mossroy Is this line the problem? I assumed this was an array of dirEntries, but maybe it isn't.

mossroy: Sorry, my bad : it's supposed to run on DirEntries, but the unit test does not (and it's unused in the actual code). You can leave it with .title for this PR.
Maybe we should remove this function and its unit test

Jaifroid:
No problem, I've reverted that commit. I'm fine about removing the unused function. A different PR?

mossroy:
It now looks good.
Yes, removing the unused function (and the corresponding unit test) would be better in a different PR

Jaifroid:
I don't think I have any further changes for this PR. When you're happy with it (after you've done any further testing you'd like to do), please approve the PR. I'll then squash/merge.

mossroy:
Sorry, I forgot to officially approve with the github feature. I did not test, but the code is OK for me

PR #490, opened 2019-05-02
Read mime type list from zim archive

Jaifroid:
~~This PR is *not* ready to merge. I'm opening it in order to test the prototype code and get help on it.~~

The code is in a working state (in jQuery mode so far), but has a problem I'd like help with. It reads the mimeTypeList correctly and stores it in a prototype Map. This works well, and prevents reading the binary data from the ZIM file multiple times per page load. Pages load successfully, and assets are assigned the correct MIME type.

~~The problem is that the map unexpectedly persists between ZIM file loads, which is not good. I don't know how to place it in the ZIM file scope. I shouldn't have to erase it explicitly on new archive load, because it's a prototype of  ZIMFile -- see zimfile.js line 114:~~

 ~~`ZIMFile.prototype.mimeTypeMapCache = new Map;`.~~

~~I'm fairly new to defining prototypes, so I'm probably doing something obviously wrong. Any clues?~~
EDIT: See below for solution

### DIFF
--- a/www/js/lib/zimArchive.js
+++ b/www/js/lib/zimArchive.js
@@ -116,16 +116,6 @@ define(['zimfile', 'zimDirEntry', 'util', 'utf8'],
     };
     
     /**
-     * Looks up the dirEntry's mimetype number in the ZIM file's MIME type list, and returns the corresponding MIME type
-     * 
-     * @param {Integer} mimetype The mimetype number stored in dirEntry.mimetype, used as a lookup value
-     * @return {String} The MIME type corresponding to mimetype in the ZIM file's MIME type list
-     */
-    ZIMArchive.prototype.getMimetype = function(mimetype) {
-        return this._file.mimeTypes.get(mimetype);
-    };
-    
-    /**
      * Looks for the DirEntry of the main page
      * @param {callbackDirEntry} callback
      * @returns {Promise} that resolves to the DirEntry
--- a/www/js/lib/zimArchive.js
+++ b/www/js/lib/zimArchive.js
@@ -116,11 +116,6 @@ define(['zimfile', 'zimDirEntry', 'util', 'utf8'],
     };
     
     /**
-     * @param {Map} mimeTypeList Defines a Map to hold the MIME Type list
-     */
-    ZIMArchive.prototype.mimeTypeList = new Map;
-
-    /**
      * Populates mimeTypeList from the ZIM archive if necessary, and looks up the MIME type string from the given dirEntry's mimetype
      * 
      * @param {Integer} mimetype The mimetype number stored in dirEntry.mimetype, used as a lookup value
@@ -128,10 +123,9 @@ define(['zimfile', 'zimDirEntry', 'util', 'utf8'],
      */
     ZIMArchive.prototype.getMimetypeString = function(mimetype, callback) {
         if (this.isReady()) {
-            if (this.mimeTypeList.size === 0) {
-                // The mimeTypeList hasn't been populated yet, so let's populate it
-                
-            }
+            this._file.mimeTypeMap().then(function(data) {
+                callback(data.get(mimetype));
+            });
         }
     };
     
--- a/www/js/lib/zimArchive.js
+++ b/www/js/lib/zimArchive.js
@@ -84,7 +84,7 @@ define(['zimfile', 'zimDirEntry', 'util', 'utf8'],
                 });
             }
         }
-    };
+    }
 
     /**
      * Searches the directory for all parts of a split archive.
@@ -113,6 +113,26 @@ define(['zimfile', 'zimDirEntry', 'util', 'utf8'],
      */
     ZIMArchive.prototype.isReady = function() {
         return this._file !== null;
+    };
+    
+    /**
+     * @param {Map} mimeTypeList Defines a Map to hold the MIME Type list
+     */
+    ZIMArchive.prototype.mimeTypeList = new Map;
+
+    /**
+     * Populates mimeTypeList from the ZIM archive if necessary, and looks up the MIME type string from the given dirEntry's mimetype
+     * 
+     * @param {Integer} mimetype The mimetype number stored in dirEntry.mimetype, used as a lookup value
+     * @param {Function} callback The function to call with the result
+     */
+    ZIMArchive.prototype.getMimetypeString = function(mimetype, callback) {
+        if (this.isReady()) {
+            if (this.mimeTypeList.size === 0) {
+                // The mimeTypeList hasn't been populated yet, so let's populate it
+                
+            }
+        }
     };
     
     /**
--- a/www/js/lib/zimArchive.js
+++ b/www/js/lib/zimArchive.js
@@ -116,7 +116,7 @@ define(['zimfile', 'zimDirEntry', 'util', 'utf8'],
     };
     
     /**
-     * Populates mimeTypeList from the ZIM archive if necessary, and looks up the MIME type string from the given dirEntry's mimetype
+     * Gets the mimeTypeList from the ZIM archive and looks up the MIME type string from the given dirEntry's mimetype
      * 
      * @param {Integer} mimetype The mimetype number stored in dirEntry.mimetype, used as a lookup value
      * @param {Function} callback The function to call with the result
--- a/www/js/lib/zimArchive.js
+++ b/www/js/lib/zimArchive.js
@@ -122,7 +122,7 @@ define(['zimfile', 'zimDirEntry', 'util', 'utf8'],
      * @return {String} The MIME type corresponding to mimetype in the ZIM file's MIME type list
      */
     ZIMArchive.prototype.getMimetype = function(mimetype) {
-            return this._file.mimeTypes.get(mimetype);
+        return this._file.mimeTypes.get(mimetype);
     };
     
     /**
--- a/www/js/lib/zimArchive.js
+++ b/www/js/lib/zimArchive.js
@@ -123,9 +123,7 @@ define(['zimfile', 'zimDirEntry', 'util', 'utf8'],
      */
     ZIMArchive.prototype.getMimetypeString = function(mimetype, callback) {
         if (this.isReady()) {
-            this._file.mimeTypeMap().then(function(data) {
-                callback(data.get(mimetype));
-            });
+            callback(this._file.mimeTypes.get(mimetype));
         }
     };
     
--- a/www/js/lib/zimArchive.js
+++ b/www/js/lib/zimArchive.js
@@ -116,12 +116,12 @@ define(['zimfile', 'zimDirEntry', 'util', 'utf8'],
     };
     
     /**
-     * Gets the mimeTypeList from the ZIM archive and looks up the MIME type string from the given dirEntry's mimetype
+     * Looks up the dirEntry's mimetype number in the ZIM file's MIME type list, and returns the corresponding MIME type
      * 
      * @param {Integer} mimetype The mimetype number stored in dirEntry.mimetype, used as a lookup value
-     * @return {String} The MIME Type corresponding to mimetype in the ZIM file's MIME Type list
+     * @return {String} The MIME type corresponding to mimetype in the ZIM file's MIME type list
      */
-    ZIMArchive.prototype.getMimetypeString = function(mimetype) {
+    ZIMArchive.prototype.getMimetype = function(mimetype) {
             return this._file.mimeTypes.get(mimetype);
     };
     
--- a/www/js/lib/zimArchive.js
+++ b/www/js/lib/zimArchive.js
@@ -119,12 +119,10 @@ define(['zimfile', 'zimDirEntry', 'util', 'utf8'],
      * Gets the mimeTypeList from the ZIM archive and looks up the MIME type string from the given dirEntry's mimetype
      * 
      * @param {Integer} mimetype The mimetype number stored in dirEntry.mimetype, used as a lookup value
-     * @param {Function} callback The function to call with the result
+     * @return {String} The MIME Type corresponding to mimetype in the ZIM file's MIME Type list
      */
-    ZIMArchive.prototype.getMimetypeString = function(mimetype, callback) {
-        if (this.isReady()) {
-            callback(this._file.mimeTypes.get(mimetype));
-        }
+    ZIMArchive.prototype.getMimetypeString = function(mimetype) {
+            return this._file.mimeTypes.get(mimetype);
     };
     
     /**
--- a/www/js/lib/zimDirEntry.js
+++ b/www/js/lib/zimDirEntry.js
@@ -31,7 +31,7 @@ define([], function() {
      * @property {File} _zimfile The ZIM file
      * @property {Boolean} redirect
      * @property {Integer} offset
-     * @property {Integer} mimetype MIME type number as defined in the MIME type list
+     * @property {Integer} mimetypeInteger MIME type number as defined in the MIME type list
      * @property {String} namespace defines to which namespace this directory entry belongs 
      * @property {Integer} redirectTarget
      * @property {Integer} cluster cluster number in which the data of this directory entry is stored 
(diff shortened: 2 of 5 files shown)


Jaifroid:
I think I solved the problem, copying the code that stores the other ZIMFile properties such as `zf.articleCount`, `zf.mainPage` etc. So I've now defined `zf.mimeTypes` as the map of MIME Types. It seems to work on initial testing.

I propose for this PR to be only the _implementation_ of the `mimeTypeMap` prototype, with a sample application of it in `app.js` for testing (see the `app.js` file in this PR). I would suggest that full _application_ of this implementation throughout `app.js`, SW mode, unit tests, etc., should be a separate PR.

www/js/lib/zimfile.js
```diff
@@ -110,6 +110,42 @@ define(['xzdec_wrapper', 'util', 'utf8', 'q', 'zimDirEntry'], function(xz, util,
...
+    ZIMFile.prototype.mimeTypeMap = function() {
+        var typeMap = this.mimeTypes;
+        // If we have already populated mimeTypes, we can just return it
+        if (typeMap.size) {
+            return Q.resolve().then(function() {
+                return typeMap;
+            });
+        } else {
+            // Otherwise, we have to populate it
+            return this._readSlice(this.mimeListPos, 256).then(function(data) {
```
Jaifroid: The function is set to read 256 bytes of data starting from `mimeListPos`. I have not found any MIME Type list that is longer than 130 bytes yet. Should we read 512 bytes, or even 1024 bytes, to be safe? Can we envisage any circumstance in which we have 1K of MIME Types defined?

www/js/lib/zimfile.js
```diff
@@ -110,6 +110,42 @@ define(['xzdec_wrapper', 'util', 'utf8', 'q', 'zimDirEntry'], function(xz, util,
...
+    /**
+     * Reads the whole MIME Type list and returns it as a populated Map
+     * 
+     * @return {Promise} A promise for the MIME Type list as a Map
+     */
+    ZIMFile.prototype.mimeTypeMap = function() {
+        var typeMap = this.mimeTypes;
+        // If we have already populated mimeTypes, we can just return it
+        if (typeMap.size) {
+            return Q.resolve().then(function() {
```
Jaifroid: Because `ZIMArchive.prototype.getMimetypeString` expects a promise, the result is here wrapped in a promise even though we could get it synchronously. Am I adding unnecessary overhead? It makes for a neat getter function in `zimArchive.js`, but maybe there's a faster way to return results synchronously, with a different architecture in `ZIMArchive.prototype`?

Jaifroid:
Tested and working on: Edge 44, Edge 75 (Chromium), Internet Explorer 11, Firefox ESR, FFOS.

Jaifroid:
This PR has now been rebased on the latest master for testing.

www/js/lib/zimfile.js
```diff
@@ -110,6 +110,42 @@ define(['xzdec_wrapper', 'util', 'utf8', 'q', 'zimDirEntry'], function(xz, util,
...
+     * The mimeTypeMap is extracted once after the user has picked the ZIM file
+     * and is stored as ZIMFile.mimetypes.
+     * 
+     * @returns {Promise} A promise for the MIME Type list as a Map
+     */
+    ZIMFile.prototype._mimeTypeMap = function() {
+        var typeMap = new Map;
+        return this._readSlice(this.mimeListPos, 256).then(function(data) {
+            // DEV: We have read 256 bytes: increase this if you encounter longer MIME type lists
+            // also change "while (pos < 255)" below
```
Jaifroid: The largest MIME type list I have found, in a recent ZIM, is 141 bytes. Is 256 bytes enough of a safety margin, or should I increase to 512?

Jaifroid:
@mossroy I think I have completed this code. The code is an implementation, but not a full application (there is only a test application in jQuery mode in app.js). This is intentional, as I think full application, especially to Service Worker code, may be complicated and will require a separate PR and testing (we'd have to pass the found MIME type to the SW via `postMessage`).

No rush, but when you are able, it would be great if you could review it. It's tested and working in FFOS simulator and all the other browsers I test on. It's also working in UWP.

www/js/lib/zimfile.js
```diff
@@ -110,6 +110,42 @@ define(['xzdec_wrapper', 'util', 'utf8', 'q', 'zimDirEntry'], function(xz, util,
...
+                var mimeString;
+                while (pos < 255) {
+                    pos++; 
+                    mimeString = utf8.parse(data.subarray(pos), true);
+                    // If the parsed data is an empty string, we have reached the end of the MIME type list, so break 
+                    if (!mimeString) break;
+                    // Store the parsed string in the Map
+                    typeMap.set(i, mimeString);
+                    i++;
+                    while (data[pos]) {
```
Jaifroid: This is safer than `while (data[pos] !== 0)` because if `pos` increments up to 256 (beyond the end of our data slice), then `data[pos]` will return `undefined`. If we just test `!==0` then we could end up with an infinite loop.

mossroy:
Sorry for the delay in looking at this PR and the other ones.
I'm quite busy for now, but will have a look at all this, as soon as I can

Jaifroid:
No problem, @mossroy , and absolutely no rush!

www/js/lib/zimArchive.js
```diff
@@ -115,6 +115,16 @@ define(['zimfile', 'zimDirEntry', 'util', 'utf8'],
         return this._file !== null;
     };
     
+    /**
+     * Looks up the dirEntry's mimetype number in the ZIM file's MIME type list, and returns the corresponding MIME type
+     * 
+     * @param {Integer} mimetype The mimetype number stored in dirEntry.mimetype, used as a lookup value
+     * @return {String} The MIME type corresponding to mimetype in the ZIM file's MIME type list
+     */
+    ZIMArchive.prototype.getMimetype = function(mimetype) {
```
mossroy: In terms of object-oriented programming, I think that resolving this integer to a string should be done in a DirEntry function. Because we have no reason to get a mimetype string from an arbitrary integer (outside of a DirEntry), and a DirEntry already knows its integer.
That's how it is done in libzim.
Each DirEntry knows its zimfile, so it should have access to the `mimeTypes` map.

BTW, we should try to be consistent in the `mimeType` or `mimetype` naming, if it's not complicated

www/js/lib/zimfile.js
```diff
@@ -110,6 +110,42 @@ define(['xzdec_wrapper', 'util', 'utf8', 'q', 'zimDirEntry'], function(xz, util,
...
     };
 
+    /**
+     * Reads the whole MIME type list and returns it as a populated Map
+     * The mimeTypeMap is extracted once after the user has picked the ZIM file
+     * and is stored as ZIMFile.mimeTypes
+     * 
+     * @returns {Promise} A promise for the MIME Type list as a Map
+     */
+    ZIMFile.prototype._mimeTypeMap = function() {
```
mossroy: This should be renamed with an action verb, as it's not a simple attribute. Something like `readMimeTypesMap`, for example

www/js/lib/zimfile.js
mossroy: In fileimpl.cpp (of libzim), they seem to know the exact size of what they need to read : getUrlPtrPos() - getMimeListPos()
We can probably do the same, which would be safer than to assume a maximum length.

www/js/lib/zimfile.js
mossroy: That's right.
Don't hesitate to copy the algorithm used in the libzim as it is the reference implementation

www/js/lib/zimfile.js
```diff
@@ -110,6 +110,42 @@ define(['xzdec_wrapper', 'util', 'utf8', 'q', 'zimDirEntry'], function(xz, util,
...
+                    typeMap.set(i, mimeString);
+                    i++;
+                    while (data[pos]) {
+                        pos++;
+                    }
+                }
+            }
+            return typeMap;
+        }).fail(function(err) {
+            console.errror('Unable to read MIME type list: ' + err);
```
mossroy: There are too many `r` here ;-)
And it's better to put the error as parameter instead of putting it in the string. It allows the browser dev tools to show it as an object.
I suggest `console.error('Unable to read MIME type list', err);`

www/js/lib/zimfile.js
```diff
@@ -238,6 +274,9 @@ define(['xzdec_wrapper', 'util', 'utf8', 'q', 'zimDirEntry'], function(xz, util,
                 zf.mimeListPos = readInt(header, 56, 8);
                 zf.mainPage = readInt(header, 64, 4);
                 zf.layoutPage = readInt(header, 68, 4);
+                zf._mimeTypeMap().then(function(data) {
+                    zf.mimeTypes = data;
```
mossroy: It would be safer to return the zf instance (through the Promise) after the `mimeTypes` map has been populated.
Else we get a race condition, where the zf could be read before we have the `mimeTypes` map

www/js/lib/zimArchive.js
Jaifroid: OK, I didn't know where to put it. `zimArchive` was a guess, because I reasoned that `zimFile` is for low-level stuff, and getting the MIME type would be a function exposed by the archive itself (as it depends on the archive being loaded). Happy to try putting it in dirEntry.

www/js/lib/zimfile.js
Jaifroid: OK

www/js/lib/zimfile.js
Jaifroid: I thought you might say that! I didn't know how to calculate it. Thanks for pointing me in the right direction in libzim. I always get lost whenever I look there (I don't know C++).

www/js/lib/zimfile.js
Jaifroid: OK. It's worth noting that there are several other places where `while (data[pos] !== 0)` is used in our code, which seems to me less safe (and more verbose) than `whlle (data[pos])`

www/js/lib/zimfile.js
Jaifroid: Ah, thanks for catching the typo. Will do.

www/js/lib/zimfile.js
Jaifroid: OK, I may need help with this. I struggled quite a lot to find a good place to put it so that the `mimeTypeMap` would only be read once and would persist, but so that it would automatically be voided when the archive is changed. Part of the problem is that we need `zf.mimeListPos` to be available and populated before `zf._mimeTypeMap()` can be called...

Jaifroid: @mossroy To do this I would need to duplicated the code that calculates `zf.mimeListPos` and `zf.urlPtrPos`, and I wouldn't be able to use `zf._readSlice`. These are needed to access the MIME type map.

In practice, the MIME type map is read and stored within a few ticks of the archive being loaded. We don't need any MIME type string until the dirEntry has been read, which is much, much later. In the meantime, the zf object is exported and available to other processes that (might) need it, which is actually an advantage. Nothing else attempts to read the MIME type map other than a dirEntry, and then only after the dirEntry has itself been read, a process that lies behind several later promises...

Does it seem OK? I have tested this code a lot!

mossroy: It's true that the risk is very very low in the current context.
But it's a race condition anyway, it's better to fix it because it might occur in a different context (later version of the app, or a unit test for example)

Could you try with the following piece of code : 
```
                return zf._readMimetypeMap().then(function(data) {
                    zf.mimeTypes = data;
                    return zf;
                });
```

www/js/lib/zimDirEntry.js
```diff
@@ -116,6 +116,15 @@ define([], function() {
...
         return this.title ? this.title : this.url;
     };
+    
+    /**
+     * Looks up the dirEntry's mimetype number in the ZIM file's MIME type list, and returns the corresponding MIME type
+     * 
+     * @return {String} The MIME type corresponding to mimetype in the ZIM file's MIME type list
+     */
+    DirEntry.prototype.getMimetype = function() {
+        return this._zimfile.mimeTypes.get(this.mimetype);
```
mossroy: For the ease of understanding, what would you think of renaming the `mimetype` attribute of a DirEntry to `mimetypeInteger` or something similar?
That would prevent from some confusion between it and this `getMimetype` function, that does not return the integer, but the corresponding string

Jaifroid: Fine by me! Will do.

www/js/lib/zimfile.js
Jaifroid: @mossroy I had already tried that, and various other combinations. It's not possible to return zf inside the second promise, because that just defers resolving zf, but JS execution still continues of course, and `ZIMArchive.prototype.isReady` returns true before zf is returned. As a result, we get an immediate exception in `ZIMArchive.prototype.getMainPageDirEntry`, which thinks the ZIM Archive is ready, before we've actually constructed the file object!

I also tried chaining the promises this way:
```
            return util.readFileSlice(fileArray[0], 0, 80).then(function(header) {
                var mimeListPos = readInt(header, 56, 8);
                var urlPtrPos = readInt(header, 32, 8);
                readMimetypeMap(fileArray[0], mimeListPos, urlPtrPos).then(function(data) {
                    var zf = new ZIMFile(fileArray);
                    zf.articleCount = readInt(header, 24, 4);
                    zf.clusterCount = readInt(header, 28, 4);
                    zf.urlPtrPos = urlPtrPos;
                    zf.titlePtrPos = readInt(header, 40, 8);
                    zf.clusterPtrPos = readInt(header, 48, 8);
                    zf.mimeListPos = mimeListPos;
                    zf.mainPage = readInt(header, 64, 4);
                    zf.layoutPage = readInt(header, 68, 4);
                    zf.mimeTypes = data;
                    return zf;
                });
            });
```
However, `ZIMArchive.prototype.isReady` *still* returns true *before* the last promise above is resolved. In fact, doing it this way introduces a race condition. (The above code does actually produce the MIME type table, with a modification in `_readMimetypeMap` to use `util.readFileSlice` instead of `this.readSlice`.)

I'm afraid I've tried everything I can think of!

Jaifroid: PS to get the above code to run, I had to refactor `_readMimetypeMap` into a normal function (not part of the prototype) which is why it appears above referenced as `readMimetypeMap`. I can push a commit with that "working" (apart from the exception in `ZIMArchive.prototype.getMainPageDirEntry`) if you think you can fix it yourself.

My head is hurting from trying to understand the interaction between Objects and Promises here...

Jaifroid: @mossroy Please ignore!! I was missing a pesky return in the above code! I think I can push a working version now...

Jaifroid: Resolved!

Jaifroid:
> It would be worth adding some unit tests on this newly implemented feature.
> For example we could check the mime-type (in string format) of :
> * one of the articles we read in an existing unit test
> * one of the images we read in an existing unit test
> I can help you on that if you wish

I think this is the last task now (assuming you approve of the new Promise chain).
I've never programmed a unit test. I would be very grateful if you could look after that part of this PR. Alternatively, I could add it as an issue to be done in due course (unless you think getting it running in the unit test is a good stress test of the code).

www/js/lib/zimfile.js
mossroy: Cool, I did not test but it looks good

mossroy:
Interesting : if I add some unit tests (on the Ray Charles archive), they all fail : 

- I/m/RayCharles_AManAndHisSoul.jpg has a MIME-type "image/gif" where it should be "image/jpeg"
- -/s/style.css has "image/svg+xml" instead of "text/css"
- -/j/local.js and A/Ray_Charles.html have undefined

So there are either bugs in this implementation, or many errors in the mime types inside this ZIM file.
This needs to be checked before merging this.

I'll commit my unit tests : they will fail, so it should make Travis complain

Jaifroid:
Hmm. I don't see how the logic could be at fault here, but I'll investigate.

mossroy:
Excellent!
Your commit fixed almost all unit tests, except one that I'll fix with another push on this PR

Jaifroid:
Thanks, @mossroy ! I was misled (or I'm a bit naïve) by the [OpenZIM specs](https://www.openzim.org/wiki/ZIM_file_format#MIME_Type_List_(mimeListPos)). Where it says "<1st MIME Type>" is at Offset 0, I took it to mean that mimetype = 1 was at offset 0. It should probably say "<0th MIME Type>" to be clear, or maybe all developers would understand that! So the map started at 1 when it shoudl have started at 0. I thought I'd tested it, but clearly not!

I'm very glad your idea of making a Unit Test picked this up.

mossroy:
Yes, that's one of the goals of a unit test. Plus the fact that it checks that nothing is broken when you refactor some code.
I think this PR can now be squashed/rebased/merged

Issue #484, opened 2019-04-02
Image map links are not parsed by parseAnchorsJQuery


Jaifroid:
Image maps work in SW mode, but in jQuery mode, they lead to an errorpage in the browser, which is ugly to recover from. An example would be the mammal image map in `Mammal.html` in English Wikipedia.

It shouldn't be difficult to parse these hrefs which have the format:

`<area title="Tasmanian devil" href="Tasmanian_devil.html" shape="rect" coords="0,68,100,136" alt="Tasmanian devil">`

mossroy:
Yes, it's also the case in the home page of some wikivoyage ZIM files (which have an image map of the earth).
It's probably easy to trap the click events on each area (in a generic way), and inject some javascript to replace the hyperlink, like we do with classical hyperlinks. See https://stackoverflow.com/questions/29921696/can-i-have-an-onclick-event-on-a-imagemap-area-element (but I did not test)

PR #492, opened 2019-05-03
Add support for area tag

Jaifroid:
This PR addresses #484. It turned out to be very easy.

Because I needed to use `e.preventDefault` (as per https://stackoverflow.com/questions/29921696/can-i-have-an-onclick-event-on-a-imagemap-area-element), it seemed best to change jQuery calls to pure JS in the `parseAnchorsJQuery` function, so as not to confuse styles. This function is now fully consistent with `insertMediaBlobsJQuery`, which also uses pure JS. Removal of jQuery makes it appear that there are more changes than there actually are, but on examination, it will be seen that they are merely stylistic changes.

Area maps and normal links are tested and working in this PR with `wikivoyage_en_all_novid_2019-03.zim` in Edge 44, Edge 75 (Chromium), IE11, Firefox ESR. I have also tested a Stackexchange ZIM to be sure that the use of subdirectories is unaffected by the removal of jQuery from `parseAnchorsJQuery`. I tested normal hyperlinks in FFOS simulator, but not a Wikiovoyage image map, becasue I don't have a Wikivoyage loaded in the fake SD card of FFOS.


Jaifroid:
Tests failed, but it looks like a problem with the dependencies in Travis, not with this PR (looks like the dependencies failed at boot up).

PR #493, opened 2019-05-05
Enable drag and drop of ZIM files

Jaifroid:
This is a (relatively) easy PR addressing an old issue #29.

The entire app is enabled as a drop target, so a file can be dropped anywhere except in the populated iframe. I experimented with trying to enable dropping onto the iframe, but it would have required injecting scripts into the frame, which seems excessive, since the file can easily be dropped anywhere above it. 

Once a file has been selected in this way, the file picker and instructions are hidden, but a link is provided to show them again if the user wants to use the file picker.

### DIFF
--- a/www/js/app.js
+++ b/www/js/app.js
@@ -602,6 +602,13 @@ define(['jquery', 'zimArchiveLoader', 'util', 'uiUtil', 'cookies','abstractFiles
             document.getElementById('downloadInstruction').style.display = 'none';
             document.getElementById('selectorsDisplay').style.display = 'inline';
         }
+        // Check for usable file types
+        for (var i = files.length; i--;) {
+            if (!/\.(?:zim\w{0,2}|dat|idx|txt)$/i.test(files[i].name)) {
+                alert("One or more files does not appear to be a ZIM file!");
+                return;
+            }    
+        }
         resetCssCache();
         selectedArchive = zimArchiveLoader.loadArchiveFromFiles(files, function (archive) {
             // The archive is set : go back to home page to start searching
--- a/www/js/app.js
+++ b/www/js/app.js
@@ -837,9 +837,13 @@ define(['jquery', 'zimArchiveLoader', 'util', 'uiUtil', 'cookies','abstractFiles
                     // The content is fully loaded by the browser : we can hide the spinner
                     $("#searchingArticles").hide();
                     // Deflect drag-and-drop of ZIM file on the iframe to Config
-                    var docBody = iframeArticleContent.contentDocument.documentElement.getElementsByTagName('body')[0];
-                    docBody.addEventListener('dragover', handleIframeDragover);
-                    docBody.addEventListener('drop', handleIframeDrop);
+                    var doc = iframeArticleContent.contentDocument.documentElement;
+                    var docBody = doc ? doc.getElementsByTagName('body') : null;
+                    docBody = docBody ? docBody[0] : null;
+                    if (docBody) {
+                        docBody.addEventListener('dragover', handleIframeDragover);
+                        docBody.addEventListener('drop', handleIframeDrop);
+                    }
                     iframeArticleContent.contentWindow.onunload = function() {
                         $("#searchingArticles").show();
                     };
@@ -981,15 +985,20 @@ define(['jquery', 'zimArchiveLoader', 'util', 'uiUtil', 'cookies','abstractFiles
             $('#articleListHeaderMessage').empty();
             $('#articleListWithHeader').hide();
             $("#prefix").val("");
+            
             // Inject the new article's HTML into the iframe
             var articleContent = iframeArticleContent.contentDocument.documentElement;
             articleContent.innerHTML = htmlArticle;
-            // Add any missing classes stripped from the <html> tag
-            var docBody = articleContent.getElementsByTagName('body')[0];
-            if (htmlCSS) docBody.classList.add(htmlCSS);
-            // Deflect drag-and-drop of ZIM file on the iframe to Config
-            docBody.addEventListener('dragover', handleIframeDragover, false);
-            docBody.addEventListener('drop', handleIframeDrop);
+            
+            var docBody = articleContent.getElementsByTagName('body');
+            docBody = docBody ? docBody[0] : null;
+            if (docBody) {
+                // Add any missing classes stripped from the <html> tag
+                if (htmlCSS) docBody.classList.add(htmlCSS);
+                // Deflect drag-and-drop of ZIM file on the iframe to Config
+                docBody.addEventListener('dragover', handleIframeDragover);
+                docBody.addEventListener('drop', handleIframeDrop);
+            }
 
             // Allow back/forward in browser history
             pushBrowserHistoryState(dirEntry.namespace + "/" + dirEntry.url);
--- a/www/js/app.js
+++ b/www/js/app.js
@@ -586,7 +586,7 @@ define(['jquery', 'zimArchiveLoader', 'util', 'uiUtil', 'cookies','abstractFiles
     function displayFileSelect() {
         document.getElementById('openLocalFiles').style.display = 'block';
         // Set the main drop zone
-        configDropZone.addEventListener('dragover', handleGlobalDragover, false);
+        configDropZone.addEventListener('dragover', handleGlobalDragover);
         configDropZone.addEventListener('dragleave', function(e) {
             configDropZone.style.border = '';
         });
@@ -595,7 +595,7 @@ define(['jquery', 'zimArchiveLoader', 'util', 'uiUtil', 'cookies','abstractFiles
             e.preventDefault();
             if (configDropZone.style.display === 'none') document.getElementById('btnConfigure').click();
             e.dataTransfer.dropEffect = 'link';
-        }, false);
+        });
         globalDropZone.addEventListener('drop', handleFileDrop);
         // This handles use of the file picker
         document.getElementById('archiveFiles').addEventListener('change', setLocalArchiveFromFileSelect);
--- a/www/js/app.js
+++ b/www/js/app.js
@@ -583,13 +583,15 @@ define(['jquery', 'zimArchiveLoader', 'util', 'uiUtil', 'cookies','abstractFiles
         document.getElementById('openLocalFiles').style.display = 'block';
         // Make the whole app a drop zone
         var dropZone = document.getElementById('search-article');
-        dropZone.addEventListener('dragover', function(e) {
-            e.stopPropagation();
-            e.preventDefault();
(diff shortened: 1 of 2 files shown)


www/js/app.js
```diff
@@ -580,17 +580,49 @@ define(['jquery', 'zimArchiveLoader', 'util', 'uiUtil', 'cookies','abstractFiles
...
     function displayFileSelect() {
-        $('#openLocalFiles').show();
-        $('#archiveFiles').on('change', setLocalArchiveFromFileSelect);
+        document.getElementById('openLocalFiles').style.display = 'block';
+        // Make the whole app a drop zone
+        var dropZone = document.getElementById('search-article');
+        dropZone.addEventListener('dragover', function(e) {
+            e.stopPropagation();
+            e.preventDefault();
+            e.dataTransfer.dropEffect = 'copy';
```
mossroy: Wouldn't the 'link' dropEffect be more appropriate?
Considering the size of ZIM files, a 'copy' icon made me reluctant to drop the files, thinking it would start copying a lot of GB.

www/js/app.js
```diff
@@ -580,17 +580,49 @@ define(['jquery', 'zimArchiveLoader', 'util', 'uiUtil', 'cookies','abstractFiles
...
+        }, false);
+        dropZone.addEventListener('drop', setLocalArchiveFromFileList);
+        document.getElementById('archiveFiles').addEventListener('change', setLocalArchiveFromFileSelect);
     }
 
     function setLocalArchiveFromFileList(files) {
+        // Check to see if the files are from file select or from drag-and-drop
```
mossroy: Maybe it would more clear to create a dedicated function like `setLocalArchiveFromDragAndDroppedFiles`, that takes the files from the dataTransfer, and then calls `setLocalArchiveFromFileList`

www/js/app.js
```diff
@@ -580,17 +580,49 @@ define(['jquery', 'zimArchiveLoader', 'util', 'uiUtil', 'cookies','abstractFiles
...
+            files.stopPropagation();
+            files.preventDefault();
+            files = files.dataTransfer.files;
+            document.getElementById('openLocalFiles').style.display = 'none';
+            document.getElementById('downloadInstruction').style.display = 'none';
+            document.getElementById('selectorsDisplay').style.display = 'inline';
+        }
+        // Check for usable file types
+        for (var i = files.length; i--;) {
+            if (!/\.(?:zim\w{0,2}|dat|idx|txt)$/i.test(files[i].name)) {
```
mossroy: I seem to remember that the dat, idx and txt extensions were for the old Evopedia archive format. Maybe we only need the .zim* files here?

www/js/app.js
Jaifroid: @mossroy Good idea, I'll try it out!

www/js/app.js
Jaifroid: OK, will see what I can do.

www/js/app.js
Jaifroid: Yes, I was wondering about them. It sounds like those extensions are obsolete. I'll remove.

Jaifroid:
> I also don't see the purpose of hiding the file select after a file drop. It adds some code and logic to show/hide it, that I don't find necessary. But maybe I'm wrong.

The purpose of hiding the file selector after the drop is:

* Any previously picked file (using picker) remains showing in the picker even after the new dropped file is loaded, and I find that misleading (but I also didn't explore much how to change that, if it's possible);
* The user is more likely, having used drag-and-drop, to continue to use that method to launch different ZIM files, so it streamlines the interface.

In reality that part is adapted from Kiwix JS Windows, where there are an awful lot of options, so I try to declutter the interface if a user has chosen drag-and-drop (or double-click on a ZIM file, which is also possible in app context, but not in browser context I think).

If you know how to clear/set what's shown in the picker, that would be a good alternative, given that there is less of an argument for streamlining in Kiwix JS, at least until we add more options (like dark mode, display scaling, etc.).

www/js/app.js

www/js/app.js
Jaifroid: Done, I hope, but please review in due course.

mossroy:
OK thanks for the explanation.
You're right about the filepicker keeping the old file name.
I did not test, but, based on https://stackoverflow.com/questions/1703228/how-can-i-clear-an-html-file-input-with-javascript, setting the value of the file picker to empty or null should be enough to reset it (it does not work on IE<=10, but we don't care).

In any case, I think that the explanations should remain visible, even after a file drop.

Another minor improvement that could be added (but it's not compulsory) would be to give another visual feedback to the user when he is dragging a file over the content. I saw some websites (JIRA Help Desk) that surround the drop zone with a large dotted line (only when the user drags a file on it) : it helps the user to know that the mouse is in the right place, and that the dropped file will be processed

Jaifroid:
Just a quick note to say that I've pushed a commit that makes drag-and-drop also possible into the iframe area in jQuery mode (turned out to be simple). I haven't attempted to implement same for SW mode because it would require similar code to attach the events to the body of the document, and I don't even know if that's possible when it's running in a Service Worker. (Should be?)

It would be nice if we could *prevent* drop into the iframe in SW mode, but I can't find a way to do that. Currently, if a user drops a ZIM into the iframe area, the app reacts (in Chromium) by opening a Save As dialogue (this must be a default action when the browser doesn't know what to do with a file). I haven't tested this in the other browsers specifically.

Jaifroid:
OK, I've now added the dotted red border for a drop zone, and I've solved the issue of dropping to the iframe, by deflecting any attempted drag over the iframe into the Config area.

I discovered that although it's possible to drop a file into the iframe when the app is using the file:// protocol, we get permission denied errors if we try to do this when the app is running from localhost (and hence also in SW mode). Therefore, I hit upon the idea of forcing a click on the Configuration button when dragging a file over the iframe area. As the dragover event must occur before the drop event, this means that even a quick drop will be made to the Config page. This gets rid of any permission errors, and makes for a consistent feel.

You will note that I have to attach the dragover event to the body of the document in the iframe in SW mode. I haven't yet been able to find any target which works that is not in the iframe's document, but with this new architecture it should be easy to experiment.

Jaifroid:
> In any case, I think that the explanations should remain visible, even after a file drop.

@mossroy I've left an instruction now after drop. Is it sufficient, or would you prefer me to get rid of the logic that hides the file selector? It could be useful if we have more features/options in the future, to declutter the interface.

www/js/app.js
```diff
@@ -768,10 +833,16 @@ define(['jquery', 'zimArchiveLoader', 'util', 'uiUtil', 'cookies','abstractFiles
...
                 $('#articleListWithHeader').hide();
                 $("#prefix").val("");
-                iframeArticleContent.onload = function () {
+                iframeArticleContent.onload = function() {
                     // The content is fully loaded by the browser : we can hide the spinner
-                    iframeArticleContent.onload = function () {};
                     $("#searchingArticles").hide();
+                    // Deflect drag-and-drop of ZIM file on the iframe to Config
+                    var docBody = iframeArticleContent.contentDocument.documentElement.getElementsByTagName('body')[0];
+                    docBody.addEventListener('dragover', handleIframeDragover);
```
mossroy: Maybe it would be safer to check that all intermediate variables are defined before using them.

www/js/app.js
```diff
@@ -914,7 +985,12 @@ define(['jquery', 'zimArchiveLoader', 'util', 'uiUtil', 'cookies','abstractFiles
             var articleContent = iframeArticleContent.contentDocument.documentElement;
             articleContent.innerHTML = htmlArticle;
             // Add any missing classes stripped from the <html> tag
-            if (htmlCSS) articleContent.getElementsByTagName('body')[0].classList.add(htmlCSS);
+            var docBody = articleContent.getElementsByTagName('body')[0];
+            if (htmlCSS) docBody.classList.add(htmlCSS);
+            // Deflect drag-and-drop of ZIM file on the iframe to Config
+            docBody.addEventListener('dragover', handleIframeDragover, false);
```
mossroy: Why is there a last `false` argument here, where there isn't on line 841?

www/js/app.js
```diff
@@ -768,10 +833,16 @@ define(['jquery', 'zimArchiveLoader', 'util', 'uiUtil', 'cookies','abstractFiles
                 $('#articleListHeaderMessage').empty();
                 $('#articleListWithHeader').hide();
                 $("#prefix").val("");
-                iframeArticleContent.onload = function () {
+                iframeArticleContent.onload = function() {
                     // The content is fully loaded by the browser : we can hide the spinner
-                    iframeArticleContent.onload = function () {};
```
mossroy: I don't remember why this line was necessary in the past. Maybe it's not necessary any more, I suppose you checked?

Jaifroid: It was necessary to prevent the onload function happening every time an article is loaded in the Service Worker. However, for this code to work, we now need that function to run. It has no negative effects, and we are not altering the html of the document loaded by the SW, but we are attaching an event to the body tag of that document, once it has completed loading. I found no other way of monitoring a dragover on the iframe (and I tried quite a few ways), attaching it to the contentWindow, attaching it to the contentDocument, attaching it to the iframe itself, none worked.

www/js/app.js
Jaifroid: OK, will do. You're right, we could get unhandled exceptions.

www/js/app.js
mossroy: All right, thanks for the explanation

www/js/app.js
Jaifroid: The code I was using as a model had these. I hadn't actually checked what it does. I see false means "The event handler is executed in the bubbling phase". However, it's also the default! So I'll experiment with removing them.

www/js/app.js
```diff
@@ -768,10 +833,20 @@ define(['jquery', 'zimArchiveLoader', 'util', 'uiUtil', 'cookies','abstractFiles
...
                     $("#searchingArticles").hide();
+                    // Deflect drag-and-drop of ZIM file on the iframe to Config
+                    var doc = iframeArticleContent.contentDocument.documentElement;
+                    var docBody = doc ? doc.getElementsByTagName('body') : null;
+                    docBody = docBody ? docBody[0] : null;
```
Jaifroid: @mossroy Is this enough checking of intermediate variables? I guess I could check that the iframe has a `contentDocument` before checking for the `documentElement`, but elsewhere we just do `iframeArticleContent.contentDocument.documentElement`, so it seems unnecessary.

I use the same code in line 994 below.

mossroy: Yes, I think it's enough. I wanted to cover at least the case where the iframe would be empty

Issue #495, opened 2019-05-10
Use a flat model for resolving hyperlinks in jQuery mode


Jaifroid:
This is a thought exercise provoked by the issues in #494. I am in two minds myself, but it occurs to me and I've tried it out in Kiwix JS Windows in a test branch.

I am beginning to wonder if we are making too much work for ourselves in trying to emulate what the browser does when resolving hyperlinks in jQuery mode. I suggest we consider reverting to a "flat" model, by which I mean that we don't forcibly try to situate the browser inside any subdirectories by using the `<base>` tag. Since we have to use JS anyway to resolve all ZIM links, we don't need the browser to calculate relative links for us. It is adding a lot of support work -- see #487, #346, #484 and generally the fact that we have to programme every anchor (e.g. `<a href="#main_topic" ...>`).

ISTM that we could simplify quite a lot if we just let the browser think it's in the location of `article.html`. ~~We know how to find a ZIM link and whether it is in the A namespace, and any other namespaces will be provided explicitly. All we need is the URL (and namespace if the link has one) with no relative links attached. It's exactly the same as what we do with images: the browser doesn't really add any useful information by adding on extra ../../../ when it's in a "subdirectory". We can throw all that away and extract the ZIM link directly from the arcihive.~~

In fact, I now realize that SW mode operates a flat model too, which is why #494 doesn't affect it. The code `titleWithNameSpace = nameSpace + '/' + title;` (line 163 of `service-worker.js`) throws away any relative link info and just asks the back end for that ZIM URL (e.g. `I/myImage.jpeg`). This caused me a lot of puzzling over why obviously incorrect relative links in `Singapore/Riverside` were working fine in SW mode. Shouldn't jQuery mode use the same principle?

~~**An important caveat:** On the other side of the argument, if we had already implemented this, then we would never have caught the errors in #494 . So there is a useful diagnostic element to doing things the way we currently do in jQuery mode.~~

Jaifroid:
This has already been discussed at some length in https://github.com/kiwix/kiwix-js/pull/309 .

Jaifroid:
The PR #496 is a generic alternative to using the base tag. It simplifies the proliferation of link handling that the base tag was causing. Contrary to my (incorrect) caveat above, this code maintains the diagnostic / reference status of Kiwix JS.

mossroy:
If I understood correctly, the goal is to remove the `base` tag, in order to simplify the code?
It sounds like a good idea.

PR #496, opened 2019-05-12
Remove dependency on base URL

Jaifroid:
This is proof of concept relating to #495 . It simplifies the parsing and programming of URLs in `parseAnchorsJQuery()`.

### DIFF
--- a/www/js/app.js
+++ b/www/js/app.js
@@ -936,7 +936,7 @@ define(['jquery', 'zimArchiveLoader', 'util', 'uiUtil', 'cookies','abstractFiles
     // Pattern to find the path in a url
     var regexpPath = /^(.*\/)[^\/]+$/;
     // Pattern to find a ZIM URL (with its namespace) - see https://wiki.openzim.org/wiki/ZIM_file_format#Namespaces
-    var regexpZIMUrlWithNamespace = /(?:^|\/)([-ABIJMUVWX]\/.+)/;
+    var regexpZIMUrlWithNamespace = /^[.\/]*([-ABIJMUVWX]\/.+)$/;
     // Regex below finds images, scripts, stylesheets and media sources with ZIM-type metadata and image namespaces [kiwix-js #378]
     // It first searches for <img, <script, <link, etc., then scans forward to find, on a word boundary, either src=["']
     // or href=["'] (ignoring any extra whitespace), and it then tests the path of the URL with a non-capturing lookahead that
@@ -978,12 +978,6 @@ define(['jquery', 'zimArchiveLoader', 'util', 'uiUtil', 'cookies','abstractFiles
         // This replacement also processes the URL to remove the path so that the URL is ready for subsequent jQuery functions
         htmlArticle = htmlArticle.replace(regexpTagsWithZimUrl, '$1data-kiwixurl$2');
 
-        // Compute base URL
-        var urlPath = regexpPath.test(dirEntry.url) ? urlPath = dirEntry.url.match(regexpPath)[1] : '';
-        var baseUrl = dirEntry.namespace + '/' + urlPath;
-
-        // Inject base tag into html
-        htmlArticle = htmlArticle.replace(/(<head[^>]*>\s*)/i, '$1<base href="' + baseUrl + '" />\r\n');
         // Extract any css classes from the html tag (they will be stripped when injected in iframe with .innerHTML)
         var htmlCSS = htmlArticle.match(/<html[^>]*class\s*=\s*["']\s*([^"']+)/i);
         htmlCSS = htmlCSS ? htmlCSS[1] : '';
@@ -1043,7 +1037,7 @@ define(['jquery', 'zimArchiveLoader', 'util', 'uiUtil', 'cookies','abstractFiles
             // Pattern to match a local anchor in an href even if prefixed by escaped url; will also match # on its own
             var regexpLocalAnchorHref = new RegExp('^(?:#|' + escapedUrl + '#)([^#]*$)');
             var iframe = iframeArticleContent.contentDocument;
-            Array.prototype.slice.call(iframe.querySelectorAll('a, area')).forEach(function (anchor) { 
+            Array.prototype.slice.call(iframe.querySelectorAll('a, area')).forEach(function (anchor) {
                 // Attempts to access any properties of 'this' with malformed URLs causes app crash in Edge/UWP [kiwix-js #430]
                 try {
                     var testHref = anchor.href;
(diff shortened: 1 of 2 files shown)


Jaifroid:
Advantages of this PR:

* It simplifies the parsing of anchors in `parseAnchorsJQuery()`;
* We do not need to inject a base tag into the incoming HTML;
* Far fewer anchors need to have a click event listener attached; 
* It is generic, as detailed in the [JSDoc](https://github.com/kiwix/kiwix-js/pull/496/files#diff-764ebb2c13f00a104681b9e5f0327d3bR89) of the `uiUtil.URL()` function;
* Code to parse a ZIM URL is only run in the click event, not in the page loading stage;
* It remains diagnostic.

Disadvantages of this PR:

* It is doing something the browser should be able to do by itself;
* It doesn't handle every imaginable format of a ZIM URL, though it flexibly handles the known ones listed in the JSDoc.

In response to these disadvantages:

* Although the browser ought to be able to do it, we have seen that this involves a lot of extra coding to account for the base URL;
* We know [the code that does the URL rewriting](https://github.com/openzim/mwoffliner/commit/dc5a9f047ba53bff695b0afe828b8defa141b88c#diff-f2248b5a374d47921fd284b8c73f64ba), at least for mwoffliner, and we have not to my knowledge encountered other formats.

mossroy:
This looks promising.
But we can't rely on how mwoffliner works to implement that.
We definitely need to handle all URLs in a generic way, regardless of how the ZIM file was generated.

I did not have time to have a close look at the code, but you seem to try to resolve a relative URL to an absolute URL, and to add a base to the result.
This can probably be done with a built-in feature of browsers, using the URL interface : https://developer.mozilla.org/en-US/docs/Web/API/URL
A few examples can be found in https://stackoverflow.com/questions/14780350/convert-relative-path-to-absolute-using-javascript or https://stackoverflow.com/questions/3943281/resolving-relative-urls-in-javascript

Jaifroid:
As you no-doubt noticed, I basically implemented my own URL interface. I did code a version using the URL feature. It is not compatible with IE11 (the browser doesn't support URL as a function). I worked out another way to do it for IE11, with [createHTMLDocument](https://developer.mozilla.org/en-US/docs/Web/API/DOMImplementation/createHTMLDocument), adding a base href to it. But then when I saw how easy it is to do with regular expressions, I thought I could produce something generic and fast that way. You know I like regular expressions :-) ... I'll add the other method to this PR, which I think you'll prefer!

Jaifroid:
New commit uses the URL function and a workaround for IE11. Not as elegant as my regular expressions!!! But it works, and it leaves the job to the browser. Works (obviously) in IE11 and Edge. I haven't tested much on the other browsers yet.

Jaifroid:
>  but you seem to try to resolve a relative URL to an absolute URL

Not quite. I take a relative URL like `../Einstein` or `../../I/im%C3%A1gen.png` and I return a *ZIM URL*, e.g. `A/Einstein` or `I/imágen.png`.

Both versions of the code ([they can be compared here](https://github.com/kiwix/kiwix-js/pull/496/commits/adec808c99884c077eac7f470922e5b7ce44d1c4#diff-764ebb2c13f00a104681b9e5f0327d3b)) do the exact same thing. Just the first one does it using regexes and the second one does it using the URL API and a "polyfill" for IE11. They both work with all types of links found in Stackexchange ZIMs as well as with Wikimedia ZIMs.

www/js/lib/uiUtil.js
```diff
@@ -84,6 +84,30 @@ define([], function() {
...
 
+    /**
+     * Derives the URL.pathname from a relative or semi-relative URL using the given base ZIM URL
+     * 
+     * @param {String} url The (URI-encoded) URL to convert (e.g. "Einstein", "../Einstein",
+     *      "../../I/im%C3%A1gen.png", "-/s/style.css", "/A/Einstein.html")
+     * @param {String} base The base ZIM URL of the currently loaded article (e.g. "A/" or "A/subdir1/subdir2/")
+     * @returns {String} The derived ZIM URL in decoded form (e.g. "A/Einstein", "I/imágen.png")
+     */
+    function deriveURL(url, base) {
```
mossroy: I think this function should be renamed to something more meaningful, like `constructZimUrlFromRelativeUrl`

www/js/lib/uiUtil.js
```diff
@@ -84,6 +84,30 @@ define([], function() {
...
+        // We use a dummy domain because URL API requires a valid URI
+        var dummy = 'http://d/';
+        var derive = function(url, base) {
```
mossroy: This function might be renamed, too

www/js/lib/uiUtil.js
```diff
@@ -181,6 +205,7 @@ define([], function() {
     return {
         feedNodeWithBlob: feedNodeWithBlob,
         replaceCSSLinkWithInlineCSS: replaceCSSLinkWithInlineCSS,
+        URL: deriveURL,
```
mossroy: It would probably be better to expose the same function name to the outside as the original function name

Jaifroid: My idea in exposing it as "URL" was that I'm doing a poorman's version of a URL polyfill for IE11. If we weren't supporting IE11, we wouldn't actually need a function to do this, we could easily just use the URL API.

Given this reasoning, would you still prefer me to expose it publicly with the new name for the function (`uiUtil.deriveZimUrlFromRelativeUrl`), or keep `uiUtil.URL` in order to show its relationship to the URL API? (The reason I can't call it `URL` inside `uiUtil.js` is because it would conflict with the URL API.)

I did look into adding this function into the prototype of the URL function for IE11, but it looked like some extra coding for the same result, and it didn't seem justified.

mossroy: I understand, but it's not really a polyfill : as you said, it also has some ZIM-specific logic.
So I'd stick to a classical function.
Another option would be to use a real polyfill, but I agree with you that it's probably overkill for this simple need

Jaifroid: OK, done!

Jaifroid:
Please note I rebased this branch on master to enable some radical code simplification of the handling of media BLOBs with `deriveZimUrlFromRelativeUrl`.

Jaifroid:
Oh, I thought I'd checked subtitle handling with this PR in jQuery mode, at least on Edge 44 and Edge 76. I might have missed Firefox. I'll investigate. Thanks for checking!

www/js/app.js
```diff
@@ -936,14 +936,14 @@ define(['jquery', 'zimArchiveLoader', 'util', 'uiUtil', 'cookies','abstractFiles
...
-    // Regex below finds images, scripts, stylesheets and media sources with ZIM-type metadata and image namespaces [kiwix-js #378]
+    var regexpZIMUrlWithNamespace = /^[.\/]*([-ABIJMUVWX]\/.+)$/;
+    // Regex below finds images, scripts, and stylesheets with ZIM-type metadata and image namespaces [kiwix-js #378]
     // It first searches for <img, <script, <link, etc., then scans forward to find, on a word boundary, either src=["']
     // or href=["'] (ignoring any extra whitespace), and it then tests the path of the URL with a non-capturing lookahead that
     // matches ZIM URLs with namespaces [-IJ] ('-' = metadata or 'I'/'J' = image). When the regex is used below, it will also
     // remove any relative or absolute path from ZIM-style URLs.
     // DEV: If you want to support more namespaces, add them to the END of the character set [-IJ] (not to the beginning) 
-    var regexpTagsWithZimUrl = /(<(?:img|script|link|video|audio|source|track)\b[^>]*?\s)(?:src|href)(\s*=\s*["'])(?:\.\.\/|\/)+(?=[-IJ]\/)/ig;
+    var regexpTagsWithZimUrl = /(<(?:img|script|link)\b[^>]*?\s)(?:src|href)(\s*=\s*["'])(?:\.\.\/|\/)+(?=[-IJ]\/)/ig;
```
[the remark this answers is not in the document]
Jaifroid: There was a reason for changing this handling, which was to do with preventing insertMediaBlobsjQuery from needlessly cycling over ever track inside the video DOM. I'll check my logic...

Jaifroid:
Just to say, I confirm the regression with subtitles in Firefox ESR. Strangely, it seems to be a Firefox-specific issue. With Edge and Chromium, it doesn't occur: subtitles work normally with this PR...

Jaifroid:
OK, this was a bizarre bug to fix. It turns out that Firefox (at least the ESR version), tries to load all text tracks with an src before the associated video is loaded. This of course fails, because we have not yet extracted the text tracks. Then, once the text tracks have been extracted, Firefox gets confused and loads several of them when a user selects a specific track.

The solution was to revert to the method of hiding the src for text tracks. Because `regexpTagsWithZimUrl` also removes any relative prefixes (`../../`) in the URL, it is no longer possible to derive the correct URL using `deriveZimUrlFromRelativeUrl()`, and text tracks with a ZIM URL have to be exempted from processing by that function. Fortunately, `regexpTagsWithZimUrl` does not process text tracks that are relative to the loading document (e.g. `Free-software-free-society-Richard-Stallman-at-TEDxGeneva-2014/es-ES.vtt`), so these do (correctly) get processed by `deriveZimUrlFromRelativeUrl()`.

So, it seems to be fixed. @mossroy , can you confirm?

mossroy:
If you're satisfied with this PR, it would be better to remove the [WIP] in its title

PR #498, opened 2019-05-15
Deal correctly with forward slashes in article URL in SW mode

### LINKED ISSUE #494: Deal correctly with article titles containing a forward slash
Currently, on master, `parseAnchorsJQuery` gets confused by titles such as `A/Singapore/Riverside`. It thinks that it is dealing with a subdirectory, when in fact it is dealing with a title `Singapore/Riverside`. Hyperlinks on that page therefore do not work, as the base html tag is incorrectly set (I believe). We perhaps are not percent-encoding the title at the right point (or maybe we percent-decode it too early).

Jaifroid:
This is a simple PR that fixes #494 in SW mode while still allowing #428 to work.
At first I thought the solution was to use `encodeURI` instead of `encodeURIComponent` as per https://stackoverflow.com/a/12650216/9727685. However, it turns out that `encodeURI`, while not encoding the forward slashes, also doesn't encode `$` and other special characters that need to be encoded in order for #428 to work. So, I have to apply `encodeURIComponent` only to text between the slashes, which is what this PR does.

Tested on Wikipedia's `Presidential $1 Coin` article, and also tested on Stackexchange ZIMs as discussed in #494. So far, only tested on Edge.

### DIFF
--- a/www/js/app.js
+++ b/www/js/app.js
@@ -888,12 +888,19 @@ define(['jquery', 'zimArchiveLoader', 'util', 'uiUtil', 'cookies','abstractFiles
 
     /**
      * Read the article corresponding to the given dirEntry
-     * @param {DirEntry} dirEntry
+     * @param {DirEntry} dirEntry The directory entry of the article to read
      */
     function readArticle(dirEntry) {
         if (contentInjectionMode === 'serviceworker') {
             // In ServiceWorker mode, we simply set the iframe src.
             // (reading the backend is handled by the ServiceWorker itself)
+
+            // We will need the encoded URL on article load so that we can set the iframe's src correctly,
+            // but we must not encode the '/' character or else relative links may fail [kiwix-js #498]
+            var encodedUrl = dirEntry.url.replace(/[^/]+/g, function(matchedSubstring) {
+                return encodeURIComponent(matchedSubstring);
+            });
+            
             var iframeArticleContent = document.getElementById('articleContent');
             iframeArticleContent.onload = function() {
                 // The iframe is empty, show spinner on load of landing page
@@ -918,7 +925,7 @@ define(['jquery', 'zimArchiveLoader', 'util', 'uiUtil', 'cookies','abstractFiles
                     };
                 };
                 // We put the ZIM filename as a prefix in the URL, so that browser caches are separate for each ZIM file
-                iframeArticleContent.src = selectedArchive._file._files[0].name + "/" + dirEntry.namespace + "/" + encodeURIComponent(dirEntry.url);
+                iframeArticleContent.src = selectedArchive._file._files[0].name + "/" + dirEntry.namespace + "/" + encodedUrl;
                 // Display the iframe content
                 $("#articleContent").show();
             };
--- a/www/js/app.js
+++ b/www/js/app.js
@@ -925,7 +925,7 @@ define(['jquery', 'zimArchiveLoader', 'util', 'uiUtil', 'cookies','abstractFiles
                     };
                 };
                 // We put the ZIM filename as a prefix in the URL, so that browser caches are separate for each ZIM file
-                iframeArticleContent.src = '../' + selectedArchive._file._files[0].name + "/" + dirEntry.namespace + "/" + encodedUrl;
+                iframeArticleContent.src = "../" + selectedArchive._file._files[0].name + "/" + dirEntry.namespace + "/" + encodedUrl;
                 // Display the iframe content
                 $("#articleContent").show();
             };
--- a/www/js/app.js
+++ b/www/js/app.js
@@ -925,7 +925,7 @@ define(['jquery', 'zimArchiveLoader', 'util', 'uiUtil', 'cookies','abstractFiles
                     };
                 };
                 // We put the ZIM filename as a prefix in the URL, so that browser caches are separate for each ZIM file
-                iframeArticleContent.src = selectedArchive._file._files[0].name + "/" + dirEntry.namespace + "/" + encodedUrl;
+                iframeArticleContent.src = '../' + selectedArchive._file._files[0].name + "/" + dirEntry.namespace + "/" + encodedUrl;
                 // Display the iframe content
                 $("#articleContent").show();
             };


Jaifroid:
Additionally I discovered we have been setting the iframe's src incorrectly when searching for an article. We had:
```
iframeArticleContent.src = dirEntry.namespace + "/" + encodedUrl;
```
when we should have had:
```
iframeArticleContent.src = '../' + dirEntry.namespace + "/" + encodedUrl;
```
This became obvious when I was experimenting with inserting an external resource (image) into the iframe, and the image could not be found when the iframe had been set by searching for an article, but could be found when following a hyperlink.

mossroy:
As explained in #506, I don't reproduce the issue that leads to this PR, so it's difficult to review it for now.

Jaifroid:
We should probably await new ZIMs that treat, e.g. "Singapore/Riverside" as a subdirectory. In the previous ZIM, all hyperlinks on the page were missing an extra "../", yet they worked in Service Worker mode because of one of the two issues corrected in this PR. The code has been corrected in mwoffliner, which I think will clearly expose the issue.

Jaifroid:
To reproduce this issue with existing ZIMs:

1. Load any Stackexchange ZIM with subdirectories in Service Worker mode;
2. Navigate to a page in a subdirectory and note the page's title (you will need to search for this page in step 4);
3. Click on a link, it will work (because page has been loaded natively by the browser).
4. Now use the UI search to search for the page in step 2 and load the page;
5. Click on the same link you clicked on earlier. It will fail.

Reason it fails is because you have loaded the page via app.js, not natively in the SW, and app.js has the above line which encodes the URL.

Here is an example of a URL in a Stackexchange article loaded natively in SW: `http://localhost/Repos/kiwix-js/A/tag/uso-de-palabras/1.html`.

Here is an example of the same URL in the page that has been loaded into SW by app.js:
`http://localhost/Repos/kiwix-js/www/tag/uso-de-palabras/1.html`

Note the extra (incorrect) www in the path in the second case. This is the issue fixed by this PR.

mossroy:
Thanks a lot for the test-case : that's what I was missing.
I can indeed reproduce the issue, and confirms the PR fixes it.

www/js/app.js
```diff
@@ -760,6 +760,9 @@ define(['jquery', 'zimArchiveLoader', 'util', 'uiUtil', 'cookies','abstractFiles
         if (contentInjectionMode === 'serviceworker') {
             // In ServiceWorker mode, we simply set the iframe src.
             // (reading the backend is handled by the ServiceWorker itself)
+            var encodedUrl = dirEntry.url.replace(/([^/]+)/g, function(_p0, part) {
```
mossroy: Maybe this piece of code could be moved closer to where this variable is used?
A comment should be added to explain what it does.
And the _p0 parameter should be renamed to something more meaningful (like `matchedSubstring`). In this particular case, the first capture group is the entire matched substring, so you could probably remove the capture group, and use this first parameter instead, like : 
```suggestion
            var encodedUrl = dirEntry.url.replace(/[^/]+/g, function(matchedSubstring) {
                return encodeURIComponent(matchedSubstring);
            });
```

www/js/app.js
```diff
@@ -773,7 +776,7 @@ define(['jquery', 'zimArchiveLoader', 'util', 'uiUtil', 'cookies','abstractFiles
                     iframeArticleContent.onload = function () {};
                     $("#searchingArticles").hide();
                 };
-                iframeArticleContent.src = dirEntry.namespace + "/" + encodeURIComponent(dirEntry.url);
+                iframeArticleContent.src = '../' + dirEntry.namespace + "/" + encodedUrl;
```
mossroy: I suppose the `../` has been added to remove the `www/` part of URLs? It seems to be a good idea

mossroy:
I've tested it on Firefox 68 and Chromium 75

www/js/app.js
Jaifroid: OK, yes this looks good. I hadn't noticed that I was selecting a part that is actually the same as the fully matched string.

www/js/app.js
Jaifroid: From memory, I think this was necessary to match the scope of the SW, so that the URL is in the same scope whether it is loaded by app.js (as the result of a search), or by the Service Worker. But it was some time ago. I tested it in different scenarios, and this is what worked.

www/js/app.js
Jaifroid: Regarding moving the code closer to where the variable is used, I put it where it is so that it wouldn't have to be run on article load. Any closer would involve putting it inside the onload function, and it seems to me there's already quite a lot going on at that time. Maybe it makes no difference, but instinctively that was my reason.

Jaifroid:
Firefox 45 Travis failure again: "Timed out while waiting for element <#articleList> to be visible for 20000 milliseconds."

Jaifroid:
Before merging this, I need to point out one thing: although this now implements the "correct" way to do it, on some existing ZIMs that have forward slashes in their filenames we  may get broken links. The reason is because mwoffliner was not correctly dealing with such cases until a recent fix that only became noticeable as a result of the shift away from using .html endings. Is this a problem? Should we look for such cases and work around them?

www/js/app.js
mossroy: OK you can keep it here, but please add a comment above to explain why we do that

mossroy:
I restarted the Travis build and it succeeded. I created #536 for those unstable builds, it's becoming annoying.
Appart from adding a comment to explain the way we encode, I think it's ready to squash/rebase/merge.
We should not try to add workarounds for mistakes in some ZIM files, especially for rare cases like this one

Jaifroid:
I'm just doing a bit of testing on Edge in different scenarios and comparing with master. If all goes well, I'll merge in about 20 mins.

Jaifroid:
OK, as predicted, this PR breaks hyperlinks on the front page in SW mode in `wikivoyage_en_all_novid_2019-06.zim`, which is the current Wikivoyage available on the download page. I'll add details in a moment.

Jaifroid:
Details of breaking change:

The URL of the landing page of this ZIM is:

`Wikivoyage%3AOffline_reader_Expedition/Home_page`

Note the forward slash. The code in this PR reads that slash as a subdirectory. However, this ZIM has not been encoded with the mwoffliner fix, so hyperlinks on the landing page are missing an extra `../` in front of them. Here is a sample hyperlink from the raw HTML on the landing page:

`<a title="Europe" href="Europe">Europe</a>`

In our code, this "correctly" resolves to (e.g.): 

`http://localhost/Repos/kiwix-js/A/Wikivoyage%3AOffline_reader_Expedition/Europe`

... but of course that doesn't exist in this ZIM. So, while we are now doing it correctly, there are no ZIM files yet produced that correspond to this correct method of doing it. Is it therefore still too soon to merge this?

The reason I'm insisting on this, is because if we merge it, I want us to be fully aware of the breaking change it involves vis-à-vis existing ZIMs, and the subsequent usability issues it can introduce.

Jaifroid:
I wonder if we need to "co-ordinate" this fix and ensure kelson and ISNIT0 are aware (I've deliberately not pinged them just yet). It seems to me that the mwoffliner fix is going to affect other readers, not just ours.

mossroy:
Thanks for your careful tests.
I thought it might only affect very specific articles that would have a slash in their title : I was wrong.

Before pinging them, we might have a look at the source code of other Kiwix implementations where the above scenario probably works, and the `wikivoyage_en_all_novid_2019-06.zim` main page probably works too.
Maybe there's another way to deal with the issue, which would be compatible with both : we should do it exactly as the other Kiwix implementations do.

Jaifroid:
> Maybe there's another way to deal with the issue, which would be compatible with both : we should do it exactly as the other Kiwix implementations do.

I have checked with kiwix-desktop_windows_x64_2.0-beta5 (the reference Kiwix Desktop application -- the new code currently in beta). This is based on libzim, so should be fully compliant. Results of a quick test:

* Links on landing page of `wikivoyage_en_all_novid_2019-06.zim` work fine;
* Links to subdirectories in Stackexchange ZIMs (I tested `spanish.stackexchange.com_eng_all_2018-08.zim` and `philosophy.stackexchange.com_en_all_2017-10.zim`) are broken.

I think this shows that readers based on libzim have the same issue as we do, and will have to deal with it once the fix in https://github.com/openzim/mwoffliner/issues/726 comes through. I'd best report this on the Kiwix Desktop github.

Jaifroid:
Just to add that I've found out that the iOS app is having problems with exactly the same issue: https://github.com/kiwix/apple/issues/157 . @kelson42 , I think this needs your attention and a co-ordinated approach between the different ZIM readers and mwoffliner. See also a closely related issue I've opened for the Desktop beta app https://github.com/kiwix/kiwix-desktop/issues/206 .

mossroy:
Thanks a lot for these tests @Jaifroid.
So this will probably take some time to discuss and implement : I'd be in favor of post-poning this fix for 2.7

Jaifroid:
Fine by me. We should, IMHO, at least wait till there is a new, fixed Wikivoyage ZIM before merging this.

kelson42:
> We should not try to add workarounds for mistakes in some ZIM files, especially for rare cases like this one

Fully support, just avoid to crash :)

Jaifroid:
> @Jaifroid I suspect that this is difficult to solve the pb at the libkiwix level, the problem seems to be a wrong handling a the reader level. But kiwix/apple#157 is about MacOS reader, what is your reproduction case for iOS?

I don't have a reproduction case for iOS, I just noticed that the described symptoms (that app cannot find the URL of articles linked on the Wikivoyage landing page) is absolutely consistent with the problems we are experiencing here too, and is linked also to the problem I found with the Kiwix Desktop Beta app. Different clients are adopting different strategies for reading article URLs that have `/` in them. It seems important, for example, that the iOS app isn't "fixed" to support the current Wikivoyage landing page if that landing page is in fact incorrectly coded.

Jaifroid:
Given that links on the landing page of `wikivoyage_en_all_novid_2019-06.zim` do not work in jQuery mode, and given that @ISNIT0 has confirmed that links on this page should have an extra `../`, and that the iOS app also fails on this ZIM, perhaps we should merge this PR as part of 2.6.0. It does fix a real bug with Stackexchange ZIMs.

Jaifroid:
I've rebased on master. @mossroy, should we merge? If we don't, we will have a release with a bug (albeit minor) on Stackexchange ZIMs, and one that has inconsistent behaviour between jQuery and SW mode (with regard to how links on the landing page of `wikivoyage_en_all_novid_2019-06.zim` are read).

Jaifroid:
Just a quick note on the last commit: pages loaded by function `readArticle` have a `www` in the URL of hyperlinks on the page, while the same page loaded entirely by the Service Worker have no `www`. This commit, as in my original branch, adds `../` to the src of pages loaded by `readArticle`, ensuring that pages appear to the browser to be in the same scope as the Service Worker (i.e., the `www` is removed). This ensures URLs are consistent.

mossroy:
Sorry, I had not followed the discussions of https://github.com/openzim/mwoffliner/issues/726. Now I've read them.
So you say that it's a bug in mwoffliner, affecting `wikivoyage_en_all_novid_2019-06.zim` (but not the previous ones).
Other Kiwix clients (at least on desktop) have the same (wrong) behavior we have on stackexchange ZIM files, and should be fixed too.
Merging this PR would fix the bug we have with stackexchange ZIM files, and would make `wikivoyage_en_all_novid_2019-06.zim` unusable, which is normal as it's broken.

Well, if you're 100% sure of that, let's merge it.

Jaifroid:
That's about right. The iOS app is already "fixed" from what I can tell: it has the same behaviour as we would have once we implement this PR (except that it shouldn't crash). I.e., links on the landing page of the current Wikivoyage fail because they are incorrectly encoded. This is the case for us in jQuery mode already (links fail), but not in SW mode until we merge this PR. So, we would be merging for consistency and to fix the minor Stackexchange bug.

The Kiwix desktop beta app does load links on landing page of current Wikivoyage, but fails with any subdirectories in Stackexchange ZIMs. So it definitely needs a fix.

mossroy:
OK you can squash/merge this PR

Issue #442, opened 2018-11-24
The search results cannot be dismissed by clicking on the article


Jaifroid:
This is a regression caused by #425 / #427 . To reproduced, open any ZIM with a range of content, search for something (ideally something with several results), and attempt to abort the search by clicking back into the article. The search results remain on screen. Likewise, emptying the search box by erasing the typing with backspace does not make the results disappear.

This should be relatively easy to fix with a focus event.

Issue #55, opened 2013-08-17
Allow keyboard selection of a title in the list


mossroy:
By using up and down arrow keys, the user might browse the results, like in a "auto-suggest" feature.
Maybe use a library like http://nicolasbize.github.io/magicsuggest/ ?

Jaifroid:
PR #501 implements this old issue.

kelson42:
Nice :)

PR #499, opened 2019-05-16
Hide article list on prefix blur or on Esc

### LINKED ISSUE #39: It (rarely) happens to have both search results and an article displayed at the same time
It's certainly due to the fact that searches are asynchronous.
It probably happened after a quick click on an article, while another search was running.

A click on an article should probably "cancel" any running search (in the sense that the search results should not be displayed in that case)

Jaifroid:
This fixes #39 and the regression in #442.

The behaviour is as follows:

* User can dismiss search results with Esc, click/tap anywhere outside of prefix (search box), or with tab;
* Search results are not lost, they are hidden;
* If user re-enters prefix (if it gains focus), then the search results are shown again if prefix is not empty.

In the process, testing on FFOS where the on-screen-keyboard causes drastic screen resizes, I discovered an obsolete value still being used, erroneously in the `resizeIframe()` function. Since we now float `articleListWithHeader` over the article, we shouldn't subtract its height from the `window.outerHeigh`. It causes a drastically shortened screen otherwise.

I have coded this with jQuery for the sake of consistency with surrounding code.

Tested in Edge 44, Edge 75, FFOS.

### DIFF
--- a/www/js/app.js
+++ b/www/js/app.js
@@ -71,7 +71,6 @@ define(['jquery', 'zimArchiveLoader', 'util', 'uiUtil', 'cookies','abstractFiles
     function resizeIFrame() {
         var height = $(window).outerHeight()
                 - $("#top").outerHeight(true)
-                - $("#articleListWithHeader").outerHeight(true)
                 // TODO : this 5 should be dynamically computed, and not hard-coded
                 - 5;
         $(".articleIFrame").css("height", height + "px");
@@ -96,8 +95,21 @@ define(['jquery', 'zimArchiveLoader', 'util', 'uiUtil', 'cookies','abstractFiles
     });
     $('#prefix').on('keyup', function(e) {
         if (selectedArchive !== null && selectedArchive.isReady()) {
+            if (/^Esc/.test(e.key)) {
+                // Hide the article list
+                $('#articleListWithHeader').hide();
+                $('#articleContent').focus();
+                return;
+            }
             onKeyUpPrefix(e);
         }
+    });
+    $('#prefix').on('focus', function(e) {
+        if ($('#prefix').val() !== '') 
+            $('#articleListWithHeader').show();
+    });
+    $('#prefix').on('blur', function() {
+        $('#articleListWithHeader').hide();
     });
     $("#btnRandomArticle").on("click", function(e) {
         $('#prefix').val("");
@@ -777,7 +789,12 @@ define(['jquery', 'zimArchiveLoader', 'util', 'uiUtil', 'cookies','abstractFiles
                 '" class="list-group-item">' + dirEntry.getTitleOrUrl() + '</a>';
         }
         articleListDiv.html(articleListDivHtml);
-        $('#articleList a').on('click', handleTitleClick);
+        // We have to use mousedown below instead of click as otherwise the prefix blur event fires first 
+        // and prevents this event from firing; note that touch also triggers mousedown
+        $('#articleList a').on('mousedown', function (e) {
+            handleTitleClick(e);
+            return false;
+        });
         $('#searchingArticles').hide();
         $('#articleListWithHeader').show();
     }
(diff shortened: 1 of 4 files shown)


Jaifroid:
Hmm, Travis seems to have fallen over on Edge, but not on any other browser. I have tested this extensively on Edge 44, and the code only affects the UI, not the content in any way. I wonder if this is a transitory issue.

www/js/app.js
```diff
@@ -92,9 +91,22 @@ define(['jquery', 'zimArchiveLoader', 'util', 'uiUtil', 'cookies','abstractFiles
     });
     $('#prefix').on('keyup', function(e) {
         if (selectedArchive !== null && selectedArchive.isReady()) {
+            if (e.which === 27) {
```
mossroy: https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/which seems to be deprecated.
Could you try to use https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key instead? (Check the example at the end of this web page, because IE uses specific values for this property, that need to be handled too)

Jaifroid:
> If you're in the mood of implementing old issues, there is #55 in the same area ;-)

You're telepathic! I just implemented that today and was waiting to get this PR approved / merged before posting the code as a new PR.

mossroy:
Regarding the UI tests that fail on Edge 40 and 42, I tried to restart Travis several times, but they still fail.
We'll need to fix that before merging. I'm trying to find out what happens

www/js/app.js
Jaifroid: Ah, OK, that's good to know. I'll try the new method, and will check the Edge / IE workarounds.

mossroy:
Well, the issue on Edge UI tests is not obvious.
As we're talking about old versions of Edge, that you manually tested, it might not be worth investing too much time on that.
I already had to skip some UI tests on some versions of Edge : maybe we'll have to skip more.

www/js/app.js
Jaifroid: Done. Let's see if that makes a difference with the tests. I tested manually on Edge 44, Edge 76 (it's incremented by one recently), Firefox ESR and IE11.

Jaifroid:
Tests still failing on Edge 40. Edge 40 is pretty much only still used on Windows Mobile, and that platform is no longer being developed (though still supported with bugfixes for a while longer). I test manually for Windows Mobile when I port things to Kiwix JS Windows. So far I haven't found incompatibilities, but haven't tested this latest code there yet. If there were an incompatibility, I'd work around it there.

I'd say it's pretty safe to drop tests for Edge 40.

mossroy:
OK, so I'll remove the UI tests for Edge 40 (but keep the unit tests on it).
I suggest to completely remove Edge 42 from the browser list, but to keep :

- Edge 44 (as it is the last version based on EdgeHTML)
- lastest Edge (for the upcoming version based on Chromium)

Jaifroid:
@mossroy I've compiled a test version of Kiwix JS Windows that uses this code for Windows Mobile (Edge 40), and have successfully run it there. So yes, it's safe to remove old versions. I have no idea why the tests are failing...

mossroy:
I've kept the beginning of UI tests on Edge (all versions), because they never failed (as far as I know) : opening the ZIM file and searching for articles are still covered. It's still better than no UI test at all.

I had to disable latest Edge, because I did not find an easy way to know its version, which is needed to distinguish Edge 44 (currently latest, which needs to disable some tests), and EdgeChromium (future latest, on which we should try to run all tests).

When EdgeChromium will be released, and available on SauceLabs, I'll re-enable latest Edge

Jaifroid:
OK, sounds good! Shall I go ahead and squash/merge this?

Jaifroid:
And if I merge, should I keep two commits, one for this issue and one for the unit tests?

mossroy:
Yes, you can go ahead.
It would be perfect to have 2 commits, if you manage to do it. But it's not compulsory...

PR #501, opened 2019-05-18
Allow keyboard selection in search list

Jaifroid:
This addresses #55 . It also has the functionality of #499 built into this PR, but #499 could be merged first in order to keep the development steps in order. This is tested on Edge 44, Edge 76, Firefox ESR, Internet Explorer 11. It also works on FFOS simulator if using the computer keyboard to select titles. However, this isn't really relevant in that context, I think.

There is one display bug on FFOS simulator only, which is that if we scroll down (with down arrow on computer keyboard) a long list of found titles, and then scroll back up (with up arrow), then the whole app scrolls up a little in the IDE, leaving an odd space at the bottom. I think this is just an artefact of the simulator, and I guess it can't be tested on a real device because a keyboard can't be attached.

The other issue on FFOS simulator is that the on-screen keyboard remains in place even after pressing the physical button. This is because I have to keep focus on the prefix field, to monitor further key presses. But again, I think this is not a "real" issue in any touchscreen app environment, just an artefact of the simulator.

### DIFF
--- a/www/css/app.css
+++ b/www/css/app.css
@@ -73,6 +73,10 @@
     background: lightblue;
 }
 
+#articleList a:hover, #articleList a.hover {
+    background: lightblue;
+}
+
 .articleIFrame {
     width: 100%;
     border: 0;
--- a/www/js/app.js
+++ b/www/js/app.js
@@ -117,7 +117,7 @@ define(['jquery', 'zimArchiveLoader', 'util', 'uiUtil', 'cookies','abstractFiles
             var activeElement = document.querySelector("#articleList .hover") || document.querySelector("#articleList a");
             if (!activeElement) return;
             // If user presses Enter, read the dirEntry
-            if (/^Enter$/.test(e.key)) {
+            if (/Enter/.test(e.key)) {
                 if (activeElement.classList.contains('hover')) {
                     var dirEntryId = activeElement.getAttribute('dirEntryId');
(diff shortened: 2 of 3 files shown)


Jaifroid:
I've rebased this branch / PR so that we can see if the tests work here.

Jaifroid:
Oh, that's strange, Chrome doesn't like this code! Will test later.

Jaifroid:
Right, not sure why Chrome failed the Travis tests for [this commit](https://github.com/kiwix/kiwix-js/pull/501/commits/de04b239b600f00d92575c81378eedc20e9a57e1). I'm not getting any error in Chromium, and the code works fine. However, the code isn't working in Firefox ESR (focus on contentWindow does nothing). Have to think again.

Jaifroid:
Ha, finally found a way that Travis approves of!

This is tested in jQuery and SW modes on Edge and Chromium, and in jQuery modes in IE11, Firefox ESR and FFOS simulator.

Jaifroid:
@mossroy I know you've approved this PR, but I shan't merge until you tell me either that you've successfully tested the fix for #503, or that you're happy with the test I've done on FFOS simulator.

mossroy:
I've quickly tested on a Firefox OS device, and on Firefox desktop (using only the keyboard). It seems to work well : #503 seems to be solved by this PR

mossroy:
I should be able to have a look at the other issues and PRs at the end of the following week.

Issue #173, opened 2016-06-21
Display images only on user request


mossroy:
Displaying a big article can be quite slow on a low-end device.
It can even crash the browser because of too much memory consumption.
Some optimizations are certainly possible, but not in the short term.

I suspect that the handling of images consumes a lot of resources.
An idea would be to do like most email clients do : first display the content without images, and allow the user to click somewhere to display them. Email clients do that for privacy reasons, we would do that to save resources.
Ideally, an option in the settings would allow the user to indicate if he always wants to display them.

Jaifroid:
@mossroy  Displaying images on user request is now implemented in the Preload stylesheets branch commit https://github.com/kiwix/kiwix-js/commit/634d01d5365135835e87b39bae39fe994504d9e0. See screenshot of configurable options below.

![image](https://user-images.githubusercontent.com/4304337/28513768-4ef5f96e-704f-11e7-8247-c9dad1cfe528.png)

Jaifroid:
@mossroy , @sharun-s : I've now fully the manual download of images in commit https://github.com/kiwix/kiwix-js/commit/6857a40d24012d74e741c605223892a4a2906aea : the code now shows a light yellow area where the image should be, and user can click on it to download just that image. Tested on Firefox, Edge and Edge Mobile. Firefox had an issue whereby the image is rendered un-clickable if no data has been downloaded for it. However, by adding an `alt="Image"` attribute, the image is clickable once again. This means that the clickable area, in Firefox, is only the width of the word "Image". Also, other "alt" text may conflict -- I haven't tested this aspect, as most images don't have alt text, but occasionally some do (if you can think of a better solution, let me know). In the other browsers tested, the clickable area is the actual width of the image to be obtained. Some screenshots from Firefox below.

![image](https://user-images.githubusercontent.com/4304337/28537110-b0b3783c-70a1-11e7-8d69-dbd310330c26.png)

![image](https://user-images.githubusercontent.com/4304337/28537186-ee1d745c-70a1-11e7-92aa-8dc1750e1c96.png)

![image](https://user-images.githubusercontent.com/4304337/28537223-02500a16-70a2-11e7-975f-caf982d8dd3e.png)

![image](https://user-images.githubusercontent.com/4304337/28537252-1255ced2-70a2-11e7-889b-56ff4826f5a5.png)

Jaifroid:
Changed lightyellow to lightblue, as it is less intrusive and fits with the standard link colour.

[...]

PR #504, opened 2019-05-28
Enable manual extraction of images

Jaifroid:
This addresses an old issue #173 from the backlog. It backports the feature from Kiwix JS Windows, but hopefully with better coding style, as I've learnt a lot more about how Kiwix JS works since I originally developed it.

Currently this is only applied to jQuery mode. Adding it to SW mode would be possible, but would involve manipulating the html coming from the ZIM. @mossroy I would only do this if you tell me you would like to see the feature provided for SW mode as well. If you would like me to produce a demo, then the method I would propose would be to substitute images with the blue area using a regex on the incoming html of an article before the SW sees it. This PR currently does the transformation via `loadImagesJQuery()` with DOM methods, but in Kiwix JS Windows I use a regex on the html which has proven to be very robust.

If you would prefer not to have this applied to SW mode, then I shall label the feature as applying only to jQuery mode, or else, I can hide the option when SW mode is selected.

This feature makes articles with many images much more usable on old, slow devices or browsers. For example, loading the "Paris" article from English Wikipedia on Internet Explorer takes nearly 2 minutes for images to load on a modern PC, and will probably crash the browser on older PCs. It has been known to crash my copy of IE.


mossroy:
In my opinion, it would be enough (and more simple) to have a general switch to display all images (if they were hidden). That's how it works on Thunderbird for example, even if I agree that it's not for the same reason.
I think the main use case is to speed up browsing a ZIM file (especially on slow devices). When the user wants to see one image, he'll probably want to see some other ones in the article, so displaying all of them with one click can be useful (and probably enough).
I agree that there might be edge cases like a huge article with many images that can crash the browser (because of lack of memory). In this case, it might be useful to display images one by one. But it seems like a very rare use-case to me.

I think we should implement the feature for both modes at the same time. And (if possible) the same way.
It should be easy in SW mode to tell the SW it should send an empty content for all images (or to tell it it should read images again, and refresh the iframe). And this would be on par with the jQuery mode.
I'm against manipulating the HTML string in SW mode to implement the blue placeholders : it won't work for 100% of the images, and seems overkill for the need.

Maybe this feature should also include videos?

NB : I found a small issue with the blue placeholders in jQuery mode. If you use for example wikipedia_fr_all_novid_2018-03.zim, there are a few placeholders in the main page, and some of them generate a `Could not find DirEntry for image: null TypeError: "dirEntry is null"` error in the console (and nothing appears)

Jaifroid:
OK, I've modified the code so that when a user clicks on any image, all images that are currently in the area of the viewport are extracted.

I think the issue you are raising, @mossroy , is that it requires too many clicks on individual images to load images in the area that the user is reading. The revised code achieves a compromise between loading all images on the page (which would involve the user having to wait a very long time in IE and other old browsers on pages such as "Paris") and loading the images the user wants to look at in any particular moment.

**Please note that I have not yet investigated or addressed your N.B. above.**

This version of the code, _merely for demonstration purposes_, applies manual extraction also to Service Worker. I may need some guidance in how to achieve it differently, i.e.:

> It should be easy in SW mode to tell the SW it should send an empty content for all images (or to tell it that it should read images again, and refresh the iframe).

Jaifroid:
I think I've worked out a way to hook up the Service Worker without manipulating the HTML coming from the ZIM. You can compare the regex method and this new method in commit https://github.com/kiwix/kiwix-js/pull/504/commits/d968a5fae46cccd9fce213e6ef052bfd4509f77e. Is this more suitable?

Jaifroid:
> NB : I found a small issue with the blue placeholders in jQuery mode. If you use for example wikipedia_fr_all_novid_2018-03.zim, there are a few placeholders in the main page, and some of them generate a Could not find DirEntry for image: null TypeError: "dirEntry is null" error in the console (and nothing appears)

@mossroy I have downloaded `wikipedia_fr_all_novid_2018-10.zim` (because `wikipedia_fr_all_novid_2018-03.zim` is no longer available on the server), and I cannot reproduce the issue with the latest commit of this PR. Could you check if the problem still persists with the latest code, and if it does, please post here the full URL of the landing page, and the URLs of affected dirEntries, so I can diagnose?

mossroy:
With your latest commits, I do not reproduce the "dirEntry is null" issue any more.
I've tested on Firefox OS and it works well too. It's very useful on my slow device.

In SW mode, it also blocks videos (because of their namespace, I suppose). It's good, but there is no way to display them afterwards (because no event is attached when you click on them).
In jQuery mode, videos are not blocked.
In both mode, I also have issues when clicking on some blue placeholders that are also links to some other pages (for example in dirtybiology ZIM files). We should let the user do 2 things independently : see the image and follow the link.

Regarding the implementation in SW mode, I much prefer this version, where you attach events instead of modifying the HTML string with regular expressions. But I don't like the fact that, when the user clicks on it, it uses the jQuery mode to display it. It will makes things more complicated for future refactorings like #507.

But I'm still thinking all this is too complex for the need.
It seems to work well on some use-cases I tested, and the implementation is clever. But it adds complexity in the code, and can not be 100% safe as images and videos could be injected through javascript for example. And I'm afraid we'll find other issues on different kinds of ZIM files.

IMHO, we would cover 95% of the need with a more simple implementation :
- hide all images and videos (if the setting is set to hide them)
- provide a way to let the user easily change that setting and reload the current page, so that to display all images of the page (not only the visible ones)
I agree with you that it's less optimal (it's a "all or nothing", where your implementation is much clever).
But it's already a very big improvement for slow devices. And it's much simpler to implement, easily compatible with any mode, and makes much less assumptions on the ZIM files structure and content. That's why I think it's a more sustainable way to handle that, following the https://en.wikipedia.org/wiki/KISS_principle

Jaifroid:
@mossroy I've repaired the video playplack, which I broke when simplifying the Service Worker.
I've also prevented bubbling better. Could you retry with the latest code?

I'll think about your other points, but I would like to iron out any gotchas with this code. I have simplified quite a lot!

Jaifroid:
PR / branch has been rebased on master.

Jaifroid:
> In SW mode, it also blocks videos (because of their namespace, I suppose).

I think this one is fixed.

> In both modes, I also have issues when clicking on some blue placeholders that are also links to some other pages

I think it's fixed. I was erroneously using `onmousedown`, but it turns out you can't cancel a click event from `onmousedown`. Using `onclick `with `preventDefault `and `stopPropagation `seems to do the trick.

> But I don't like the fact that, when the user clicks on it, it uses the jQuery mode to display it. 

I've fixed this in line 1304 of app.js. It now does it by setting the src of the image to the image's ZIM URL, which triggers a fetch event. That event is then captured by the Service Worker. No more jQuery, yay!

> it can not be 100% safe as images and videos could be injected through javascript

It works completely offline and never accesses the network. I don't see how any malicious injection could occur, especially now I'm using the Service Worker API to trigger extaction of an asset from the ZIM.

> But I'm still thinking all this is too complex for the need.

Overall the PR adds 66 lines (after subtracting deleted lines) across app.js and service-worker.js, which doesn't seem such a huge amount. 

> IMHO, we would cover 95% of the need with a more simple implementation

My worry with the idea of loading all images at once (which could be done) is that it negates the usefulness of this feature where it is most needed, i.e. in long articles full of images. The wait for all images to be loaded in "Bolivia" or "Paris" is over a minute in IE11, whereas if we just load the images in the area the user is viewing, the wait is at most two or three seconds in IE11 and instantaneous in other browsers. I don't think any user would hang around for a minute. They'd just give up and go view something else.

BTW, I got great feedback about this feature from a user of kiwix-js-windows who said the "blue box show pictures on demand setting" was a "brilliant idea"....

Jaifroid:
For a demonstration of the problem, spin up Internet Explorer, and compare image loading times (with and without display of images enabled in Config) in the article "Trigonometric functions" in `wikipedia_en_all_novid_2018-06.zim` or the much smaller `wikipedia_en_maths_novid_2018-06.zim`. The difference in usability is quite dramatic. This problem is rife for many many articles in `wikipedia_en_maths_novid_2018-06.zim`.

Jaifroid:
In an attempt to address the concerns in #507, I've refactored this PR to put all the new image processing code in a new module called `images.js`. This should declutter `app.js`. The only slightly clunky thing is the need to pass `selectedArchive` to `images.js` in order to read images, but I guess this is the discipline imposed by `require.js`. There is also a slightly clunky workaround for passing `contentInjectionMode` to the new module. It would be better to define a global variable in `init.js`, so that we don't have to pass these variables and large objects around.

Jaifroid:
Rebased branch on master.

Jaifroid:
Just to say that I've coded this feature because I have for a long time now wanted to refactor the clunky image-loading routines I developed very early on for Kiwix JS Windows, and I wanted to do it an a way compatible with Kiwix JS (hence I have to do it on Kiwix JS). Rest assured I'm not going to press for features that are not wanted in Kiwix JS, but I'm trying to make this into a feature-complete replacement for what I already have in Kiwix JS Windows. Of course I would be delighted if any of it is useful here, and I do think that progressive image display has several benefits on all browsers, and manual image display is of particular benefit to IE11 users and to people using mobile or old machines. However, I'm happy to reduce the feature set to what is appropriate for Kiwix JS, though I do want to get the full feature to an acceptable level that is fully compatible with the Kiwix JS way of doing things first.

mossroy:
@Jaifroid, sorry for the delay but I wanted to take some time to think about it, and to explain my opinion.
I'm sorry it's probably frustrating as you already spent quite some time on it.
I'm still available to discuss all this, and I'm sure we'll find a solution together.

To me, this PR is an optimization. A workaround for the fact that our backend is not very fast.
It's a very good optimization in terms of UX (user experience), I agree with the feedback you got from a user. And the implementation is smart.

On the other hand, the initial tests I did with libzim in https://github.com/kiwix/kiwix-js/issues/116#issuecomment-423538950 let us hope a _very_ significant performance increase for this backend (x2 or x3).
It should make this optimization much less desirable.
Unfortunately, libzim for kiwix-js is not ready yet.

So, in the short term, this PR might look worth it.
The problem is that implementing it has many drawbacks, in my opinion :
- the main one is that it makes a few assumptions and modifications in the ZIM file content. It browses the DOM to find the images, and adds some event handlers on them. It's already the case in jQuery mode (because we have no other choice, and we already know we can not cover all the cases in this mode), but it must be avoided in SW mode. If we want to be sure we're compatible with any past/present/future ZIM file, we must not do that
- second one : while we're working on this PR for the short term, we do not work for the mid/long term instead. Our available time to work on that project is not unlimited : it seems to me that we're putting too much effort on enhancing the current version (sometimes in a disposable way) instead of implementing the next one
- we're optimizing a use-case that seems very specific to me. Hiding images will probably be activated only by people who have a slow/old device/browser, like my Firefox OS device. It's true that big pages can take a very long time to load with all their images. But I felt it was a reasonable tradeoff to be able to browse much faster without images (most of the time), and to wait for all images to load when you're on a specific article where you want to see some images
- it's implementing behaviors that should be handled by the browser itself. For example, in SW mode, Firefox displays the page even if all images are not loaded (and finishes loading them in the background). And Chrome will soon be able to do image lazy-loading by itself (in SW mode) : https://addyosmani.com/blog/lazy-loading/
- as any optimization, it increases the technical debt, that we might pay later. Either by having to maintain it, or by having frustrated users if we remove it


As a reply to previous comments :
> > it can not be 100% safe as images and videos could be injected through javascript
> 
> It works completely offline and never accesses the network. I don't see how any malicious injection could occur, especially now I'm using the Service Worker API to trigger extaction of an asset from the ZIM.

My concern is not about security, sorry if I was unclear. It's about compatibility with ZIM files.
What happens if the blue placeholders break the layout of the page, so that some content or features are not visible/available?
What happens if there is already some code on the event handlers of some images? For example, the ZIM file might include its own javascript implementation of image lazy-loading like https://appelsiini.net/projects/lazyload/
Think about some ZIM files like PhET, which have a much more dynamic content than the ones from mwoffliner

Generally speaking, the only way to make sure we do not mess up the content is to not touch it at all (which is only possible in SW mode, of course)

> > But I'm still thinking all this is too complex for the need.
> 
> Overall the PR adds 66 lines (after subtracting deleted lines) across app.js and service-worker.js, which doesn't seem such a huge amount.

Github seems to say it's currently 220-31=189 lines, but maybe I'm wrong. It's true that some of these lines are also needed if we use a simpler implementation (hide/show all images) : handling the setting, storing it in a cookie etc

> > IMHO, we would cover 95% of the need with a more simple implementation
> 
> My worry with the idea of loading all images at once (which could be done) is that it negates the usefulness of this feature where it is most needed, i.e. in long articles full of images. The wait for all images to be loaded in "Bolivia" or "Paris" is over a minute in IE11, whereas if we just load the images in the area the user is viewing, the wait is at most two or three seconds in IE11 and instantaneous in other browsers. I don't think any user would hang around for a minute. They'd just give up and go view something else.

It might be true. But hiding all images would already be a big enhancement compared to the current kiwix-js state, that would give us some time to see if/how the libzim enhances the situation (or not)

> BTW, I got great feedback about this feature from a user of kiwix-js-windows who said the "blue box show pictures on demand setting" was a "brilliant idea"....

You do an amazing work, really. And I have no doubt this enhancement is significant for this user.

But we need to look at the "big picture" of the application, with all its users and use-cases, and with its roadmap/maintenance.

To sum up (sorry for the length of this comment), I'm against merging this PR as it is now.
No problem for keeping it to make your implementation feature-complete.
When you're done, if you agree, I would suggest that you implement the simple version (hiding/showing all images) in kiwix-js (maybe in a different PR), keep what is common, and choose entry-points where you could easily extend the features for kiwix-js-windows (with my help).

Jaifroid:
Thank you @mossroy for explaining the issues so carefully. I'm of course very happy to be guided by you as to what works in the road plan for Kiwix JS and what doesn't.

> What happens if the blue placeholders break the layout of the page, so that some content or features are not visible/available?
> What happens if there is already some code on the event handlers of some images? For example, the ZIM file might include its own javascript implementation of image lazy-loading like https://appelsiini.net/projects/lazyload/
> Think about some ZIM files like PhET, which have a much more dynamic content than the ones from mwoffliner

I understand, but of course the feature would be optional and would only be turned on by the user when they feel they need it. And if the layout is messed up in a specific case, there is an easy remedy: turn the feature off. FWIW, I only squash the placeholders so the user doesn't have to see large empty space in the browser. The layout can be kept exactly as it is originally, though IMHO it looks terrible!

> Github seems to say it's currently 220-31=189 lines, but maybe I'm wrong

The 66 lines I mentioned (lines added minus lines subtracted) was before merging and rebasing on #519. This PR was the place I originally implemented #519, which cut out a lot of code from `service-worker.js`. I then realized it should be a separate PR.

> it seems to me that we're putting too much effort on enhancing the current version (sometimes in a disposable way) instead of implementing the next one

I understand, but I don't feel capable of working on the libzim port due to lack of knowledge of anything C++. So what is left for me to do while waiting for that is to optimize the working of the current codebase. I kind of feel that once libzim is implemented it's a whole new ball-game, and much of the current codebase will become redundant. I'm a bit impatient, and I want the app to be the best it can be now as far as UX is concerned... Once libzim is ready, it might actually be best to build up a new Kiwix JS 3.0 or 4.0 using a modern UI framework like [ReactJS](https://reactjs.org/).

> I would suggest that you implement the simple version (hiding/showing all images) in kiwix-js (maybe in a different PR), keep what is common, and choose entry-points where you could easily extend the features for kiwix-js-windows

I'm happy to do this. However, in the light of the design requirements you have explained, there are still some issues and options that need to be decided. I'll present these options as different posts on the original issue thread #173, and will close this PR for now.

mossroy:
I should be able to have a look this weekend

mossroy:
HI @Jaifroid.
Sorry for the delay, I've been quite busy.

> I'm of course very happy to be guided by you as to what works in the road plan for Kiwix JS and what doesn't.

Thanks for your understanding

> > What happens if the blue placeholders break the layout of the page, so that some content or features are not visible/available?
> > What happens if there is already some code on the event handlers of some images? For example, the ZIM file might include its own javascript implementation of image lazy-loading like https://appelsiini.net/projects/lazyload/
> > Think about some ZIM files like PhET, which have a much more dynamic content than the ones from mwoffliner
> 
> I understand, but of course the feature would be optional and would only be turned on by the user when they feel they need it. And if the layout is messed up in a specific case, there is an easy remedy: turn the feature off.

You're right, it's less critical when it's an optional feature (maybe we should warn the user with a mention next to the setting)

> FWIW, I only squash the placeholders so the user doesn't have to see large empty space in the browser. The layout can be kept exactly as it is originally, though IMHO it looks terrible!

It's a detail, but, depending on the ZIM file, it's not always possible to keep the original layout. If there is an `` that does not precise the width/height of the image (with attributes of the HTML tag), we can not know the image dimensions (except if we read it in the backend, but that's what we want to save)

> > it seems to me that we're putting too much effort on enhancing the current version (sometimes in a disposable way) instead of implementing the next one
> 
> I understand, but I don't feel capable of working on the libzim port due to lack of knowledge of anything C++. So what is left for me to do while waiting for that is to optimize the working of the current codebase. I kind of feel that once libzim is implemented it's a whole new ball-game, and much of the current codebase will become redundant. I'm a bit impatient, and I want the app to be the best it can be now as far as UX is concerned... Once libzim is ready, it might actually be best to build up a new Kiwix JS 3.0 or 4.0 using a modern UI framework like [ReactJS](https://reactjs.org/).

I understand.
BTW, it might be worth backporting in kiwix-js the top bar you implemented in kiwix-js-windows, to save some vertical space. Grouping the search bar, random button, buttons to go to settings etc looks like a good idea, if it's still compatible with very small screens.

Regarding a UI framework, I've always hesitated on that : we don't have many features/pages in our UI, so I'm wondering if it would be worth it or not. And sometimes this kind of UIs can add some constraints on the language, code structure, and/or the browser support

> > I would suggest that you implement the simple version (hiding/showing all images) in kiwix-js (maybe in a different PR), keep what is common, and choose entry-points where you could easily extend the features for kiwix-js-windows
> 
> I'm happy to do this. However, in the light of the design requirements you have explained, there are still some issues and options that need to be decided. I'll present these options as different posts on the original issue thread #173, and will close this PR for now.

Great, I'll answer there

Issue #500, opened 2019-05-18
Apply use of dirEntry.getMimetype() throughout the code


Jaifroid:
Now that we have merged #490 , we need to apply it throughout the code.

Jaifroid:
It looks to me that we can apply the use of `dirEntry.getMimetype()` to Service Worker code by sending the derived MIME type along with the post message from `app.js`, as part of the message object. This should simplify quite a lot of code in `service-worker.js`.

Would you like me to tackle it, @mossroy , or do you prefer to do it?

mossroy:
Yes, that's how I would have implemented it, too.
It would be great if you can do it, else I would implement it later.

PR #519, opened 2019-06-02
Apply dirEntry.getMimetype throughout code

Jaifroid:
This is a PR to apply the use of `dirEntry.getMimetype()` throughout the code as per #500.

So far I have done Service Worker. I need to check if there are other places remaining where we could use this function.

### DIFF
--- a/service-worker.js
+++ b/service-worker.js
@@ -95,19 +95,12 @@ function fetchEventListener(event) {
                 messageChannel.port1.onmessage = function(event) {
                     if (event.data.action === 'giveContent') {
                         // Content received from app.js
-                        var contentLength;
+                        var contentLength = event.data.content ? event.data.content.byteLength : null;
                         var contentType = event.data.mimetype;
                         var headers = new Headers ();
-                        if (event.data.content && event.data.content.byteLength) {
-                            contentLength = event.data.content.byteLength;
-                            headers.set('Content-Length', contentLength);
-                        }
-                        if (contentType) {
-                            headers.set('Content-Type', contentType);
-                        }
+                        if (contentLength) headers.set('Content-Length', contentLength);
+                        if (contentType) headers.set('Content-Type', contentType);
                         // Test if the content is a video.
-                        // String.prototype.startsWith is not supported by IE11, but IE11 does not support service workers, so it's safe to use it here.
-                        // See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/startsWith
                         if (contentLength >= 1 && /video|mp4|webm|ogg/i.test(contentType)) {
                             // In case of a video, Chrome and Edge need these HTTP headers else seeking doesn't work
                             // (even if we always send all the video content, not the requested range, until the backend supports it)
--- a/service-worker.js
+++ b/service-worker.js
@@ -100,8 +100,10 @@ function fetchEventListener(event) {
                         var headers = new Headers ();
                         if (contentLength) headers.set('Content-Length', contentLength);
                         if (contentType) headers.set('Content-Type', contentType);
-                        // Test if the content is a video.
-                        if (contentLength >= 1 && /video|audio|mp4|webm|og[gmv]|mpeg/i.test(contentType)) {
+                        // Test if the content is a video or audio file
+                        // See kiwix-js #519 and openzim/zimwriterfs #113 for why we test for invalid types like "mp4" or "webm" (without "video/")
+                        // The full list of types produced by zimwriterfs is in https://github.com/openzim/zimwriterfs/blob/master/src/tools.cpp
+                        if (contentLength >= 1 && /^(video|audio)|(^|\/)(mp4|webm|og[gmv]|mpeg)$/i.test(contentType)) {
                             // In case of a video (at least), Chrome and Edge need these HTTP headers else seeking doesn't work
                             // (even if we always send all the video content, not the requested range, until the backend supports it)
                             headers.set('Accept-Ranges', 'bytes');
--- a/service-worker.js
+++ b/service-worker.js
@@ -108,7 +108,7 @@ function fetchEventListener(event) {
                         // Test if the content is a video.
                         // String.prototype.startsWith is not supported by IE11, but IE11 does not support service workers, so it's safe to use it here.
                         // See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/startsWith
-                        if (contentLength && contentLength >= 1 && contentType && contentType.startsWith('video/')) {
+                        if (contentLength && contentLength >= 1 && /video|mp4|webm|ogg/i.test(contentType)) {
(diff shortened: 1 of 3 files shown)


Jaifroid:
A few more instances done. I couldn't find anything else. @mossroy are you aware of any other places we can apply `dirEntry.getMimetype()`?

Jaifroid:
Here is the MIME type table from a recent TEDx ZIM:

image

As you can see, the entry for `mp4` has no `video` in front, so that explains the regression. It looks like an abbreviated MIME type. I suppose it must be valid. Looks like we'll have to test for each possible type: mp4, webm, etc, permissively.

Jaifroid:
Latest commit allows seeking in Edge Chromium on an mp4 video, but please test in other flavours. One gotcha is that I'm testing for `ogg` becasue `video/ogg` is the MIME type, but there is also of course `audio/ogg`. Would it do any harm if an audio file is caught by this? Or in fact might it be beneficial to include audio formats, because some players allow seeking audio too? If so, I should add audio to the regex and any common audio formats.

By the way, I've tested that a value of undefined (but declared) for `contentType` will *not* produce an exception with `regex.test()`.

Jaifroid:
I think the test on this line can be simplified (see latest commit). Try copying and pasting this into console in any browser:

`setTimeout(function() {var contentLength; console.log(contentLength > 1)})`

You'll see that the comparison `contentLength > 1` doesn't produce an exception and prints `false` even though `contentLength` is `undefined` (but declared). Now remove `var contentLength;`, and an exception will be thrown.

Jaifroid:
Travis has failed. It's been doing that a lot with the following error for Firefox:

`Timed out while waiting for element <#articleList> to be visible for 20000 milliseconds. - expected "visible" but got: "not visible"`

Usually, adding another minimal commit (chanigng a space in a comment) will cause Travis to run again, and sometimes succeed, or sometimes fail again. It's a bit pot luck as to whether it completes. Is this a known issue, or have we introduced a race condition somewhere?

Jaifroid:
Here's another MIME type map from `wikipedia_en_100_2019-01.zim`:

image

As you can see, `webm` is also missing the `video`. Looking at https://developer.mozilla.org/en-US/docs/Web/HTML/Supported_media_formats, this actually looks invalid. I wonder if the problem is in parsoid or in mwoffliner. Should we raise a ticket? We still have to work around it.

mossroy:
Thanks for the investigation.
Yes, it looks like an invalid mime-type in the ZIM file, that we should report.
I agree with you that we should try to detect it and recognize at as audio/video in this case. I don't see how these headers could be harmful, even if (in worst scenario) for other types of content.
Yes, it might be a good idea to put the headers on audio mime types too. I think there are a few ZIM files that have audio content (I seem to remember that @sharun-s tested one of them)

Regarding Travis failures, I've seen that often in the past too. I doubt it comes from a recently introduced race condition on our side. I usually restart the Travis build

Jaifroid:
I've added the audio types, and have tested with the national anthem audio in the article "United States" from `wikipedia_en_100_2019-01.zim`. On both Chromium (Edge's flavour) and Edge 44, the audio files on that page play with this PR, and it is possible to seek within the file, backwards and forwards. I don't know whether it was possible to do that before, but having these headers (it's for an `.ogg` file) certainly do no harm.

Jaifroid:
For future reference, it may be useful to know that the full list of possible MIME types produced by zimwriterfs is here: https://github.com/openzim/zimwriterfs/blob/master/src/tools.cpp . For example, it shows that searching for "audio" in our regex would not have found ogg files, because they are listed as `application/ogg`, so we will always have to search for ogg.

service-worker.js
```diff
@@ -167,20 +95,14 @@ function fetchEventListener(event) {
...
-                            headers.set('Content-Type', contentType);
-                        }
+                        if (contentLength) headers.set('Content-Length', contentLength);
+                        if (contentType) headers.set('Content-Type', contentType);
                         // Test if the content is a video.
-                        // String.prototype.startsWith is not supported by IE11, but IE11 does not support service workers, so it's safe to use it here.
-                        // See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/startsWith
-                        if (contentLength && contentLength >= 1 && contentType && contentType.startsWith('video/')) {
-                            // In case of a video, Chrome and Edge need these HTTP headers else seeking doesn't work
+                        if (contentLength >= 1 && /video|audio|mp4|webm|og[gmv]|mpeg/i.test(contentType)) {
```
mossroy: It would be good to add a comment here to explain why we test for mp4, webm, og*, and mpeg, even if they should be matched by video, with a link to the issues you created for the invalid mime-types.
It would also be better to be more specific in the regexp, to match only the content types that _start_ with one of these strings

mossroy:
I quickly tested on Firefox 67 and Chromium 73 (in jQuery and SW mode) on some ted videos, dirtybiology, and gutenberg.
Everything seems to work well.

service-worker.js
Jaifroid: OK, I've made the regex more specific, and have tested it carefully in regexbuddy for all possible combinations. It will only match primary types (audio, video) at the beginning of the string, and subtypes either at the beginning of the string or after an /, and subtypes must also be at the end of the string. Have added comment.

Jaifroid:
As the regex has changed, I'll do more testing and will merge tomorrow if all OK. However, Travis seems to be stuck (could be server load).

service-worker.js
mossroy: I think we should not match subtypes beginning with /.
IMHO, we should only match : 
- the regular audio and video mime-types (with ^video or ^audio)
- the known unregular mime-types (^mp4$, ^webm$, ^og[gmv]$, ^mpeg$)

It should not match something like application/mp4encoder (this probably does not exist)

mossroy:
All right, I think you can squash/rebase/merge if your tests are successful.

service-worker.js
Jaifroid: It can't match application/mp4encoder because then mp4 wouldn't be at the end of the string. As I've written the regex, it can **only** match `^mp4$` or `/mp4$`, i.e., mp4 on its own, or mp4 after a forward slash and with no further characters after it. Is that OK?

Jaifroid: image

mossroy: You're right.
But why should it match application/ogg, for example? It's not a known exception.
I would have used `^(video|audio)|^(mp4|webm|og[gmv]|mpeg)$`, to match only what is necessary, but it's minor.

Jaifroid: I know it looks strange, and I queried it with kelson, but it has to match `application/ogg` because that's the MIME type used in zimwriterfs for OGG audio or video. See the code here: https://github.com/openzim/zimwriterfs/blob/master/src/tools.cpp#L75 . If we don't match `/ogg` we wouldn't match this MIME type.

mossroy: OK sorry, I had not seen.
In this case, your regexp is perfectly fine.

Issue #520, opened 2019-06-04
Reading a property of null in some cases in SW mode


mossroy:
When doing some tests in SW mode on Chromium, I had the following error twice : 
`Cannot read property 'documentElement' of null`
which happened on line 909 of app.js

It's a line we recently added. We also added a few null tests, but it looks like we need to add another one

![Capture du 2019-06-04 20-30-16](https://user-images.githubusercontent.com/890923/58908041-cf148b00-870f-11e9-9d96-ff5d717365d2.png)

mossroy:
I don't remember the exact scenario, but I was switching from a git branch to another branch.
I suspect it might happen when the SW has been modified, and needs to be reloaded by the browser

Jaifroid:
Unfortunately I don't think this issue is fixed. In trying to deduplicate code between app.js and service-worker.js in #556, I came across this error, and it seems reproducible on master in Chromium. I cannot reproduce the error on Edge classic or on Firefox Quantum. Therefore I believe it's a timing issue / race condition due to the way Chromium loads the Service Worker. Steps to reproduce on Chromium (running master from localhost):

**Simple version**

* Switch to Service Worker mode
* Open DevTools, load app.js if necessary
* Completely reload browser by pressing Ctrl-Shift-R inside DevTools, with DevTools open
* You should see error in screenshot below

**Steps to reproduce while completely clearing and reloading Service Worker (to be ultra-sure)**

* Switch to jQuery with DevTools shut
* Exit browser
* Open browser and go back to localhost/.../kwix-js/...
* Ensure you really are in jQuery mode, if not, repeat above, until browser opens from cold start in jQuery mode
* Open DevTools and unregister any Service Worker for kiwix-js on localhost
* Completely exit browser, re-open, open DevTools and be sure there is no SW registered
* Load app.js in DevTools if necessary, and leave it open
* Switch to Service Worker mode
* Reload browser completely with Ctrl-Shift-R while DevTools has the focus
* Error should be reproduced  

![image](https://user-images.githubusercontent.com/4304337/64319284-dcae0b00-cfb3-11e9-9661-eb5735d68318.png)

Jaifroid:
@mossroy Maybe this error could be solved with code from https://developer.mozilla.org/en-US/docs/Web/API/ServiceWorkerContainer/ready , e.g.

`navigator.serviceWorker.ready.then(function(serviceWorkerRegistration) { ... });
`

Jaifroid:
[...]

PR #524, opened 2019-06-25
Add further tests for existence of content

Jaifroid:
This is a simple PR to address #520.

### DIFF
--- a/www/js/app.js
+++ b/www/js/app.js
@@ -906,8 +906,7 @@ define(['jquery', 'zimArchiveLoader', 'util', 'uiUtil', 'cookies','abstractFiles
                     // The content is fully loaded by the browser : we can hide the spinner
                     $("#searchingArticles").hide();
                     // Deflect drag-and-drop of ZIM file on the iframe to Config
-                    var doc = iframeArticleContent.contentDocument;
-                    doc = doc ? doc.documentElement : null;
+                    var doc = iframeArticleContent.contentDocument ? iframeArticleContent.contentDocument.documentElement : null;
                     var docBody = doc ? doc.getElementsByTagName('body') : null;
                     docBody = docBody ? docBody[0] : null;
                     if (docBody) {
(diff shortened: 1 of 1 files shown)


www/js/app.js
```diff
@@ -906,14 +906,15 @@ define(['jquery', 'zimArchiveLoader', 'util', 'uiUtil', 'cookies','abstractFiles
                     // The content is fully loaded by the browser : we can hide the spinner
                     $("#searchingArticles").hide();
                     // Deflect drag-and-drop of ZIM file on the iframe to Config
-                    var doc = iframeArticleContent.contentDocument.documentElement;
+                    var doc = iframeArticleContent.contentDocument;
+                    doc = doc ? doc.documentElement : null;
```
mossroy: It would be better to use different variable names here for the contentDocument and its documentElement to avoid confusion

mossroy:
Maybe I spoke too soon : Firefox 45 UI tests seem to fail on this PR, even after restarting the build.
I don't see why for now...

Jaifroid:
I’ve been getting that Firefox failure a lot, sometimes repeatedly, and then randomly it’s fixed. Nothing material has changed with this code, and I don’t see any typos. However, I’m about to push a commit that doesn’t re-use one variable, though it really shouldn’t make any difference...

Sent from Mail<https://go.microsoft.com/fwlink/?LinkId=550986> for Windows 10

Jaifroid:
OK, nothing to do with that variable, then (failed again). This is Travis, not this PR. I was getting precisely that failure: `Timed out while waiting for element <#articleList> to be visible for 20000 milliseconds.` for FF45 multiple times on previous PRs. It even failed once on `master` after we had merged a PR that had itself had already passed. You can still see that failure in the Commits for Master (see screenshot below), and you can visit the failed Travis log and you'll find the error is exactly the same as this one...

image

mossroy:
All right.
In this case you can squash and merge this PR.
We should open a new issue for these failures on Travis.
After a few restarts, some builds manage to pass : it looks like a random issue

Issue #526, opened 2019-06-29
Upgrade Bootstrap to version 4


Jaifroid:
I've worked on upgrading Bootstrap to version 4. It's rather different. For example, panels have been replaced with cards. I don't think this should be in Release 2.6 because we need to test and there is still some adapting of html elements to do. However, I have managed mostly to reproduce our current design in version 4.

Issue #470, opened 2019-01-13
Find a way to make the html for Bootstrap alerts persistent, and avoid putting HTML code in uiUtil.js


Jaifroid:
In line with https://github.com/kiwix/kiwix-js/pull/467/files#r247345420, we need to find a way to make Bootstrap alerts remain in the DOM (and hide them when not shown) rather than allowing Bootstrap to remove them from the DOM (which is the native Bootstrap behaviour). This avoids having to re-create the alert in JavaScript each time we wish to display it.

mossroy:
More generally, we should not generate HTML strings in javascript code. It makes maintenance complicated (for example if we re-design the UI)

Jaifroid:
@mossroy, you may have missed it, but in #527 I pointed out that this issue could be closed very easily by adopting one of the commits from that PR, and it may be worth doing it for #525 release. I copy the comment below for convenience:

> FYI Commit https://github.com/kiwix/kiwix-js/pull/527/commits/c463ba40e7f5ca462685d3d7f2be8e4f1db70ce4 is a proposed fix for #470, i.e., we no longer have to create the alert boxes' html in code each time we want to display them.
>
> However, as you can see, it doesn't save any code; on the contrary, it adds to the amount of code, because we have to implement an override on Bootstrap's own `.alert('close')` function. The only advantage of this code is that it maintains the separation between html and code, but at the cost of a little bit more complexity.
>
> This commit (as opposed to the whole PR) could be added to a separate branch and we could have a PR on it as a solution to #470, if you'd prefer to do it that way, @mossroy . Just let me know, and also let me know if you think this way of doing it is better than what it replaces (just focus on commit https://github.com/kiwix/kiwix-js/pull/527/commits/c463ba40e7f5ca462685d3d7f2be8e4f1db70ce4).

Jaifroid:
This issue was fixed in #527. Closing.

PR #527, opened 2019-06-29
Upgrade Bootstrap to version 4

Jaifroid:
This implements #526. I will document the reasons why certain changes had to be made for this version.


Jaifroid:
These are the main changes.

* The built-in colours in Bootstrap 4 have become very garish. See https://getbootstrap.com/docs/4.0/utilities/colors/#background-color . I tried the standard colours, and frankly, they look horrible! So I have included a code block in app.css that overrides them with the current colours we use.

* Panels have been replaced with cards. See https://getbootstrap.com/docs/4.3/components/card/ . They are pretty similar, but you have to do border styling, etc., yourself. Basically, Bootstrap has separated out all the colouring elements, probably because people wanted to customize them anyway.

* Glyphicons have been removed from Bootstrap 4, and are no longer licensed. It is recommended to use Fontawesome instead. I've done this, but haven't yet added the licence info for Fontawesome (it is compatible with Open Source). I have only used the solid icons, as it minimizes the size, and we only use a small subset.

* Headings no longer play nicely with plain text that is not wrapped in a `<p>` or similar, as the margin-bottom is provided by the `<p>` element. Otherwise there is no spacing between text and the next header. This mostly affects our About page. I've inserted `<p>` instead of `<br>` where it makes a difference. I made some incidental updates (e.g. new Ray Charles), but About could generally do with a rewrite. That's a separate issue.

Jaifroid:
NB I haven't done the alerts yet or other dynamic Bootstrap content. With the alerts, we may be able to tackle #470 at the same time, but I haven't looked into the changes yet.

Jaifroid:
FYI Commit https://github.com/kiwix/kiwix-js/pull/527/commits/c463ba40e7f5ca462685d3d7f2be8e4f1db70ce4 is a proposed fix for #470, i.e., we no longer have to create the alert boxes html in code each time we want to display it.

However, as you can see, it doesn't save any code; on the contrary, it adds to the amount of code, because we have to implement an override on Bootstrap's own `.alert('close')` function. The only advantage of this code is that it maintains the separation between html and code, but at the cost of a little bit more complexity.

This commit (as opposed to the whole PR) could be added to a separate branch and we could have a PR on it as a solution to #470, if you'd prefer to do it that way, @mossroy . Just let me know, and also let me know if you think this way of doing it is better than what it replaces (just focus on commit https://github.com/kiwix/kiwix-js/pull/527/commits/c463ba40e7f5ca462685d3d7f2be8e4f1db70ce4).

kelson42:
@Jaifroid Version of Bootstrap in `www/index.html` should be updated too.

www/index.html
```diff
@@ -156,27 +154,28 @@ <h3>Credits</h3>
...
+                    <br>along with Kiwix (file LICENSE-GPLv3.txt).  If not, see <a href="http://www.gnu.org/licenses/">http://www.gnu.org/licenses/</a>
                     </i>
-                    <br />
-                    <br />Main libraries and ressources used :
+                    <br>
+                    <br>Main libraries and ressources used :
                     <ul>
                         <li><a href="http://jquery.com/" target="_blank">jQuery</a> 3, released under the <a href="http://jquery.org/license" target="_blank">MIT license</a></li>
-                        <li><a href="http://getbootstrap.com/" target="_blank">Bootstrap</a> 3, released under the <a href="http://www.apache.org/licenses/LICENSE-2.0" target="_blank">Apache license 2.0</a>, which includes icons from <a href="http://glyphicons.com/">Glyphicons</a></li>
+                        <li><a href="http://getbootstrap.com/" target="_blank">Bootstrap 4</a>, released under an <a href="https://github.com/twbs/bootstrap/blob/master/LICENSE" target="_blank">MIT Licence</a></li>
```
Jaifroid: @kelson42 Do you mean this line (licence info)? If so, it's already in the PR, but the diff was hidden due to large number of (superficial) changes.

Jaifroid:
I've rebased this branch/PR on master.

www/index.html
kelson42: @Jaifroid Yes, Sorry missed it.

mossroy:
I don't forget this PR, but will need more time to look at it.

Jaifroid:
No rush! After your holiday is fine.

mossroy:
I finally took a bit of time to have a look at this PR.
It seems to work fine : nice job!
I like the fact that the current field or the clicked/tapped button is highlighted. It also indicates in which section we just jumped, which is convenient.
And the flat design is trendy.

I'm OK with fixing #470 in this same PR, and happy to see that we do not put HTML strings in javascript files any more. I find that much better than the previous version, even if it adds a few lines of code.

There are a few minor issues that I noticed :
- there is a small useless gap between the top of the window and the top bar. It seems to come from the margin of `[role=region] > header` in app.css
- on Firefox OS, in the Configure section, there is a hyperlink to rescan the SD cards `Click here to rescan...` but the `here` hyperlink is not visible : it looks like some standard text (see screenshot below)
- on Firefox OS, when you open the menu, the items are surrounded like buttons, which is not very beautiful (see screenshot below). But, if it's FirefoxOS-specific, do not bother to try to fix it, it's very minor

- on any browser, if the screen is small enough to have the sandwich menu and you click on an item, the menu stays expanded, where it was automatically collapsed in previous versions. I found the previous behavior more convenient

Jaifroid:
Thanks @mossroy . I'll take a look at those issues. One thing to note, however, is that if the intention is to use this as a stepping-stone towards a similar interface to that provided in Kiwix JS Windows, then it may not be worth trying to fix these issues yet, but to tidy up once we have the full new interface. I'm not sure. On the other hand, I think most of the issues are probably pertinent to functionality we'll want with the new interface. I don't currently use a collapsing toggle, but it may be necessary to continue supporting very narrow screens. We'll see.

www/index.html
```diff
@@ -53,81 +57,76 @@
...
-                  <!-- Collect the nav links, forms, and other content for toggling -->
-                  <div class="collapse navbar-collapse navbar-ex1-collapse">
-                    <ul class="nav navbar-nav">
-                      <li id="liHomeNav" class="active"><a href="#" id="btnHome">Home</a></li>
-                      <li id="liConfigureNav"><a href="#" id="btnConfigure">Configure</a></li>
-                      <li id="liAboutNav"><a href="#" id="btnAbout">About</a></li>
-                    </ul>
-                  </div><!-- /.navbar-collapse -->
-                </nav>
+                    <!-- Toggler/collapsibe Button -->
```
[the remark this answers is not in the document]
Jaifroid: Yes. Have corrected in my local copy and will push with other changes.

www/js/lib/uiUtil.js
```diff
@@ -111,17 +111,17 @@ define([], function() {
...
-                '<a href="#" class="close" data-dismiss="alert" aria-label="close">&times;</a>' +
-                '<strong>Unable to display active content:</strong> This ZIM is not fully supported in jQuery mode.<br />' +
-                'Content may be available by searching above (type a space or a letter of the alphabet), or else ' +
-                '<a id="swModeLink" href="#contentInjectionModeDiv" class="alert-link">switch to Service Worker mode</a> ' +
-                'if your platform supports it. &nbsp;[<a id="stop" href="#displaySettingsDiv" class="alert-link">Permanently hide</a>]' +
-            '</div>';
-        document.getElementById('alertBoxHeader').innerHTML = alertHTML;
+        var alertActiveContent = document.getElementById('activeContent');
+        alertActiveContent.style.display = 'block';
+        if (activeContentWarningSetup) {
```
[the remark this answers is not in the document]
Jaifroid: Fixed in local copy.

Jaifroid:
> on any browser, if the screen is small enough to have the sandwich menu and you click on an item, the menu stays expanded, where it was automatically collapsed in previous versions. I found the previous behavior more convenient

Fixed in local copy

Jaifroid:
> there is a small useless gap between the top of the window and the top bar. It seems to come from the margin of [role=region] > header in app.css

I think this is also fixed (but I found it hard to see, so not entirely sure). I've pushed all fixes so far.

I'm still missing: 

* Menu button surround. This seems to be specific to FFOS: in browsers, the buttons are only surrounded when selected, but we could probably replace the buttons with links. However, as we're going to replace the whole interface, I think this one should be fixed later if it persists.
* SD card rescan. This could be turned into a button perhaps.
* Selection of ZIM in FFOS: we probably need to remove or add a style.

I propose to fix the above after replacing the text-based system, because lots of things could change. Please let me know if you're happy for me to squash rebase.

Jaifroid:
Hmm. "CodeFactor" doesn't seem to like the bootstrap library code. Can it be marked as library code, so it doesn't get checked?

mossroy:
The code of your latest commits looks good.
I did not have time to test, but I think you can rebase/squash/merge if you did.
Don't bother for FirefoxOS. I was mentioning theses minor issues in case they would affect other platforms too.
Replacing the hyperlink to scan for ZIM files by a button is a valid workaround.
Regarding the button surround, I think it's not worth investing time on it.

I did not do anything to activate CodeFactor. I suppose it's a general modification done at the hackathon. It should not prevent you from merging this PR.

Jaifroid:
I changed the FFOS rescan link to a button, and it now looks like this in the simulator (NB I intentionally opened the menu here):

image

As you see here, the code I pushed yesterday means the Configure button no longer remains selected, so no longer has the surround.

mossroy:
Great!

Jaifroid:
Hmm, the code yesterday seems to have a race condition (I guess), with jQuery not being ready for a bootstrap function. I'm getting this in SW mode on Edge. Need to do more testing and fix before I merge this.

image

Jaifroid:
OK, so it's sufficient to make the menu collapsing functions async by using a setTimeout with no number. This has the effect of running them on the next "tick" of the JS event loop, by which time bootstrap has been able to patch its functions into jQuery.

What do you think (see commit https://github.com/kiwix/kiwix-js/pull/527/commits/9a19ef2f725bb802b946d3db2c10508ff3c607fc)? Is it a bit ugly? This is what we had before (which wasn't working with Bootstrap 4):
```
if ($('#navbarToggle').is(":visible") && $('#liHomeNav').is(':visible')) {
    $('#navbarToggle').click();
}
```
Thoughts?

Jaifroid:
PS If it's OK, I'll add a comment to explain. Also, we could probably get away with only having this timeout on the btnConfigure funciton (I think), as it's the one that runs immediately before bootstrap has finished loading.

kelson42:
@Jaifroid @mossroy Not sure if I have already said something about that, but a lot neater would be to have bootstrap 4 handled properly as an external dependency. Mixing your own source code with the external libraries in your code repository is a bad approach which has many drawbacks.

Jaifroid:
@kelson42, I guess you mean something like using node packet management? This project isn't currently built with node/npm/yarn, though we do use `require.js` to manage dependency loading and unloading. Actually, your comment has made me wonder if it's `require` that is causing the issue here. @mossroy, do you think it might be solved by changing the order in which require loads the dependencies? (I don't really understand the details of how require.js works...)

kelson42:
@Jaifroid Yes, usually for this kind of project, npm + webpack does the job. It is a bit of work to setup, but this is worth it.

Jaifroid:
@mossroy I experimented with changing the order of dependency loading in bootstrap, but I couldn't make it solve this issue. So, I've pushed a minimal version that does solve the issue: a single setTimeout on the btnConfigure function. We don't need to setTimeout on the other UI click functions, because none of them are run early in the load process.

I've also included a fix for a long-standing minor irritation that is made more obvious in Bootstrap 4: the fact that some UI elements retain focus after an article is loaded (e.g., the random button remains "pressed" if it was used to load an article). It's a simple fix added to the article onload events (jQuery and SW) placing the focus on the article content.

mossroy:
@Jaifroid : if I understand correctly, we need to make sure that jQuery and Bootstrap are loaded before app.js is.
That's the job of require.js.
Did you try to add 'jquery' on line 41 of init.js (before bootstrap)?

In any case, using a setTimeout is not safe (and it's probably why it fails on Chrome 58 in Travis builds, even after restarting it)

@kelson42 : I created #554 for npm+webpack

Jaifroid:
@mossroy, The problem is that by the time app.js runs, bootstrap hasn't inserted its custom alert box and menu toggling functions into jQuery's object. Clearly this is done asynchronously for performance reasons, so I tried running that one command asynchronously too. We ran into this issue before with the `.alert('close')` methods. In those cases we decided to use standard jQuery hide statements instead, but in this case the action is more complex than hiding.

I assume by setTimeout not being "safe", you mean that it just masks a race condition? I agree, but I'm a bit at a loss with how to get require to do its job. I've tried:

* systematically putting bootstrap before jquery in init.js --> no effect;
* rewriting line 41 to `requirejs(['jquery', 'bootstrap', '../app']);` --> no effect;
* rewriting line 41 to `requirejs(['bootstrap', 'jquery', '../app']);` --> this reduces the frequency of the error, but it's still there (in both Edge and Chromium);
* combinations of the above --> no improvement.

I shall experiment with using a click event, as we originally had, but this was in our code and wasn't working. Explicitly closing the menu does seem neater than blindly clicking. I'll report back, but if you have further ideas about requirejs, let me know.

Jaifroid:
I've come up with an alternative method https://github.com/kiwix/kiwix-js/pull/527/commits/dd3d6881555ae21323a780647546b9b6d40ef3ca using CSS. It has the disadvantage that the menu doesn't close smoothly. So I've put a quick test for the existence of the Bootstrap function (in jQuery). So setTimeout has gone completely from this. **However, it's still failing with Chrome 58.** 😒 I can't see a good reason why... Help!

Jaifroid:
(Latest commit is a test of Travis...)

Jaifroid:
Nope. 😒

mossroy:
If bootstrap runs some code asynchronously, there are chances that it provides a callback, or some other way to run our code only when it's ready.
Maybe I'll find some time to have a look tomorrow.

Jaifroid:
> If bootstrap runs some code asynchronously, there are chances that it provides a callback, or some other way to run our code only when it's ready.

But I've already bypassed any asynchronous code with:
```
if (typeof jQuery().collapse === 'function') $('.navbar-collapse').collapse('hide');
else document.getElementById('collapsibleNavbar').classList.remove('show');
```
And Travis is still failing, even though the code now works perfectly fine in all browsers I can test...

www/js/init.js
```diff
@@ -26,7 +26,7 @@ require.config({
     baseUrl: 'js/lib',
```
mossroy: It's probably not the cause of our issues, but loading the fontawesome javascript files should be done by requirejs too. If you remove the `script` tags in index.html and apply the suggestion below, it should be enough.

I did not find any reference of asynchronous things done by bootstrap, so I suspect it could come from the way we configure requirejs.
This piece of code makes sure bootstrap is loaded before running app.js (in the sense of requirejs). I think it was not the case before, and requirejs was loading bootstrap and app.js in parallel.
Could you try this (without the CSS workaround)?

```suggestion
    baseUrl: 'js/lib',
    paths: {
        'jquery': 'jquery-3.2.1.slim',
        'bootstrap': 'bootstrap.bundle',
        'fontawesome': 'fontawesome/fontawesome',
        'fontawesome-solid': 'fontawesome/solid'
    },
    shim: {
        'jquery' : {
            exports : '$'
        },
        'bootstrap': {
            deps: ['jquery', 'fontawesome', 'fontawesome-solid']
        }
    }
});

requirejs(['bootstrap'], function (bootstrap) {
    requirejs(['../app']);
});

```

Jaifroid: @mossroy Thank you! This code seems to solve the async loading issue, and means that the bootstrap functions are ready and available for use as soon as app.js runs. I've removed all workarounds and I don't see the error appearing on quick tests.

However, as you can see, Travis is still failing with the same error on Chrome 58... 😒

mossroy:
On Chrome 58, what happens is that the search field is empty when trying to search for an article. So it does not find the article "Ray Charles" it's looking for.
image
I'll send you an email to give you the credentials and details to be able to restart Travis builds and see the screenshots on Sauce

Jaifroid:
> On Chrome 58, what happens is that the search field is empty when trying to search for an article. So it does not find the article "Ray Charles" it's looking for

Hmm. This might be fixed by #563, though it's odd that the text has disappeared (and only in this browser). It could be a race condition produced by the rate-limiter we have for typing in the search field, but that should have been picked up elsewhere. So most likely to be a combination of timing issues introduced by a change in the loading sequence in the new bootstrap.

mossroy:
Not sure if the text disappears or never reaches the search field.

Jaifroid:
OK, I'll rebase on Master shortly to see if the (possible) fix to the Firefox error makes any difference here.

Jaifroid:
So I've rebased on master, and it unfortunately doesn't cure the Chrome test failure. I've checked in sauce labs, but as you said, @mossroy, the keys R-a-y are never registered by the search field. This only occurs in Chrome. They are clearly visible in the other browsers. The http response (visible in the screenshot) is the same as in Firefox (status 200 and null). So it's a bit of a mystery. I'm back to wondering if our timeout delay (to avoid a race condition when typing fast) is interacting with sendkeys and Bootstrap in a weird way in Chrome. Could Chrome have its own form of rate limiter? I'll see if I can test more at the weekend.

image

Jaifroid:
Looking at my Commit list, one other idea occurs to me: we remove the focus from UI elements as soon as the document is loaded in order to fix one of the issues with FFOS display in the collapsed menu. It could be that, due to the async "optimizations" in Chrome, this blurring occurs after Travis has selected the prefix field, and before the keypresses are sent (maybe exacerbated by the setTimeout on the prefix field). It seems like a plausible explanation that needs testing.

Jaifroid:
So that was indeed the problem. The diagnostic commit comments out the line that removes focus from the UI elements, and it now passes.

So now we have to decide how to deal with this. In some ways this error can be considered an artefact of the testing system, because no user will select the field and type so fast after article loads. These possibilities occur to me:

* Introduce a delay in Travis between checking that the article has loaded and selecting the prefix field, so that it gets selected after we have removed focus;
* Introduce a delay (a setTimeout) on the command to blur the UI elements. This would be a race condition, and I'm not sure it's a good idea, as we'd have to hit the sweet-spot between Travis working and the user not being inconvenienced by a delayed blur while they are typing;
* Do not remove focus from the UI elements on article load. The effect of this on mobile devices would be that the keyboard would pop up always at least on the landing page. I know we used to have that behaviour, but at some point it appears to have been dropped. There probably is some other effect, because the test error occurred after I fixed one of the FFOS issues mentioned above.

Any other thoughts?

mossroy:
What if you moved this piece of code, so that it is ran synchronously, when starting to read the article, instead of running asynchronously in the onload event (after the article is displayed)?

Jaifroid:
Thanks for the helpful suggestion. Although it didn't solve the issue with Chrome directly, it did help simplify the code a little, and to understand what we're doing. It looks like we have a solution. Below is the explanation in case you're curious! But, subject to a bit more testing, I think the code could be ready to merge now, if you agree @mossroy.

**Explanation:**

In response to your noticing that in FFOS some buttons remained pressed, I introduced a line to blur any focused UI element on page load. It returns the interface to a neutral state once an article is loaded. In most browsers this worked fine, but in Chrome it caused a failure because the blur conflicts with the sendkey routine for an article search. In fact, the reason there is no failure in the other browsers may be accidental: many browsers defer or batch UI rendering, and maybe Chrome was being the "speedy guy" here, blurring "too soon".

Whatever the case, it turns out that the Unit Test expects the search field to be selected when the landing page loads, and doesn't actively select it before sending the search keys. This is consistent with expected behaviour, as traditionally this app selects the search field for the user on pressing the Home button. So, the solution is to blur only if we are not loading the landing page. We already have an `isLandingPage` parameter, so it's easy.

A very rare side effect of this is that it's possible that the Random key will remain pressed if hitting it results in the user loading the Landing page randomly. It's rather unlikely and very minor, with no negative effect.

www/js/app.js
```diff
@@ -899,17 +887,20 @@ define(['jquery', 'zimArchiveLoader', 'util', 'uiUtil', 'cookies','abstractFiles
      * @param {DirEntry} dirEntry The directory entry of the article to read
      */
     function readArticle(dirEntry) {
+        // If we are not loading the landing page, remove focus from the UI elements (for both jQuery and SW modes)
```
mossroy: It would be worth explaining a bit more why we have to do that

PR #529, opened 2019-07-08
Update changelog for version 2.6.0

mossroy:


mossroy:
Note that I did not write the release date yet

Jaifroid:
I wonder if we should add a note to the changelog to say that users will need to search manually for the links on the Wikivoyage landing page until a new Wikivoyage is released? I know it's a bit specific, but might be helpful to users upgrading their apps?

mossroy:
I've updated the changelog for that

Jaifroid:
Looks good, thanks!

mossroy:
I rephrased for the 2019-06 wikivoyage ZIM files, to say that it's fixed in 2019-07 ZIM files.
Now I think we're ready to rebase/squash/merge this PR and start the release

Issue #138, opened 2016-01-06
Move the service worker (or its scope), so that it catches all the request from the iframe, and not the ones from the UI


mossroy:
It's a blocker for some ZIM files (for example wikipedia_fr_all_09_2012.zim), which use absolute URLs like /A/Article.html.
And the service-worker.js file should be placed with the other javascript files (if possible)

mossroy:
@kelson42 tells me that the norm is now to have relative URLs everywhere (since the end of 2014).
So this issue is not crucial, but it would be much cleaner.

kelson42:
I would not recommend to handle cases which are not respecting the ZIM specification. That only makes the code more complicated.... for no added value. I strongly recommend to close this ticket as wont-fix.

mossroy:
No.
The goal here is not to handle faulty ZIM files, but to make the ServiceWorker only handle what it should handle.
Currently, it technically catches ALL the HTTP requests made by the browser, including the ones to load the resource of kiwix-js itself (like fetching www/js/app.js).

There is a regular expression at the beginning of the ServiceWorker that checks if the URL matches a ZIM URL. It ignores the rest, so that everything works.

But it would be much better if we managed to make the ServiceWorker catch only the requests that come from the iframe.
Now that we implemented #534, we might catch only the URLs with our prefix.
The problem is that this prefix is the ZIM file name, which we do not know when the ServiceWorker is initialized.

A solution might be to change our prefix : instead of having only the ZIM file name, we might use something like `"articleContent/" + zimFileName`, and make the ServiceWorker catch only the URLs starting with "articleContent/"
This should not be very difficult.

mossroy:
I close this issue as "won't fix" because #764 needs the ServiceWorker to catch all the requests (including the ones from the UI)

PR #534, opened 2019-07-14
Add the ZIM file name as a prefix of the iframe URL, in SW mode.

### LINKED ISSUE #530: Broken content when switching between video ZIM files in Chromium extension
The behavior is the same with kiwix-js 2.5.0 and the latest code.
It happens only in SW mode (but these files are not really working in jQuery mode)

Steps to reproduce : 
- open the kiwix-js extension in Chromium (tested with version 75)
- open the main page of ted_en_global_issues_2018-07.zim : it is displayed properly
- open the main page of dirtybiology_fr_all_2018-06.zim : you can see thumbnails of TED (and they don't work if you click on it because the link does not exist in the ZIM file)
- close and reopen Chromium, and the kiwix-js extension
- open the main page of dirtybiology_fr_all_2018-06.zim : it is displayed properly

I suspect Chromium keeps some content in cache. I suppose these ZIM files have been generated by the same code, so they might try to call the same URLs.

I did not notice this issue in older versions of Chromium.
I think we should try to fix this before releasing a version 2.6 (#525)

mossroy:
So that the browser cache does not mix content between different ZIM files.
Fixes #530 (as long as only relative URLs are used).

@kelson42 : in #138, you told me that the ZIM files should only use relative URLs (since the end of 2014). Can we rely on that?
Because, if ZIM files use absolute URLs in some hyperlinks, the prefix disappears and there might still be some cache issues.
I suspect that it still happens on some recent ZIM files

### DIFF
--- a/service-worker.js
+++ b/service-worker.js
@@ -70,7 +70,8 @@ self.addEventListener('message', function (event) {
 });
 
 // Pattern for ZIM file namespace - see https://wiki.openzim.org/wiki/ZIM_file_format#Namespaces
-var regexpZIMUrlWithNamespace = new RegExp(/(?:^|\/)([-ABIJMUVWX])\/(.+)/);
+// In our case, there is also the ZIM file name, used as a prefix in the URL
+var regexpZIMUrlWithNamespace = /(?:^|\/)([^\/]+\/)([-ABIJMUVWX])\/(.+)/;
 
 function fetchEventListener(event) {
     if (fetchCaptureEnabled) {
@@ -82,8 +83,9 @@ function fetchEventListener(event) {
                 var title;
                 var titleWithNameSpace;
                 var regexpResult = regexpZIMUrlWithNamespace.exec(event.request.url);
-                nameSpace = regexpResult[1];
-                title = regexpResult[2];
+                var prefix = regexpResult[1];
+                nameSpace = regexpResult[2];
+                title = regexpResult[3];
 
                 // We need to remove the potential parameters in the URL
                 title = removeUrlParameters(decodeURIComponent(title));
@@ -121,7 +123,7 @@ function fetchEventListener(event) {
                         resolve(httpResponse);
                     }
                     else if (event.data.action === 'sendRedirect') {
-                        resolve(Response.redirect(event.data.redirectUrl));
+                        resolve(Response.redirect(prefix + event.data.redirectUrl));
                     }
                     else {
                         console.error('Invalid message received from app.js for ' + titleWithNameSpace, event.data);
--- a/www/js/app.js
+++ b/www/js/app.js
@@ -917,7 +917,8 @@ define(['jquery', 'zimArchiveLoader', 'util', 'uiUtil', 'cookies','abstractFiles
                         $("#searchingArticles").show();
                     };
                 };
-                iframeArticleContent.src = dirEntry.namespace + "/" + encodeURIComponent(dirEntry.url);
+                // We put the ZIM filename as a prefix in the URL, so that browser caches are separate for each ZIM file
+                iframeArticleContent.src = selectedArchive._file._files[0].name + "/" + dirEntry.namespace + "/" + encodeURIComponent(dirEntry.url);
                 // Display the iframe content
                 $("#articleContent").show();
             };


Jaifroid:
@mossroy I think this PR would in theory have interactions with #498, and that one should probably be merged first. We are currently not dealing correctly with relative links in SW mode, as documented in that PR and in the associated issue #494. Having said that, I haven't yet checked the code here for conflicts.

mossroy:
Once #498 is merged, I'll rebase this PR on it

mossroy:
As discussed in #498 , let's try to merge this PR before #498.

Jaifroid:
OK, I've had a bit of a test with this using a Stackexchange ZIM, and the prefix doesn't always remain. After surfing a bit, it disappears. I'll try to document a precise procedure.

Jaifroid:
Procedure to observe the issue:

1. Load a Stackexchange ZIM. I used `mathoverflow.net_eng_all_2018-08.zim`
2. From the landing page, click on a question;
3. Hover over one of the tags, and you will see that the ZIM prefix is no longer in the URL.

kelson42:
> @kelson42 : in #138, you told me that the ZIM files should only use relative URLs (since the end of 2014). Can we rely on that?

Yes, you can assume that. Extract from the openZIM spec:
```
Absolute URLs, ie. with a leading slash (/) are forbidden, because this avoid including the ZIM contents in any HTTP sub-hierachy. ZIM contents URLs must consequently be relative. Be careful, <article_url> may itself contain slashes (for example "BMW_501/502"). 
``

**Comment by kelson42** [comment]
@Jaifroid Not sure to understand your last comment, but please report a but on [Sotoki](https://github.com/openzim/sotoki) repository if something is not OK with the `mathoverflow.net_eng_all_2018-08.zim`.

**Comment by Jaifroid** [comment]
> Be careful, <article_url> may itself contain slashes (for example "BMW_501/502").

@kelson42 What does "be careful" mean here? Since we removed `.html` from URLs, we are no longer able to distinguish between a title `BMW_501/502` in the A namespace, and a file `502` inside a subdirectory `BMW_501`, because `dirEntry.title` is now empty (it is the same as `dirEntry.url`).

What I think we need to know for sure is that any hyperlinks in the article `BMW_501/502` will have `../` added to them by mwoffliner (this was the fix in https://github.com/openzim/mwoffliner/issues/726), i.e. will be fully relative. Are you happy that will be the case? (It is in fact the problem with the Wikivoyage landing page, and is the reason the iOS app is crashing on that page: it's expecting `../` and not finding it).

**Comment by Jaifroid** [comment]
> @Jaifroid Not sure to understand your last comment, but please report a but on Sotoki repository if something is not OK with the mathoverflow.net_eng_all_2018-08.zim.

@kelson42 Stackexchange ZIMs are working perfectly and are well coded. They have subdirectories. Files in subdirectories have relative links. So I use Stackexchange ZIMs to test the relative link/subdirectory support of our clients, that's all.

**Comment by kelson42** [comment]
@kelson42 I have removed the comment which was not clear at all. I just meant that if you have absolute URLs then you don't care about how the URL needed to looks like is. But if you have relative URLs, then they depend of the the URL of the document iself... exactly the kind of problem we try to resolve properly these last months (years?) on both ZIM sides: creation & readers. This is more complex that what you write in your comment, if the content is available independently if the slash is encoded or not in the article url, the webrender which will render the corresponding article HTML is sensitive to that (with the unencoded slash you gain a level in document path depth, otherwise not).

**Comment by Jaifroid** [comment]
> if the content is available independently if the slash is encoded or not in the article url, the webrender which will render the corresponding article HTML is sensitive to that (with the unencoded slash you gain a level in document path depth, otherwise not)

OK, that's new information which should be documented. So we have to check whether a `/` has been URI encoded or not, to decide which level we're at in the path? Is that always consistent, e.g. that dirEntry.url has unencoded `/` in the path, but encoded in the title? Even in the examples you gave of 'Singapore/Marina_Bay' (https://github.com/openzim/mwoffliner/issues/726#issuecomment-500404796) or 'Singapore/Riverside'? There you seemed to suggest that the `/` should not be percent-encoded, if I understood correctly. In any case, I believe the fix in https://github.com/openzim/mwoffliner/issues/726 implemented treating hyperlinks in the article 'Singapore/Riverside' as if they were in a subirectory 'Singapore/', and it added `../` to the links, but maybe that's actually wrong...

**Comment by Jaifroid** [comment]
I should add that the *Wikipedia URL* for Singapore/Riverside is https://en.wikivoyage.org/wiki/Singapore/Riverside , which would support treating it as a subdirectory.

**Comment by mossroy** [comment]
@Jaifroid : I've commited a fix in this PR for the SW, that was necessary to keep the prefix. They were lost when there was a redirect.
I could not test a lot but it seems to be enough to fix the scenario you described in https://github.com/kiwix/kiwix-js/pull/534#issuecomment-511363774

File: service-worker.js [hunk 1]
```diff
@@ -70,7 +70,8 @@ self.addEventListener('message', function (event) {
 });
 
 // Pattern for ZIM file namespace - see https://wiki.openzim.org/wiki/ZIM_file_format#Namespaces
-var regexpZIMUrlWithNamespace = new RegExp(/(?:^|\/)([-ABIJMUVWX])\/(.+)/);
+// The ZIM file name (used as a prefix in the URL) is ignored
+var regexpZIMUrlWithNamespace = new RegExp(/(?:^|\/)([^\/]+\/)([-ABIJMUVWX])\/(.+)/);
```
Jaifroid: There's no real advantage of using the new RegExp constructor function here. You could just use:

`var regexpZIMUrlWithNamespace = /(?:^|\/)([^\/]+\/)([-ABIJMUVWX])\/(.+)/;`

But it's a matter of style, not substance.

mossroy: You're right, I'll change that.

mossroy:
Regarding adding the prefix in jQuery mode, I'm not sure. I suppose you mean put that in the browser URL, where we currently put `?title=... `?
This URL parameter is currently only to give a user feedback. The title really used for back/forward buttons is read in the `state` object of history.
We might add the ZIM file name in this `state` object, but the main issue is still to be able to programmatically open a ZIM file (which is allowed only in some specific contexts)

Jaifroid:
Yes, sorry, I wasn't thinking very clearly about what I meant by adding the prefix in jQuery mode. You're right, that we're only really talking about the iframe here in this PR. It would be completely different issue to decide how to add info to the URL that would allow launching the app (in some contexts) and then reading the ZIM file and article to open from the URL.

This PR is looking good now, and I think it can be merged.

PR #546, opened 2019-07-26
Add support for opening details sections in jQuery mode

Jaifroid:
This is a simple PR to open all details sections in jQuery mode. It supports recent ZIMs that use the details-summary schema to open and close sections in a document. In browsers that support the details-summary elements, the sections remain closed because the piece of included JavaScript in the ZIM that opens them is not able to run in jQuery mode. This PR adds the `open` attribute to all unopened `details` elements, and does it by adding a simple query and test to our existing routine for opening all sections. The PR creates consistency of approach with previous ZIM types.

We can consider adding logic here to open all details sections only if the screen size is larger than mobile width. Personally I prefer to start with all sections open even in mobile, and as a user I would like an option to override automatic decisions taken by the browser that I would find annoying (!). Therefore, if we implement opening all sections only in desktop mode, I'd vote for adding a simple configuration option "Always open all sections in Wikimedia articles".


Jaifroid:
I've added an issue for this PR #547. An additional point is that it is impossible to jump to references / endnotes when the notes section is closed, therefore we are obliged at least to open that section.

Jaifroid:
Please note that this PR is simply a revision to our **current workaround** for opening sections. It adds support for newer ZIMs, and allows endnotes to work in jQuery mode (otherwise clicking on an endnote does nothing). ISTM that until the issue is resolved in mwoffliner (and I have reported it) we should keep the workaround up-to-date and not let it 'rot'.

Jaifroid:
OK, I understand. So I'll close this for now. I haven't had a response to my report about the issue, but I do see that the current feature (details-summary tags) is looking a bit broken. See https://github.com/openzim/mwoffliner/issues/952 .

PR #551, opened 2019-08-18
Fix regex issues identified by Codefactor

### LINKED ISSUE #550: Fix 'Unnecessary escape character: \/.' issue in www\js\app.js
As identified by Codefactor. Forward slashes do not need to be escaped inside regex character groups [/]. For example (there are several instances):

[CodeFactor](https://www.codefactor.io/repository/github/kiwix/kiwix-js/overview/master) found an issue: Unnecessary escape character: \/.

It's currently on:
[www\js\app.js:1109
](https://www.codefactor.io/repository/github/kiwix/kiwix-js/source/master/www/js/app.js#L1109)

Jaifroid:
Fixes #550

### DIFF
--- a/www/js/app.js
+++ b/www/js/app.js
@@ -596,7 +596,7 @@ define(['jquery', 'zimArchiveLoader', 'util', 'uiUtil', 'cookies','abstractFiles
         if (archiveDirectory && archiveDirectory.length > 0) {
             // Now, try to find which DeviceStorage has been selected by the user
             // It is the prefix of the archive directory
-            var regexpStorageName = /^\/([^\/]+)\//;
+            var regexpStorageName = /^\/([^/]+)\//;
             var regexpResults = regexpStorageName.exec(archiveDirectory);
             var selectedStorage = null;
             if (regexpResults && regexpResults.length>0) {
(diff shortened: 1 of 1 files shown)


www/js/app.js
```diff
@@ -995,9 +995,9 @@ define(['jquery', 'zimArchiveLoader', 'util', 'uiUtil', 'cookies','abstractFiles
     
     // Compile some regular expressions needed to modify links
     // Pattern to find the path in a url
-    var regexpPath = /^(.*\/)[^\/]+$/;
+    // var regexpPath = /^(.*\/)[^/]+$/; // Kept for reference
```
mossroy: Why do you keep this commented line? (and not for the other similar modifications?)

Jaifroid: Codefactor identified this line as redundant: we never use `regexPath` any more in our code. I thought, however, that it might be useful to have the regex in the code (but commented out) for reference, in case we ever need it. However, I can delete. It's not exactly a hard regex to write if we need it again.

Jaifroid:
I added a comment to your in-code query above, but not sure it pinged you. We can just delete that line.

Yes, I often forget about not needing to escape some special characters in regex character-classes. According to https://www.regular-expressions.info/charclass.html, the only ones that need to be escaped, if they are used literally, are `^-]\` .

www/js/app.js
mossroy: You can remove it, as it's unused.
At worst, we still have it in git history.

Jaifroid:
Travis is in a bad mood. This simple PR is failing on the infamous Firefox 45 error, so I'm going to go ahead and squash/merge because we know that's not serious. However, #527 is failing with Chrome, so I suspect a more serious issue, but it's odd that it passes all the other browsers. (Need your help with that one, @mossroy..).

PR #555, opened 2019-08-21
WIP: Add native asset caching to Service Worker

Jaifroid:
This addresses #441. It adds the native Cache API to `service-worker.js` for the caching of css and js assets. It significantly speeds up SW mode. The cached entries can be checked in the Cache entry of devtools in Chromium or Edge (I didin't try with Firefox). Example of running this on the most recent Ray Charles is shown in the image below (Ifrom Chromium / Edge 78). This is not for merging yet, but I would appreciate testing and any comments.

image

### DIFF
--- a/service-worker.js
+++ b/service-worker.js
@@ -23,38 +23,49 @@
  */
 'use strict';
 
-self.addEventListener('install', function(event) {
+var CACHE = 'kiwixjs-cache';
+
+self.addEventListener('install', function (event) {
     event.waitUntil(self.skipWaiting());
 });
 
-self.addEventListener('activate', function(event) {
+self.addEventListener('activate', function (event) {
     // "Claiming" the ServiceWorker is necessary to make it work right away,
     // without the need to reload the page.
     // See https://developer.mozilla.org/en-US/docs/Web/API/Clients/claim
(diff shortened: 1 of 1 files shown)


Jaifroid:
I think the main TODO is to provide a mechanism for deleting / pruning entries. We could just do it on change of ZIM, but as the quota on most devices is quite generous, and as the entries are unique to each ZIM, a more subtle mechanism could be found. We could provide usage statistics on the Configuration page and a button to clear the cache if it is quite large, for example. This would need to be decided.

Jaifroid:
Info for testing: To delete the cache and Service Worker after testing (in Chromium), delete `kiwixjs-cache` (right-click and delete) and unregister the Service Worker (under Application -> Service Workers ... ).

If there are code updates, a similar procedure would need to be followed, then a full page refresh (e.g. Ctrl-Shift-R), before testing again.

Jaifroid:
I'm closing and re-creating this PR because for some very strange reason GitHub Web is no longer tracking commits to the branch. I've tried squashing and rebasing, and it completely ignores changes, even though they're appearing in the branch. So something has come uncoupled. Strange.

Issue #414, opened 2018-09-18
Use the Cache API of ServiceWorkers to cache content


mossroy:
If #411 can't be implemented, we'll have to fallback to cache the content ourself.

One approach would be to use the Cache API of ServiceWorkers
https://developer.mozilla.org/en-US/docs/Web/API/Cache , that has been designed for this purpose.
See examples from https://serviceworke.rs/caching-strategies.html

Issue #507, opened 2019-06-01
Refactor app.js to reduce its size/complexity


mossroy:
The file app.js now has around 1400 lines.
It's way too much IMHO.
We should find a way to reduce its size and complexity, so that newcomers can get into it more easily. It would also be easier for us to find what we're looking for.

One approach would be to split it in different .js files.
For example, we might move some functions that are specific to one mode (jQuery or SW) in a specific file for each mode. But it's just an idea.

It might be not so easy because of some global variables that we use in app.js, that would not be accessible from other files. Maybe passing them as parameters could be enough in most cases (if the functions do not modify them)

Jaifroid:
Good idea. It may be worth pointing out that any variables set in `init.js` become global to the whole app. It is suitable, for example, for setting global parameters retrieved from cookies, e.g. a global params object.

ykabusalah:
Can I try refactoring this?

Jaifroid:
@ykabusalah Unless you're familiar with the codebase, I would suggest that this is quite a complex task and not a good first issue.

Issue #415, opened 2018-09-18
Cache resources through localStorage/indexedDB on the backend side


mossroy:
This would be an alternative to #414.
@Jaifroid has already done some hard work in branch https://github.com/kiwix/kiwix-js/tree/Persist-cache-in-localStorage

We have to decide if we prefer to implement this or #414.
An advantage for this solution is that it could probably be made to work the same for both jQuery and ServiceWorker modes.

mossroy:
On the other hand, there are some drawbacks : 
- the cached content would have to go through app.js, instead of being directly retrieved by the ServiceWorker

And some things to pay attention to :
- I think we need to refactor the code, so that this cache is handled by an extra layer between app.js and the ZIM backend. It would make things less complicated in app.js, while allowing to replace our current ZIM backend with something compiled from libzim (#116). It would also allow to use the same cache for any kind of content (and to remove the existing CSS cache implementation)
- the branch code is perfect for comparing different cache implementations (memory, localstorage, indexeddb). But, if we want to be able to cache binary objects (like images), I suppose indexeddb is more suitable. https://caniuse.com/#search=indexeddb seems to say that there are only a few limitations in Microsoft browsers : it might be worth a try. If not, something like https://hacks.mozilla.org/2014/02/localforage-offline-storage-improved/ might help us having only one cross-browser implementation. There are interesting pointers for the cache topic in https://developers.google.com/web/fundamentals/instant-and-offline/web-storage/offline-for-pwa

And things to do, that are the same if we choose #415 :
- we need to choose and implement a cache retention strategy, so that the cache does not grow up forever. I suppose we should not try to reinvent the wheel. It looks like IndexedDB and the Cache API of ServiceWorkers implement it for us, if we use a "temporary" storage. See https://developer.mozilla.org/en-US/docs/Web/API/IndexedDB_API/Browser_storage_limits_and_eviction_criteria
- we also need to be able to separate caches coming from different ZIM files. Each ZIM file should have its own cache

Jaifroid:
> I think we need to refactor the code, so that this cache is handled by an extra layer between app.js and the ZIM backend

Do you mean something different from [cache.js](https://github.com/kiwix/kiwix-js/blob/Persist-cache-in-localStorage/www/js/lib/cache.js), which is the current cache layer I'm using? I tried to put most of the code in there, but what remains in app.js are mostly the intercept hooks. Maybe more of the code can be pushed to cache.js?

[...]

Issue #411, opened 2018-09-17
Make use of the browser standard cache in ServiceWorker mode


mossroy:
We have a ServiceWorker that traps the HTTP requests and reads the content inside the ZIM file.
It works fine, but we would greatly improve performance if we allowed the browser engine to make use of its standard HTTP cache.

I thought it would be simple to add an `ETag` in the responses coming from the ServiceWorker to the browser, so that next time a resource is needed, the ServiceWorker would get the request with a `If-None-Match` HTTP header. And the ServiceWorker could answer that the resource is unchanged.

But I did not manage to make it work so far, because I can not get the `If-None-Match` request header in the ServiceWorker.
It looks like the `event.request.header` object does not have all the headers, but only a few of them. I've created https://stackoverflow.com/questions/51696766/how-to-make-browsers-pass-all-http-headers-to-a-serviceworker in the hope to have some help.
The code of my initial (non-working) tests is in branch use-browser-cache-in-serviceworker-mode

mossroy:
Based on https://github.com/w3c/ServiceWorker/issues/212 , I suppose that, inside the ServiceWorker spec, there are some rules to not expose all the request headers. At least when the request is not for the same origin.
Could we be in such a case?

I tested by putting the stackoverflow test-case on a distant https web server (so that I'm sure everything comes from that same origin), it did not change the result. So it's probably not a question of origin.

mossroy:
Maybe another way to achieve what we want is to use the fetch API with the "only-if-cached" parameter, in order to grab a cached version if it exists?
See https://developer.mozilla.org/en-US/docs/Web/API/Request/cache

Jaifroid:
I don't have the knowledge to contribute to an httpd solution, though it certainly sounds like the best way to do it if it works with localhost. I wonder if some browsers disable caching for localhost?

Without wishing at all to distract you from doing it this way, I did develop back in May (intending to use this cleaner code to replace my current filesystem caching in Kiwix JS Windows) a caching solution that uses indexeddb, localStorage or memory cache, according to the capabilities of the browser. It can persist (if the user wishes) the cache across browsing sessions (if the browser can use indexeddb or localStorage). I applied it to Service Worker, but don't know if I'm catching Service Worker requests the right way. Maybe something like my solution could be used for jQuery mode, alongside an httpd version for Service Worker?

[...]

PR #556, opened 2019-08-22
Add native asset caching to Service Worker

Jaifroid:
This addresses ~~#441~~ #414. It adds the native Cache API to `service-worker.js` for the caching of css and js assets. It significantly speeds up SW mode. The cached entries can be checked in the Cache entry of devtools in Chromium or Edge (I didin't try with Firefox). Example of running this on the most recent Ray Charles is shown in the image below (from Chromium / Edge 78). This is not for merging yet, but I would appreciate testing and any comments.

image

I think the main TODO is to provide a mechanism for deleting / pruning entries. We could just do it on change of ZIM, but as the quota on most devices is quite generous, and as the entries are unique to each ZIM, a more subtle mechanism could be found. We could provide usage statistics on the Configuration page and a button to clear the cache if it is quite large, for example. This would need to be decided.

Info for testing: To delete the cache and Service Worker after testing (in Chromium), delete `kiwixjs-cache` (right-click and delete) and unregister the Service Worker (under Application -> Service Workers ... ).

If there are code updates, a similar procedure would need to be followed, then a full page refresh (e.g. Ctrl-Shift-R), before testing again.

### DIFF
--- a/service-worker.js
+++ b/service-worker.js
@@ -29,10 +29,7 @@ var cachedContentTypesRegexp = /text\/css|text\/javascript|application\/javascri
 // DEV: add any URL schemata that should be excluded from caching with the Cache API to the regex below
 // As of 08-2019 the chrome-extension: schema is incompatible with the Cache API
 // 'example-extension' is included to show how to add another schema if necessary
-// You can test this code by temporarily changing 'example-extension' to 'http' and running on localhost
 var excludedURLSchema = /^(?:chrome-extension|example-extension):/i;
-// This Map will be used as a fallback volatile cache for the URL schemata not supported above
-var assetsCache = new Map();
 
 // Pattern for ZIM file namespace - see https://wiki.openzim.org/wiki/ZIM_file_format#Namespaces
 // In our case, there is also the ZIM file name, used as a prefix in the URL
@@ -73,7 +70,7 @@ self.addEventListener('fetch', function (event) {
                     return fetchRequestFromZIM(event).then(function (response) {
                         // Add css or js assets to CACHE (or update their cache entries) unless the URL schema is not supported
                         if (cachedContentTypesRegexp.test(response.headers.get('Content-Type')) &&
-                                !excludedURLSchema.test(event.request.url)) {
+                            !excludedURLSchema.test(event.request.url)) {
                             console.log('[SW] Adding ' + event.request.url + ' to CACHE');
                             event.waitUntil(updateCache(event.request, response.clone()));
                         }
@@ -149,13 +146,6 @@ function fetchRequestFromZIM(fetchEvent) {
                     headers: headers
                 };
 
-                // If we are dealing with an excluded schema, store the response in assetsCache instead of Cache
-                // NB we have to store the data in its constitutent format, otherwise the Response is expired by the system 
-                if (excludedURLSchema.test(fetchEvent.request.url) && cachedContentTypesRegexp.test(contentType)) {
-                    console.log('[SW] Adding EXCLUDED schema URL ' + fetchEvent.request.url + ' to assetsCache');
-                    assetsCache.set(fetchEvent.request.url, [msgPortEvent.data.content, responseInit]);
-                }
-
                 var httpResponse = new Response(msgPortEvent.data.content, responseInit);
 
                 // Let's send the content back from the ServiceWorker
@@ -188,21 +178,14 @@ function removeUrlParameters(url) {
  * @returns {Response} The cached Response (as a Promise) 
  */
 function fromCache(request) {
-    // If the response has been stored in assetsCache, it is an excluded URL schema
-    if (assetsCache.has(request.url)) {
-        var data = assetsCache.get(request.url);
-        var response = new Response(data[0], data[1]);
-        return Promise.resolve(response);
-    } else {
-        return caches.open(CACHE).then(function (cache) {
-            return cache.match(request).then(function (matching) {
-                if (!matching || matching.status === 404) {
-                    return Promise.reject("no-match");
-                }
-                return matching;
-            });
+    return caches.open(CACHE).then(function (cache) {
+        return cache.match(request).then(function (matching) {
+            if (!matching || matching.status === 404) {
+                return Promise.reject("no-match");
+            }
+            return matching;
         });
-    }
+    });
 }
 
 /**
--- a/service-worker.js
+++ b/service-worker.js
@@ -23,16 +23,34 @@
  */
 'use strict';
 
-// We do not define CACHE here in order to avoid duplication; it will be given a value on init
+/**
+ * The name of the Cache API cache in which assets defined in regexpCachedContentTypes will be stored
+ * The value is defined in app.js and will be passed to Service Worker on initialization (to avoid duplication)
+ * @type {String}
+ */
 var CACHE;
+
+/**
+ * A global Boolean that governs whether CACHE will be used
+ * Caching is on by default but can be turned off by the user in Configuration
+ * @type {Boolean}
+ */
 var useCache = true;
-// DEV: add any Content-Types you wish to cache to the regexp below, separated by '|'
-var cachedContentTypesRegexp = /text\/css|text\/javascript|application\/javascript/i;
-// DEV: add any URL schemata that should be excluded from caching with the Cache API to the regex below
-// As of 08-2019 the chrome-extension: schema is incompatible with the Cache API
-// 'example-extension' is included to show how to add another schema if necessary
-// See equivalent regex in app.js function refreshCacheStatus() and ensure both are the same
-var excludedURLSchema = /^(?:chrome-extension|example-extension):/i;
+
+/**  
+ * A regular expression that matches the Content-Types of assets that may be stored in CACHE
+ * Add any further Content-Types you wish to cache to the regexp, separated by '|'
+ * @type {RegExp}
+ */
+var regexpCachedContentTypes = /text\/css|text\/javascript|application\/javascript/i;
+
+/**
+ * A regular expression that excludes listed schemata from caching attempts
+ * As of 08-2019 the chrome-extension: schema is incompatible with the Cache API
+ * 'example-extension' is included to show how to add another schema if necessary
+ * @type {RegExp}
+ */
+var regexpExcludedURLSchema = /^(?:chrome-extension|example-extension):/i;
 
 // Pattern for ZIM file namespace - see https://wiki.openzim.org/wiki/ZIM_file_format#Namespaces
 // In our case, there is also the ZIM file name, used as a prefix in the URL
@@ -71,8 +89,8 @@ self.addEventListener('fetch', function (event) {
                     // and add it to the cache if it is an asset type (css or js)
                     return fetchRequestFromZIM(event).then(function (response) {
                         // Add css or js assets to CACHE (or update their cache entries) unless the URL schema is not supported
-                        if (cachedContentTypesRegexp.test(response.headers.get('Content-Type')) &&
-                            !excludedURLSchema.test(event.request.url)) {
+                        if (regexpCachedContentTypes.test(response.headers.get('Content-Type')) &&
+                            !regexpExcludedURLSchema.test(event.request.url)) {
                             event.waitUntil(updateCache(event.request, response.clone()));
                         }
                         return response;
@@ -227,7 +245,7 @@ function updateCache(request, response) {
  * @returns {Promise} A Promise that resolves with an array of format [cacheType, cacheDescription, assetCount]
  */
 function testCacheAndCountAssets(url) {
-    if (excludedURLSchema.test(url)) return Promise.resolve(['custom', 'Custom', '-']);
+    if (regexpExcludedURLSchema.test(url)) return Promise.resolve(['custom', 'Custom', '-']);
     if (!useCache) return Promise.resolve(['none', 'None', 0]);
     return caches.open(CACHE).then(function (cache) {
         return cache.keys().then(function (keys) {
--- a/service-worker.js
+++ b/service-worker.js
@@ -140,7 +140,7 @@ function fetchRequestFromZIM(fetchEvent) {
                 // See kiwix-js #519 and openzim/zimwriterfs #113 for why we test for invalid types like "mp4" or "webm" (without "video/")
                 // The full list of types produced by zimwriterfs is in https://github.com/openzim/zimwriterfs/blob/master/src/tools.cpp
                 if (contentLength >= 1 && /^(video|audio)|(^|\/)(mp4|webm|og[gmv]|mpeg)$/i.test(contentType)) {
-                    // In case of a video (at least), Chrome and Edge need these HTTP headers else seeking doesn't work
+                    // In case of a video (at least), Chrome and Edge need these HTTP headers or else seeking doesn't work
                     // (even if we always send all the video content, not the requested range, until the backend supports it)
                     headers.set('Accept-Ranges', 'bytes');
                     headers.set('Content-Range', 'bytes 0-' + (contentLength - 1) + '/' + contentLength);
--- a/service-worker.js
+++ b/service-worker.js
@@ -23,38 +23,58 @@
  */
 'use strict';
 
-self.addEventListener('install', function(event) {
+var CACHE = 'kiwixjs-cache';
+
+self.addEventListener('install', function (event) {
     event.waitUntil(self.skipWaiting());
 });
 
-self.addEventListener('activate', function(event) {
+self.addEventListener('activate', function (event) {
     // "Claiming" the ServiceWorker is necessary to make it work right away,
     // without the need to reload the page.
     // See https://developer.mozilla.org/en-US/docs/Web/API/Clients/claim
     event.waitUntil(self.clients.claim());
 });
 
-var regexpRemoveUrlParameters = new RegExp(/([^?#]+)[?#].*$/);
-
-// This function is duplicated from uiUtil.js
-// because using requirejs would force to add the 'fetch' event listener
-// after the initial evaluation of this script, which is not supported any more
-// in recent versions of the browsers.
-// Cf https://bugzilla.mozilla.org/show_bug.cgi?id=1181127
-// TODO : find a way to avoid this duplication
-
-/**
- * Removes parameters and anchors from a URL
- * @param {type} url
- * @returns {String} same URL without its parameters and anchors
- */
-function removeUrlParameters(url) {
-    return url.replace(regexpRemoveUrlParameters, "$1");
-}
-
 var outgoingMessagePort = null;
 var fetchCaptureEnabled = false;
-self.addEventListener('fetch', fetchEventListener);
+
+self.addEventListener('fetch', function (event) {
+    if (fetchCaptureEnabled &&
+        regexpZIMUrlWithNamespace.test(event.request.url) &&
+        event.request.method === "GET") {
+
+        // The ServiceWorker will handle this request either from CACHE or from app.js
+
+        event.respondWith(
+            // First see if the content is in the cache
+            fromCache(event.request).then(
+                function (response) {
+                    // The response was found in the cache so we respond with it 
+                    console.log('[SW] Supplying ' + event.request.url + ' from CACHE...');
+                    return response;
+                },
+                function () {
+                    // The response was not found in the cache so we look for it in the ZIM
+                    // and add it to the cache if it is an asset type (css or js)
+                    return fetchRequestFromZIM(event).then(function(response) {
+                        // Add css or js assets to CACHE (or update their cache entries)
+                        if (/(text|application)\/(css|javascript)/i.test(response.headers.get('Content-Type'))) {
+                            console.log('[SW] Adding ' + event.request.url + ' to CACHE');
+                            event.waitUntil(updateCache(event.request, response.clone()));
+                        }
+                        return response;
+                    }).catch(function(msgPortData, title) {
+                        console.error('Invalid message received from app.js for ' + title, msgPortData);
+                        return msgPortData;
+                    });
+                }
+            )
+        );
+    }
+    // If event.respondWith() isn't called because this wasn't a request that we want to handle,
+    // then the default request/response behavior will automatically be used.
+});
 
 self.addEventListener('message', function (event) {
     if (event.data.action === 'init') {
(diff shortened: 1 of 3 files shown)


mossroy:
I suppose this addresses #414, not #441 (wrong link in the description).
As you know, I believe the use of emscripten version of kiwix-lib/libzim is the right way to improve performance (and libzim already has its own cache).
On the other hand, implementing #414 might be easy, maybe easier than #411 or #415. And it could still be useful with kiwix-lib/libzim.
So why not, if it's easy to unplug later (it seems to be the case), if we feel it becomes useless in the future.

It's hard to see the changes in github, that tends to say that everything has been modified.
But I saw the most important parts, which look good.

I quickly tested on Firefox, and it seems to work well too (I see the cache content in devtools, and the log properly says it uses it). This is promising.
I tested in a Chromium extension, but it gives some errors :

> Uncaught (in promise) TypeError: Request scheme 'chrome-extension' is unsupported
>    at service-worker.js:192

It fails because we can not `cache.put` a request with this scheme.
I don't know if it would be the same in Firefox (with a schema 'moz-extension') because, for now, the SW mode is not available at all with Firefox extensions.
I don't know what is the best way to handle that : blacklist this scheme in the SW cache code? But maybe Google might allow it in the future, who knows.
If we put a whitelist instead, for specific schemes (like `file`, `http` and `https`), we might miss the opportunity to activate the cache if Mozilla allows it (or in some other context).
Maybe we could trap the error, keep inside the SW a list of schemes that failed, add an info log, and then do not attempt to cache any of these schemes. But maybe it becomes a bit complicated.

The Cache API is not available on all platforms, see https://developer.mozilla.org/en-US/docs/Web/API/Cache#Browser_compatibility. I think we don't have any supported platform that supports SW, but not Cache API. But, to be cleaner, it would be better to test for this API availability before using it. If it's not available, we can simply fallback to standard behavior (without cache), and put an information message in the console.

Regarding the cache eviction strategy, things are quite open.
I agree that, in any case, we should add a button in settings to manually clear the SW cache. If it's easy to add the statistics, it would be great.
I also agree that we can keep cache from different ZIM files (not necessary to clear the cache on ZIM change)
If we use the cache only for javascript and CSS, maybe it's not necessary to add an eviction strategy, and simply rely on the standard strategy of the browser (it can clear the cache if it feels it's necessary, and can ask the user if he allows to use more cache in case it becomes big). Because these files are not numerous and won't use a lot of space, it might be enough. Except for PhET, that we should test more.

But we also might activate the cache for some small images (that could be useful for some icons), for which a specific eviction strategy could be used (with a smaller available space). But well, that could be for later, and maybe too complicated for the need.

Jaifroid:
I'm pleased you're interested in this, @mossroy ! It really does make a relatively easy performance jump in SW mode, while we wait for #116. Up till now, jQuery mode has been much faster than SW mode, due to the fact that we have memory caching for jQuery, but none for SW (and also because SW decompresses scripts). This equalizes things, and makes using SW mode much more feasible for daily use.

It's a real shame that the Chrome Extension schema doesn't work with Cache, as Chrome Extension is a major use-case for this app. I wonder if there's a solution for that, or whether it's just outright not supported. It's worth investigating.

Regarding an interface to control the Cache, I did develop one when I worked on a comprehensive cache based on indexDB, and it looks like the image below. This could be adapted to our needs.

I didn't know libzim comes with a cache. But it'll take some time for that code to be developed, so I think it's worth using this optimization.

I agree PhET is a problem. JS is unique to each experiment and shouldn't be cached. I'm sure we can find a way to distinguish.

Thanks for the detailed and thoughtful comments, which all seem very sensible.

image

Jaifroid:
I've checked support in chrome-extension, and unfortunately it's not supported full stop. Chrome-extension has its own caching strategy that devs can use via `chrome.storage`, but this is only available in the background script, which seems like a huge complication. People seem to recommend using localStorage or indexedDB as an alternative.

Because localStorage is synchronous, it cannot be accessed from a Service Worker. IndexedDB is, however, asynchronous and apparently can be used inside Service Worker.

**So the question is: is it worth adding a fallback of indexedDB caching in SW specifically for Chrome extension usage?** I have a ready-made indexedDB function.

It should be noted that the Cache API does *not* need to be accessed from the SW. It is also available in the window object, and can be accessed from the main thread. However, it seemed to me to be neater to keep it all in the SW (also running on a separate process, so maybe faster).

Jaifroid:
For now, I've added a (volatile) memory cache (`assetsCache`) to the code for excluded URL schemata. This works well to cache assets during a session, just like cssCache in the main thread, which should be a gain for Chrome extensions.

To test, see the comment in the new code (commit https://github.com/kiwix/kiwix-js/pull/556/commits/5f37306e1690609b02cc169a22dcc09899a0588f) about changing `example-extension` to `http` and running the code on localhost. Remember to clear any existing Service Worker first, as it won't auto-update. You should then get messages in console log showing "[SW] Adding EXCLUDED schema URL...". Shown below is the snapshot of `assetsCache` (which is just a Map). Note we can't actually store a Response in a Map because Responses are special objects that get annulled easily (I wasn't successful storing and retrieving them). Instead I store the constituents of the Response we construct.

image

Jaifroid:
Ah, the dreaded Firefox 45 Travis error...

mossroy:
No, it's not necessary to add anything when inside the chrome-extension scheme, because Chromium already has a built-in in-memory cache (see https://github.com/kiwix/kiwix-js/issues/411#issuecomment-511199358) when inside an extension.
It should even not go through the SW more than once, and bypass it to read its in-memory cache.
No need to make this more complicated.

I think we simply need to properly avoid the console error messages in this case

Jaifroid:
Ah OK, I'd forgotten about that comment (haven't used Chrome extension much). So should I remove the memory chache I've added in the last commit? (It only took me 30mins or so to add it, so no great loss if you don't see a use for it.) The only use I can think of might be to speed up cache access generally.

Jaifroid:
What I mean by that is that, when I was experimenting with IndexedDB last year, I found it was noticeably slower to access assets stored there than from cssCache. So I opted to keep cssCache for fast retrieval, and use IndexedDB to store assets between sessions. But I'm not sure it's the case with Cache API. It seems "fast enough".

mossroy:
IMHO you should remove this last commit and only keep the Cache API (without introducing IndexedDB either)

Jaifroid:
> IMHO you should remove this last commit and only keep the Cache API (without introducing IndexedDB either)

Done. I've kept the structural improvements to the code: providing the relevant regexes at the top of the code with a comment for developers on what they do. It's better than having regexes buried further down the code which would be difficult to maintain.

NB this code excludes `^chrome-extension:` from caching, so we don't get the console log message (as you suggested above) and unnecessary CPU cycles. It can easily be changed in the future if extensions start to support Cache. See comments at top of file.

Jaifroid:
Just to say that it would be best for me to delay working on adding a cache control panel (like that above) until we've merged #527, because otherwise I'd have to do double work (Bootstrap 3 panels are rather different from Bootstrap 4 cards). But in due course it would be good to know whether that design seems right, and whether you'd be interested in having the "Return to last visited page when you open an archive" functionality. That functionality is useful, for example, in a NWJS app. It's less useful if the user still has to pick the archive, which is the case in most contexts. The Cache used field would show "Memory cache" for jQuery mode and "Cache API" for SW mode.

Jaifroid:
I had some time yesterday evening, so added the cache evacuation infrastructure and UI anyway. The checkbox for remembering the last page between sessions is currently *non-functional* and can be removed if that functionality is not needed / wanted, in which case the panel title would change from "Performance and privacy settings" to "Performance settings". Asset counting works for both the memory cache and Cache API, but only the assets in the respective cache being used are shown. However, on deleting assets, the sum of all assets deleted is displayed.

Jaifroid:
As a side issue, the "random" Firefox 45 Travis failures must I guess be caused by a race condition somewhere in our code. Is there a way to debug Travis builds? I suppose the Chrome failure for #527 is also a race condition (but a different one).

mossroy:
It would be better to discuss remembering the last page in a different PR

Jaifroid:
> It would be better to discuss remembering the last page in a different PR

OK, I've added issue #561 for this.

Jaifroid:
Remember last page checkbox removed. This is what the interface looks like straight after the user turns caching off:

image

Jaifroid:
I've rebased this PR on master and have adapted the cache control panel to use Bootstrap 4.

Jaifroid:
I think this PR is ready for review, unless there are any further features that need to be added. The only thing that occurs to me is that, in accordance with #507, I could put the new functions in `app.js` into a new module, called `cache.js`. Would that be worthwhile?

mossroy:
I quickly tested on Firefox and Chromium and like the settings UI.
If it's technically possible, it would be better to add how much space the cache takes (and the quota of the browser). For the Cache API, https://developer.mozilla.org/en-US/docs/Web/API/StorageEstimate should do it. For the Memory cache, it's less necessary as it is emptied at the end of each session.
I did not have a deep look at the code, but I'm puzzled by the fact that you are able to display the "Caching _filename_ ..." overlay in SW mode. It's consistent with the jQuery mode, which is cool, but it probably means you're modifying the DOM when getting some content for the SW, which is not cool.
The code that is ran at this moment can not know if the asset will be put in cache or not, which has a side-effect when running as a chromium extension : on every page it says it is caching the CSS and javascript files (which is wrong).
In this extension context, the settings UI says it uses a "Memory" cache, but it always has zero assets, which can be confusing. The truth is that kiwix-js does not use any of its own cache (memory or CacheAPI) BUT Chromium caches all the files by itself (in another memory region, and it can probably not be detected in javascript). Maybe it's not crucial if we don't give a very precise information here, but we must try to not give a wrong one.

IMHO, we should not display "Caching ..." in SW mode because it breaks the software layers, and can not be reliable. Never mind if it's not consistent with jQuery mode in this case.
And the Settings UI should be fixed for the case of Chromium extension.
The difficulty is that we would have to duplicate the logic that is in the SW (blacklisting chromium extensions). I'm wondering if a better approach would be to test the SW instead of duplicating some of its code : asking for a dummy content, and see if the SW serves it from cache or not (this information could be added in the response headers by the SW). But maybe it's too complicated for the current need.

mossroy:
Thinking again about that, I changed my mind.
I think almost nobody will care about this setting, and people who care probably can use the devtools to have more info.
I would prefer that we keep this implementation as simple as possible.

So I suggest that : 
- we don't try to get the space usage of caches (only keep the count you already have)
- we change a bit the behavior of the UI so that we don't need to tell the user which cache is in use : we could simply always display both caches count in the UI, and empty both of them when the user clicks to disable them
- we don't try to guess which one will be in use

Jaifroid:
> I'm puzzled by the fact that you are able to display the "Caching filename ..." overlay in SW mode. It's consistent with the jQuery mode, which is cool, but it probably means you're modifying the DOM when getting some content for the SW, which is not cool

He he, nothing gets past you! ;-) It's very easy: it's just [this one line](https://github.com/kiwix/kiwix-js/pull/556/files#diff-bca463ed8ccf5e2058dfe307a95e8501R1065) linking to the same function that updates the cache status messages in jQuery mode. I can remove the line if you really think it's problematic -- personally I find it helpful to have feedback about why a load is taking so long. It helps to see that the app hasn't got stuck (which it still does occasionally, especially when switching between jQuery and SW modes).

Jaifroid:
PS The code doesn't need to know if the cache is being used or not: if it is being used, the file won't be extracted by readBinaryFile and the message won't be shown. If the file is being extracted by `readBinaryFile`, then no cache (not even Chrome extension's built-in cache) has been used for the file. It would be easy to exempt showing the message if we are in a Chrome extension. But I'd like to investigate why its in-built caching is clearly not working in that case, as we don't want unnecessary extractions from the ZIM.

Jaifroid:
PPS Maybe some confusion is caused by the fact that the message says "Caching ...". More accurately, it should say "Extracting ..." or "Getting ..." (regardless of whether we keep it for SW mode or not). The message is not registering a cache, it's registering extraction of the file from the ZIM archive, but only uncached files are extracted.

Jaifroid:
So, with the latest updates (which I've tested in extension form on Edge Chromium) we have the following simplified behaviour:

* In both jQuery and SW modes, the Cache count shown is always the sum of cssCache.size and CacheAPI count;
* If the total is zero and we're in a Chrome extension, the asset count is shown as '-', so as not to mislead the user, but if the caches we control do have something in them, that amount is shown;
* No caching overlay is shown in SW mode in a Chrome extension;
* Caching overlay is currently shown in other contexts in both jQuery and SW modes, but removing it from SW mode is just a case of removing one line.

The logic for working out which cache we're using is already written, and it's not complicated or a maintenance headache IMHO. If chrome-extension starts using the standard Cache, it's a simple edit to standardize it, and I've added comments in the appropriate places in SW mode and jQuery mode (with cross-referencing) for future Devs (most likely ourselves) to remember what to do.

So hopefully that addresses your points. It remains to be decided if I should change "Caching ..." to "Extracting ..." or "Getting ..." (latter a bit shorter), and for you to confirm that you really, really don't want the pretty overlay in SW mode 😉.

mossroy:
To me, we have unneeded logic and assumptions, where we could simply have a UI like this (instead of the "Cache status" box) :

```
Cache usage (# of assets) : 
Memory cache : 18      Cache API : 23
```
Without telling the user which cache is currently in use. It avoids making some assumptions (that might become wrong in the future) and some duplicated code/logic with the SW.

On the left of this box, when the user clicks on "Do not cache assets", it empties both caches, and refreshes the cache usage box, so that both show zero assets (no need to tell how much have been removed)

When the user switches between jQuery and SW mode, nothing special happens regarding this cache status

It seems much simpler to me, and much easier to maintain. The drawback is that the user does not know which cache region is currently in use. But we don't know for sure either (we did not test every context where it can be run). And there is always a workaround to know for sure : open a page, come back there and see which figures have increased.

Regarding the wording, "Getting ... " is fine for me, but you're the expert on that topic ;-)
And yes, I still think we must not modify the DOM while getting some content for the SW.

service-worker.js
```diff
@@ -23,117 +23,191 @@
  */
 'use strict';
 
-self.addEventListener('install', function(event) {
+var CACHE = 'kiwixjs-cache';
```
mossroy: This should be renamed to indicate that it's a cache of assets, used by the SW

service-worker.js
```diff
@@ -23,117 +23,191 @@
...
+                    return response;
+                },
+                function () {
+                    // The response was not found in the cache so we look for it in the ZIM
+                    // and add it to the cache if it is an asset type (css or js)
+                    return fetchRequestFromZIM(event).then(function (response) {
+                        // Add css or js assets to CACHE (or update their cache entries) unless the URL schema is not supported
+                        if (cachedContentTypesRegexp.test(response.headers.get('Content-Type')) &&
+                            !excludedURLSchema.test(event.request.url)) {
+                            event.waitUntil(updateCache(event.request, response.clone()));
```
mossroy: Why is it necessary to wait for the cache to be updated before returning the response? (2 lines below).
It should not be a problem that the cache can be updated a bit later, and should speed up the request/response timing?

service-worker.js
Jaifroid: `kiwixjs-assetCache`? NB The Cache API could perfectly well be used in jQuery mode. The fact that we only implement it in SW mode is an arbitrary decision (albeit a sensible one for us at this moment in time). Cache API is available in `window`, so accessible from app.js.

service-worker.js
Jaifroid: @mossroy It doesn't wait. `event.waitUntil` may look like async `await`, but in fact it is merely an instruction to the browser not to let the event expire until the cache has been updated (events are expired quickly by browsers to optimize memory usage). The next line (`return response;`) will run immediately, before the `updateCache` promise completes, and we can't `return` before the `event.waitUntil` statement or else that code will become unreachable.

Jaifroid:
> To me, we have unneeded logic and assumptions, where we could simply have a UI like this (instead of the "Cache status" box) :
> Cache usage (# of assets) : 
> Memory cache : 18      Cache API : 23

This would be possible. But the reason that I originally implemented it in a more dynamic way was because I wanted a concise display that could adapt easily to whatever cache had been implemented or was supported by the current browser: Memory / LocalStorage / IndexedDB / Cache API. My original implementation had Memory, LocalStorage and IndexedDB. I'd quite like to keep compatibility of the codebase with future versions of Kiwix JS Windows (2.0) where I'd like to adopt this interface (I haven't ported it there yet) but where Service Worker is not available and Cache API is available only on some devices, so I will most likely use IndexedDB, (and Cache API in environments that support it).

If you feel strongly the interface needs to be simplified, it wouldn't be hard to change it to your preferred model above, but it would complicate the relationship between the two code bases as I'm envisioning it.

> And yes, I still think we must not modify the DOM while getting some content for the SW.

Personally I would distinguish *conceptually* between "modifying the DOM" and "modifying the UI", even though they are *technically* the same thing in HTML5 development. The former sounds scary, but all we're doing is providing UI feedback. However, I respect your strong views on this and shall remove the offending line ;-)

service-worker.js
mossroy: All right. This name looks fine

service-worker.js
mossroy: All right, thanks for the explanation

service-worker.js
Jaifroid: I've changed the name. Be sure to delete the old Cache, either before downloading the commit, or in dev tools.

Jaifroid:
> no need to tell how much have been removed

I've deleted the functions / div that display the number of deleted assets. Instead, the asset count (which is always now a total of assets in cssCache and CACHE) is simply zeroed.

> Regarding the wording, "Getting ... " is fine for me

Since we're no longer showing this for SW mode, maybe we don't need to change the wording. In jQuery mode, I think we do actually only show this when caching, or at least very close to the `Map.set()` method, so I think it's not misleading. If we do change it, however, I think I would prefer "Fetching ..." because "Getting ..." sounds a bit like we dumbed things down (!).

mossroy:
Choose the wording you prefer, no problem.

About the cache status, my goal is mainly to avoid making the code more complicated, duplicated and take the risk to give a wrong information to the user.
Besides that, no "normal user" cares about how the cache behaves and is implemented. So it's no use duplicating code to give this kind of information.

Having an option to disable the cache can still be a good idea for debugging purpose (or in some edge cases), but it's still an "expert setting", and could be moved in the "expert settings" zone.

I understand your concern about code diverging between kiwix-js and kiwix-js-windows.
What about, in the case of kiwix-js-windows, using something like this on the right : 
```
Cache usage (# of assets) : 
Memory cache : 18
Cache API : 23
IndexedDB : 12
LocalStorage : 26
```
When some API is unavailable, or unusable, you could simply put a "N/A" instead of the asset count
But maybe I'm missing something, as I don't really know what it is like (I did not see it on https://kiwix.github.io/kiwix-js-windows)

Regarding PhET, it's a good use case to test what happens when many assets are put in the cache (at some point, the browser might ask if the user wants to grow the cache or not). But we must not do anything that is PhET specific.
I don't currently see a generic and reliable way to distinguish the javascript files that should be kept for long (like the ones of wikimedia) and the ones that should not (like the ones of PhET).
Except by implementing our own LRU eviction strategy, or something similar.
But I think it's not necessary to do that for now. Let's see how it behaves.

Based on https://developers.google.com/web/fundamentals/instant-and-offline/web-storage/offline-for-pwa , the browsers have their own safe-guards when too much cache is used : it might be enough to rely on them.

Jaifroid:
OK, I'll have a think about what can be done to simplify. As it is, we currently have three different caching strategies: "Memory", "Cache API", and "Custom" (Chrome-Extension, which is a black box). Unless we go for some automatic cache evacuation strategy, I'd like users to be aware that items are being cached potentially for a long time in the case of Cache API and that they can end up taking significant space.

Jaifroid:
PS This system is not implemented in Kiwix JS Windows yet. My plan is to replace the file-based caching currently used there (hard to maintain) with this automatic one (much easier to maintain), when it's ready. But I also have to provide indexedDB for jQuery mode because of the lack of general support for Service Worker in UWP JS apps (SW support only exists for content served from https: , i.e. if I implement it as a PWA inside UWP).

Jaifroid:
So this is my best attempt to prevent duplication of variables, regexes, and functions across app.js and SW, and to simplify the logic. The latest code works like this:

* When app.js needs to display cache details, it queries the Service Worker, if it is available, to determine the attributes of the cache in use (type, description, assetCount);
* If we are in jQuery mode, then it merely returns the attributes of cssCache (because it cannot presume that it can access SW, even if sometimes it technically could);
* When user turns off caching, all caches are deleted (except the internal Chrome extension cache of course);
* When user switches from jQuery mode to SW mode, cssCache is emptied, as it is no longer needed;
* When user switches from SW mode to jQuery mode, CACHE is emptied, as it will no longer be used, and we should not have items taking significant space that are useless (and that the user may not be aware of) [I realize this is an assumption, but I think it's correct from a user perspective, as users will not switch modes frequently, and the app should not hog memory or storage space];
* **Therefore**, the cache count only shows the count of the cache in use, or '-' if it cannot be detemined (i.e. Chrome extension). It would be useless to show any other counts, because they will always be **0**;
* CACHE is defined in app.js, and is passed to Service Worker early in its initialization. It has to be this way round, because otherwise we have no way to delete CACHE from app.js when Service Worker has been deactivated (we *need* the CACHE constant). **But we only define CACHE once.**

I think this simplifies the decision logic significantly. However, the amount of code in app.js has not decreased by much, because messaging the Service Worker and receiving a response is not completely trivial. **The big benefit is that we no longer have to tell DEVs to ensure that duplicated CACHE definitions and regexes have to be kept in sync, because there is no duplication.** So the main benefit is maintainability.

Please let me know what you think @mossroy.

service-worker.js
```diff
@@ -23,117 +23,240 @@
...
+
+/**
+ * Looks up a Request in CACHE and returns a Promise for the matched Response
+ * @param {Request} request The Request to fulfill from CACHE
+ * @returns {Response} The cached Response (as a Promise) 
+ */
+function fromCache(request) {
+    // Prevents use of Cache API if user has disabled it
+    if (!useCache) return Promise.reject('disabled');
+    return caches.open(CACHE).then(function (cache) {
```
mossroy: You should test that the Cache API is available before using it, so that it does not break if it's not.
It's probably a rare case, anyway

www/js/app.js
```diff
@@ -311,7 +346,62 @@ define(['jquery', 'zimArchiveLoader', 'util', 'uiUtil', 'cookies','abstractFiles
         apiStatusPanel.classList.add(apiPanelClass);
 
     }
-    
+
+    /**
+     * Queries Service Worker if possible to determine CACHE capability and returns an object with cache attributes
+     * If Service Worker is not available, the attributes of the memory cache are returned instead
+     * @returns {Promise} A Promise for an object with cache attributes 'type', 'description', and 'count'
```
mossroy: The `type` does not seem be currently used. I suppose it's for some future usage?
If you have something in mind, it's fine : I just want to check it's not a mistake

service-worker.js
```diff
@@ -23,117 +23,240 @@
  */
 'use strict';
 
-self.addEventListener('install', function(event) {
+/**
+ * The name of the Cache API cache in which assets defined in regexpCachedContentTypes will be stored
+ * The value is defined in app.js and will be passed to Service Worker on initialization (to avoid duplication)
+ * @type {String}
+ */
+var CACHE;
```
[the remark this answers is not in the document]
Jaifroid: OK, will do.

service-worker.js
```diff
@@ -23,117 +23,240 @@
...
+            // On 'init' message, we initialize the outgoingMessagePort and enable the fetchEventListener
+            outgoingMessagePort = event.ports[0];
+            fetchCaptureEnabled = true;
+        } else if (event.data.action === 'disable') {
+            // On 'disable' message, we delete the outgoingMessagePort and disable the fetchEventListener
+            outgoingMessagePort = null;
+            fetchCaptureEnabled = false;
+        }
+    }
+    if (event.data.useCache) {
```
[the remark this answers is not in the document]
Jaifroid: I did consider that, but the problem was (if I remember correctly) that event.data.action is a string (or at least it is tested as a string) and I wanted a psuedo-Boolean. However, I can think of a solution that uses event.data.action without having to have two code blocks, so will try that, and if it's easy, will do it in this PR.

service-worker.js
Jaifroid: OK, will do.

www/js/app.js
Jaifroid: Ah, my editor (vscode) shows me details for the Promise interface as if it were a valid type. But I think it's heavily orientated towards TypeScript, where strong typing is mandatory. I'll just change this to `{Object}` and make clear that it's a promise for an object in the description. We do have some custom non-JS types in use -- e.g. we have `@type Integer` elsewhere, whereas strictly there is no such primitive JS type, and we should use `{Number}`.

service-worker.js
Jaifroid: Actually, thinking about this, a test might be redundant as it seems Cache API is a *required part* of the Service Worker spec. According to the [CacheStorage part of the Service Worker spec](https://w3c.github.io/ServiceWorker/#cachestorage-interface), "The user agent _must_ create a CacheStorage object when a Window object or a WorkerGlobalScope object is created and associate it with that global object."  So if we are in Service Worker the Cache API is available, even if we can't actually save a request in it due to having the wrong URL schema. Do you agree @mossroy?

service-worker.js
```diff
@@ -23,117 +23,240 @@
...
+            return matching;
+        });
+    });
 }
+
+/**
+ * Stores or updates in CACHE the given Request/Response pair
+ * @param {Request} request The original Request object
+ * @param {Response} response The Response received from the server/ZIM
+ * @returns {Promise} A Promise for the update action
```
Jaifroid: Here we have another {Promise} return type, but it's trickier to know what to substitute, because it's a promise for an action. Looking at the https://developer.mozilla.org/en-US/docs/Web/API/Cache/put , they say that `cache.put()` returns a Promise. Is it OK to leave this here, @mossroy ? Where an obvious Object or Array is returned *via* a Promise, I'm changing the return types, but here it doesn't seem possible.

service-worker.js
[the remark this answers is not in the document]
mossroy: It's really not crucial, and my initial implementation might not be the best

service-worker.js
mossroy: You seem to be right. In this case, it's only necessary to test when in app.js

www/js/app.js
mossroy: I was not mentioning this annotation (which seems right), but the attribute`type` of the returned object (next to `description` and `count`)

service-worker.js
mossroy: I agree

service-worker.js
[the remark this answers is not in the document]
Jaifroid: > It's really not crucial, and my initial implementation might not be the best

It turns out to be very easy to do (I'm about to push a commit with it) and it does make sense to have all "actions" under this key.

www/js/app.js
Jaifroid: Ah, OK. I'm not using "type" as a technical term. It's the name of a key, so if I assign this returned object to a variable like `cache` I can use `cache.type`, `cache.description` and `cache.count`. The string `cache.type` could be reported as the strings `cacheAPI`, or `memory`, or `custom` or even `none`, depending on context and user options. Are you OK with `type` as the name of this object key? Type and description are subtly different. Type would be `cacheAPI` and description is `Cache API` with a space and initial capital (a future DEV might want a fuller description, for example).

Jaifroid: Yes, you're right I don't now use it. I did before, and that code has been cut in the simplification. Should I keep?

Jaifroid:
Commit https://github.com/kiwix/kiwix-js/pull/556/commits/dd18a67e5e75a33d8acc0d947f408a9bd1e6d34e implements what was discussed in code review. Given that some changes arose from a small misunderstanding over return types, what's your view on whether a Promise should always be described as a `{Promise}` in a JSDoc return type annotation, or whether it is better, if we are returning an Object via a Promise to describe the return type as `{Object}` and then make clear in the description that it's via a Promise. I read people saying we should only use "Primitive JavaScript types", but that may be outdated, and I can't find JSDoc official advice.

www/js/app.js
mossroy: I'm OK with the `type` name, and understand the purpose. But I did not see where this attribute was used.
It's not a big issue to keep it if you think it might be useful in the future

Jaifroid:
Last question: should I remove console.log lines (not console.error) from the code before merging, or is it useful to have them for a while and remove them later?

www/js/app.js
Jaifroid: OK, I'll keep it. It should be useful for extensibility and I'm averse to having data structures that use descriptions as primary keys (from working with relational databases)!

Jaifroid:
It seems the syntax `{Promise<T>}` has become the accepted usage. (And VSCode JSLint recognizes this as valid.) See https://github.com/kiwix/kiwix-js/pull/556/commits/301bee8e72e763c6dbe20e8520ec56aceb9558b1 .

mossroy:
Maybe we could keep the `console.debug` lines for now, and remove them later.
Regarding the way to write annotations for Promises, I don't know. Our existing code usually uses `@return {Promise}` followed by a textual description of what it will return.
If there is a more common and standardized way of doing it, please do (and maybe open another issue to refactor the existing code)

service-worker.js
```diff
@@ -208,26 +208,27 @@ function removeUrlParameters(url) {
 }
 
 /**
- * Looks up a Request in CACHE and returns a Promise for the matched Response
- * @param {Request} request The Request to fulfill from CACHE
- * @returns {Promise<Response>} A Promise for the cached Response 
+ * Looks up a Request in CACHE_NAME and returns a Promise for the matched Response
+ * @param {Request} request The Request to fulfill from CACHE_NAME
+ * @returns {Promise<Response>} A Promise for the cached Response
+ * @throws {Promise<String>} A rejected Promise with strings 'disabled' or 'no-match'
```
mossroy: I don't think we should use the `@throws` annotation for the case of a rejected Promise.
I suppose it is meant only for exceptions that are synchronously sent by the function (like in Java) : https://jsdoc.app/tags-throws.html

In this case, we might add this information (about what is rejected by the promise) as a comment at the end of the `@returns` line

Jaifroid: OK! Will remove before merging. It's somewhere in that long thread mentioned in #568, but you're right that it might not be the best way to do it, and it's not properly documented anywhere.

Jaifroid:
After a bit more testing (including in Chrome extension, but not Firefox extension because I don't know how), can I squash and merge?

mossroy:
Yes you can

Issue #558, opened 2019-08-25
Simplify article loading in SW mode


Jaifroid:
In SW mode, when searching for an article or using the random button (but not when clicking on a hyperlink) we currently load `article.html`, wait for it to be loaded, then set the iframe content to the URL of the page we want to load, wait for that to be loaded, then turn off the spinner, etc.

This is convoluted and unnecessary. It can be simplified. From the outset, we simply set the iframe.src to the URL of the article, and let the Service Worker do its thing. I've tested this, and it seems to work fine with no issues. It's a quick and easy PR.

PR #559, opened 2019-08-25
Simplify article loading in SW mode

Jaifroid:
This PR addresses #558.


Jaifroid:
Manually tested on Edge, Firefox and Chromium in SW mode on a variety of ZIMs.

Issue #536, opened 2019-07-14
The Travis build is unstable with Firefox 45


mossroy:
UI tests sometimes fail with :

> Timed out while waiting for element <#articleList> to be visible for 20000 milliseconds.

mossroy:
What I can see in Saucelabs (for example on https://app.saucelabs.com/tests/4d28dbb03b6b4d80acbd62b11bb70b23) is that, after typing "Ray" in the search field, the browser immediately displays the results, before we click on the search button.
For some reason, it makes the result list disappear, so it fails to find the article.
Maybe we should modify the script (or kiwix-js) to avoid this race-condition.

Jaifroid:
It is by design that results show without clicking the search button. However, clicking search should not then hide the results. It could be since we introduced the code to hide results if the search field loses focus. We could probably fix this by adding code to check that the focus is not the search button (if my hunch here is correct).

Jaifroid:
I confirm that clicking the search button after typing "ray" causes the results to disappear briefly before being shown again. Moreover, if we do search by clicking the search button, it is no longer possible to dismiss the results by clicking elsewhere in the app, because the onlostfocus event of the search button does not have a listener attached to it. It seems we need to add code to deal with this situation, as it does come across as a little buggy (even if it's a minor detail most users will never encounter).

Jaifroid:
Since commit https://github.com/kiwix/kiwix-js/commit/b954b323b2f8ea77abf5839dd604cc3f47056b66 this issue is no longer occurring. So I'm closing for now (but we can re-open if the issue re-occurs).

PR #563, opened 2019-08-29
Prevent hiding of articleList on clicking searchArticles button

### LINKED ISSUE #562: Tapping search button causes any displayed results to disappear briefly
Tapping or clicking the search button after typing a search causes the already displayed results to disappear briefly before being shown again. It seems this is the cause of #536.

Moreover, if we search by clicking the search button, it is no longer possible to dismiss the results by clicking elsewhere in the app, because the onlostfocus event of the search button does not have a listener attached to it.

It seems we need to add code to deal with this situation, as it does come across as a little buggy (even if it's a minor detail most users will never encounter).

_Originally posted by @Jaifroid in https://github.com/kiwix/kiwix-js/issues/536#issuecomment-526078200_

Jaifroid:
Let's see if this fixes #562 and #536.

### DIFF
--- a/www/js/app.js
+++ b/www/js/app.js
@@ -168,7 +168,7 @@ define(['jquery', 'zimArchiveLoader', 'util', 'uiUtil', 'cookies','abstractFiles
         if ($('#prefix').val() !== '') 
             $('#articleListWithHeader').show();
     });
-    // Hide the search resutls if user moves out of prefix field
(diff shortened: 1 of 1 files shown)


Jaifroid:
So far so good. 🙂

Jaifroid:
Second build also passed.

PR #564, opened 2019-08-29
Add badges to the README.md

kelson42:


Jaifroid:
Could they be placed either at the very top of the README or at the very bottom? They look a bit odd in the middle of the page, but if you've standardized that position across all kiwix repos, that's OK by me.

kelson42:
@Jaifroid I have indeed standardised a bit this presentation. To me they are important, but not more important than the explanation about what does the tool. This is why they are put just after the introduction. But, this is pretty much a subjective thing, I agree.

Jaifroid:
@kelson42 OK.

Jaifroid:
Whoops, sorry, meant to comment on the other PR!

Issue #523, opened 2019-06-24
Rework User Interface like Kiwix JS Windows


Jaifroid:
This issue is a combination of #267, #304 and #337 (all of which are issues concerning the UI). The aim is to backport a version of the Bootstrap UI used in Kiwix JS Windows (originally used by @sharun-s in his fork of Kiwix JS). I think this work needs to be done before any more progress can be made on #127 and #173, for example. I'm opening a dedicated issue for this, so that we can set out a road-plan of what is desired and what is not desired before starting to backport, with the aim of avoiding undertaking unnecessary work.

Jaifroid:
Let's break down the features of the reworked interface in Kiwix JS Windows. Refer to screenshot below:

1. Colour: in "light mode", the colour I adopted is light blue (both for top and bottom bar); in "dark mode", this changes to a very dark grey. The bottom bar is slightly translucent, so that upcoming content can be glimpsed through it as in current Kiwix JS. I don't propose to backport dark mode just yet, so the main decision here is the colour of the bars in light mode.

2. Mutating Home icon (top bar): The idea of the top-left icon is to provide some visual "branding" for the app, but it doubles as a Home icon. Because Kiwix JS has separate apps for Kiwix, WikiMed and Wikivoyage, I have code that mutates the Home icon, showing specific branded icons when WikiMed and Wikivoyage ZIMs are loaded. This means I don't have to maintain different code bases. For all other ZIM types, currently, the neutral Kiwix icon is used. Note that we are only using icons in this interface, there is no room for text as in the current Kiwix JS. How should I proceed? Shall we have a single neutral icon top left, or shall we mutate it? Note that the bottom bar also has a Home icon, but in this case, it is the standard "Home" symbol. An alternative design would be for the top-left icon to be a hamburger (sandwich) menu. Top-left is the traditional place in apps for hamburger menus.

3. Mutating Print/About icon: Due to lack of space, the last icon top right is a Print icon when an article is loaded in the iframe and is displaying, and is an About icon when there is no article loaded or when the Configuration page is showing. We have not yet implemented #350. I suspect we want a more elegant solution for Kiwix JS with regard to extensibility of features, as per #521. There @mossroy mentioned using a hamburger menu for extra features/content, such as About, Printing, Load all images. Traditional location for hamburger is top-left rather than top-right. We should decide this in advance of any coding.

[...]

Issue #386, opened 2018-05-30
In old Stackexchange files (at least) some content is still hidden under the bottom bar


Jaifroid:
I noticed this on several (all?) of the tag question lists on `stackoverflow.com_eng_all_2017-05.zim` (in Firefox, but that's just because it's what I was using). It affects mostly non-mobile screen widths: sometimes actual content is hidden, or a "Next" link becomes impossible to click (you can see it beneath the transparency, but not click on it) sometimes part of the Kiwix info footer is hidden. Mobile-width displays are not so affected.

This may be better addressed through #120 or #337.

kelson42:
@Jaifroid I don't think I can reproduce the problem anymore on recent FF with recent Stackoverflow ZIM file. That said, it is true that if the bottom bar can not be removed, or the some kind of page padding introduced, then if something is clickable at the really bottom of the page, it won't be anymore.

Jaifroid:
It seems less of a problem now. In any case, we have #523 that covers this, as one of the ideas there is to make the bottom bar disappear a few seconds after scroll stop. So I'm closing this.

Issue #337, opened 2018-01-17
Rework the bottom bar


mossroy:
This bottom bar has a few problems : 
- It can hide some content at the end of an article. For example, we had to put a few carriage returns in the About section, just to allow clicking on the links at the bottom
- It needed some extra work in the automated UI tests for the same reason : sometimes you can not click on a link if it's hidden behind the bottom bar
- It should be hidden when the user is in the Configuration or About section (see kiwix/kiwix-js-windows#43)
- The user can not click on the bottom right of the screen (for example on the scroll bar) : it clicks on the "go to top" button instead

mossroy:
I'm starting to wonder if the solution would not be to completely remove this bottom bar.
There's no such thing in the Android application. They have found other solutions to provide the same features.
@kelson42 says that the dev version now has a bottom bar that disappears after a few seconds. We might try to align with the same ergonomic ideas (as long as they are compatible with a desktop version).
Is there a doc/wiki explaining the ergonomic strategy of this app?

Jaifroid:
The scrolling viewport (iframe) gets sized to fill the entire screen space before the bottom bar is added in as an absolutely positioned fixed overlay, so hiding come content beneath the bar. I've experimented with different ways of calculating the viewport size (but I was doing this early on, before I had a proper understanding of the layout) and didn't come up with a satisfactory solution other than adding padding to the HTML. Even if we reduce the viewport size, then the transparency effect of the bottom bar becomes useless, and it might as well just be a solid block (I think I encountered this when experimenting and it was visually less satisfying, in part because without the transparency there are no visual clues that there is further content to read).

Note that ergonomically, bottom bars are much easier to reach on mobile devices than top bars, as they can be controlled by the thumb without having to stretch, especially with large screen sizes which otherwise have to be operated with two hands. IMHO therefore the bottom bar should be the last thing to go and hiding it should be optional unless another completely satisfactory way of navigating one-handed can be found...

Having said that, hiding the bottom bar in the Config and About screens is a simple `document.getElementById('footer').style.display = "none";`, and can be added to the Config button and About button onclick routines. A very easy PR (implemented in https://github.com/kiwix/kiwix-js-windows/issues/43).

practik:
> I'm starting to wonder if the solution would not be to completely remove this bottom bar.

[...]

Issue #120, opened 2015-08-22
Rework the behavior and layout of the screen (iframe version), to optimize the available space for article content


mossroy:
The iframe currently takes all the available space (including height).
But the top area always stays visible, and does not let enough space for the article.

IMHO, it would be ideal to implement the same behavior as in the browser of Firefox OS >= 2.1 :
- when the user scrolls up, the top area shrinks (or even disappears). There are options in jquery hide() function to hide them smoothly, but we also need to resize the iframe at this moment
- when the user scrolls down, the top area reappears (and the iframe should be resized)
- the top area also contains the back/forward buttons
- the bottom bar and buttons become useless and might be removed

... but it might be hard to implement

mossroy:
Maybe using something like https://medium.com/@mariusc23/hide-header-on-scroll-down-show-on-scroll-up-67bbaae9a78c would do what we need for the header BUT the difficulty will be to expand/reduce the iframe without visibly moving the content inside

kelson42:
Yes, would be great to have it that way (it works like this on Kiwix for iOS for example) but I would pospone this after the first release. For example Kiwix for Android has always the top bar visible, ant this is OK, users do not complain too much about it.

mossroy:
You're right in the sense that we first need a functional application, and we can leave some enhancements for next versions.
But there is still a minimum that needs to be done in any case : 
- reduce the size of top area : the article content is really in a too small area for now
- make the back/forward/top buttons work again

mossroy:
OK I did what I considered to be the minimum for a first release with the iframe. So I push this enhancement for the following version

Jaifroid:
On small screens, at least on Windows Mobile, the top area does disappear when scrolling down the page, so the article becomes fullscreen apart from the very narrow navbar and buttons at the bottom.  I think this should be a behaviour based on *window size* and in particular on *pixel height* of the window, rather than absolute screensize. Effectively it would be good to have the app/page fully responsive, so that it doesn't matter if it is showing on a 720 x 1024 screen or in a window of 720 x 1024, but it will adopt the same behaviour based on the available display space.

Jaifroid:
This would be superseded by #523.

PR #567, opened 2019-09-03
Remove vertical scrollbar

### LINKED ISSUE #566: There is a vertical scrollbar since bootstrap 4
It did not show up in previous versions.

I'm not talking about a vertical scrollbar inside the iframe (which is normal), but one on the whole window.
It appears on both Firefox and Chromium.

Jaifroid:
Fixes #566.

### DIFF
--- a/www/css/app.css
+++ b/www/css/app.css
@@ -26,7 +26,7 @@
     bottom: 0;
     left: 0;
     position: fixed;
-    overflow-x: hidden;
+    overflow: hidden;
 }
 [role=region] > header {
      margin-bottom: 1rem;


Jaifroid:
Hmm, I think that's a different issue which is not related to #566 as far as I can tell. This PR is about the extra uselesss scrollbar. What you're reporting is a version of a longstanding issue covered (perhaps indirectly) by #120, #337 or #386. I've checked on master, and the cutoff (worse on older ZIMs, better on recent ZIMs of type "maxi") is the same before and after this PR. The only difference is that the extra vertical scrollbar shows before this PR, but that extra scrollbar doesn't help to scroll the iframe further down beyond the cutoff (it is a completely useless scrollbar). So I think this PR does address the reported issue, and we should fix the longstanding issue of content cut off by the bottom bar as part of #523, or else in a specific PR if you think there's a different viable solution. (You may remember we once had to add some `<br />` padding to the end of the content to work around this issue in jQuery mode, but that doesn't work in SW mode, and we weren't ever satisfied with that kludge...).

Jaifroid:
PS The extra scrollbar is produced by the "search-article" div, which has no bearing on the iframe, and therefore on the content in the iframe.

Jaifroid:
Hey @mossroy , I wondered if you had further thoughts about this PR (reading my comments above). As I say there, I don't think the issue you report pertains to this PR, which is merely about removing the extra vertical scrollbar. The extra scrollbar in question doesn't appear to govern the amont of content shown in the iframe, evinced by the fact that you can't actually scroll it at all.

mossroy:
Sorry that I did not have the time to work on that.

Before the bootstrap upgrade, it' true that we already had issues with the bottom content.
But having some content behind the bottom bar is not the same as having some content not visible at all.
When it's behind the bottom bar, it's still mostly visible as the bottom bar is transparent (even if links can not be clicked). And the content can become completely available if we remove (or hide) the bottom bar.
With this PR, the content seems lost to me.
Using `overflow: hidden` hides the content that does not fit (see examples in https://developer.mozilla.org/fr/docs/Web/CSS/overflow )

But, after testing a bit more, I noticed that this issue was already present before the bootstrap upgrade, a least on Chrome (in SW mode). So it might not be a regression caused by the upgrade.

mossroy:
@Jaifroid : I suggest to close this PR for now, as both of us did not have the time to work on it in the last month

PR #570, opened 2019-09-22
Stablebot configuration

kelson42:
This is the configuration for the Stablebot. The Stalebot set a "stalled" label on issues/PRs which are not active a certain time. This is a good system to help people to not forget about things and carry on with improvements.

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -1,0 +1,15 @@
+daysUntilClose: false
+staleLabel: stale
+
+issues:
+  daysUntilStale: 60
+  markComment: >
+    This issue has been automatically marked as stale because it has not had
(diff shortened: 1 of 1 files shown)


Jaifroid:
@kelson42 Would you be OK with us significantly increasing the default time periods before issues and PRs are marked stale? As we are both hard-pressed for the foreseeable future, the timescales look a bit unrealistic for us. The main reason we have so many open issues is lack of person-power to deal with them, not that we have forgotten about things that need doing. Arguably some long-standing issues could be closed as won't fix, but often they are longer term aspirations rather than immediate bug-fix problems.

mossroy:
This PR looks "stalled" to me (even with no automatic feature ;-) )
Feel free to re-open it if necessary

PR #572, opened 2019-10-30
Fixed extra vertical scrollbar appearing

### LINKED ISSUE #566: There is a vertical scrollbar since bootstrap 4

ab1nash:
A vertical scrollbar was appearing along with the one in the iframe
component which was unnecessary. This PR removes it by changing a small attribute in _app.css_.

Fixes #566.

### DIFF
--- a/www/js/app.js
+++ b/www/js/app.js
@@ -89,11 +89,10 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'cookies','abstractFilesystemAcc
      * Resize the IFrame height, so that it fills the whole available height in the window
      */
     function resizeIFrame() {
-        var height = $(window).outerHeight()
-                - $("#top").outerHeight(true)
-                // TODO : this 5 should be dynamically computed, and not hard-coded
-                - 5;
-        $(".articleIFrame").css("height", height + "px");
+        var height = window.innerHeight
+            - document.getElementById('top').getBoundingClientRect().height
+            - 6; // TODO: These 6 pixels appear to be padding or margins added to some element, but it is not clear what
+        document.getElementById('articleContent').style.height = height + 'px';
     }
     $(document).ready(resizeIFrame);
     $(window).resize(resizeIFrame);
--- a/www/js/app.js
+++ b/www/js/app.js
@@ -90,11 +90,15 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'cookies','abstractFilesystemAcc
      */
     function resizeIFrame() {
         var headerStyles = getComputedStyle(document.getElementById('top'));
-        var height = window.innerHeight
-            - parseFloat(headerStyles.height)
-            - parseFloat(headerStyles.marginBottom)
-            - 6; // TODO: These 6 pixels appear to be padding or margins added to some element, but it is not clear what
-        document.getElementById('articleContent').style.height = height + 'px';
+        var headerHeight = parseFloat(headerStyles.height) + parseFloat(headerStyles.marginBottom);
+        var iframe = document.getElementById('articleContent');
+        var searchArticle = document.getElementById('search-article');
+        if (iframe.style.display === 'none') {
+            searchArticle.style.height = window.innerHeight + 'px';
+        } else { 
+            searchArticle.style.height = window.innerHeight + headerHeight + 'px';
+        }
+        document.getElementById('articleContent').style.height = window.innerHeight - headerHeight + 'px';
     }
     $(document).ready(resizeIFrame);
     $(window).resize(resizeIFrame);
@@ -275,6 +279,7 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'cookies','abstractFilesystemAcc
         $('.alert').hide();
         refreshAPIStatus();
         refreshCacheStatus();
+        resizeIFrame();
         return false;
     });
     $('#btnAbout').on('click', function(e) {
@@ -298,6 +303,7 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'cookies','abstractFilesystemAcc
         $('#articleListWithHeader').hide();
         $("#searchingArticles").hide();
         $('.alert').hide();
+        resizeIFrame();
         return false;
     });
     $('input:radio[name=contentInjectionMode]').on('change', function(e) {
--- a/www/js/app.js
+++ b/www/js/app.js
@@ -95,7 +95,7 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'cookies','abstractFilesystemAcc
         if (iframe.style.display === 'none') {
             // We are in About or Configuration, so we only set the region height
             region.style.height = window.innerHeight + 'px';
-            region.style.overflowY = 'auto';
+            // region.style.overflowY = 'auto';
         } else { 
             // IE cannot retrieve computed headerStyles till the next paint, so we wait a few ticks
             setTimeout(function() {
@@ -104,7 +104,7 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'cookies','abstractFilesystemAcc
                 iframe.style.height = window.innerHeight - headerHeight + 'px';
                 // We have to allow a minimum safety margin of 10px for 'iframe' and 'header' to fit within 'region'
                 region.style.height = window.innerHeight + 10 + 'px';
-                region.style.overflowY = 'hidden';
+                // region.style.overflowY = 'hidden';
             }, 100);
         }
     }
--- a/www/js/app.js
+++ b/www/js/app.js
@@ -101,7 +101,9 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'cookies','abstractFilesystemAcc
                 // Get  header height *including* its bottom margin
                 var headerHeight = parseFloat(headerStyles.height) + parseFloat(headerStyles.marginBottom);
                 iframe.style.height = window.innerHeight - headerHeight + 'px';
-                region.style.height = window.outerHeight + 'px';
+                // We have to allow a minimum safety margin of 5px for 'iframe' and 'header'
+                // to fit within 'region' without displaying an overflow
+                region.style.height = window.innerHeight + 5 + 'px';
             }, 100);
         }
     }
@@ -286,6 +288,7 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'cookies','abstractFilesystemAcc
         $('.alert').hide();
         refreshAPIStatus();
         refreshCacheStatus();
+        // Use a timeout of 400ms because uiUtil.applyAnimationToSection uses a timeout of 300ms
         setTimeout(resizeIFrame, 400);
         return false;
     });
@@ -310,6 +313,7 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'cookies','abstractFilesystemAcc
         $('#articleListWithHeader').hide();
         $("#searchingArticles").hide();
         $('.alert').hide();
+        // Use a timeout of 400ms because uiUtil.applyAnimationToSection uses a timeout of 300ms
         setTimeout(resizeIFrame, 400);
         return false;
     });
--- a/www/js/app.js
+++ b/www/js/app.js
@@ -91,15 +91,17 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'cookies','abstractFilesystemAcc
     function resizeIFrame() {
         var headerStyles = getComputedStyle(document.getElementById('top'));
         var iframe = document.getElementById('articleContent');
-        var searchArticle = document.getElementById('search-article');
+        var region = document.getElementById('search-article');
         if (iframe.style.display === 'none') {
-            searchArticle.style.height = window.innerHeight + 'px';
+            // We are in About or Configuration, so we only set the region height
+            region.style.height = window.innerHeight + 'px';
         } else { 
-            // IE cannot compute headerStyles fast enough, so we have to wait a few ticks
+            // IE cannot retrieve computed headerStyles till the next paint, so we wait a few ticks
             setTimeout(function() {
+                // Get  header height *including* its bottom margin
                 var headerHeight = parseFloat(headerStyles.height) + parseFloat(headerStyles.marginBottom);
                 iframe.style.height = window.innerHeight - headerHeight + 'px';
-                searchArticle.style.height = window.outerHeight + 'px';
+                region.style.height = window.outerHeight + 'px';
             }, 100);
         }
     }
--- a/www/js/app.js
+++ b/www/js/app.js
@@ -89,8 +89,10 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'cookies','abstractFilesystemAcc
      * Resize the IFrame height, so that it fills the whole available height in the window
      */
     function resizeIFrame() {
+        var headerStyles = getComputedStyle(document.getElementById('top'));
         var height = window.innerHeight
-            - document.getElementById('top').getBoundingClientRect().height
+            - parseFloat(headerStyles.height)
+            - parseFloat(headerStyles.marginBottom)
             - 6; // TODO: These 6 pixels appear to be padding or margins added to some element, but it is not clear what
         document.getElementById('articleContent').style.height = height + 'px';
     }
--- a/www/js/app.js
+++ b/www/js/app.js
@@ -95,15 +95,16 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'cookies','abstractFilesystemAcc
         if (iframe.style.display === 'none') {
             // We are in About or Configuration, so we only set the region height
             region.style.height = window.innerHeight + 'px';
+            region.style.overflowY = 'auto';
         } else { 
             // IE cannot retrieve computed headerStyles till the next paint, so we wait a few ticks
             setTimeout(function() {
                 // Get  header height *including* its bottom margin
                 var headerHeight = parseFloat(headerStyles.height) + parseFloat(headerStyles.marginBottom);
                 iframe.style.height = window.innerHeight - headerHeight + 'px';
-                // We have to allow a minimum safety margin of 5px for 'iframe' and 'header'
-                // to fit within 'region' without displaying an overflow
-                region.style.height = window.innerHeight + 5 + 'px';
+                // We have to allow a minimum safety margin of 10px for 'iframe' and 'header' to fit within 'region'
+                region.style.height = window.innerHeight + 10 + 'px';
+                region.style.overflowY = 'hidden';
             }, 100);
         }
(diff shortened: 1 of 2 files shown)


Jaifroid:
That's an interesting change, @ab1nash . I've added myself as a reviewer, and will try to test later today (but probably not till this evening). Thanks for the PR.

Jaifroid:
An additional problem with this code is that it removes the vertical scrollbar on the Config page, so it is no longer possible to scroll down to settings that are below the fold.

image

ab1nash:
Thanks for the review. On it.

ab1nash:
@Jaifroid it was supposed to be `-1px` . Seems to be working now. Please check

ab1nash:
Screenshot from 2019-11-01 19-09-54

Jaifroid:
Hi @ab1nash Thank you for this. When you add further commits, please don't rebase your branch, because then my local copy gets out of sync with the server. Instead, just push any further commitsto the same PR on your repo, and it will get synchronized here.

Jaifroid:
Hi @ab1nash , I've taken a quick look at this, but I'm still seeing an extra scrollbar with smaller screen sizes. This is from Chromium with your latest change now synchronized:

image

ab1nash:
@Jaifroid I am fixing up my commit so that I don't have to make multiple commits for such a small issue. And hence the force push. Probably that is why you are facing problems. Do you want me to continue like that, or do you want me to make multiple commits?

> Hi @ab1nash Thank you for this. When you add further commits, please don't rebase your branch, because then my local copy gets out of sync with the server. Instead, just push any further commitsto the same PR on your repo, and it will get synchronized here.

Jaifroid:
@ab1nash It's best if you make multiple commits to the same branch, even if a commit (intentionally) undoes previous work. At the end, when all/any changes are approved, I can then do a simple squash and merge into a single commit on master.

ab1nash:
Please review now

Jaifroid:
@ab1nash This seems to be working, according to tests done so far, which is great news!
However, this PR is helpful in pointing us to a deeper underlying problem, which I think should be fixed at source rather than by changing the frame dimensions. If we are producing miscalculated content size, we need to fix the original miscalculation rather than arbitrarily patching the size of the frame. We should at least explore this.

I suggest starting by looking at the [`resizeIFrame` function](https://github.com/kiwix/kiwix-js/blob/master/www/js/app.js#L89) in `app.js`. You'll see that there's a fudged value there of -5 pixels which, as the comment states, should be dynamically calculated. Maybe this function is consistently calculating a size that is too large for the region, causing the overflow to appear. I think it's worth experimenting on this, as it would be a better fix. Are you able to look at it?

mossroy:
I fully agree with @Jaifroid : we need to understand the underlying problem and fix it

ab1nash:
@Jaifroid changing the -5 to -10 seems to fix the problem. But do you have any suggestions on how to make it dynamic? Also, decreasing that height property by some value does this:
Screenshot from 2019-11-03 20-25-45

Jaifroid:
OK, thanks, that's good news. It would be great if it's possible to implement a dynamic calculation here, so that we don't have to go around hunting again in the future if some other parameter changes the amount of space available to the iframe (in this case, it seems it was changed by the upgrade from Bootstrap 3 to 4).

At the moment, the only suggestion I have is to check if it's possible to derive the missing 10 pixels by subtracting values from different parts of the box model. You could try to use vanilla JavaScript properties `element.clientHeight` (equivalent to innerHeight) and `element.offsetHeight` (sort of equivalent to outerHeight) on different elements until you find the missing pixels (you can also use the CSS box model in DEV tools to see if you can spot what we're missing in the calculation). Note that the vanilla JS versions are properties, not functions like in jQuery.

ab1nash:
`body` seems to be missing a length vertically. Other elements have this attribute, as well as `body` of other webpages. I believe this is causing the problem. On it.

Screenshot from 2019-11-15 00-20-11
Screenshot from 2019-11-15 00-19-56
Screenshot from 2019-11-15 00-22-21

ab1nash:
I don't get it. Why is height of the `<body>` element `0` ?

ab1nash:
Screenshot from 2019-11-15 00-20-11

Jaifroid:
@ab1nash I'm not sure I follow you. Is this the `<body>` of the html document in the iframe, or the `<body>` of the outer html document, outside the iframe?

ab1nash:
@Jaifroid I am referring to the `body` of the outer html document

ab1nash:
Also, from what I found out, the height of `top` seems to be fine. `.view-content` is the one which seems to have an extra height of about 6px.
Is there a way to match the height of `.view-content` class with the iframe inside it? Height of the iframe and `top` add up to the height of the window

Jaifroid:
What specific element has the .view-content class? It's not really a class that has a height, it'll be the html element that is referenced by that class.

ab1nash:
The `<article>` element that follows the header with ID `top`

kelson42:
@jaifroid @ab1nash This PR seems stalled. What should we do?

Jaifroid:
@kelson42 @ab1nash A number of changes have been made in #582 that affect the display of UI elements. That PR is awaiting approval. I think once that has been merged, we can come back to this one and a) see if the problem still persists; and b) if so, implement a dynamic calculation on the affected element. I have been working on some of these issues in Kiwix JS Windows, and have developed ways of calculating UI elements' sizes so that we do not cause overflows.

kelson42:
> That PR is awaiting approva

Whose approval is this PR awaiting? Does work from @ab1nash is still needed?

Jaifroid:
@kelson42 This PR is lacking a dynamic calculation of some screen element size which @ab1nash was going to determine. However, I'm suggesting that we could merge #582 first because it contains some UI fixes that may supersede this PR. #582 is awaiting approval from @mossroy .

If @ab1nash wishes to complete this PR without waiting for #582, then that is also fine.

ab1nash:
Hey!
apologies for the inactivity.
No I wasn't able to find a feasible dynamic calculation for this. Although I did make a static solution for this it is fine someone else wishes to take this up.

mossroy:
I just approved #582

kelson42:
@Jaifroid This PR should be probably closed then if @ab1nash does not plan to work on ot anymore.

Jaifroid:
I'll take a look this weekend, because the remaining work for this PR is very minor.

Jaifroid:
@mossroy I've updated the PR using vanilla JS. Effectively all this does is update the value we were subtracting from the calculations (prior to this PR). Previously we subtracted a mysterious 5 pixels. Subtracting 6 pixels instead seems to make the scrollbar disappear on quick tests. I think these pixels are padding or margins, possibly auto-calculated by Bootstrap, on some element in the UI.

Is this OK as a fix? When we complete #523 , we can revisit this calculation.

Jaifroid:
Agreed, it's not ideal. I need to test on all browsers and screen formats. Will open another issue before merging.

Jaifroid:
On wide screens, the scrollbar is still showing. :-(  Needs some refining...

Jaifroid:
Interestingly, there is a difference in computed height between the `<article>` element and the `<iframe>` element of 5.5px. This explains the need to set the iframe to a height of `window.innerHeight` minus the `<header>` size _minus_ 6px for the overflow not to show. Well, it either explains it, or it is the result of us setting the iframe 6px less...

Jaifroid:
Well, I've had a go at computing these values. I'm not so sure how much is gained, as it turns out to be complicated, involving computedStyles, and having to set timeouts due to the timeouts added by the slide transition. And then there is a bug in Internet Explorer that returns "auto" for a computed style height, but then resolves it a few milliseconds later into a an actual pixel value, meaning we end up with trying to set a box height to NaN because the value hasn't been calculated yet. We could use a timeout here too, but I'm wondering if all this is worth it just to avoid hard-coding a value... Thoughts?

Jaifroid:
PS I suspect Bootstrap of mutating some styles, and because IE11 is slow, it can't resolve the calculation fast enough to provide us with a pixel value in time... a race condition, I guess.

Jaifroid:
So, I've cleaned up this code, documented it, and have tested on FFOS, IE11, Edge Legacy and Edge Chromium. I now understand the layout issues. There is an interaction between the `search-article` region (not a particularly good name, because it encompasses the entire app window, not just search), and the iframe. If `search-article` is smaller than the iframe + header + header margin, an overflow will show. So we have to compute the latter value. This is what jQuery computes with its `outerHeight(true)` function (we weren't using `true`, which is why we had to add the mysterious extra pixels). Additionally, we have to run `resizeIframe()` each time we switch between Home, About, Configuration, because we are now setting the height of `search-article` ('auto' does not work, I tried).

It was only by working this out in vanilla JS that I could understand the problem. JQuery runs a *lot* of code to compute the outerHeight including margin. The vanilla version I've come up with runs a lot less code, but looks more complex at first sight. I've documented it, nevertheless. Up to you, @mossroy, whether you prefer me to revert to jQuery here. We will still need to run `resizeIframe()` when switching app pages, and we'll need to do this in a timeout due to the swipe transitions.

Personally I prefer vanilla because it is less of a black box, and allows devs to understand what is really going on.

www/js/app.js
```diff
@@ -274,6 +286,7 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'cookies','abstractFilesystemAcc
         $('.alert').hide();
         refreshAPIStatus();
         refreshCacheStatus();
+        setTimeout(resizeIFrame, 400);
```
mossroy: It would be worth adding the same comment here

www/js/app.js
```diff
@@ -297,6 +310,7 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'cookies','abstractFilesystemAcc
         $('#articleListWithHeader').hide();
         $("#searchingArticles").hide();
         $('.alert').hide();
+        setTimeout(resizeIFrame, 400);
```
mossroy: And here too

Jaifroid:
Darn, the one browser I didn't test on, because I have Firefox running the old version that still supports FFOS. I did have the latest version running in WSL Ubuntu, but it's stopped working due to some dependency not supported by WSL...

mossroy:
Maybe running the latest Firefox inside a Virtual Machine could be helpful (with Virtualbox for example, inside an Ubuntu VM, or any other OS)

Jaifroid:
@mossroy I installed the latest Firefox (73.0.1 64bit) in Windows Sandbox, together with Git. However, on Ray Charles I'm not seeing the extra overflow (see below). Do you see it on Ray Charles, and if not there, do you see it on another ZIM / page?

image

mossroy:
I see it on Ray Charles ZIM file, and with any ZIM file, with my usual Firefox 73.0.1 (on Ubuntu 18.04). Either in jQuery or SW mode.
image

But I don't see it on Firefox 69.0.2 or 73.0.1 in an Ubuntu 18.04 VM.
I need to investigate.
It does not seem to depend on the screen size

mossroy:
It seems to come from an interaction with CanvasBlocker extension : https://addons.mozilla.org/en-US/firefox/addon/canvasblocker/
When installed (with default settings), the scrollbar appears. When disabled (or uninstalled), the scrollbar disappears.

Jaifroid:
OK, could it be blocking the correct calculation of the screen size? I think I read somewhere about that being blocked to prevent fingerprinting based on screen size.

mossroy:
Certainly.
The extension gives the following notification : 

    Faked screen readout on localhost

So it probably returns an intentionaly wrong value to one (or some) of the APIs we use.
The corresponding extension code seems to be https://github.com/kkapsner/CanvasBlocker/blob/master/lib/modifiedScreenAPI.js
It might be the use of `outerHeight` property, but I can't say for sure.

Now that we know it only happens with this extension activated, I don't think it should be considered as a blocker.
Anyway, it would be better if we used only APIs that are not considered "privacy unsafe" by this extension, if it's possible. Because Firefox (and maybe other browsers) is starting to implement similar ways to prevent fingerprinting (in the standard browser, without any extension).

Jaifroid:
Yes, outerHeight makes sense, as it might be a way to calculate the screen size if the browser screen were maximized (which TOR browser warns the user against doing). It may be possible to do without that. I'll check.

Jaifroid:
I have removed outerHeight by (guess what) using `window.innerHeight + 5 + 'px'`. 5px is the minimum that will not produce the overflow. The difference from the previous code is that we are now correctly calculating the header size including margin, and we are also correctly setting the region height when viewing about and config pages.

Could you test if this now works with the canvasBlocker extension? It works on Firefox 73 and Edge Legacy. I haven't yet tested on other browsers.

I have documented the 5px as a "safety margin" in a comment, which is what I believe it is. If you don't like reverting to hard-coding 5px, I could set the region size to `window.innerheight + 1px`, and hide the overflow, then re-show the overflow when it is needed (in About and Config). This would guarantee that we will never see the scrollbar, and we can be sure that the iframe will still fit in the region, because its size is precisely calculated.

Jaifroid:
Given the failure of Unit Tests due to this branch being initiated from a remote repository, it might be a good idea if, once we have agreed the code, I open a test branch/PR with the same diff from master, purely for the purpose of automated testing, before we merge this branch.

Jaifroid:
Interesting: creating a duplicate PR has caused the test result to carry over to this branch... I guess for Git, it just sees a successful test complete on that diff hash.

mossroy:
On my computer, 5 pixels are not enough (scrollbar visible), but 6 pixels are (no scrollbar).
I would suggest that :
- we use the 100% computed value in the usual case (no anti-fingerprinting feature)
- we use some hard-coded value only when we have no other choice

The difficulty is to find how to detect if the outerHeight is a fake one.
Based on https://github.com/kkapsner/CanvasBlocker/blob/master/lib/modifiedScreenAPI.js#L199 and my tests, `window.innerHeight` equals `window.outerHeight` when CanvasBlocker is activated.

So this works on my computer :

                if (window.outerHeight !== window.innerHeight) {
                    console.debug("outerHeight=" + window.outerHeight + " innerHeight=" + window.innerHeight);
                    region.style.height = window.outerHeight + 'px';
                } else {
                    console.debug("outerHeight=innerHeight=" + window.innerHeight + " => faked outerHeight?");
                    region.style.height = window.innerHeight + 6 + 'px';
                }

BUT I don't know if outerHeight and innerHeight can be equal in a "normal" case (no fake value)? Maybe in case of a WebView or another embedded browser (with no menu, tab, status bar etc)

Anyway, I'm surprised that we have to use outerHeight here. In my tests, its value is significantly different than innerHeight : 745px vs 438px. And, when I see what it corresponds to (see snapshot in https://developer.mozilla.org/en-US/docs/Web/API/Window/outerHeight ), I don't see why we have to use it.

Jaifroid:
Hmm. I was using `outerHeight` simply because that is what was used in the original code, as you can see. But I did not check that `$(window).outerHeight()` might mean something different from `window.outerHeight`. I should have checked.

To be honest, we could set the "region" it to an arbitrarily large number - I just tried `region.style.height = window.innerHeight + 10000 + 'px';`, and it works. The issue comes when region is smaller than the total space taken up by the iframe and the header and all margins and padding. We could even do `window.innerHeight * 2`. The important thing is that we reset region to `window.innerHeight` when we display Config and About.

A belt-and-braces approach would be to set it to something comprehensible like `window.innerHeight + 10` and hide `overflowY`, re-showing `overflowY` when displaying About & Config (together with resetting it to `window.innerHeight`). I think you talked yourself out of the code you gave above! I agree, with what we know now, that we should do without `outerHeight` completely.

Jaifroid:
@mossroy Does the latest code, which explicitly hides the overflow while ensuring no content is clipped, meet your queries above (subject to further testing)?

mossroy:
I'm sorry I could not have a look, yet.
Things are getting a bit crazy those days in Europe

Jaifroid:
Yes, I know, don't worry and stay safe! It's getting bad in UK too...

mossroy:
It works fine on the devices I tested (including Firefox OS).
It's better now, without the use of `outerHeight`, as anti-fingerprinting features will not interfere.
I still don't like the fact that we manually hide the `overflowY`, and add the arbitrary 10 pixels. I feel it might break on next UI change.
On the other hand, it works fine and I don't have time to work/review more for now.

So let's merge this PR, so that we can release our version.
It might be improved later.

Jaifroid:
I could comment out the overflowY, it's not needed if the code works correctly, it's just there as a redundancy.

mossroy:
Then yes, I would suggest to remove it

Jaifroid:
OK, but can't do till this evening as I'm currently in a car (navigating) to go and retrieve our children from London.

If you want to release today, please feel free to make that change, test and merge, otherwise I'll do it this evening.

mossroy:
No rush! Take care of your family

mossroy:
It seems to work fine with the commented lines about `overflowY`. I see no extra scrollbar on Firefox and Chromium.
Tested on Firefox OS too.

Jaifroid:
It seems to be working (ignore the Travis failure -- it's because I pushed to this external PR before pushing to the equivalent internal one, I'll sort that when I remove the commented lines).

However, I've noticed one instance where hiding the `overflowY` might be useful (see screenshot below). As can be seen, the extra scrollbar shows when the collapsed menus are opened on a narrow screen, due to the extra vertical height that is added suddenly inside `region`.

So, I have a question:

a) is it acceptable to show the extra scrollbar briefly in this instance? - it will disappear once the user clicks on one of the options;
b) or, if it should not show in this instance, would it be better to:
    i) hide it explicitly? or
   ii) recalculate the needed pixels when the user presses the hamburger menu?

image

mossroy:
My opinion is that the scrollbar appearing in this particular use-case is acceptable

Jaifroid:
Unfortunately, I seem to be getting the overflow again on a test with Edge Chromium in SW mode with Spanish wikipedia... I need to investigate more...

image

Jaifroid:
Hi @mossroy I found the issue in SW mode. It's curious that it only manifests on Chromium (unless I failed to test widely enough in SW mode), but we know that browsers use different async algorithms for rendering loaded content.

The solution is to add an extra `resizeIframe()` at the end of the code that runs after each SW-mode page load. Running this in Edge Chromium (SW mode), I do briefly (for maybe a tenth of a second) see the extra scrollbar, which then disappears immediately. This "flash" is only on the very first page load, not on subsequent page loads. If we want never to see it even for that brief moment, we'll need to hide `overflowY` as per the lines I commented out (now removed). I'm assuming that's not necessary.

I've now tested this on Edge Legacy, Edge Chromium, IE11, Firefox Quantum and FFOS.

If you're happy with it, could you please approve changes (you're currently still requesting changes), and then I'll merge.

Issue #304, opened 2017-09-13
User-interface (navigation and search placement)


b16r05:
The top panel has Kiwix version #, Home, Configure and About links, and to the right, a vast blank space exists which could house Search AND Navigation buttons found elsewhere.

Consider placing search (Search box and Scope along with Random Article button) on the top right to free up screen real estate. Also, consider placing bottom navigation panel buttons (home, left, right and up) in the middle of the top panel. 

Further, Navigation could be the first [left] item, followed by Search in the center and lastly, kiwik info+config on the right as that is usually changed/checked every now and then. Being able to move the three sets of items [left, center, right] by users would be nice.

Top Panel example:
HOME LEFT RIGHT UP.................................. SEARCH RANDOM..................................KIWIX HOME CONFIG ABOUT

Jaifroid:
I've only just seen this @b16r05 . I'd be interested in your view on the alternative arrangement we have in Kiwix JS Windows, which looks like this:

![image](https://user-images.githubusercontent.com/4304337/33700879-c3324ca4-db13-11e7-99f2-45b94aba6056.png)

On very narrow displays, the Search field become extremely small (but is still functional, because it opens to a wider field when typing in it):

![image](https://user-images.githubusercontent.com/4304337/33701515-fb6727ae-db16-11e7-8962-6954ee552abf.png)

This suggests to me that it's unlikely we could fit more buttons on this row without using the hamburger solution that Kiwix JS uses for narrow displays. The hamburger makes navigation very cumbersome, as it adds a significant number of extra taps just to change a setting.

b16r05:
It has been a while and I am embarrassed to say that I have forgotten how to open a sample zim page in kiwix-Firefox addon :-( However, I managed to review the pictures of kiwix 2.0 addon from the addons page (https://addons.mozilla.org/en-US/firefox/addon/kiwix-offline/#&gid=1&pid=3) and it has all come back to me :-)

Yes, what you have from Kiwix Windows interface looks good. However, I don't see navigation buttons, I think this is part of the application menu itself. What takes up one line space in Kiwix Windows takes up three line spaces in kiwix 2.0 (addon) minus the navigation. You could consider using a similar menu with BACK, FRONT and UP navigation buttons placed between the HOME (kiwix logo) and SEARCH box...

Jaifroid:
[...]

PR #574, opened 2019-11-12
Nav and search ui

kazerthekeen:
Hello, we were able to complete the first two recommended PR stages as in #304: changing text to icons and moving the search bar to the top menu. We still need to fix formatting for the collapsible hamburger menu on smaller screens as a next step but are looking for feedback on our current progress, feel free and test it out. Due to mobile issues it is not advised to merge this yet.

### DIFF
--- a/www/index.html
+++ b/www/index.html
@@ -57,44 +57,29 @@
                     <!--<a class="navbar-brand">Kiwix 2.7-WIP</a> -->
                     <img src="img/icons/kiwix-32.png">
                     <!-- Toggler/collapsible Button -->
+                    
+                    <input type="search" id="prefix" placeholder="Search..." class="form-control" />
+                    <button type="button" class="btn btn-outline-secondary" id="searchArticles"><i class="fas fa-search"></i></button>
+
                     <button class="navbar-toggler" type="button" data-toggle="collapse"
                         data-target=".navbar-collapse">
                         <i class="fas fa-bars"></i>
                     </button>
 
+
+
                     <!-- Collect the nav links, forms, and other content for toggling -->
                     <div class="collapse navbar-collapse" id="collapsibleNavbar">
                         <ul class="navbar-nav">
                             <li id="liHomeNav" class="nav-item"><button type="button" class="btn btn-light" id="btnHome"><i class="fas fa-home"></i></button></li>
-                            <input type="search" id="prefix" placeholder="Search..." class="form-control" />
-                            <button type="button" class="btn btn-outline-secondary" id="searchArticles"><i class="fas fa-search"></i></button>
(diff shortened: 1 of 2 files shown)


Jaifroid:
@kazerthekeen Thank you for this. I'll take a look as soon as I can and will give you some feedback.

www/index.html
```diff
@@ -54,44 +54,32 @@
         <section id="search-article" role="region">
             <header id="top">
                 <nav class="navbar navbar-expand-md bg-light" role="navigation">
-                    <a class="navbar-brand">Kiwix 2.7-WIP</a>
-
+                    <!--<a class="navbar-brand">Kiwix 2.7-WIP</a> -->
```
mossroy: Replacing the text by the icon is a good move.
On the other hand, it means that we're now missing the version information. Maybe it could simply be added in the About/Information section?

By the way, the whole line can be removed instead of commented

kelson42:
@kazerthekeen Thank you very for what I believe is your first PR here. Welcome on board!

kazerthekeen:
So we were looking into icons being off screen and it seems to only happen with narrow views on the desktop version of Firefox for at least version (70.0.1). Mobile seems to work fine on actual mobile devices for all browsers we tested. 
We tested with the following: 
Desktop: IE, Edge, Chrome, Safari and Firefox
Mobile: Firefox, Safari and Chrome

mossroy:
Thanks for the version in about section.

On Firefox (70.0.1 desktop and on an old deprecated Firefox OS 2.5 device, based on Gecko 44), there seems to be an issue with the search form width.
On Chromium 78, when you open the sandwich menu on a small screen, the width of the search form is reduced, so that to leave space to display the icons on the right.
But on Firefox, the width does not reduce below 234 pixels. Where it can reach 26 pixels on Chromium.
On my Firefox OS mobile device (ZTE Open C, with a 480x800 screen), it makes the sandwich menu almost unusable : 
image
When you tap on the sandwich menu, nothing seems to happen because the search form can't be reduced. So the icons appear off-screen, on the right. You can reach them by scrolling but it's hard to guess you should do that.

It might be related to some different CSS interpretation on these browsers.
You can reproduce and debug that with the dev tools of Firefox, by reducing the screen size in responsive mode : https://developer.mozilla.org/en-US/docs/Tools/Responsive_Design_Mode

NB : it's not crucial to support Firefox OS, but this issue needs to be fixed as it also affects modern versions of Firefox, and might affect some other browser engines too.

mossroy:
@kazerthekeen : do you intend to finalize this PR?

kelson42:
@mossroy Proably a good idea to close that PR now.

mossroy:
Closing this PR, as we have no news from the developer

Issue #575, opened 2019-11-13
Broken display when search results have special characters


mossroy:
Example below (on stackoverflow_eng_all_2017-05.zim, after typing "Why should")
![image](https://user-images.githubusercontent.com/890923/68788631-12224500-0644-11ea-80cc-3b0cc3bfd534.png)
In this example, the issue certainly comes from the < and > characters, that are interpreted as HTML.

We should probably apply some HTML encoding before doing some string concatenation

Jaifroid:
Ouch, that looks a bit nasty!

ayushiitr2901:
Hi, I am new contributing to Kiwix and this issue seems to be great headstart. Can you please help me with how I can start with it, I would like to solve this issue.

Jaifroid:
@ayushiitr2901 Thank you for offering to contribute!
First you will need to be able to reproduce the problem, which may need you to download the latest version of the Stackoverflow ZIM.
Once you are able to reproduce it, start debugging with
`function populateListOfArticles(dirEntryArray)`
in `app.js`.
Once you have tested a solution in your own clone of this repo, you can make a Pull Request, and one of us will take a look at your code.

vasusharma7:
Sir
I was also interested in solving the issue and I looked for the zim files of stackoverflow and even downloaded a few.But I was not a able to find the exact zim file with the issue.....
If it doesnt bother you, Can you please send me the name or the link of the exact file as some of the files are very large in size and downloading all of them may not be feasible....

With reagrds
Vasu Sharma

________________________________

mossroy:
@vasusharma7 : I found another case in a (very) smaller zim file.
If you download https://download.kiwix.org/zim/stack_exchange/matheducators.stackexchange.com_en_all_2019-02.zim (81MB) and type "imp" in the search field, you'll see a broken search result on `Importance of "Calculus->Analysis Transition Books"?` : 
![image](https://user-images.githubusercontent.com/890923/69994833-0cd35e80-154f-11ea-91cb-6c6d17918a79.png)

vasusharma7:
Thanks a lot sir for quick help...It would really help me in solving the issue...

________________________________

kelson42:
It think this has been fixed by #578

PR #578, opened 2019-12-05
 Solved : Broken display when search results have special characters (#575)

vasusharma7:
Solved the issue of "Broken display when search results have special characters (#575)" on kiwix-js
Screenshot (355)

### DIFF
--- a/www/js/app.js
+++ b/www/js/app.js
@@ -939,7 +939,8 @@ define(['jquery', 'zimArchiveLoader', 'util', 'uiUtil', 'cookies','abstractFiles
         var listLength = dirEntryArray.length < MAX_SEARCH_RESULT_SIZE ? dirEntryArray.length : MAX_SEARCH_RESULT_SIZE;
         for (var i = 0; i < listLength; i++) {
             var dirEntry = dirEntryArray[i];
-            articleListDivHtml += '<a href="#" dirEntryId="' + dirEntry.toStringId().replace(/'/g, '&apos;') +
+            var dirEntryStringId = uiUtil.htmlEscapeChars(dirEntry.toStringId());
+            articleListDivHtml += '<a href="#" dirEntryId="' + dirEntryStringId +
                 '" class="list-group-item">' + dirEntry.getTitleOrUrl() + '</a>';
         }
         articleListDiv.html(articleListDivHtml);
--- a/www/js/lib/uiUtil.js
+++ b/www/js/lib/uiUtil.js
@@ -211,7 +211,28 @@ define([], function() {
         else 
             return rect.top < window.innerHeight && rect.bottom > 0 && rect.left < window.innerWidth && rect.right > 0;
     }
+    /**
+     * Encodes the html escape characters in the string before using it as html class name,id etc.
+     * 
+     * @param {String} string The string in which html characters are to be escaped
+     * 
+     */
+    function htmlEscapeChars(string){
+        var escapechars = {
+            '&': '&amp;',
+            '<': '&lt;',
+            '>': '&gt;',
+            '"': '&quot;',
+            "'": '&#39;',
+            '/': '&#x2F;',
+            '`': '&#x60;',
+            '=': '&#x3D;'
+          };
+          string = String(string).replace(/[&<>"'`=/]/g, function (s) {
+            return escapechars[s];});
+          return string;
 
+    }
     /**
      * Functions and classes exposed by this module
      */
@@ -222,6 +243,7 @@ define([], function() {
         removeUrlParameters: removeUrlParameters,
         displayActiveContentWarning: displayActiveContentWarning,
         displayFileDownloadAlert: displayFileDownloadAlert,
-        isElementInView: isElementInView
+        isElementInView: isElementInView,
+        htmlEscapeChars : htmlEscapeChars
     };
 });


kelson42:
@mossroy I'm of course cautious because of my limited knowledge about Javascript... but I had a look to this PR and would have written the same but proposing a slightly other approach for the solution which seems even more generic. See https://medium.com/@tertiumnon/js-how-to-decode-html-entities-8ea807a140e5

vasusharma7:
Many thanks for your guidance and suggestions.I will definitely look into it and implement it as soon  as possible.

mossroy:
@kelson42 : your approach is indeed close to another answer in the stackoverflow link I gave.
It's certainly working, and gives a more concise code.
On the other hand, I was worried about its performance. This code is to be run in a loop for each search result. For each title it would create a temporary DOM object, which I presume to be slower than a regular expression search/replace.
But maybe a performance test could proove me wrong : @vasusharma7 if you're comfortable with https://jsperf.com/ or some other way to compare the performance of both approaches, it would be great.

vasusharma7:
Thanks you sir for your alternative solutions and short analysis. I will implement both and try to compare the efficiency for both.
But the comment section of the jQuerry approach doesn't seem to be quite inspiring (the jQuerry technique has some issues like-
Annotation 2019-12-05 224500
 returning new lines etc. on some browsers). Anyway I will try and see both which is better.

mossroy:
@vasusharma7 : an article title should be on a single line so it's probably not affected.
Regarding which implementations to compare, I would exclude the jQuery way.
I would only compare the performance of escapeHtml (from mustache.js) and the "Encode HTML-entities" implementation of the medium.com link

mossroy:
@vasusharma7 : after reading the comments from stackoverflow, it should be checked if quotes and double-quotes are escaped with the medium.com implementation.
If they're not, it's probably not worth using it.

vasusharma7:
The First method of htmlEscapeChars looks good as double quotes are also  encoded by it where as htmlEntities method doesn't encode the double quotes. On top of that we can define more of the custom encodings in the EscapeChars method if needed in future.This is a plus point.
i haven't done the performance check yet..
But if still needed,I will do it and get back to you soon

www/js/app.js
```diff
@@ -939,7 +939,9 @@ define(['jquery', 'zimArchiveLoader', 'util', 'uiUtil', 'cookies','abstractFiles
         var listLength = dirEntryArray.length < MAX_SEARCH_RESULT_SIZE ? dirEntryArray.length : MAX_SEARCH_RESULT_SIZE;
         for (var i = 0; i < listLength; i++) {
             var dirEntry = dirEntryArray[i];
-            articleListDivHtml += '<a href="#" dirEntryId="' + dirEntry.toStringId().replace(/'/g, '&apos;') +
+            var dirEntryStringId = dirEntry.toStringId();
+            dirEntryStringId = uiUtil.htmlEscapeChars(dirEntryStringId);
```
mossroy: Very minor : these 2 lines could be combined into a single one, so that the `dirEntryStringId` can not have two different values (escaped and unescaped) : 
```suggestion
            var dirEntryStringId = uiUtil.htmlEscapeChars(dirEntry.toStringId());
```

vasusharma7:
Thanks for the approval sir..

www/js/app.js
```diff
@@ -939,7 +939,9 @@ define(['jquery', 'zimArchiveLoader', 'util', 'uiUtil', 'cookies','abstractFiles
         var listLength = dirEntryArray.length < MAX_SEARCH_RESULT_SIZE ? dirEntryArray.length : MAX_SEARCH_RESULT_SIZE;
         for (var i = 0; i < listLength; i++) {
             var dirEntry = dirEntryArray[i];
-            articleListDivHtml += '<a href="#" dirEntryId="' + dirEntry.toStringId().replace(/'/g, '&apos;') +
+            var dirEntryStringId = dirEntry.toStringId();
```
mossroy: My code proposal was intended to replace both lines. So this one should be removed now.

vasusharma7: Ya I have done that and am going to upload it....I just pressed the commit suggested changes and it did that.. sorry for that

vasusharma7: Please review now..I have done which you had suggested.

www/js/lib/uiUtil.js
```diff
@@ -211,7 +211,28 @@ define([], function() {
...
+            '&': '&amp;',
+            '<': '&lt;',
+            '>': '&gt;',
+            '"': '&quot;',
+            "'": '&#39;',
+            '/': '&#x2F;',
+            '`': '&#x60;',
+            '=': '&#x3D;'
+          };
+          string = String(string).replace(/[&<>"'`=\/]/g, function (s) {
```
mossroy: Sorry for another minor remark : CodeFactor says the backslash is unnecessary here : https://www.codefactor.io/repository/github/kiwix/kiwix-js/pull/578

vasusharma7: Ya sir...I have just removed that.....please review now.

mossroy:
IMHO, it's ready for merge.
If you know how to rebase and squash your changes in this PR, please do it. Else I'll squash through github.

vasusharma7:
I was new to rebase and squash but i learned about them and have done it.Please have a review if I have done it correctly or not.....

mossroy:
The rebase/squash was fine.
I've merged your PR : thanks for this contribution

vasusharma7:
Thanks a lot for accepting my PR . I learned a lot from this PR and whole of our conversation. Keep guiding in future also...

PR #580, opened 2019-12-31
ray charles zim has different url, which will change in 2 months again

soloturn:
not sure what to do, maybe a latest url on the webserver for all files?


kelson42:
@soloturn Do you plan to fix this PR?

Jaifroid:
@mossroy We could simply merge this update to the latest Ray Charles right now, before releasing 2.7 and close this, pending a better documentation update later. Have just pushed correct updated link.

kelson42:
@soloturn @Jaifroid I think you should use here a permalink which always points to the last version, ie. https://download.kiwix.org/zim/wikipedia/wikipedia_en_ray_charles_maxi.zim

Jaifroid:
Thanks, @kelson42, I'd forgotten about permalink. Very good solution! The actual link is:
https://download.kiwix.org/zim/wikipedia_en_ray_charles_maxi.zim
If pushed this.

Jaifroid:
EDIT: Sorry, that is the same link you gave! Not sure why I got a not found first time. Working now anyway.

PR #587, opened 2020-01-31
Attempt to fix issue # 89

### LINKED ISSUE #89: When going quickly through history (back/forward buttons), it displays some parts of random articles
It's certainly because the callbacks are not ignored after the user went to another page in history

zengchu2:
Fixes #89

Add a new global variable in app.js which tracks the current curl.  

Add early return in the **displayArticleContentInIframe()** , **readArtile()** if the article's title does not match current url's title.


www/js/app.js
```diff
@@ -85,6 +85,9 @@ define(['jquery', 'zimArchiveLoader', 'util', 'uiUtil', 'cookies','abstractFiles
     var globalDropZone = document.getElementById('search-article');
     var configDropZone = document.getElementById('configuration');
     
+    // Title for current url.
+    var curTitle = "A/index";
```
Jaifroid: @zengchu2 Thank you for this PR. I haven't tested yet, but can I just clarify why you assume here that the current URL is named `A/index`? Is this just a placeholder with no significance that will be overwritten, in which case, should it just be `curTitle = ""`?

zengchu2: I hard coded it to be **"A/index"** to enable **readArticle()** load the index page. (In the case of **curTitle = ""**,  the index page will not be loaded due to the early return)

mossroy: I understand the purpose, but you should not hard-code the name of the index page (because it depends on the ZIM file).
A better way to do that would be to set this variable in `goToMainArticle` function (before calling `readArticle`) with the content of `dirEntry` variable.
And the same thing should be done in `goToRandomArticle`.

There also seems to be a confusion between the URL (with or without namespace) and the title of an article (it's my fault and the code is not always clear on that).
When I was referring to a title in #89 , I should have referred to the URL (with namespace, so that it is unique).
It's already what you're putting in the variable, but you should rename the variable to something more meaningful, to indicate that it's the URL we expect to see appearing. Maybe `futureArticleURL` or `expectedArticleURLToBeDisplayed`? @Jaifroid is the master of wording. And an explanation might be put above on what is our intention (which problem we are solving with that)

www/js/app.js
```diff
@@ -1017,6 +1020,10 @@ define(['jquery', 'zimArchiveLoader', 'util', 'uiUtil', 'cookies','abstractFiles
         // but we should not do this when opening the landing page (or else one of the Unit Tests fails, at least on Chrome 58)
         if (!params.isLandingPage) document.getElementById('articleContent').contentWindow.focus();
 
+        // Early return if article's title not equal current url's title
+        var matchUrl = dirEntry['namespace'] + "/" + dirEntry["url"];
```
mossroy: We usually use the syntax `dirEntry.namespace` instead of dirEntry['namespace'] (this applies everywhere you used this syntax)

www/js/app.js
```diff
@@ -1017,6 +1020,10 @@ define(['jquery', 'zimArchiveLoader', 'util', 'uiUtil', 'cookies','abstractFiles
...
+        if(matchUrl !== curTitle) return;
```
mossroy: It would be worth adding a `console.debug` here so that this kind of (rare) case could be easily detected and understood

zengchu2:
Thanks for your patience. I managed to fix the mentioned issues I had so far and will keep my eyes on my code.

www/js/app.js
```diff
@@ -1017,6 +1021,14 @@ define(['jquery', 'zimArchiveLoader', 'util', 'uiUtil', 'cookies','abstractFiles
         // but we should not do this when opening the landing page (or else one of the Unit Tests fails, at least on Chrome 58)
         if (!params.isLandingPage) document.getElementById('articleContent').contentWindow.focus();
 
+        // Early return if article's title not equal current url's title
```
mossroy: I think this piece of code should be moved right before line `iframeArticleContent.src = `

zengchu2: Hi mossroy, I have some trouble with the `selectedArchive.resolveRedirect(dirEntry, readArticle);` part in the `readArticle()` function. 

For now I just update the global variable `expectedArticleURLToBeDisplayed`(I exchanged the name of curTitle and expectedArticleURLToBeDisplayed as suggested from your previous comment) everytime before calling `readArticle()`.  In the case of viewing documents where there is a redirect,  `selectedArchive.resolveRedirect(dirEntry, readArticle);` will trigger the `readArticle()` without updating the `expectedArticleURLToBeDisplayed` and cause an early return. Is setting another global variable `isRedirectFromOtherEntry` gonna to help with fixing this unexpected early return caused by redirecting? Say `isRedirectFromOtherEntry` is set to true every time the `resolveRedirect()` is called and the `readArticle()` will skip early return if `isRedirectFromOtherEntry` is true and set it back to false.

mossroy: Could you please push to github the current state of your code? I do not see the variable names exchanged, for example.
I don't think we should introduce a new global variable for that. It would be better that a redirect would update the global variable to the new URL, before starting reading it in the backend

zengchu2: Code updated. I tested my version locally and it seems fine now.

www/js/app.js
```diff
@@ -998,16 +1001,47 @@ define(['jquery', 'zimArchiveLoader', 'util', 'uiUtil', 'cookies','abstractFiles
             document.getElementById('articleContent').contentWindow.focus();
             $("#searchingArticles").show();
             if (dirEntry.isRedirect()) {
-                selectedArchive.resolveRedirect(dirEntry, readArticle);
+                // Update expectedArticleURLToBeDisplayed before redirect. 
+                selectedArchive.resolveRedirect(dirEntry, function (resolvedDirEntry) {
+                    updatedExpectedURLOnRedirect(resolvedDirEntry, readArticle);
```
mossroy: That should work (I did not have time to test for now, sorry).
On the other hand, I think it could be simplified by removing the updatedExpectedURLOnRedirect function (which was fine), and move the `expectedArticleURLToBeDisplayed = dirEntry.namespace + "/" + dirEntry.url;` at the beginning of `readArticle`.
It should replace several places where you call that function or do the same : lines 1006, 1010, 1103, 1540 and 1563.

www/js/app.js
```diff
@@ -1017,6 +1051,14 @@ define(['jquery', 'zimArchiveLoader', 'util', 'uiUtil', 'cookies','abstractFiles
         // but we should not do this when opening the landing page (or else one of the Unit Tests fails, at least on Chrome 58)
         if (!params.isLandingPage) document.getElementById('articleContent').contentWindow.focus();
 
+        // Early return if article's title is not expected to be displayed
+        // Cause an issue if the current dirEntry is redirected from another dirEntry since the urlExpecedToBedisplayed is not updated
+        // before redirecting.
+        if(! entryIsToExpectedBeDisplayed(dirEntry)){
```
mossroy: It's too early here to do the check, as the content has not been read in the ZIM file yet.
See https://github.com/kiwix/kiwix-js/issues/89#issuecomment-561666315 : it should be done right before displaying the content (at two different lines : one for jQuery mode, one for ServiceWorker mode)

www/js/app.js
```diff
@@ -998,16 +1001,47 @@ define(['jquery', 'zimArchiveLoader', 'util', 'uiUtil', 'cookies','abstractFiles
...
         } else {
             alert("Data files not set");
         }
     }
 
+    /**
+     * Check whether the given URL from given dirEntry equals the expectedArticleURLToBeDisplayed
+     * @param {DirEntry} dirEntry The directory entry of the article to read
+     */
+    function entryIsToExpectedBeDisplayed(dirEntry) {
```
mossroy: The name of this function seems difficult to understand for me.
I would suggest `isDirEntryExpectedToBeDisplayed`

www/js/app.js
zengchu2: This is a good point. I use a helper function cause it is easier for me to log the debug messages while testing.

www/js/app.js
```diff
@@ -998,16 +1001,36 @@ define(['jquery', 'zimArchiveLoader', 'util', 'uiUtil', 'cookies','abstractFiles
             document.getElementById('articleContent').contentWindow.focus();
             $("#searchingArticles").show();
             if (dirEntry.isRedirect()) {
-                selectedArchive.resolveRedirect(dirEntry, readArticle);
+                // Update expectedArticleURLToBeDisplayed before redirect. 
+                selectedArchive.resolveRedirect(dirEntry, function (resolvedDirEntry) {
+                    expectedArticleURLToBeDisplayed = resolvedDirEntry.namespace + "/" + resolvedDirEntry.url;
```
mossroy: As said in my previous review, it would simply the code to move this line at the beginning of `readArticle` (and remove all the same ones that are right before calling `readArticle`).
Here, it would also avoid creating this new inline function. In fact, the original code could be left

www/js/app.js
```diff
@@ -1051,16 +1074,30 @@ define(['jquery', 'zimArchiveLoader', 'util', 'uiUtil', 'cookies','abstractFiles
                     $("#searchingArticles").show();
                 };
             };
+
+            if(! isDirEntryExpectedToBeDisplayed(dirEntry)){
+                console.debug(dirEntry, " is not expected to be displayed. Early retrun in readArticle().")
```
mossroy: This debug line seems unnecessary to me, because `isDirEntryExpectedToBeDisplayed` already outputs a log.
Or maybe you could merge them?

www/js/app.js
```diff
@@ -1051,16 +1074,30 @@ define(['jquery', 'zimArchiveLoader', 'util', 'uiUtil', 'cookies','abstractFiles
...
+
             // We put the ZIM filename as a prefix in the URL, so that browser caches are separate for each ZIM file
             iframeArticleContent.src = "../" + selectedArchive._file._files[0].name + "/" + dirEntry.namespace + "/" + encodedUrl;
         } else {
             // In jQuery mode, we read the article content in the backend and manually insert it in the iframe
             if (dirEntry.isRedirect()) {
-                selectedArchive.resolveRedirect(dirEntry, readArticle);
+                // Update expectedArticleURLToBeDisplayed before redirect then call readArticle.
+                selectedArchive.resolveRedirect(dirEntry, function (resolvedDirEntry) {
+                    expectedArticleURLToBeDisplayed = resolvedDirEntry.namespace + "/" + resolvedDirEntry.url;
```
mossroy: Same remark as in function `findDirEntryFromDirEntryIdAndLaunchArticleRead` : the original code could be left if the variable assignment was done at the beginning of `readArticle`

www/js/app.js
```diff
@@ -1051,16 +1074,30 @@ define(['jquery', 'zimArchiveLoader', 'util', 'uiUtil', 'cookies','abstractFiles
...
+                    readArticle(resolvedDirEntry)
+                });
             } else {
                 // Line below was inserted to prevent the spinner being hidden, possibly by an async function, when pressing the Random button in quick succession
                 // TODO: Investigate whether it is really an async issue or whether there is a rogue .hide() statement in the chain
                 $("#searchingArticles").show();
+                if(! isDirEntryExpectedToBeDisplayed(dirEntry)){
```
mossroy: It's still too early to do that.
Let me re-explain our goal : we need to assign the variable as soon as possible (when we know the URL that is requested), and we need to check if this variable did not change right before displaying the article content. So that the main asynchronous part (reading the content in the ZIM file) is between these 2 steps.
Here, `selectedArchive.readUtf8File` is the asynchronous part, so you need to move that at the beginning of `displayArticleContentInIframe`

www/js/app.js
zengchu2: The reason why I do update here is that I have no places to do the param udpate before `resolveRedirect()` call the `readArticle()` with a new dirEntry in this file except using this new inline function.   Please let me know if there is any other places to do the update after calling the resolveRedirect() method.

www/js/app.js
zengchu2: Let me try to rephrase this problem to see if I truly understand the design here. 

Consider the following situation with without refactoring the too early early return in the code. 
Assumes the user clicked on article **1**.  The user does not click on article **2** until function `displayArticleContentInIframe()` with dirEntry for article **1** as it's input into the asyn queue.  With the too early early return,  article **1** is still loaded before article **2** which is not expected in this rare condition under jQuery mode.  This can be fixed by placing the early return right at the begining of `displayArticleContentInIframe`. 

As a side note, there is no such problem in serviceworkder mode since there is no asyn loading.

www/js/app.js
```diff
@@ -1051,15 +1072,23 @@ define(['jquery', 'zimArchiveLoader', 'util', 'uiUtil', 'cookies','abstractFiles
...
             if (dirEntry.isRedirect()) {
-                selectedArchive.resolveRedirect(dirEntry, readArticle);
+                // Update expectedArticleURLToBeDisplayed before redirect then call readArticle.
+                selectedArchive.resolveRedirect(dirEntry, function (resolvedDirEntry) {
+                    expectedArticleURLToBeDisplayed = resolvedDirEntry.namespace + "/" + resolvedDirEntry.url;
+                    readArticle(resolvedDirEntry)
+                });
             } else {
                 // Line below was inserted to prevent the spinner being hidden, possibly by an async function, when pressing the Random button in quick succession
-                // TODO: Investigate whether it is really an async issue or whether there is a rogue .hide() statement in the chain
```
mossroy: Why did you remove this comment?
It was a comment from @Jaifroid, please leave it unless he asked you to remove it

www/js/app.js
```diff
@@ -998,16 +1001,34 @@ define(['jquery', 'zimArchiveLoader', 'util', 'uiUtil', 'cookies','abstractFiles
...
 
+    /**
+     * Check whether the given URL from given dirEntry equals the expectedArticleURLToBeDisplayed
+     * @param {DirEntry} dirEntry The directory entry of the article to read
+     */
+    function isDirEntryExpectedToBeDisplayed(dirEntry) {
+        var curArticleURL = dirEntry.namespace + "/" + dirEntry.url;
+
+        if (expectedArticleURLToBeDisplayed !== curArticleURL) {
+            return false;
```
mossroy: I think it's necessary to keep a debug log in this case.
So that, when testing the app, we can see that there has been an early return

www/js/app.js
```diff
@@ -1152,6 +1181,11 @@ define(['jquery', 'zimArchiveLoader', 'util', 'uiUtil', 'cookies','abstractFiles
             if (regexpActiveContent.test(htmlArticle)) uiUtil.displayActiveContentWarning();
         }
 
+        if(! isDirEntryExpectedToBeDisplayed(dirEntry)){
```
mossroy: This should be at the very beginning of `displayArticleContentInIframe`.
Else we could be in a situation where we display a warning about an article that we did not display.

www/js/app.js
```diff
@@ -1017,7 +1038,7 @@ define(['jquery', 'zimArchiveLoader', 'util', 'uiUtil', 'cookies','abstractFiles
         // but we should not do this when opening the landing page (or else one of the Unit Tests fails, at least on Chrome 58)
         if (!params.isLandingPage) document.getElementById('articleContent').contentWindow.focus();
 
-        if (contentInjectionMode === 'serviceworker') {
+        if (contentInjectionMode === '1') {
```
mossroy: Why this change??
It looks like a mistake that breaks the ServiceWorker mode.

Please review your code before asking for a review from someone else.

www/js/app.js
zengchu2: Thanks for the hints. 
I over conplicated my code. The udpates are duplicated if we put the early return in the asynchronous part.

zengchu2:
Sorry for the abruptly fall back of my code. I managed to go back to commit 44ab9d0 to get a cleaner version of code (Without modifying the main part of the code).

I have tested this version of code by tests.html and play around under jQuery mode and serviceworker mode.  The pages can loaded normally and there is no errors shown from tests.html.

The only thing I am not sure is that there are some errors logged by 'jquery.js', 'jsConfigVars.js' and 'site.js' saying things like 'unexpected token .' and 'mw is not defined'.

mossroy:
Sorry, I did not have time to test yet. Please give me some time to do so.
I had a quick look at the code, and it seems right now.

It's not crucial but there are a few minor changes on app.js that are probably not intentional and should be avoided.
When you have a look at the changes (https://github.com/kiwix/kiwix-js/pull/587/files), you'll see that 2 lines are marked as modified, only because of extra spaces. Could you please revert these lines to their original state?
Another strange change is that the attributes (mode) of this file app.js has been changed : you seem to have added the executable bit (-rwxr-xr-x instead of -rw-r--r--) : a chmod -x app.js should fix that.

I'll try to find some time to test in the following week.

zengchu2:
Thank you sir, will fix the space and permission problem soon.

Issue #588, opened 2020-02-01
Update deprecated Promise patterns and methods


Jaifroid:
We have a number of non-spec-compliant deprecated Promise patterns and methods in the code. For example, `.fail()` is used instead of the ES6-compliant `.catch()` (no reason for this as the latter is fully supported by Q). We also have a few cases of the [`.defer` anti-pattern](https://github.com/petkaantonov/bluebird/wiki/Promise-anti-patterns#the-deferred-anti-pattern). Correcting/re-factoring these will enable an easier transition to ES6, for example if/when we implement #554 .

mossroy:
Good idea, and I suppose it's not risky

kelson42:
@Jaifroid Should we close this ticket?

Jaifroid:
Yes!

Issue #554, opened 2019-08-20
Introduce npm+webpack to build the app


mossroy:
It would have some advantages. At least : 
- ability to use recent javascript APIs (like ES6), and use a transpiler like https://babeljs.io/ to generate a javascript code compatible with all browsers. It can be handy in some cases, but won't magically add features like ServiceWorkers to browsers that don't support it
- ability to code with TypeScript (or some other language) instead of javascript
- running the code through a local webserver is a standard feature of npm (and it's now needed on Firefox and Chrome)
- webpack can handle the dependencies without putting them in our source code (for the common ones like jQuery or Bootstrap), and could certainly replace requirejs
- it would probably enable some automatic reload of the browser window when modifying the source code
- it could probably replace the dirty shell scripts I use to generate the packages for each platform when doing a release

I thought about implementing this several times, but the following drawbacks kept me from doing it, so far : 
- first of all, it's not a trivial change to implement : it would take quite some time
- it would be a significant change in the architecture, and would also imply to update the unit tests, Travis CI, github pages etc
- we would need to implement the source maps to be able to debug, and check that the dev tools of our browsers support them (it's probably not the case for some old browsers)
- some pros of using this stack (like minification, handling of browser cache by creating a new javascript file name on each build etc) are not relevant for us, and could on the contrary make our daily work more complicated. But we probably can disable the features we don't want
- setting up a development environment will be a bit more complicated as it involves installing nodejs, letting npm download the libraries etc.

kelson42:
I strongly support this

Jaifroid:
This guide seems useful (much appreciated in the comments):

https://gist.github.com/xjamundx/b1c800e9282e16a6a18e

Bam92:
This is very interesting. Moderne JS development requires at least these two technologies. What do you think?

Jaifroid:
@Bam92 Yes. Personally I think the time has come for this, not least because it would enable use of frameworks that now expect this structure, automatic updating of dependencies, and possibility of using supersets of JS such as TypeScript, compiling for all targeted browsers with Babel (which makes use of antiquated dependencies like jQuery unnecessary). We believe it is a huge amount of work, however...

Bam92:
@Jaifroid Yes, it is a huge amount of work. But I'd suggest migrating little by little.

mossroy:
This might be some work for an upcoming hackathon

Bam92:
Which hackathon do you mean here?

[...]

PR #589, opened 2020-02-01
Update Q and deprecated Promise patterns and methods

Jaifroid:
Implements #588 for almost all of the deprecated methods and patterns. I have not touched the patterns in `abstractFilesystemAccess.js` because they are FFOS-specific, and I guess this code will not be transferred to any ES6-webkit-npm version. I have kept Q, however, for backwards compatibility with old devices. To dispense with Q, we would simply need to change `Q.Promise()` patterns to `new Promise()`.

I have standardized on using Q with a capital letter (we have some variation).

There is one case of a `Q.when` which is not in the Promise spec. I haven't touched it (yet?) because I don't know what the difference is between when and then in Q.

### DIFF
--- a/www/js/app.js
+++ b/www/js/app.js
@@ -862,7 +862,7 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'cookies','abstractFilesystemAcc
      * @returns {Promise<Blob>} A promise for the requested file (blob)
      */
     function readRemoteArchive(url) {
-        // DEV: This deferred can't be standardized to an ES6 Promise pattern (using Q) because
+        // DEV: This deferred can't be standardized to a Promise/A+ pattern (using Q) because
         // IE11 is unable to scope the callbacks inside the Promise correctly. See [kiwix.js #589]
         var deferred = Q.defer();
         var request = new XMLHttpRequest();
@@ -879,8 +879,7 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'cookies','abstractFilesystemAcc
                 }
             }
         };
-        request.onabort = deferred.reject;
-        request.onerror = deferred.reject;
+        request.onabort = request.onerror = deferred.reject;
         request.send();
         return deferred.promise;
     }
--- a/www/js/app.js
+++ b/www/js/app.js
@@ -26,8 +26,8 @@
 // This uses require.js to structure javascript:
 // http://requirejs.org/docs/api.html#define
 
-define(['jquery', 'zimArchiveLoader', 'util', 'uiUtil', 'cookies','abstractFilesystemAccess','q'],
- function($, zimArchiveLoader, util, uiUtil, cookies, abstractFilesystemAccess, q) {
+define(['jquery', 'zimArchiveLoader', 'uiUtil', 'cookies','abstractFilesystemAccess','q'],
+ function($, zimArchiveLoader, uiUtil, cookies, abstractFilesystemAccess, Q) {
      
     /**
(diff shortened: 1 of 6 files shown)


Jaifroid:
Yes, I think you're probably right about which function is causing the problem, though I did seem to notice it working on some browsers which would be odd. Maybe a timing issue. Will check.

Jaifroid:
With the new commit, it's passing tests on all browsers except IE11... The previous one was failing on all browsers, so it's an improvement at least.

Jaifroid:
With Q updated to 1.5.1, the standard Promisify pattern for XHR (as recommended by Mozilla, for example) now works on all browsers except IE11. I think the problem is the interaction between callbacks inside a Promise and Q.all. I'll try a few more things, but it may be necessary to revert back to the Q.defer() antipattern in this case, and then try again with babel if/when we achieve #554 (npm and webpack).

Jaifroid:
Although IE11 tests are now passing, IE11 is failing to load more than three pages in succession with this PR - it seems to have a major memory leak. Clearly IE11 does not do too well with standard Promises implemented by Q...

Working on FFOS, FF, Edge Legacy. But I will need to debug the further IE11 issues before this will be ready to merge.

Jaifroid:
The noted issue with IE11 seems to exist on master also (see #594), and may not be a problem with this PR, though it appeared to be exacerbated by using Promise/A+ patterns especially those that are iterated with `Q.all`. It needs careful testing.

Jaifroid:
I have rebased this PR on master, as it had got out of hand with many reverts, and master had changed from the base used to set up the PR.

Jaifroid:
OK, @mossroy, this code is now ready for final review.

Please note that I've included an update to the latest version of the [Q library (v1.5.1)](https://github.com/kriskowal/q/releases/tag/v1.5.1) . Our copy was copyright 2012, the new one is copyright 2017.

The latest commits are passing Travis. The CodeFactor issues are all from the Q library itself, and these will disappear once merged, as we have that file marked as a library file (to be ignored in master once merged).

Where it was not possible to convert a function to the Promise/A+ spec patterns due to blocking by IE11, I have left a comment to explain. I expect that when we achieve #554, we will use [babel](https://babeljs.io/) or typescript, or both, to provide minimal polyfills, and will no longer need Q. Similarly, I have not touched the promises in the functions that are only currently used by FFOS, as I imagine these may not be merged into #554.

The latest commit https://github.com/kiwix/kiwix-js/pull/589/commits/3fa47b00ac98506ad2811e25d4d09dccc73b660d has been tested in jQuery and SW modes (where appropriate) on: Edge Chromium, Edge Legacy, Internet Explorer 11, Firefox LTS, FFOS.

Jaifroid:
Thanks, no rush @mossroy. It's a good idea to check it on a physical device. I'll squash once you've approved (since I already rebased recently on master).

Incidentally, I've been thinking we may have enough for a new release, since you had signalled end of February for the next release. The main addition is dark mode, but I think it's a big-enough and requested feature to be worth a release. I guess it could be discussed under an issue rather than here.

Jaifroid:
I saw that and was sure I'd fixed it... Let me check!

mossroy:
Maybe I did not update the branch properly, I'll check too

mossroy:
Sorry, it was certainly my mistake.
I'll test again with your updated branch

Jaifroid:
OK, thanks @mossroy . Yes, I had rebased the branch precisely because of that issue, but the comment about that is buried somewhere above.

PR #592, opened 2020-02-04
Fix missing footer and footer style with light contentTheme

### LINKED ISSUE #591: Missing footer toolbar when switching themes using Show applied theme link
This is a minor bug (regression) I've accidentally introduced with #582 . It can easily be corrected with a small PR. It only happens when switching themes and using the provided "[ Show article with applied theme ]" link. The issue is that the footer toolbar is not shown again (after it was hidden in Config/About pages). If the user presses Home instead, the bug is not present (footer is correctly shown). This bug is a one-line fix.

There is an associated minor issue: if the user selects to have the app in dark mode, but not invert content, then the footer toolbar is almost invisible (though it is present). This is also a simple fix that involves adding (probably) one css rule.

I am sorry I somehow missed these despite testing #582 extensively on different browsers. Strange, because it's almost the first thing I noticed when trying out master after merging...

Jaifroid:
Fixes #591. Because this PR is essentially a missing piece of #582, I will merge it on Friday 7th if there are no observations raised.

### DIFF
--- a/www/css/app.css
+++ b/www/css/app.css
@@ -93,7 +93,7 @@
 /* App theme: dark */
 
 .dark header, .dark #cachingAssets, .dark #about, .dark #configuration, .dark #welcomeText, .dark #articleListWithHeader,
-.dark #alertBoxHeader, .dark footer, .dark img {
+.dark #alertBoxHeader, .dark footer:not([class=_light]), .dark img {
     filter: invert(1) hue-rotate(180deg);
 }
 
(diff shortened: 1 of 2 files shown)


Jaifroid:
Tested on FFOS simulator, Edge Legacy and Edge Chromium (latter in SW mode as well as jQuery). Tested also on IE11, but as noted previously it lacks required inversion filters.

Issue #184, opened 2016-06-27
Race conditions in title search, leading sometimes to wrong results, or results displayed on top of article


mossroy:
The title search is launched automatically while the user is typing. It's an easy way to implement suggestions, but can lead to race conditions.
I've faced already two different situations : 
- you launch the app, quickly type a few letters, wait one second (it launches a search), then type a few more letters (it launches a second search). If the first search takes more time than the second one, its results might replace the other ones, which is not what you would expect : you always want the latest results, corresponding to what is typed. This happens more often with the cache optimization from #183
- I don't have a precise scenario, but if you choose an article while another search is still running, its results might be displayed on top of the article.

This can be reproduced relatively easily on a smartphone, but not on desktop (a computer is more powerful : the searches are finished much quicker so these scenario become very unlikely)

Jaifroid:
I've noticed something like this on the Windows Mobile version too. I wonder if we can introduce a minor delay in starting the search if the user has typed less than three characters (or maybe four). This wouldn't interfere with people who want to search for letters of the alphabet or for two-lettered words, the delay would be very small, but it would stop the app getting into a tizzy and searching through the index for a single character which is mostly not what the user wants. (I write this comment without having looked at the relevant code.)

sharun-s:
I haven't looked at this code recently, but I have MAX_SEARCH_RESULT_SIZE set to 20 on my local repo just to speed things up a little. Once the binary search cache, as discussed in #130, is in place this process should speed up and also be more controllable (as in being able to terminate the request midway).

mossroy:
This simply comes from the fact that we populate the screen asynchronously.
If I do action A, and it takes a long time to end, I have the time to do action B, and the result of both actions are displayed on the screen.
An approach would be to keep a kind of state for the screen (in a global javascript variable) that stores which action is the current one. And each asynchronous result should first check if it is still the current one : if it's not, the result should be discarded.

I did not investigate that much on that issue, as it is somehow a consequence of the poor performance we have. If we manage to have a significant performance gain, it might become useless (or a minor issue that occurs very rarely).

sharun-s:
> An approach would be to keep a kind of state for the screen (in a global javascript variable) that stores which action is the current one.

You have got the right idea. This is what the makeIterator/startChain code in my worker does. It provides a way to capture the state and control the rate of async task execution.

mossroy:
If someone wants to work on this, it would be worth starting with the work already done in #593

PR #593, opened 2020-02-18
Add prefix checking before displaying list of matched articles

zengchu2:
Attempt to address the issue described in issue #184 .

Add an early return if either the follow conditions holds:

- The search prefix input box is empty. (Which prevents loading search result after redirecting)
- The dirEntry array is non-empty and is not the result from latest prefix search. (Will handle the empty array in the origin code)

### DIFF
--- a/www/js/app.js
+++ b/www/js/app.js
@@ -82,8 +82,9 @@ define(['jquery', 'zimArchiveLoader', 'util', 'uiUtil', 'cookies','abstractFiles
     var globalDropZone = document.getElementById('search-article');
     var configDropZone = document.getElementById('configuration');
     
-    // Unique identifier of the article expected to be displayed
-    var expectedArticleURLToBeDisplayed = "";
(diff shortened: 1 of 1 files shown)


zengchu2:
Thanks for the hints in addressing this issue. 

I managed to use the `latestUserAsynAction` as the state to store user's latest action that might cause false displayed. I compare the state to perform early return in `readArticle` and `searchDirEntriesFromPrefix` since these are the two callbacks that may cause false displays. 

The only problem I found while testing is after searching and clicking an article, the search result might come back too fast and block the action of going to the clicked article but I am assuming this rarely happens.

mossroy:
@zengchu2 : do you intend to finalize this PR?

kelson42:
No feedback from the developer. I think this PR will have to be closed or finished by someone else.

mossroy:
Closing this PR. The branch should be kept because I think the implementation was on the right track anyway

PR #599, opened 2020-03-10
Remove vertical scrollbar (for unit tests)

Jaifroid:
This is a test PR, a duplicate of #572, merely for the purposes of Unit Testing and CI.


Jaifroid:
OK, builds have passed on all browsers. Though we don't test for the underlying issue here, it's re-assuring that the timeouts don't seem to be causing any problems.

PR #605, opened 2020-03-23
Extraction of Images #173

Hardik500:
@Jaifroid I did the changes which @mossroy mentioned in https://github.com/kiwix/kiwix-js/issues/173#issuecomment-529898051.

### DIFF
--- a/service-worker.js
+++ b/service-worker.js
@@ -37,6 +37,11 @@ var CACHE_NAME;
  */
 var useCache = true;
 
+/**
+ * A global Boolean that governs whether images are displayed
+ * app.js can alter this variable via messaging
+ */
+var imageDisplay;
 /**  
  * A regular expression that matches the Content-Types of assets that may be stored in CACHE_NAME
  * Add any further Content-Types you wish to cache to the regexp, separated by '|'
@@ -59,11 +64,11 @@ var regexpExcludedURLSchema = /^(?:chrome-extension|example-extension):/i;
  */
 var regexpZIMUrlWithNamespace = /(?:^|\/)([^\/]+\/)([-ABIJMUVWX])\/(.+)/;
 
-self.addEventListener('install', function (event) {
+self.addEventListener('install', function(event) {
     event.waitUntil(self.skipWaiting());
 });
 
-self.addEventListener('activate', function (event) {
+self.addEventListener('activate', function(event) {
     // "Claiming" the ServiceWorker is necessary to make it work right away,
     // without the need to reload the page.
     // See https://developer.mozilla.org/en-US/docs/Web/API/Clients/claim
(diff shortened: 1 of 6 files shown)


Jaifroid:
Thanks @Hardik500 , I'll take a look!

service-worker.js
```diff
@@ -73,31 +78,45 @@ self.addEventListener('activate', function (event) {
...
         // The ServiceWorker will handle this request either from CACHE_NAME or from app.js
 
+        // If the user has disabled the display of images, and the browser wants an image, respond with empty SVG
+        // A URL with "?kiwix-display" query string acts as a passthrough so that the regex will not match and
+        // the image will be fetched by app.js  
+        // DEV: If you need to hide more image types, add them to regex below and also edit equivalent regex in app.js
+        if (imageDisplay !== 'all' && /(^|\/)[IJ]\/.*\.(jpe?g|png|svg|gif)($|[?#])(?!kiwix-display)/i.test(event.request.url)) {
+            var svgResponse;
+            if (imageDisplay === 'manual')
+                svgResponse = "<svg xmlns='http://www.w3.org/2000/svg' width='1' height='1'><rect width='1' height='1' style='fill:lightblue'/></svg>";
```
Jaifroid: @Hardik500 I can't see that this blue fill is ever used.

www/css/app.css
```diff
@@ -260,11 +298,11 @@ button {
...
     /* Reduce size of the bottom buttons */
-    .btn-lg, .btn-group-lg > .btn {
+    .btn-lg,
+    .btn-group-lg>.btn {
         padding-top: 1px !important;
         padding-bottom: 1px !important;
         font-size: 16px !important;
     }
-}
+}
```
Jaifroid: @Hardik500 You seem to have done a whole load of style changes in CSS which have nothing to do with this PR! I suggest you revert all style changes on this file, and if the file hasn't changed in any meaningful way, don't include it in the PR.

www/js/app.js
```diff
@@ -26,1533 +26,1559 @@
 // This uses require.js to structure javascript:
 // http://requirejs.org/docs/api.html#define
 
-define(['jquery', 'zimArchiveLoader', 'uiUtil', 'cookies','abstractFilesystemAccess','q'],
```
Jaifroid: @Hardik500 It's really hard to see what's changed and what hasn't in this file, because it seems as if everything has changed. This is because of the whitespace changes, I think. You need to re-do the PR so that only the actual code changes show up in the diff.

www/js/lib/images.js
```diff
@@ -0,0 +1,183 @@
+/**
```
Jaifroid: @Hardik500 There seems to be a lot of unused code in this file!

Jaifroid:
@Hardik500 Thanks very much for this PR. There are a few issues:

It's really difficult to review this PR because you've changed a lot of lines that are unrelated to the functionality of the PR. There are a lot of whitespace changes, and in the GitHub diff, it looks as if the entirety of app.js has changed. Could I please ask you to re-do the PR and make changes only to lines that have actual code changes? Otherwise it's impossible for me to read what you've done.

Additionally, there seems to be a lot of unused functionality from my previous code. You need to strip out what you are not using. Just as an example, the code to substitute the image with a light blue svg is still in service-worker.js, but it is never used as far as I can tell (I never did see the blue placeholders).

Finally, in terms of functionality, ideally we need to make it possible for the user to return to the same page when they turn on images. At the moment, the user is forced to go back to the home page and then search for their page again. Since you throw up a warning message, you could include a link to go back to the page, like we do with the theme code (see screenshot: `[ Show article with applied theme ]`.

image

Hardik500:
Ok @Jaifroid the lines must have changed because of prettier. I will try to redo the PR with the changes you have mentioned.

Jaifroid:
@Hardik500 Yes, if you used a prettifier, then it will change lots of things and make the PR impossible to review.  Take a look at the link circled in this screenshot:
image
(you'll find that at the top of this page) - you'll see the problem. What we want is **only** the code that has meaningfully changed to be showing under the "Files changed" link.

Hardik500:
As for the link back to the user page, can I link it to the same function of the theme using 
`uiUtil.applyAppTheme(cookies.getItem('appTheme'));` or should I create a different function for it.

Jaifroid:
@Hardik500 It's only something like this code from `uiUtil.js` and the function it runs - `showReturnLink()` -- that you need to use:
```
        // If we are in Config and a real document has been loaded already, expose return link so user can see the result of the change
        // DEV: The Placeholder string below matches the dummy article.html that is loaded before any articles are loaded
        if (document.getElementById('liConfigureNav').classList.contains('active') &&
            doc.title !== "Placeholder for injecting an article into the iframe") {
            showReturnLink();
        }
```
However, I don't think you necessarily want to show the same return link. It might be best to have one embedded in your red warning panel.

Hardik500:
@Jaifroid Of course I will embed the link the link the red warning panel. However the issue still arises in that function it doesn't reload the webpage and the image settings won't have effect until the reload.
So how do I tackle that?

Jaifroid:
@Hardik500 I would suggest that in Service Worker mode, you simply reload the page. This is easy in SW mode because all you need to do is to hide Config and display the iframe, then navigate the iframe to the URL (like any internal hyperlink on a page).

In jQuery mode, you don't need to reload the page, and it's better not to from a performance perspective. In that mode,  you just need to hide Config and show the iframe contents, then run the image parsing/loading routine: `loadImagesJQuery()`.

Hardik500:
@Jaifroid I tried implementing what you told me, and I am facing some issues. First regarding the Jquery one where I am supposed to call the `loadImagesJQuery()` function which is inside another function `displayArticleContentInIframe()`, so I tried calling these functions but can't seem to get them to work. Also for the SW I wasn't to reload the page using the same way as that of other URL's  so instead I wrote `history.back(); history.forward();` instead and can't think of a better way than this.
So could you help me out a bit again.

Jaifroid:
@Hardik500 It's hard to advise without seeing the code you've tried in a PR. Could you submit, either to this branch or to a new PR (you could close this one, especially as there are conflicts now in this branch)?

Hardik500:
@Jaifroid Ok I will close this one and create a new PR.

Issue #604, opened 2020-03-22
Update application logo


Popolechien:
There is a new, slightly remodeled logo for Kiwix 
https://commons.wikimedia.org/wiki/File:Kiwix_logo_v3.png
The app's icon needs to be updated accordingly

ykabusalah:
Can I try this?

Jaifroid:
@ykabusalah Yes, please do have a go.

ykabusalah:
I have submitted a solution for this through #607 .

FalakZ:
@Jaifroid Hello, I tried making a new logo for KIWIX .


![logo](https://user-images.githubusercontent.com/46991207/77821359-c0b8b500-710f-11ea-8306-a77d789a1a49.png)

FalakZ:
Hello @Jaifroid ,
I created a logo for kiwix. Can you check it out.


![ff37d763-33ca-46ed-a412-ab119debc5ac_200x200](https://user-images.githubusercontent.com/46991207/77821833-3b370400-7113-11ea-8342-0df9e987ab0c.png)

Jaifroid:
@FalakZ Nice! But I'm just a dev. I don't know if Kiwix Association is interested in a new logo, given that they have their own logos with wide recognition. Pinging @kelson42 in case Kiwix Association might be interested.

Popolechien:
@FalakZ Hi, thank you very much but we already have an updated version as indicated in the first issue comment - we are just making sure it's consistent across all apps (see [here ](https://github.com/kiwix/kiwix-android/issues/1943) and [here](https://github.com/kiwix/kiwix-desktop/issues/388)). Maybe I should have been a little clearer, sorry about that.

mossroy:
Implemented by #608

PR #607, opened 2020-03-25
fix issue #604

ykabusalah:
I proposed this solution for fixing issue #604 . Thank you for your time.


ykabusalah:
This doesn't seem to be an issue with my code, correct?

ykabusalah:
The other pull requests seem to fail for the same reason as this one for the fact that saucelabs is currently down. Are there any updates on that situation?

Jaifroid:
Don't worry, it's because this PR was made from an external repository. What we can do is push a branch from one of our repositories with the identical commits, and that will allow the tests to run. But the code can still be reviewed.

ykabusalah:
Alright. Do I have to do anything for that?

PR #608, opened 2020-03-26
fix issue #604

ykabusalah:
There was an issue with my previous pull request, so I decided to make a new one and put out my solution.


ykabusalah:
@mossroy I believe I am replacing the files with the new logo named the same so I would not need to do any code changes, not just deleting them. 

Would you like me to rename and do code changes?

https://github.com/ykabusalah/kiwix-js/tree/master/www/img/icons

Screen Shot 2020-03-28 at 11 44 19 AM
Screen Shot 2020-03-28 at 11 44 07 AM

mossroy:
Thanks for your contribution

PR #609, opened 2020-03-27
Image Extraction #173

Hardik500:
@Jaifroid I have made the changes which I mentioned in #605. Have a look at it.

### DIFF
--- a/www/js/app.js
+++ b/www/js/app.js
@@ -103,22 +103,11 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil','images', 'cookies','abstractFile
      * Resize the IFrame height, so that it fills the whole available height in the window
      */
     function resizeIFrame() {
-        var headerStyles = getComputedStyle(document.getElementById('top'));
-        var iframe = document.getElementById('articleContent');
-        var region = document.getElementById('search-article');
-        if (iframe.style.display === 'none') {
-            // We are in About or Configuration, so we only set the region height
-            region.style.height = window.innerHeight + 'px';
-        } else { 
-            // IE cannot retrieve computed headerStyles till the next paint, so we wait a few ticks
-            setTimeout(function() {
-                // Get  header height *including* its bottom margin
-                var headerHeight = parseFloat(headerStyles.height) + parseFloat(headerStyles.marginBottom);
-                iframe.style.height = window.innerHeight - headerHeight + 'px';
-                // We have to allow a minimum safety margin of 10px for 'iframe' and 'header' to fit within 'region'
-                region.style.height = window.innerHeight + 10 + 'px';
-            }, 100);
-        }
+        var height = $(window).outerHeight()
+                - $("#top").outerHeight(true)
+                // TODO : this 5 should be dynamically computed, and not hard-coded
+                - 5;
+        $(".articleIFrame").css("height", height + "px");
     }
     $(document).ready(resizeIFrame);
     $(window).resize(resizeIFrame);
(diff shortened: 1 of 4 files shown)


www/js/app.js
```diff
@@ -1034,6 +1076,7 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'cookies','abstractFilesystemAcc
      * @param {DirEntry} dirEntry The directory entry of the article to read
      */
     function readArticle(dirEntry) {
+        cookies.setItem("lastReadDir", JSON.stringify(dirEntry), Infinity);
```
Jaifroid: @Hardik500 I would call this cookie item "lastReadDirEntry" to be clearer on what is being stored.

www/js/app.js
```diff
@@ -1132,6 +1186,8 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'cookies','abstractFilesystemAcc
                         selectedArchive.readBinaryFile(dirEntry, function (fileDirEntry, content) {
                             var mimetype = fileDirEntry.getMimetype();
                             // Let's send the content to the ServiceWorker
+                            var message = { 'action': 'giveContent', 'title' : title, 'content': content.buffer, 
+                                'mimetype': mimetype, 'imageDisplay': params.imageDisplayMode };
                             var message = { 'action': 'giveContent', 'title': title, 'content': content.buffer, 'mimetype': mimetype };
```
Jaifroid: @Hardik500 Why is `message` defined twice here? Clearly defining it the second time will overwrite the first definition.

Jaifroid:
@Hardik500 Could you kindly explain where you are at with this PR? On a quick test, I could not get the switch to work in jQuery mode. When I toggle the setting in the screenshot, and click on the return link, nothing happens on the first click. On the second click, I get returned to the article, but the images do not show.

image

Jaifroid:
Additionally, your branch is behind master. You should rebase it, as it has conflicts with master.

Hardik500:
> @Hardik500 Could you kindly explain where you are at with this PR? On a quick test, I could not get the switch to work in jQuery mode. When I toggle the setting in the screenshot, and click on the return link, nothing happens on the first click. On the second click, I get returned to the article, but the images do not show.
> 
> image

@Jaifroid  That is what I mentioned in https://github.com/kiwix/kiwix-js/pull/605#issuecomment-604512532. Apparently I am not able to make a call to `loadImageQuery` as you mentioned.
As for the other issues I will try to fix them.

Jaifroid:
@Hardik500 Looking at this, I don't see why `loadImagesJQuery()` couldn't be extracted and placed in a higher-level scope. You might have to pass a needed variable or two when calling it, or else reconstruct the missing variables in the function (e.g. `iframeArticleContent` can easily be defined internally in the function).

Hardik500:
@Jaifroid I will try to apply what you suggested. Although I tried it with passing the needed variables. But let me give it a few more tries.

Jaifroid:
@Hardik500 You will need to move the function to a different scope (outside `displayArticleContentInIframe()`) to be able to call it from a different context.

Hardik500:
@Jaifroid I have made the changes as you suggested. I had to remove the blue cover for images for Jquery mode. As since the height was fixed so the images that would load, will also have the same height. As for SW one I still can't think of a proper way to reload a page, so as for now I modified the reload by modifying the history.

kelson42:
@Jaifroid This is a really old PR and I believe @Hardik500 has never got the feedback he was expecting. What is the status?

Jaifroid:
This was never passing tests, and needed quite a lot more work especially on the Service Worker side. I'm not sure if the OP still wants to work on it. It would need to be rebased and fixed, then I'd be happy to provide feedback if there are remaining issues.

kelson42:
@Jaifroid I guess @Hardik500 does not want to finish this PR,

Jaifroid:
I think this is now hopelessly out of date, given changes in the codebase. I'll close for now.

PR #610, opened 2020-03-29
Update changelog for v2.7.0

Jaifroid:
@mossroy Please note suggested new changelog format (proposed to use only for this and future releases) as suggested by @kelson42 in https://github.com/kiwix/kiwix-js-windows/issues/29 

Do you like this?

Milestone closed log is here:
https://github.com/kiwix/kiwix-js/milestone/21?closed=1 
but note a lot of issues there were closed without fixes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ Then it was renamed Kiwix-html5 (and uses ZIM files), and then again was renamed
 ## Kiwix-JS v2.7.0
 Released on *2020-03-29*
 
-* NEW: Selectable dark mode for app UI and for articles
+* NEW: Selectable dark mode for app UI and for articles (does not work on IE11)
 * NEW: Optional transition between app pages
 * NEW: Native asset caching in Service Worker mode - considerably speeds up article load times
 * UPDATE: Privacy Policy
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,9 +5,22 @@ It was first called Evopedia (and was using the file format of Evopedia).
 Then it was renamed Kiwix-html5 (and uses ZIM files), and then again was renamed to Kiwix-JS.
 
 ## Kiwix-JS v2.7.0
-Released on TODO
+Released on *TODO*
 
-Detailed changelog : TODO
+ENHANCEMENT: Selectable dark mode for app UI and for articles
+ENHANCEMENT: Optional transition between app pages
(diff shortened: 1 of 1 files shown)


mossroy:
It's cool.
But you need to add one empty line between each line, else the Markdown renderer puts everything on a single line. See https://github.com/kiwix/kiwix-js/blob/Update-changelog-for-v2.7.0/CHANGELOG.md
The prefix are a bit different in https://github.com/kiwix/kiwix-android/blob/master/CHANGELOG : maybe you could use NEW instead of ENHANCEMENT?
I'm wondering if we should add these prefixes on the previous versions too. It's not crucial.
Maybe you can put today as the 2.7 release date?

Jaifroid:
Happy to make those changes.

Jaifroid:
Bullet points neater than extra line?

image

Jaifroid:
And I just pushed a sample of formatting the previous entry. I suggest we do in the order: NEW, UPDATE, BUGFIX. This means re-ordering some previous entries. (I won't do any more until we've agreed the format.)

Jaifroid:
Also, not sure whether to add fix to "2.6.0 does not run properly on Ubuntu Touch" bug, because I'm not sure we did anything to fix that, or if the platform was updated.

mossroy:
It looks great.
Don't add a line about Ubuntu Touch : we did not change anything

Jaifroid:
I've updated the format as far down as I have been involved in this project (and down to just after the name change to Kiwix JS). Before that, I can probably guess as to what was new and what was an update, but I'm not so familiar, and maybe not necessary to go that far back, or can be left for a future update focusing on documentation?

Jaifroid:
OK, done. I used "most contexts" rather than "some devices", because I think the Chrome extension can be run on more than one device type (e.g., PCs, Macs).

mossroy:
To be precise, I don't think "most contexts" is true.
For now, the main context for kiwix-js (excluding kiwix-js-windows) is Firefox and Chrome browsers extensions. And both are not affected by this enhancement : Firefox extensions do not support SW mode, and Chrome extensions use the built-in browser cache

Jaifroid:
Oh, sorry! I was a bit hasty in merging the commit, but can do a single update as an extra commit to master. I don't think it's about devices, however, which is why I changed it. So I think we can say "some contexts"?

mossroy:
No problem. Yes, "some contexts" seems better IMHO

Jaifroid:
OK, will commit that straight to master.

Issue #612, opened 2020-04-04
Can Kiwix remember user settings of Dark theme?


alextooter:
I like the Dark theme in v2.7,it's beautiful.After I close Kiwix page,then reopen it,Kiwix not remember my choose of Dark theme.

Can we store this option in somewhere,don't let me choose everytime?

Jaifroid:
Hi @alextooter, it should be able to. Which platform are you using, so I can trobuleshoot?

alextooter:
> Hi @alextooter, it should be able to. Which platform are you using, so I can trobuleshoot?
I am using macOS 10.15.4,Firefox 74.0.1,kiwix 2.7.0
My way of using kiwix:click the icon on toolbar,kiwix open with light theme,then I choose Dark+invert(wikimedia*),it turn to dark theme.Close this tab,open it,using Light theme again.

Jaifroid:
OK, thanks, @alextooter . I presume that's in the Firefox extension?

Jaifroid:
@alextooter I reproduce the bug in Firefox Extension on Windows 10. However, it is not just with this feature, it is with any feature that stores its settings in a cookie. Something must have been deprecated. Very sorry not to have picked this up in testing.

@mossroy can you reproduce? In the Firefox extension, try changing any setting in Config that should persist between tab sessions, and they are lost the next time the Kiwix JS tab is opened.

Jaifroid:
Proof that cookies are disabled in Firefox browser extension:

In the extension tab devtools type `document.cookie="username=John Doe"` and then `document.cookie` Result:

![image](https://user-images.githubusercontent.com/4304337/78437020-b433df80-766e-11ea-8ccd-dc844aab6ee2.png)

In a normal tab opening `file:///.../kiwix-js/www/index.html`:

![image](https://user-images.githubusercontent.com/4304337/78438036-fe1cc580-766e-11ea-8437-43337b9ef247.png)

Jaifroid:
Looking at https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Work_with_the_Cookies_API, I see we should have a `permissions` entry in our `manifest.json` to use the Cookies API in Firefox Extensions:

![image](https://user-images.githubusercontent.com/4304337/78450980-75a42200-767a-11ea-80b4-fb3be07f5c47.png)

We don't have this.

mossroy:
[...]

PR #613, opened 2020-04-04
Support localStorage for storing settings

Jaifroid:
This is a possible solution for #612 in contexts where the `localStorage` API is supported. It works in Firefox running from the file:// URL. I haven't tested in extension context. It should also enable Chromium to store settings from file:// context when the flag allowing file access is used.

This solution has been implemented in Kiwix JS Windows for a year or so, to solve some other issue that I forget, so I simply copied it over here to test.

How can I test this in an extension on Firefox?


Jaifroid:
Note on how this works:

`cookies.js` runs on app launch and returns a cookies object with various methods. I have added code at the head of this file to test a) whether cookie support is available in the context and b) whether `localStorage` support is available. If `localStorage` is available, it is preferred. If it is not available, it falls back to using `document.cookie`.

The test only runs once on app launch, not every time a cookie method is used (you can see this in console.log).

The reason we have to prefer `localStorage` to solve the reported issue is that in file:// context, `document.cookie` appears to work and receive values, but does not remember them between tab launches. Therefore the provided test does not provide useful information. In extension context, the test will probably show that cookie support is disabled, as per the console.log test I did in #612.

mossroy:
In https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/storage, Mozilla says : 

> it is recommended that you don't use Window.localStorage in the extension code to store extension-related data. Firefox will clear data stored by extensions using the localStorage API in various scenarios where users clear their browsing history and data for privacy reasons, while data saved using the storage.local API will be correctly persisted in these scenarios

They suggest to use this https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/storage API instead, that should work in other browser extensions too.

We have to keep in mind that, if we use these other storage APIs AND set them as default storage in some circumstances, we'll have to deal with data migration (move the cookie preferences to the other storage, then remove the cookies), and to deal with cases where some preferences would be set in different places (and decide which one we should use). Keeping cookies as the default preference storage would avoid this difficulties

Jaifroid:
> Firefox will clear data stored by extensions using the localStorage API in various scenarios where users clear their browsing history and data for privacy reasons, while data saved using the storage.local API will be correctly persisted in these scenarios

That's exactly the same scenario as cookies. If a user clears browsing data cookies will be clear. In fact I think it's an advantage that localStorage would be cleared in the same circumstances, as it allows the user to "reset" all settings to default (say, where a particular setting is causing an issue due to an API change).

Jaifroid:
I think we should respect a user's desire to clear their data for privacy reasons, and shouldn't try to circumvent that by using a more persistent API.

Jaifroid:
PS Can easily add code to migrate settings to localStorage if supported and document.cookie is accessible. On the other hand, a user might not be too annoyed by having to choose two or three settings again in Config. Still, it's a few lines of code to migrate either way. Would you like me to add?

mossroy:
OK, you're right. Our extension is more like an offline website with some preferences than an extension where we would need to keep settings even when the user clears the history/private data.
So localStorage could be OK.
And yes, it might not be worth adding too much code to migrate so few settings.
If we're lucky this would be in the same released version as #196

Jaifroid:
Yes, I agree we should do #196 straight after. #196 would need some migration code of its own, because we will need to update the stored setting to Service Worker on a one-time basis, and let users know we've done it.

Jaifroid:
Tested and working on Firefox OS. It uses `localStorage`.
Not working on IE11 from file:// protocol. It thinks `localStorage` is available, but attempting to write to it causes an exception. Therefore the test needs to attempt to write to `localStorage` to be sure it's supported.

Jaifroid:
* Now fixed for IE11. It uses cookies from file:// protocol, and localStorage when running from localhost.
* Works in Chromium (Edge). It uses localStorage when running from localhost.
* Works in Edge legacy. It uses cookies from file:// protocol, and localStorage when running from localhost.

Jaifroid:
Tested and working in Firefox 73.0.1 in an Extension installed using tools in `about:debugging`.

image

`Almacenamiento local` is Spanish for `Local Storage`. You can see the keys stored bottom right.

Jaifroid:
I've just realized that if localStorage is preferred but it's not supported, there is no point checking for cookie support, because we don't have any other option, we have to fall back to something, and the most universal is likely to be cookie. Therefore logic dictates that we remove the test for cookie support and just assume it. If a cookie test were to fail, there'd be no other storage type to use anyway. In that hypothetical situation (no localStorage support and no cookie support), defaults will be used and the user will have to re-do settings between sessions.

www/js/lib/cookies.js
```diff
@@ -21,11 +21,43 @@ define([], function() {
...
+      localStorage.setItem('kiwixStorage', '');
+      localStorage.removeItem('kiwixStorage');
+    }
+  } catch (e) {
+    localStorageTest = false;
+    console.log('LocalStorage is not supported!');
+  }
+  // Fall back to cookie if localStorage fails
+  if (!localStorageTest) type = 'cookie';
+  console.log('Storage test: type: ' + type);
```
mossroy: Maybe it would be better to report this status in the Settings section of the app (just like the other APIs status).
I realize it's more complicated : it might be a separate enhancement github issue if you prefer.

www/js/lib/cookies.js
```diff
@@ -21,11 +21,43 @@ define([], function() {
...
+  return type;
+}
+
 var docCookies = {
```
mossroy, line 51: I think this variable and the file itself (with its header) should be renamed to something more generic. Maybe `settingsStorage`?

www/js/lib/cookies.js
```diff
@@ -21,11 +21,43 @@ define([], function() {
...
+  // DEV: In FF extensions, cookies are blocked since at least FF 68.6 but possibly since FF 55 [kiwix-js #612]
+  var type = 'local_storage';
+  // First test for localStorage support
+  var localStorageTest = false;
+  try {
+    localStorageTest = 'localStorage' in window && window['localStorage'] !== null;
+    // DEV: Above test returns true in IE11 running from file:// protocol, but attempting to write a key to
+    // localStorage causes an exception; so to test fully, we must now attempt to write and remove a test key
+    if (localStorageTest) {
+      localStorage.setItem('kiwixStorage', '');
```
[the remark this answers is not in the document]
Jaifroid: Done.

www/js/lib/cookies.js
```diff
@@ -21,11 +21,43 @@ define([], function() {
...
+      localStorage.removeItem('kiwixStorage');
+    }
+  } catch (e) {
+    localStorageTest = false;
+    console.log('LocalStorage is not supported!');
```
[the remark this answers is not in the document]
Jaifroid: Deleted.

www/js/lib/cookies.js
Jaifroid: I can do this.

www/js/lib/cookies.js
Jaifroid: @mossroy I can do this, but let me check that you mean that throughout the app where "cookies" are set, we should in now use, instead of `cookies.getItem('mySetting')`, something like `settingsStorage.getItem('mySetting')`?

Jaifroid: Note to self: in doing this, I think we should include #438 (i.e. remove deprecated `escape` and `unescape`, using `encodeURIComponent` etc. instead).

mossroy: Yes, when we mention setting/reading cookies in the app, we should instead talk about setting/reading some stored settings.
And yes, it would be the right time to implement #438, I had forgotten this one

www/js/lib/cookies.js
Jaifroid: Done.

www/js/lib/cookies.js
Jaifroid: I've added the latest Mozilla framework code (albeit from 2014) that fixe #438.

Jaifroid:
@mossroy Please note that to be able to see and compare the changes to `cookies.js` it will be necessary to look at commit https://github.com/kiwix/kiwix-js/pull/613/commits/6394eae890b001cd06aad56dd4bba52a1fb934fc, because the next commit will rename the file and then everything will look as if it's changed.

Jaifroid:
Latest commit introduces 9 cofefactor issues in the Mozilla code, all of them unnecessary escape characters in regex character classes (a common error). 

image

Should I correct these? It is Mozilla code, not ours.

www/js/lib/cookies.js
Jaifroid: Done.

Jaifroid:
NB do not review again until I've logged all the manual tests I do here as successful. A lot has changed since I did the tests above, so they need to be re-done.

Jaifroid:
## Tests:

* Edge Legacy from file:// protocol uses Cookie
* Edge Legacy from localhost uses Local Storage
* Firefox 74.0.1 from file:// protocol uses Local Storage
* Firefox 74.0.1 from local host uses Local Storage 
* Firefox Extension uses Local Storage
* Firefox ESR from file:// protocol uses Local Storage
* Firefox ESR from localhost uses Local Storage
* Firefox OS uses Local Storage
* Edge Chromium from file:// protocol uses Local Storage
* Edge Chromium from localhost uses Local Storage
* IE11 from file:// protocol uses Cookie
* IE11 from localhost uses Local Storage

**Chrome Extension untested**

www/js/lib/settingsStore.js
```diff
@@ -0,0 +1,129 @@
...
+    // Now test for document.cookie API support
+    document.cookie = 'tempKiwixCookieTest=working;expires=Fri, 31 Dec 9999 23:59:59 GMT';
+    var kiwixCookieTest = /tempKiwixCookieTest=working/.test(document.cookie);
+    // Remove test value by expiring the key
+    document.cookie = 'tempKiwixCookieTest=;expires=Thu, 01 Jan 1970 00:00:00 GMT';
+    if (kiwixCookieTest) type = 'cookie';
+    // Prefer localStorage if supported due to some platforms removing cookies once the session ends in some contexts
+    if (localStorageTest) type = 'local_storage';
+    // Update API panel with API name
+    var settingsStoreStatusDiv = document.getElementById('settingsStoreStatus');
```
mossroy, line 54: The settingsStore.js file must not access the DOM. It's not its responsibility : it should only deal with preferences storage.
What could be done instead is expose a function here, that returns the storeType variable. This function could be called by app.js in the existing refreshAPIStatus function, where it would display what we want.

www/js/lib/settingsStore.js
```diff
@@ -0,0 +1,129 @@
...
+    var apiName = type === 'cookie' ? 'Cookie' : type === 'local_storage' ? 'Local Storage' : 'None';
+    settingsStoreStatusDiv.innerHTML = 'Settings Storage API in use: ' + apiName;
+    settingsStoreStatusDiv.classList.remove('apiAvailable', 'apiUnavailable');
+    settingsStoreStatusDiv.classList.add(type === 'none' ? 'apiUnavailable' : 'apiAvailable');
+    // To simplify code below, this function returns either 'local_storage' or 'cookie', but does not return 'none'
+    // This is because setting a cookie key does not cause an exception even if it is not stored (e.g. in an Extension)
```
mossroy, line 60: I don't understand this comment.
If `kiwixCookieTest==false` and `localStorageTest==false`, this function will return `none` (even if we don't know any browser that would behave like that)

www/js/lib/settingsStore.js
Jaifroid: Ah OK! No problem. Will put the code into app.js and get the value from settingsStore.js.

www/js/lib/settingsStore.js
Jaifroid: Ah yes, sorry, this comment was left over from a previous implementation and I must have thought it was still returning just the two values. I'll modify the tests below to ensure that the cookie code is used even if the storeType is 'none'. I haven't tested Chromium extension yet, so don't know how that behaves. I think we don't know how much more locked down cookies will become in the future. Already some browsers have recently stopped them working using the file:// protocol whereas they used to, and I can imagine scenarios where they are locked down for non-secure protocols like http:// as well.

www/js/lib/settingsStore.js
Jaifroid: Done.

Jaifroid:
Need to re-do tests before final review. Will log here when I've done the tests.

Jaifroid:
Quick question, @mossroy: Does it matter that if we store a truthy value as a setting, settingsStore.js will return 'true' (string) if it was stored in a cookie, whereas it will return `true` (Boolean) if it was stored in localStorage? It might be better if it returned exactly what we sent it, which would mean checking for strings 'true', 'false' and possibly 'null' (not sure about that one) and returning Booleans from the cookies part. Thoughts? Obviously, if changing it, I'd go through the points in app.js where we read a cookie value and ensure it's getting what it expects.

mossroy:
I'm not sure I understand what you're trying to achieve.
In any case, I think each function should only do what it's intended to do : getting/setting/removing an item, or giving the current storage API in use for example.
I think it would be very confusing to return 'true' (string) or true (boolean) in different cases.

mossroy:
On another subject, it might be not very difficult to do the preference migration, and would be a better user experience on upgrades.

It could be a new function in this file, run each time the application is started by app.js.
If the current storage type is cookie, it would do nothing.
Else it would check if our preferences cookies are set and, if some are, use `setItem` to set it with the current storage type, and remove the cookie.

If you prefer to move that into a different issue, it's fine with me.

Jaifroid:
> I'm not sure I understand what you're trying to achieve.

Sorry, I didn't explain clearly. I'm pointing out an issue with the code.

**Issue:** If sKey was stored with value `true` (Boolean), then Mozilla's `cookies.getItem(sKey)` will return 'true' (string), while `localStorage.getItem(sKey)` will return `true` (Boolean). So the code is inconsistent according to the store in use.

My question is, should I modify Mozilla's code to return Booleans if the value stored in the cookie is 'true' or 'false', so that it is consistent with the native behaviour of `localStorage.getItem()`? Personally, I would recommend doing this, rather than changing the native behaviour of `localStorage` methods.

If I do modify, I will check all the instances of `settingsStore.getItem()` in app.js to ensure the new behaviour is accounted for.

Jaifroid:
> On another subject, it might be not very difficult to do the preference migration, and would be a better user experience on upgrades.

I can do this. Presumably you think it's not necessary to move settings the other way (from localStorage to cookies)? I guess it is far less likely (or maybe impossible) that on any one platform we will have localStorage suddenly unavailable.

Jaifroid:
Please ignore comments about setting and retrieving Boolean values. After further testing, I found out that localStorage behaves exactly like cookies in converting Booleans to strings.

Jaifroid:
This is ready for testing. I have added console log messages so we can see what exactly is migrated (it shows just the keys that are migrated).

In the contexts I've tested quickly, the migration is one-off. However, in FFOS simulator, the following key is recreated in document.cookie every time the app starts:

- permanent/fake-sdcard/wikipedia_en_medicine_novid_2018-08.zim

This causes the migration to run on every startup of the app. Is this just an artefact of the simulator?

Jaifroid:
## Tests

* Firefox 75.0 from localhost: uses localStorage and migrates settings once only
* Edge Chromium from localhost: uses localStorage and migrates settings once only
* Firefox ESR 52.5 from localhost uses localStorage and migrates settings once only
* Firefox OS simulator in Firefox ESR uses localStorage but migrates settings on each app launch due to fake-sdcard appearing in document.cookie even after it is expired
* Firefox ESR from file:// protocol uses localStorage
* Edge Legacy from file:// protocol uses cookies (and does not migrate settings obviously)
* Edge Legacy from localhost uses localStorage and migrates settings once only
* IE11 from file:// protocol uses cookies
* IE11 from localhost uses localStorage and migrates settings once only

Methodology: first load the app using master branch to ensure document.cookie is populated. Then switch to Use-localStorage branch and reload app with console open to see the migration. Reload again to see if migration is repeated.

Jaifroid:
Now also tested in an Edge Chromium extension using developer mode. It uses localStorage and does not migrate settings because the extension loads with document.cookie empty. Settings are remembered between tab loads.

I think this is ready for your testing and review, @mossroy .

mossroy:
Just starting the review : data migration worked perfectly on my Firefox OS device : 

> Migrating Settings Store from cookies to localStorage...  settingsStore.js:126:5
> - lastContentInjectionMode  settingsStore.js:133:7
> - listOfArchives  settingsStore.js:133:7
> - lastSelectedArchive  settingsStore.js:133:7
> - imageDisplayMode  settingsStore.js:133:7
> - showUIAnimations  settingsStore.js:133:7
> - appTheme  settingsStore.js:133:7
> - hideActiveContentWarning  settingsStore.js:133:7
> Migration done.  settingsStore.js:135:5

and on Chromium 80 (desktop) too.

I noticed on Chromium that there were some other settings in the localStorage : 
image

The volume is the default volume of some videos (like the ones of dirtybiology).

So maybe we should add a prefix in our localStorage keys, to clearly distinguish them with the ones that might come from the ZIM contents, and avoid conflicts with them. It seems to be what MediaWiki does BTW. This prefix could be a constant, that we automatically add in the getItem/setItem/etc functions

Jaifroid:
> So maybe we should add a prefix in our localStorage keys, to clearly distinguish them with the ones that might come from the ZIM contents, and avoid conflicts with them.

Yes, I was thinking about that, as I'd seen some non-Kiwix settings being migrated too. I guess we can add a `kiwixjs-` prefix. But it doesn't solve the one-time migration problem, because the migration code won't find any keys prefixed with `kiwixjs-`, so won't migrate anything. The only solution to that is to add a `kiwixjs-migrated` key which would tell us that the migration has been run already, so we do not need to look for any non-prefixed keys. Note that we will then end up with spurious keys in localStorage like `kiwixjs-volume` (to use your example), but at least it wouldn't keep running the migration every time it finds `volume` in the cookies. But if we do have a `kiwixjs-migrated` key, is there any point in having the prefix? We would simply not run the migration if it's already been done.

The only way I can think of to be absolutely sure we don't migrate spurious keys would be to supply a list of known keys to migrate...

mossroy:
As you seem to intend at the end of your last comment, we know which cookies kiwix-js uses, so we could look for and migrate only these specific cookies, and do not touch the other ones. No need to add `kiwixjs-migrated` setting in this case.

Jaifroid:
OK, I've added the list of keys to migrate. I could only find seven in our app by searching for `settingsStore.setItem`. There is one in your list above from your FFOS device that is from a branch that has not been merged (manual image extraction / progressive image display), so I did not include that.

www/js/lib/settingsStore.js
```diff
@@ -0,0 +1,163 @@
...
+   *
+   *  * settingsStore.setItem(name, value[, end[, path[, domain[, secure]]]])
+   *  * settingsStore.getItem(name)
+   *  * settingsStore.removeItem(name[, path[, domain]])
+   *  * settingsStore.hasItem(name)
+   * 
+   */
+
+  /**
+   * An RegExp of the settings keys used in the cookie that should be migrated to localStorage if the API is available
```
Jaifroid, line 29: Correct typo "A RegExp".

www/js/lib/settingsStore.js
```diff
@@ -0,0 +1,163 @@
...
+    }
+    console.log('Migration done.');
+  }
+
+  return {
+    getItem: settingsStore.getItem,
+    setItem: settingsStore.setItem,
+    removeItem: settingsStore.removeItem,
+    hasItem: settingsStore.hasItem,
+    testStorageSupport: testStorageSupport
```
Jaifroid, line 161: Consider changing at least public function name to 'getStorageAPI()'?

mossroy: I would change the function name (public and private) to something like `getBestAvailableStorageAPI`.

Jaifroid:
Ah, OK, I had misunderstood about the prefix (I thought we no longer needed it as we were specifying which keys to migrate). But I agree that it makes sense just in case we get key collisions with code inside the ZIMs that might also use localStorage. Will do.

Jaifroid:
Done, and tested so far only on Firefox 75. Methodology for testing:

* Switch to master
* Load Kiwix JS from localhost
* Go to Storage settings in dev console and delete the current localStorage entirely (e.g. right-click on http://localhost under Local Storage and delete), because it will probably have old key names already in it
* Add some cookie settings by toggling various settings in the app and loading a ZIM
* Switch to Use-localStorage branch
* With console log open, fully reload the app
* Finally, open the Storage tab in devtools and check what is in Local Storage. Experiment by toggling some settings in the app, and at least in Firefox, the values should change.

We'll need to test this in several different browsers/contexts before merging in case of any incompatibilities with key names (shouldn't be a problem, but you never know).

Jaifroid:
Interesting. I always associate `const'  and `let` with ES6, and thought IE11 therefore didn't support them (like it doesn't support arrow notation). Interesting to see we could have been using some of the features. We actually have quite a few constants that could use this, though I guess the effect on memory usage/speed is tiny, it's more for readability. 

Happy to experiment with that.

Jaifroid:
I've implemented that and have tested on Firefox 75 with file:// protocol, and on IE11 from localhost. It works fine on the latter, i.e. `const` works as expected (which is good to know).

I'll do tests on the other platforms before squashing/merging (but not today).

Jaifroid:
Summary of final tests conducted:

FFOS simulator: working
Firefox 75 from file://: working
IE11 from localhost: working
Edge Chromium developer extension: working
Firefox temporary extension: working

So I'll squash and merge.

Issue #615, opened 2020-04-18
params.isLandingPage remains false when history is used to navigate


bhargeymehta:
To reproduce:
Open kiwix-js
Go to some article
Navigate to home page using back button (using history.back())

Issue:
Currently on landing page but params.isLandingPage remains 'false'

PR #616, opened 2020-04-18
speedread init

bhargeymehta:
Initial integration of the speed read feature. 

Note:
Suffers a bug where using history to navigate to home page causes the speed read button to appear on the landing page.
Can possibly be fixed when Issue ( https://github.com/kiwix/kiwix-js/issues/615 ) is fixed.


bhargeymehta:
@mossroy 

The zethos behaviour was still unreadable but fortunately I was able to replicate it so I built that myself. The speedread text now first finds a focus point and then does the appropriate padding around the word to keep the focus point always in the center and colored. 

I didn't change the speed of the text but it is as simple as changing a number in the configuration object so that can be worked upon later without much need to look at the core speed reading code.
 
The speedread button is now always present and clicking it now unhides a new bar with the (play/pause toggle now) buttons and the other controls.

mossroy:
Thanks a lot @bhargeymehta for these improvements. The UI is now much better IMHO.

On the other hand, I still don't manage to read faster with that. Do you?
If I use for example this Chrome extension on the same text, I do read faster : https://chrome.google.com/webstore/detail/spreed-speed-read-the-web/ipikiaejjblmdopojhpejjmbedhlibno
@Jaifroid : could you test and give us your opinion?

I'm not a speed-reading specialist, but I think it's not that easy to make it efficient.
It seems to me that, at least, hyphenation should be improved, and also that we lack some small pauses when there are punctuation marks. But probably some other techniques are involved to improve the result.
You did a great job in this implementation, but I think it will be too hard to re-implement and maintain such feature, so that it's really efficient.

Using a third-party library would be much better. But I understand it might be not so easy to find an efficient and maintained one, and to integrate it in our code.

Zethos would have been my first choice because it seemed easier to reuse the code, but I might be wrong. It might not be the most efficient one, though (because the aim was a 1k challenge).
https://github.com/ReedyReader/Reedy-for-Chrome could be another one. It seems to have more advanced features, but it might be more difficult to reuse their code.
Reasy seems to be another one (but it's now unmaintained, and I don't know how efficient it is) : https://github.com/robbiehinch/reasy
https://github.com/vicrazumov/React.Spritz might be another one, too.

mossroy:
@Jaifroid : could you give us your opinion on this PR?

Jaifroid:
Apologies, will test. I've tried speed-reading apps a few times for fun, but never found them faster than I can scan well-formatted lines and paragraphs! But I understand it's a skill that can be learnt, at least for text that is primarily informational, so could be useful in a Wikipedia-like context.

Jaifroid:
This is good work, congratulations!

But like @mossroy, I find the speed pretty slow and I think it should be possible to increase/slow it by pressing up or down arrow keys, and touch equivalent (a couple of touch-size buttons), when the reading pane is focused.

Another problem is that it keeps hyphens that split words at ends of lines, and it would be better to eliminate these (even if that sometimes eliminates a hyphen that is not splitting a word), because it cuts the reading time in half for split words. Another speedup would be to display two or three words at a time, as a user selectable configuration). I have seen this in speed readers I've experimented with in the past.

An optional improvement would be for the reading pane to be put in an overlay in the middle of the viewport, and to fade the underlying article to a dark greay, to create a distraction-free reading experience.

kelson42:
What should we do with this PR?

mossroy:
As we have no news, I close this PR

Issue #179, opened 2016-06-23
Minor bug on upper/lowercase variants in title search : the util.ucEveryFirstLetter function does not work properly when accents or other non-ASCII characters are used


mossroy:
For example, if you type "amérique" as a search string, the uppercase on first letter of each word variant is wrong : it looks for AméRique instead of Amérique.

This is because the function util.ucEveryFirstLetter uses a regular expression with word boundary \b, that only works on ASCII characters : it considers the é as a word boundary

kelson42:
That's a really funny one :)

mossroy:
I thought this one was easy to fix, but it's not.
As it's a minor issue, I'll postpone it for now.
I've commited my work-in-progress on branch fix-international-ucEveryFirstLetter

kelson42:
@mossroy @Jaifroid I think this would be fixed in an acceptable manner by detecting the spaces instead of detecting the words. You could use \s for example and any other separator you judge appropriate. Would that be good for you?

Jaifroid:
@kelson42 Sorry for slow reply. Yes, I'll take a look at that regex. It's also possible that with widespread UTF8 support in JavaScript, the bug has gone away, i.e. the internal definition of a word boundary may no longer have that bug.

Jaifroid:
I confirm this bug still exists in Fiefox 75.0.

Jaifroid:
There is a broader issue with the current regex, which is that `\w` only matches `[a-zA-Z]` in JavaScript, i.e. only works with the ASCII alphabet. It's possibly this could be changed with the `/u` (Unicode) switch, but that is not supported in IE11.

I tested an initial solution to the bug, but still using `\w`, and the outcome can be seen in Regex Buddy here (note that I'm using a placeholder U to represent uppercase). As can be seen below, neither `á` in `ágata` nor any of the first letters of the Greek words are selected for conversion to uppercase.

![image](https://user-images.githubusercontent.com/4304337/80279084-24c0a000-86f3-11ea-8417-01cbe3885184.png)

Jaifroid:
If instead we select *any* character (`.`) after the start of the string or after a space or any number of spaces, as @kelson42 suggested, then we would get what you see below. My only doubts are that this will pick up numbers and punctuation marks with spaces before them. However, I suspect that converting these to uppercase will just return the original value, but haven't tested yet.

[...]

PR #617, opened 2020-04-26
Better handling of case transformations in search

Jaifroid:
This is an initial go at #179. It handles case transformations of the first letter of search terms after a space or any Unicode punctuation character.

Limitations:

* It does not (yet?) produce an array of search strings with every possible combination of upper- and lower-case first-letter transformations. For example, given the search string `estéril (Ciru`, the algorithm currently only produces these variants:
image
clearly the variant `Estéril (ciru` is missing, and in this case, then, the search fails to find the result for Sterile Surgery (`Estéril (cirugía)`) in the Spanish Wikipedia.
* I think the functions `lcFirstLetter()` and `ucFirstLetter()` in `util.js` may be redundant; I guess they were attempts to produce some measure of mixed-case transformations on the string, but they fail in the case mentioned, and it would be better to use a matrix to transform every word's first letters and recombine the strings.

In sum, I don't think this is ready for merging. It is an improvement in behaviour, I think, given that it works on Unicode strings, but it can be improved.. It would still be worth testing in a few scenarios and languages.


kelson42:
@Jaifroid Thank you. To me this is the perfect candidate for a unit test.

Jaifroid:
The latest commit finds every combination of upper- and lower-case first letters of words in the phrase. It removes the separate functions that did upper- and lower-case transforms in limited cases, and instead splits the string into its component number of "words" (`n`), ignoring any punctuation at the beginning of a word, and then iterates through `2 ^ n` combinations using bitwise `&` to set the case (`upper|lower`) of each word's first letter.

So, it can deal with this sample search, for example:

image

The Unit Tests currently contain a set of simple tests transforming a phrase to all upper-case first letters and all lower-case first letters. I will need to write a different test using the new algorithm. It will fail the current tests. I will write the new test if this is approved by @mossroy .

NB I want to test this thoroughly before removing the do-not-merge label. It is not ready for review.

kelson42:
@Jaifroid [FYI] Small remark about usual PR workflow. Github proposes a "draft" PR mode which can be chosen at the time your create the PR and since a few weeks even afterward. I recommend to use it in place of the homebrew "do-not-merge" tag. An other traditional/old way to show that the PR is not finished is to put "WIP" (for Work In Progress) at the head of the PR title. Of course, not assigning a reviewer is important too (no assignee for this PR, this is good). All of this is not super important because you mostly work only together with @mossroy, so just FYI. I have put the PR in draft.

Jaifroid:
@kelson42 Thanks, that's useful to know!

Jaifroid:
I have added "whole word" case variations, not for mixed case within a word, but for cases such as searching for `Config.sys` and finding `CONFIG.SYS`. The downside of this is that it doubles the number of searches to be matched with a concomitant halving of search performance. On a PC this is not noticeable, but it could be quite heavy on mobile. It could be made an option in Configuration. You can see this in action in the search below, and the precise array of combinations the function returns below that. Duplicates are subsequently removed from this array.

image

image

Jaifroid:
Latest commit preserves the search exactly as typed, and then normalizes it by reducing any spacing to a single space, making entire string lowercase, and then calculating the variation matrix from the normalized string. It enables pseudo case-insensitivity while also searching on the case-sensitive string as typed by the user (in case that is important). See screenshot running in FFOS simulator. Case transformations could be an option in Config, because in some contexts it could be slow. In browser/extension contexts performance is fine.

image

Jaifroid:
I say "pseudo case insensitivity" because of course it cannot transform `évoid` into `éVoid` (an actual search term in full English Wikipedia). The latter can only be found by typing case exactly.

Tested and working on:

* Firefox 75
* FFOS simulator
* Internet Explorer 11 (a little slow on full English Wikipedia)
* Edge 81 (Chromium)
* Edge 44 (Legacy)

mossroy:
Regarding the use of XRegExp vs copying its giant regexp, I'd be in favor of using their js file (only it's not too complicated). If we need to modify/extend it, it would be worth searching/creating issues on their github. They're probably better than us on dealing with unicode regexps, and maintaining them in the long term.
Based on the README of https://github.com/slevithan/xregexp , requirejs is officially supported. They also suggest to use https://www.npmjs.com/package/xregexp/v/4.3.0 (that has support for requirejs built-in, but also includes many addons that we probably don't need). It might be possible afterwards to remove these addons, or simply add the few header/footer lines add requirejs support to the non-"all" version.

Some unit tests are failing on Travis because they're still using the previous API. Don't forget to fix them (and add some that check the issues we're trying to fix).

Regarding the performance impact, I also agree that it's a concern.
Maybe there could be 2 versions (as you suggested) : 
- a simple and fast version, that only tests the same variants as the kiwix-lib fallback (https://github.com/kiwix/kiwix-lib/blob/8f07689c57356264c1ef145ba82c1d60a97b786e/src/reader.cpp#L788) : original title, ucFirst, lcFirst and toTitle (that seems to correspond to our previous ucEveryFirstLetter function). This version should not introduce any noticeable performance regression compared to version 2.7.0.
- a smarter (and more complete) version, that could use more variants on faster devices

The choice between these versions could be a new settings.
IMHO, ideally, when the setting is not set yet, we would use the smarter version, and would measure the time the first search takes. If it's not fast enough, we would display a warning to the user, suggesting to switch to the simpler version. If it's too complicated, it could be a separate enhancement

Regarding the smarter version, I'm a bit concerned by the 2^n variants. It looks overkill for most cases. Maybe a compromise would be that, for each word, we only test : 
* the originally typed word
* an uppercase version
* a lowercase version
* a version with only the first letter uppercase, and the rest lowercase
Even multiplied by the number of words, it would make a lot less variants than 2^n (especially if we remove duplicates between these versions, which should be frequent)

On the other hand, all this might be considered too complicated, considered that it would be removed if/when we manage compile kiwix-lib with emscripten. Up to you.

Jaifroid:
OK, thanks for the feedback @mossroy . I like these ideas. I might put XRegExp integration as a separate issue. Measuring search time sounds interesting, and could also determine the number of search results returned.

I definitely plan to fix the Unit Tests.

Jaifroid:
Latest commit adds this UI to the Performance Settings panel, and implements the Basic and Full search type. The difference is noticeable in IE11, but not so much in other browsers. This commit does not include any AI to time the search and warn the user. If user hovers over control, further explanation pops up (not sure what happens in mobile yet, only tested in IE11, Firefox 75 and Edge Legacy).

image

Jaifroid:
Just a quick sample of the kind of difference between the old search (top) and the new "comprehensive" search (bottom) in English Wikipedia searching for "the past is a". As you can see "The Past Is a Foreign Land" (a distinct article) is missed in the old search (together with some variants for the other two distinct articles).

image

image

kelson42:
@Jaifroid ... and I thought first this PR will be quick/short... How wrong I was :)

Jaifroid:
Latest commit returns results dynamically to the user, as they are found, and optimizes for returning common search patterns first. Clicking on a found results stops further search. I think this obviates the need for complex automation (timing searches and prompting user to change settings, or auto-changing settings).

Need to review performance on old/slow systems to see whether the new behaviour obviates the need for the "basic" search type. I suspect it will still be useful, as search can be excruciatingly slow in **some** contexts.

Will now attempt Unit Tests.

One more sample of a search that returns nothing in "basic" search (and in existing app), and 7 results with "full":

image

mossroy:
I did not look at the code, but I like your suggestion to dynamically add results, if it might simplify the code and the UI.
Tell me when I can test a pre-version on my Firefox OS device (which is slow enough to give an idea)

Jaifroid:
Code can be tested now. I haven't marked Ready for review because Unit Tests are still failing... Though I'm not sure they're failing on the new tests I've put in, seems to be something else.

Jaifroid:
Is passing Unit Tests now.

Jaifroid:
@mossory, progressive search is currently not very granular, because it reports back on each search variant that has found something, and quite a lot of variants don't of course find anything. It is, however, not difficult to report back for *each* result found. It looks like the demo gif below. I'm holding back on pushing these changes because I wonder if it is *too* granular. Thoughts?

Progressive Search Demo

Jaifroid:
I broke the tests I fixed. 😒
Have run out of time to fix them today I think. I know what I did, and can fix tomorrow... (It's to do with interim reporting of results...)

Jaifroid:
After further testing, the granular reporting works well on fast systems, and gives more informative feedback to the user on slower systems. So I've pushed it in a single commit https://github.com/kiwix/kiwix-js/pull/617/commits/1e10807594c98b11e5eaefd9c42bee681a295a4d which can easily be reversed if necessary.

Jaifroid:
Just for fun... **_searching with regular expressions_**! ༼ つ ◕_◕ ༽つ
(But it would be separate issue, because it would require significant refactoring: I just did a quick and very dirty hack to see if it was possible...)

image

mossroy:
I quickly tested on my Firefox OS device.
It works pretty well.
In terms of UX, I find that the progressive display of search results is a huge improvement!
I agree that it makes the new settings unnecessary. We could also keep the hard-coded 25 results.
If I choose a search result while the search is still in progress, I do not notice a slow-down, which is cool.

However, it sometimes creates some "flickering" on my device.
One reason is #184, for which PR #593 had been started (but not finished nor merged). It is a separate issue.
When there are not a lot of results, some lines of result sometimes appear, then disappear, then re-appear. It's harder to reproduce. I initially thought it was another issue, but maybe it was simply a side effect of #184.

Also, when the search string is long, it can take a long time to search all variants. Avoiding the 2^n variants can be a separate improvement.

I'm very pleased to see you fixed the unit tests, and added new ones.
They could be improved by avoiding to hard-code the position in the array, for the expected string. The expected string could be searched at any position in the array, with something like `assert.ok(somethingThatReturnsTrueIfOurExpectedStringIsInTheArray(), "The specific variant should be in the results")`

Regarding the regexp search, well... it's fun, but does not seem useful for a regular user ;-)

Jaifroid:
Thanks, @mossroy. It's cool to know that it's an improvement on the physical FFOS device!

I've coincidentally just pushed a commit that tries to deal with the problem that any keyup initiates a new search, even if the search string hasn't changed. That deals with some of the flickering. The other issue is that I need to cancel any old search remnants that may still be in the pipeline if the the user adds another letter or edits the search string before all the results have returned. I think that is what you were noticing: even adding one letter, launches a new search (after the timeout) but on old devices the old one is not cancelled, so you have two separate searches reporting interim results.

I agree that I should remove the Basic search type: it's no longer needed. However, I feel it would be good to keep the slider for selecting the number of results to return, because it affects the search depth. In other words, the algorithm will search further in the index on each variant the higher the number, and more results are returned with higher numbers, at the expense of slower search.

I suggest keeping a new default of 25 (it used to be 50), but I think some users could be unhappy at the loss of search depth if they are searching for a very common term. I might even suggest increasing the range of the slider from 5 to 100, default 25. Do you agree?

So I think I'm going to mark this as draft again, as there is some work to do.

Jaifroid:
Just to be _sure_ that you think it's OK to remove the "Basic" search type, here is some further info about the two types. For both search types, _the most common search patterns are searched first for speed_.

* For the Basic search type the variants are **2^n** where n is the number of **_*words*_** (not letters). So even typing five words for a search will only produce 32 variants, which are crunched through very fast on most systems.

* For the Full search type, we have **3^n** variants (see comments in code for explanation of why), i.e. 9, 27, 81 and 243 searches for five words. It is quite rare (and usually unnecessary) for someone to type out a five-word search, but it will produce quite a long search, clearly, if the user waits till the end.

The reason I suggested removing the Basic type is that a user does not have to wait for the search to finish if the result they want has been found. However, you mention a future PR preventing exponential searches. If you're thinking such a future PR may be necessary, then it might actually be better to keep both search types, as it provides a rational way of limiting the number of searches while still being thorough and clear to the user what we're doing.

Jaifroid:
**Alternative proposal**: Remove basic type from the UI, but switch to it automatically if user types five or more words?

mossroy:
Sorry, I thought we were still on a 2^n (n being the number of letters). I did not have the time to look at the code.
The future PR I was thinking of seems to be what you already implemented : try to change the case of the first letter of each word, not the case of every letter.
I agree that it should be very rare to type many words in the search field.

So I think this confirms I'm still OK with the removal of the Basic search.

mossroy:
Let's keep that simple, as all this will probably be useless and/or removed if we manage to switch to libzim/kiwix-lib (which should search a lot faster). It's already a great improvement.
Let's keep only one search algorithm, and keep an hard-coded value of number of search results (I agree that 25 is enough)

Jaifroid:
I think the latest code may also have fixed #184 (subject to testing on slow devices - I'll try it on an old Windows Mobile soon).

I track search state with a global array of search objects, each of which has a state field that allows for immediate cancellation (return of Promise) of a search that may still be running in the back end. A filter also ensures any interim results already queued from a search with a cancelled state are not reported to the UI.

Each search can have several states: `init`, `interim`, `cancelled`, `complete`. Each search object also has a `type` property (`full|basic`) which is set automatically to prevent exponential number of searches. The UI for setting the search type has been removed.

The QUnit tests have been made generic, returning `true` if the expected string is found anywhere in the results array, rather than at a specific position.

I'll do a bit more testing and will report test results below.

www/js/app.js
```diff
@@ -1060,7 +1092,9 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
      * @param {DirEntry} dirEntry The directory entry of the article to read
      */
     function readArticle(dirEntry) {
-	    // Only update for expectedArticleURLToBeDisplayed.
+        // Reset state.searches.lastPrefix to allow users to search the same string again if they want to
```
Jaifroid, line 1095: Note to self: comment is incorrect, should read "Reset search prefix to allow...."

tests/init.js
```diff
@@ -22,6 +22,13 @@
  */
 'use strict';
 
+// Define global params needed for tests to run on existing app code
+var params = {};
+var search = {
+    'prefix' : '',
+    'state' : ''
+};
+
```
Jaifroid, line 31: NB This is the `tests/init.js` file, and these are dummy variables allowing the tests to run.

www/js/app.js
```diff
@@ -965,30 +993,32 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
...
+        if (stillSearching) {
+            message = 'Searching [' + search.type + ']... found: ' + nbDirEntry;
+        } else if (nbDirEntry >= params.maxSearchResultsSize) {
+            message = 'First ' + params.maxSearchResultsSize + ' articles found (refine your search).';
         } else {
-            message = nbDirEntry + ' articles found.';
-        }
-        if (nbDirEntry === 0) {
-            message = 'No articles found.';
+            message = 'Finished. ' + (nbDirEntry ? nbDirEntry : 'No') + ' articles found' + (search.type === 'basic' ? ': try fewer words for full search.' : '.');
```
Jaifroid, line 1014: Split line for legibility.

www/js/lib/zimArchive.js
```diff
@@ -142,30 +142,71 @@ define(['zimfile', 'zimDirEntry', 'util', 'utf8'],
...
      * So, as workaround, we try several variants of the prefix to find more results.
      * This should be enhanced when the ZIM format will be modified to store normalized titles
      * See https://phabricator.wikimedia.org/T108536
      * 
-     * @param {String} prefix
-     * @param {Integer} resultSize
-     * @param {callbackDirEntryList} callback
+     * @param {Object} search The current state.searches object
+     * @param {Integer} resultSize The number of dirEntries to find
+     * @param {callbackDirEntryList} callback The funciton to call with the result
```
Jaifroid, line 153: Typo: "function" instead of "funciton"

www/js/lib/zimArchive.js
```diff
@@ -142,30 +142,71 @@ define(['zimfile', 'zimDirEntry', 'util', 'utf8'],
...
+     * @param {Boolean} noInterim A flag to prevent callback until all results are ready 
      */
```
Jaifroid, line 155: Add "(used in testing)" to `noInterim` comment above.

www/js/lib/zimArchive.js
```diff
@@ -142,30 +142,71 @@ define(['zimfile', 'zimDirEntry', 'util', 'utf8'],
...
+            // Dynamically populate list of articles
+            search.state = 'interim';
+            if (!noInterim) callback(dirEntries, search);
             var prefix = prefixVariants[0];
             prefixVariants = prefixVariants.slice(1);
-            that.findDirEntriesWithPrefixCaseSensitive(prefix, resultSize - dirEntries.length, function (newDirEntries) {
-                dirEntries.push.apply(dirEntries, newDirEntries);
-                searchNextVariant();
+            that.findDirEntriesWithPrefixCaseSensitive(prefix, resultSize - dirEntries.length, search, function (newDirEntries, interim) {
+                if (search.state === 'cancelled') return;
```
Jaifroid, line 201: It would be better to `callback ([], search)` here, i.e. with an empty array, and let the callback in app.js handle the cancellation.

www/js/lib/zimArchive.js
```diff
@@ -190,12 +232,15 @@ define(['zimfile', 'zimDirEntry', 'util', 'utf8'],
...
-                if (index >= firstIndex + resultSize || index >= that._file.articleCount)
+                if (search.state === 'cancelled' || index >= firstIndex + resultSize || index >= that._file.articleCount)
                     return dirEntries;
                 return that._file.dirEntryByTitleIndex(index).then(function(dirEntry) {
                     var title = dirEntry.getTitleOrUrl();
-                    if (!title.indexOf(prefix) && dirEntry.namespace === "A")
+                    if (~title.indexOf(prefix) && dirEntry.namespace === "A") {
                         dirEntries.push(dirEntry);
+                        // Report interim result
+                        callback(dirEntry, true);
```
Jaifroid, line 242: We should call back `[dirEntry]` here, i.e. dirEntry should be in an array, not a bare object, even though it works with concat.

www/js/lib/zimArchive.js
```diff
@@ -190,12 +232,15 @@ define(['zimfile', 'zimDirEntry', 'util', 'utf8'],
         }, true).then(function(firstIndex) {
             var dirEntries = [];
             var addDirEntries = function(index) {
-                if (index >= firstIndex + resultSize || index >= that._file.articleCount)
+                if (search.state === 'cancelled' || index >= firstIndex + resultSize || index >= that._file.articleCount)
                     return dirEntries;
```
Jaifroid, line 236: @mossroy I think I found (one?) cause of the race condition in #184: this return statement (which was in the original code like this) should *not* return an array of dirEntries: I think it should return a resolved Promise with the array.... As can be seen from the next line, `addDirEntries` is a Promise that should only return inside the `.then` function, or by Resolving the Promise.

Jaifroid:
Marking as draft again -- a few needed code changes have accumulated.

www/js/lib/zimArchive.js
Jaifroid: OK, I'm wrong here, sorry... Have found this:

Inside a `then` handler function, when x is a value (number, string, etc):
* `return x` is equivalent to `return Promise.resolve(x)`
* `throw x` is equivalent to `return Promise.reject(x)`

https://stackoverflow.com/a/47168159/9727685

Jaifroid:
Found another small bug.... Grrr.

mossroy:
I quickly tested on my Firefox OS device. It's working great!
I did not notice any flickering this time. And the user experience is really improved compared to previous versions.
It does fix #179, and probably fixes #184 too.

I did not have a deep look at the implementation. Please try not to add too much complexity as it might become useless with a future emscripten version of libzim.

In #593 , I tried to find a common implementation that would handle both #89 and #184, but it's not necessary for now, and could be a separate issue.

Jaifroid:
Thanks @mossroy . As #89 was already fixed, I didn't pay much attention to it here, although I did test the implementation with history back/forward.

I can say with pretty much certainty that the code will not allow any search results to bubble up to the UI if the search in question has a `state` of "cancelled". Search is cancelled by: launching another search, clicking on a search result even while search is still running, or on `blur` of the `prefix` field. Cancellation is tested for in `zimArchive` code before testing each prefix variation *and* at each retrieval of a dirEntry title for prefix matching, so a search is cancelled, and the Promises returned empty, pretty much as soon as possible after the user leaves the search in the UI. So I'm pretty confident that #184 is fixed.

The minor bug I thought I'd found was not real. So I think the code is ready. Will continue to test.

I completely understand that once we achieve libzim compilation, much of our existing code becomes redundant, but that's a good thing, and an opportunity to rebuild the app with contemporary JS coding practices and toolchains (e.g. #554).

Jaifroid:
## Manual tests

* Tested on Windows Mobile (slow device): working
* FFOS simulator: working
* Internet Explorer 11: working
* Edge Legacy (jQuery & SW modes): working
* Edge 81.0 (Chromium): working
* Firefox 76.0.1: working

www/js/app.js
```diff
@@ -965,30 +993,34 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
         }
     }
 
-  
     /**
      * Display the list of articles with the given array of DirEntry
-     * @param {Array} dirEntryArray The array of dirEntries returned from the binary search
+     * @param {Array} dirEntryArray The array of dirEntries returned from the binary
```
Jaifroid, line 998: Add missing word "search" back to end of comment.

kelson42: Nice, would be good as well to write in the PR descripion which tickets exactly will closed by tis PR.

kelson42:
Nice, would be good as well to write in the PR descripion which tickets exactly will closed by tis PR.

Jaifroid:
@kelson42 I've linked the issues that will be closed. I'll also use the "fixes" keyword for both issues in the squashed PR description.

kelson42:
@Jaifroid Sorry, had forgotten about the linked issues. Not sure the “fixes” keywords works if given in further comments that said...

Jaifroid:
@kelson42 When I squash a PR, I usually add "fixes" or "closes" in the description of the squashed PR, and it works then, though the new linked feature also does it.

kelson42:
@Jaifroid Just to be sure, we are waiting review of @mossroy here?

Jaifroid:
@kelson42 Yes, but no rush. I've now ported this same search algorithm to Kiwix JS Windows, and it seems to be working very well in the apps published to date that incorporate it (the 0.9.9.99 series on https://github.com/kiwix/kiwix-js-windows/releases including the vanilla store app, the WikiMed store app and the portable WikiMed executable for older Windows).

kelson42:
@Jaifroid Thank you for the comfirmation. To me this is just a PR with a lot of value and want to secure there is no deadlock. But of course, take all the time needws to continue the good job :)

mossroy:
I should be able to have a look this week

mossroy:
I think I found a small bug : use https://download.kiwix.org/zim/wikinews/wikinews_fr_all_maxi_2020-05.zim and type the search string "ir". The first search results do not correspond to this prefix :

image

mossroy:
I quickly tested on a Firefox OS device and on Chromium 81 and saw no other issue. It makes a real difference on slow devices, in terms of user experience.
And #184 seems to be solved with it.

I'm having a look at the code : it's a huge change, and not easy to get into it.

Jaifroid:
> Same kind of issue with https://download.kiwix.org/zim/wikivoyage/wikivoyage_en_all_maxi_2020-05.zim if you type the search string "l" : the first search results start with Á, Ä etc.

Hmm. That certainly shouldn't happen. Strange! I'll check it out. Thanks, @mossroy .

www/js/lib/util.js
```diff
@@ -23,57 +23,65 @@
...
+            for (var i = numCombos; i--;) {
+                mixedTypeCaseStr = '';
+                bitmask = 1;
+                for (var j = 0; j < strParts.length; j++) {
+                    // Get modulus of division (this is equivalent to bitwise AND for different bases)
+                    // caseBit will be 0, 1 or 2 (latter only for 'full' case calcualation)
+                    caseBit = ~~(i / bitmask % base);
+                    if (caseBit === 2) {
+                        // All uppercase
+                        typeCase = strParts[j].toLocaleUpperCase();
```
mossroy, line 68: According to https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/toLocaleUpperCase , using `toLocaleUpperCase` without passing a Locale uses the "host environment’s current locale".
Could it be an issue if my device is configured with a locale that is different from the one of the ZIM file? If so, we might need to pass the locale found in the ZIM metadata?

www/js/lib/zimArchive.js
```diff
@@ -142,31 +142,68 @@ define(['zimfile', 'zimDirEntry', 'util', 'utf8'],
...
+        var prefix = search.prefix;
         var that = this;
-        var prefixVariants = util.removeDuplicateStringsInSmallArray([prefix, util.ucFirstLetter(prefix), util.lcFirstLetter(prefix), util.ucEveryFirstLetter(prefix)]);
+        // Establish array of initial values that must be searched first
+        var startArray = [];
+        // Ensure a search is done on the string exactly as typed
+        startArray.push(prefix);
+        // Normalize any spacing and make string all lowercase
+        prefix = prefix.replace(/\s+/g, ' ').toLocaleLowerCase();
+        // Add lowercase string with initial uppercase (this is a very common pattern)
```
mossroy, line 165: Isn't this very common pattern already present in the `util.allCaseFirstLetters` results? (basic or full type)

www/js/lib/zimArchive.js
```diff
@@ -142,31 +142,68 @@ define(['zimfile', 'zimDirEntry', 'util', 'utf8'],
...
+        // Put cap on exponential number of combinations (five words = 3^5 = 243 combinations)
+        search.type = fullCombos.length < 200 ? 'full' : 'basic';
+        // We have to remove duplicate string combinations because util.allCaseFirstLetters() can return some combinations
+        // where uppercase and lowercase combinations are exactly the same, e.g. where prefix begins with punctuation
+        // or currency signs, for languages without case, or where user-entered case duplicates calculated case
+        var prefixVariants = util.removeDuplicateStringsInSmallArray(
+            startArray.concat(
+                // Get basic combinations first for speed of returning results
+                util.allCaseFirstLetters(prefix).concat(
+                    search.type === 'full' ? fullCombos : []
```
mossroy, line 180: When the search type is `full` (which will be the most frequent scenario), it calls `util.allCaseFirstLetters(prefix)` twice (with and without the `full` parameter), concatenates and de-duplicates the result.
Wouldn't be possible to use only the result of `util.allCaseFirstLetters(prefix,'full')`, and modify this `allCaseFirstLetters` function, so that the basic combinations are at the beginning of its result?
It would probably save some CPU

tests/init.js
Jaifroid: @mossroy You're right that for tests we don't need to track variables in the same global way, because we don't have cancellation, but these variables are dummies that exist merely so that the code can run. with minimal variations. I can look to simplify. This was a kind of "patch" to get the tests to run on the same codebase...

www/index.html
```diff
@@ -305,6 +305,20 @@ <h3>Performance settings</h3>
                                         </div>
                                     </div>
                                 </div>
+                                <br />
+                                <div class="row">
+                                    <div class="col-sm-4">
+                                        <p>Select max number of search results:</p>
```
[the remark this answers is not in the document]
Jaifroid: @mossroy I think that users of Kiwix JS are maybe used to seeing 50 results hard-coded, but this is definitely slower to run than 25. I thought some users might complain about the sudden reduction. Now, there are some marginal cases where a search result you're interested in can be found with 50 searches but not with 25. You could probably always find it by refining your search, but you might not know *how* to refine your search if you are "browsing". Maybe this is all a bit marginal, but I'm in favour of giving users options! We don't have a very cluttered UI, so I can't see the downside in having the option (given that it's already coded)?

www/js/app.js
```diff
@@ -72,13 +66,25 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
...
+    document.getElementById('titleSearchRangeVal').innerHTML = params.maxSearchResultsSize;
     // A global parameter that turns caching on or off and deletes the cache (it defaults to true unless explicitly turned off in UI)
     params['useCache'] = settingsStore.getItem('useCache') !== 'false';
     // A parameter to set the app theme and, if necessary, the CSS theme for article content (defaults to 'light')
     params['appTheme'] = settingsStore.getItem('appTheme') || 'light'; // Currently implemented: light|dark|dark_invert|dark_mwInvert
     document.getElementById('appThemeSelect').value = params.appTheme;
     uiUtil.applyAppTheme(params.appTheme);
 
+    // Define global state
+    // An array to hold searches and their state (allows modules to tell which search is current and cancel old searches)
```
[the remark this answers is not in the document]
Jaifroid: @mossroy The problem as I see it is that we immediately need to set a flag on the currently running search _from the UI_ (i.e. from `app.js`) to indicate that it has been cancelled, and that flag needs to be immediately available to the subroutines in zimArchive that are operating on that search so they can stop initiating new Promise loops. A new search can be initiated with anew state of 'init', even while the previous search is cancelling. There are queued up Promises which will keep returning for a while, because there is no way to cancel an a queued Promise. So I thought I needed to keep an open handle on the search until we can be sure all Promises have returned, and the returning functions need to see that the state of the old search is 'cancelled', even though the new search has a different state. The problem here is the async nature of search. Only when all searches are definitively over can we prune the list of searches.

I don't have a technical understanding of "state" or how to manage it. Hence my solution here is probably unorthodox...

www/js/app.js
```diff
@@ -72,13 +66,25 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
...
+    state['searches'] = [{
+        'prefix': '', // A field to hold the original search string
+        'state': '',  // The state of the search: ''|'init'|'interim'|'cancelled'|'complete'
```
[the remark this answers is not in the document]
Jaifroid: I agree, it is confusing. Will think of something else.

www/js/lib/util.js
Jaifroid: This isn't a change from the original code, which already used `toLocaleUpperCase`. But I'll investigate.

www/js/lib/zimArchive.js
Jaifroid: Indeed, but we want it to be searched first in the array of searches. It considerably speeds up the returning of results if the most common patterns are searched before the less likely (but still possible) combinations. The `removeDuplicateStringsInSmallArray` removes any duplicates that are lower down in the array.

www/js/lib/zimArchive.js
Jaifroid: I considered this, but it seemed like quite a lot of coding to get the order right when just running it twice does this for us. Basic search combinations are much more probable than full combinations, hence results are returned much faster to the UI if we search basic followed by any full combinations that were not already in basic. Because of the way the loop runs in the full type, it is impossible to guarantee that all the basic ones are returned first, whether you count up from 0000 to 1111 or whether you count down from 1111 to 0000.

And CPU usage here is tiny compared to binary search itself, so this is never going to be the bottleneck...

Jaifroid:
@mossroy Thank you for your detailed and valuable review. I'll work on the things I've indicated, and will let you know when I've done those. If you disagree with any of the explanations where I'm less convinced a change is needed, don't hesitate to let me know in the threads in the code.

www/index.html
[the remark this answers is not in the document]
mossroy: All right, you convinced me. These cases look very marginal to me, but it's not a lot of code, and not complicated to maintain.

www/js/app.js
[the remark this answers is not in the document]
mossroy: I fully agree with the need to keep a global variable, for the reasons you explained.
What I would suggest to investigate is to keep only the latest search in this global variable, instead of a list of searches. Something like : 

    state['latestSearch'] = { 'prefix': '', 'state': '', 'type': '' };

Instead of passing the current search as a parameter to some functions, these functions could directly read/modify this global variable.
These functions usually need to check that _their_ search has not been cancelled by another one. They could do that by checking `mySearchPrefix !== state.latestSearch.prefix` : if the search prefix has been changed, it means that their search can be aborted.

www/js/lib/zimArchive.js
mossroy: All right. This explanation could be added as a comment in the code.

www/js/lib/zimArchive.js
mossroy: All right, I trust you

www/js/app.js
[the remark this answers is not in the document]
Jaifroid: OK, thanks for the explanation. I'll try.

www/js/lib/zimArchive.js
Jaifroid: Will do.

Jaifroid:
Branch/PR has been rebased on master, so I can work on identified issues.

Jaifroid:
> I think I found a small bug : use https://download.kiwix.org/zim/wikinews/wikinews_fr_all_maxi_2020-05.zim and type the search string "ir". The first search results do not correspond to this prefix :
> 
> Same kind of issue with https://download.kiwix.org/zim/wikivoyage/wikivoyage_en_all_maxi_2020-05.zim if you type the search string "l" : the first search results start with Á, Ä etc.

This bug should be fixed now. After binary search had returned titles where `prefix <= title`, the filter was including results where the prefix could be found *anywhere* in the title. It did not consider a case where `prefix <= title` might not actually begin with prefix (or one of its forms). This has been rectified.

Jaifroid:
PS still working on the other issues identified.

www/js/app.js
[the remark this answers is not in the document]
Jaifroid: @mossroy You were right. There was no need to keep a list of searches. The search cancellation propagates through to the local searches in zimArchive and it seems to work just fine. It's a bit of a mystery to me which objects become local copies, independently modifiable, and which are merely pointers to a global object.

tests/init.js
Jaifroid: I have removed this from init.js. It now runs locally in tests.js.

www/js/app.js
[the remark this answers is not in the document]
Jaifroid: @mossory The only thing I could think of was to rename the global state object to "global". I think it eases comprehension, because then we have, e.g. `global.search.state`.

I did think of "mode" instead of state, but it's not very accurate. I also thought of "stage", but again "cancelled" cannot really be described as a "stage" of a search.

www/js/lib/util.js
Jaifroid: @mossroy So, looking into this, mostly `toLocaleUpperCase` will return the same result as `toUpperCase`, but in "languages like Turkish" it may vary. So the issue would only present itself if a user were accessing a Turkish (e.g.) ZIM on a machine configured for, say, en-GB locale. Then they might get some funny uppercase  Turkish letters based on UTF-8 mappings, and not on standard Turkish fonts. So, we can't resolve this without more info: does Wikipedia use UTF-8 mappings for upper-lower-case on Turkish ZIMs? Or does it use some locale-specific mapping?

I don't think we should be trying to work out locale from the ZIM metadata, because the language codes in the ZIM do not map one-to-one onto JavaScript/HTML (or Windows/Mac/Linux???) locale strings. Possibly for the currently loaded page there *might* be an HTML5 locale string we could scrape, but it's not a universal solution because we don't know that every ZIM article has locale set.

Finally, the user does have a remedy: type a more exact version of search string.

All-in-all, this is, I think, a minor issue. It's always been in Kiwix JS, it is not part of this PR, and a solution would be a separate issue. Is that OK? If you would like the issue resolved, maybe you could open it as a separate issue? Otherwise I'm resolving this for now!

Jaifroid:
I think I've addressed all the issues raised in your review, @mossroy . The two issues that *may* still be pending are:

* Do you agree with renaming the global `state` variable to `global`? I think this improves intelligibility because we then have an object property called `global.search.state`. That is much easier to comprehend than `state.search.state` or `state.search.stage` or `state.search.mode`.
* I propose to leave `toLocaleUpperCase()` in the code, because we lack the knowledge to pass a locale string, and maybe using UTF mappings is the right thing to do anyway. But it could be opened as a separate localization issue. NB it was already in Kiwix JS code, this PR did not introduce it.

If you are now happy with the code, subject to my final tests on it, please mark it as approved. I'll add test results below in the meantime. If you get to test it, please double check search cancellation on a slow device. From my tests (involving pausing debugger straight after cancellation), the cancellation propagates to zimArchive immediately. What is a bit harder to check is whether starting a new search while the old one is cancelling then re-opens the old search. If that were happening, I would expect to see erroneous results turning up in search on slow devices.

www/js/app.js
```diff
@@ -926,33 +948,35 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
...
      * with a binary search inside the index file)
-     * @param {String} prefix
+     * @param {String} prefix The string that must appear at the start of any title searched for
      */
     function searchDirEntriesFromPrefix(prefix) {
         if (selectedArchive !== null && selectedArchive.isReady()) {
+            // Cancel any previous search that may still be running before creating new search
+            global.search.state = 'cancelled';
+            // Store the new search term in the state.search global object and initialize
+            global.search = {'prefix': prefix, 'state': 'init', 'type': ''};
```
Jaifroid, line 977: @mossroy Can you please advise me on the above four lines? This is the bit that's worrying me (and the reason why I originally kept a list of searches). Here, we cancel a running search, and the cancelled state is reflected immediately in zimArchive.js. However, in the very next line, we store a new object in `global.search`. I don't have clear in my head what happens to the previously cancelled object that had been passed to `zimArchive.js`. Is that version somehow delinked from `global.search` but still exists in a delinked state?

Jaifroid:
Manual tests:

* **Firefox OS with WikiMed ZIM:** able to interrupt a full search in mid search with new pasted search, and no extraneous search results appear;
* **IE11 from file:// and http://localhost**: same;
*  **Edge Legacy from file:// and http://localhost**: same in jQuery mode;
* **Edge Legacy (from localhost) in SW mode: same;
* **Edge Chromium in jQuery and SW modes**: same.
* **Firefox 77.0.1 in jQuery and SW modes**: same.

kelson42:
@Jaifroid I believe we require a new review from @mossroy now? I have request a new one at least.

Jaifroid:
@kelson42 Yes, thanks, it's just a check that my solutions satisfactorily address issues raised by previous review.

www/js/lib/zimArchive.js
```diff
@@ -190,12 +231,16 @@ define(['zimfile', 'zimDirEntry', 'util', 'utf8'],
         }, true).then(function(firstIndex) {
             var dirEntries = [];
             var addDirEntries = function(index) {
-                if (index >= firstIndex + resultSize || index >= that._file.articleCount)
+                if (search.state === 'cancelled' || index >= firstIndex + resultSize || index >= that._file.articleCount)
```
mossroy, line 234: To be able to match a search that has been replaced by another one, we could replace this line by :
`                if (search.state === 'cancelled' || search.prefix !== prefix || index >= firstIndex + resultSize || index >= that._file.articleCount)`

(and `prefix` might be renamed to `originalPrefix` or whatever same parameter name used in other functions)

www/js/lib/zimArchive.js
```diff
@@ -142,31 +142,71 @@ define(['zimfile', 'zimDirEntry', 'util', 'utf8'],
...
-     * @param {String} prefix
-     * @param {Integer} resultSize
-     * @param {callbackDirEntryList} callback
+     * @param {Object} search The current state.search object
+     * @param {Integer} resultSize The number of dirEntries to find
+     * @param {callbackDirEntryList} callback The function to call with the result
+     * @param {Boolean} noInterim A flag to prevent callback until all results are ready (used in testing) 
      */
-    ZIMArchive.prototype.findDirEntriesWithPrefix = function(prefix, resultSize, callback) {
+    ZIMArchive.prototype.findDirEntriesWithPrefix = function (search, resultSize, callback, noInterim) {
```
mossroy, line 156: As explained above, I think we should pass both the originalPrefix (that never changes) AND the search object, to be able to detect a search replaced by another one.

www/js/lib/zimArchive.js
```diff
@@ -142,31 +142,71 @@ define(['zimfile', 'zimDirEntry', 'util', 'utf8'],
...
+                util.allCaseFirstLetters(prefix).concat(
+                    search.type === 'full' ? fullCombos : []
+                )
+            )
+        );
         var dirEntries = [];
+
         function searchNextVariant() {
+            // If user has initiated a new search, cancel this one
+            if (search.state === 'cancelled') return callback([], search);
```
mossroy, line 191: Should be replaced by :
```
            if (search.state === 'cancelled' || search.prefix !== originalPrefix) return callback([], {state : 'cancelled'});
```
so that we also trap the case of a search replaced by another one.
And, instead of passing the current `search` object, we can pass a new one that is indeed cancelled, so that `populateListOfArticles` does nothing.

www/js/lib/zimArchive.js
```diff
@@ -142,31 +142,71 @@ define(['zimfile', 'zimDirEntry', 'util', 'utf8'],
...
+            if (!noInterim) callback(dirEntries, search);
             var prefix = prefixVariants[0];
             prefixVariants = prefixVariants.slice(1);
-            that.findDirEntriesWithPrefixCaseSensitive(prefix, resultSize - dirEntries.length, function (newDirEntries) {
-                dirEntries.push.apply(dirEntries, newDirEntries);
-                searchNextVariant();
-            });
+            that.findDirEntriesWithPrefixCaseSensitive(prefix, resultSize - dirEntries.length, search,
+                function (newDirEntries, interim) {
+                    if (search.state === 'cancelled') return callback([], search);
```
mossroy, line 203: As above, should be replaced by :
```
                    if (search.state === 'cancelled' || search.prefix !== originalPrefix) return callback([], {state : 'cancelled'});
```

www/js/lib/util.js
mossroy: I agree

www/js/app.js
```diff
@@ -70,13 +64,25 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
...
+    document.getElementById('titleSearchRangeVal').innerHTML = params.maxSearchResultsSize;
     // A global parameter that turns caching on or off and deletes the cache (it defaults to true unless explicitly turned off in UI)
     params['useCache'] = settingsStore.getItem('useCache') !== 'false';
     // A parameter to set the app theme and, if necessary, the CSS theme for article content (defaults to 'light')
     params['appTheme'] = settingsStore.getItem('appTheme') || 'light'; // Currently implemented: light|dark|dark_invert|dark_mwInvert
     document.getElementById('appThemeSelect').value = params.appTheme;
     uiUtil.applyAppTheme(params.appTheme);
 
+    // Define global state
+    // An object to hold the current search and its state (allows cacncellation of search across modules)
```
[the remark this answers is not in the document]
Jaifroid: Done (in next commit).

www/js/app.js
```diff
@@ -70,13 +64,25 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
...
+    global['search'] = {
```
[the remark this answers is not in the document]
Jaifroid: Thanks!

www/js/app.js
Jaifroid: OK, @mossroy , yes that was a very rookie bit of code I'm embarrassed about now I see it. It arose from my removing the list of searches and not checking through the logic.... doh. Thanks for code suggestions, I see how that works. Will hopefully finally get this right...

www/js/app.js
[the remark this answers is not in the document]
Jaifroid: @mossroy I might be a native speaker of English, but not a native speaker of JavaScript! I think `globalstate.search.status` works well. Will use it.

tests/tests.js
```diff
@@ -24,6 +24,11 @@ define(['jquery', 'zimArchive', 'zimDirEntry', 'util', 'uiUtil', 'utf8'],
 
     var localZimArchive;
 
+    // Dummy search object so that we can simulate UI searches
+    var search = {
+        'prefix' : '',
+        'state' : ''
+    };
```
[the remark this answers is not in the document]
Jaifroid: Done

www/js/lib/zimArchive.js
Jaifroid: @mossroy This has been done by declaring a `localPrefix` as a constant, which obviously is not changed at any point in the code, and which is then used as the comparator.

Jaifroid:
I've added a commit which I consider "merge candidate", which follows the suggestion of comparing with the original prefix to add another cancellation test. Thank you for that. I'll do testing and self review, then will let you know when ready for final check.

www/js/app.js
```diff
@@ -70,13 +64,25 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
...
     // A parameter to set the app theme and, if necessary, the CSS theme for article content (defaults to 'light')
     params['appTheme'] = settingsStore.getItem('appTheme') || 'light'; // Currently implemented: light|dark|dark_invert|dark_mwInvert
     document.getElementById('appThemeSelect').value = params.appTheme;
     uiUtil.applyAppTheme(params.appTheme);
 
+    // Define global state (declared in init.js)
+    // An object to hold the current search and its state (allows cancellation of search across modules)
+    globalstate['search'] = {
+        'prefix': '', // A field to hold the original search string
+        'status': '',  // The state of the search: ''|'init'|'interim'|'cancelled'|'complete'
```
Jaifroid, line 82: Change "state of the search" to "status of the search.

www/js/app.js
```diff
@@ -926,33 +948,35 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
...
     /**
      * Search the index for DirEntries with title that start with the given prefix (implemented
      * with a binary search inside the index file)
-     * @param {String} prefix
+     * @param {String} prefix The string that must appear at the start of any title searched for
      */
     function searchDirEntriesFromPrefix(prefix) {
         if (selectedArchive !== null && selectedArchive.isReady()) {
+            // Cancel any previous search that may still be running before creating new search
+            globalstate.search.status = 'cancelled';
```
Jaifroid, line 975: Remove these useless lines, doh!

www/js/app.js
```diff
@@ -963,30 +987,34 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
...
 
-  
     /**
      * Display the list of articles with the given array of DirEntry
      * @param {Array} dirEntryArray The array of dirEntries returned from the binary search
+     * @param {Object} reportingSearchPrefix The prefix of the reporting search
      */
-    function populateListOfArticles(dirEntryArray) {
+    function populateListOfArticles(dirEntryArray, reportingSearchPrefix) {
+        // Do not allow cancelled or changedsearches to report
```
Jaifroid, line 996: Add space changedsearches -> changed searches

www/js/lib/zimArchive.js
```diff
@@ -190,12 +233,16 @@ define(['zimfile', 'zimDirEntry', 'util', 'utf8'],
...
             var dirEntries = [];
             var addDirEntries = function(index) {
-                if (index >= firstIndex + resultSize || index >= that._file.articleCount)
+                if (search.status === 'cancelled' || search.prefix !== originalPrefix || index >= firstIndex + resultSize || index >= that._file.articleCount)
                     return dirEntries;
                 return that._file.dirEntryByTitleIndex(index).then(function(dirEntry) {
                     var title = dirEntry.getTitleOrUrl();
-                    if (!title.indexOf(prefix) && dirEntry.namespace === "A")
+                    // Only return dirEntries with titles that actually begin with prefix
+                    if (dirEntry.namespace === "A" && title.indexOf(prefix) === 0) {
```
Jaifroid, line 241: Maybe use single quotes here to normalize the stile used in this function.

Jaifroid:
## Tests

I'll update this post as these tests are done:

- [ ] Code tests for cancellation (using console.log)
- [ ] Firefox OS search and cancellation test
- [ ] IE11 search and cancellation test
- [ ] Edge Legacy search and cancellation tests jQuery and SW modes
- [ ] Edge Chromium 83.0 search and cancellation tests jQuery and SW modes
- [ ] Firefox 78.0.1 search and cancellation tests jQuery and SW modes

Jaifroid:
@mossroy we need to take a decision on how to proceed with this PR given that there is more work that could be done to optimize it. Let me explain:

Code tests for cancellation using console log show that despite all the cancellation code, all the Promises have already been set up long before the user actually cancels a search by entering a new one. Because the code runs async, it very quickly sets up Promises for searches on all the variants, before the user starts typing a new search. The Promises carry on searching. We save some CPU by returning early before [this line](https://github.com/kiwix/kiwix-js/pull/617/files#diff-1916c6001d87418e72ffda9cfa7b2d82R238), and we successfully suppress the display of any results returned from the Promises. But some work is still being done, and a number of dirEntries for old searches are returned (and suppressed) even after a new search has started.

This is very similar to #426 (Prevent redundant extraction of assets after navigation), which we haven't yet fixed either. I see three ways to deal with this:

* Wait for me to find a way to return early from binary search without looking up any dirEntries and using minimal CPI on the Promises already set up -- I can't work on this now till the weekend;
* Or, decide that releasing with the working code as it is a good enough improvement for now, and add an issue like #426 to work on optimizing the cancellation later, maybe together with #426;
* Or do a simple bugfix release without this PR (which I think would be a shame, but it is now urgent to do a bugfix release given users are not able to read latest ZIMs without errors due to MIME type table issue).

To be clear, the code for this PR is working well, searches are cancelled effectively without extraneous results displayed, and the new search starts returning results fast, but CPU **is** still being consumed on redundant binary searches in Promises that are hard to cancel, for a while after the new search results are coming in. Nevertheless, IMHO the UX is still a big improvement.

Please advise on best course of action @mossroy .

mossroy:
I agree with you that this PR is already a big UX improvement.
Let's do things step by step, I vote for your second option : finalize testing and merging this PR, release a 2.8, and create another issue for a possible improvement in the future

Jaifroid:
In doing IE11 testing I found and fixed a small issue: a user could relaunch the same search while it is already in progress by pressing the search button or pressing enter. I have added https://github.com/kiwix/kiwix-js/pull/617/commits/8a7d740ac06cedc9c1d9eb01bbd5ffd015a04c85 to prevent this if the prefix has not changed and the search status is NOT 'cancelled' or 'complete' (i.e. it is in progress). A user can now press enter, e.g. by mistake or from habit, or click the search icon, without it relaunching or interrupting the search in progress. However, if search is complete and user wants to search the same terms again, then a new search on the same prefix *will* launch, as this would be a deliberate action by the user.

Jaifroid:
OK @mossroy  I've made all the manual tests indicated in https://github.com/kiwix/kiwix-js/pull/617#issuecomment-656071233 and have fixed the bug mentioned in previous comment. I've added a separate issue #637 for optimizing cancellation.

If you're happy for me to squash and merge, please approve the review. I'm of course happy to fix anything further as necessary.

Jaifroid:
I'm trying something out I discovered this morning. Please wait...

Jaifroid:
OK, sorry for last-minute commits. Having slept on it, a simple way of returning cancelled searches early from binary search came to me, and it works, preventing a lot of redundant CPU use. But it only works if we test on the globalstate.search variable rather than on the local copy of search. But this involves reinstating the globalstate variable in tests, or else they fail.

I'd be happy to revert those two commits and deal with them under #637 if you prefer, but having tested carefully with lots of console log statements, I'm confident they go a long way towards fixing #637 (but I'll keep it open for more testing subsequently).

Jaifroid:
**This is just for information** (and to document the issue for #637):

Yesterday I thought I was going mad... But this Stackexchange post explains the weirdness we're seeing here:

[JavaScript Pointer Reference craziness](https://stackoverflow.com/questions/8318357/javascript-pointer-reference-craziness-can-someone-explain-this)

So it seems that in fact the global search object and the search passed to the functions point to the same underlying object **until** the global search object is reassigned to a different object. At that point, the search object passed to the functions continue to point to the old object, and the global search object points to a new object.

It explains why I was seeing the passed searches linked to the global search, but then suddenly they weren't. We can actually exploit this, but I propose to do it for #637 and #426 at a later date, not to delay this PR anymore (which is working well now).

www/index.html
```diff
@@ -305,6 +305,20 @@ <h3>Performance settings</h3>
...
+                                    </div>
+                                    <div class="col-sm-8">
+                                        <div class="range">
+                                            <label>
+                                                <input type="range" class="form-control-range" min="5" max="50" id="titleSearchRange" value="25">
+                                                <span>Value: <span id="titleSearchRangeVal"></span> (<i>increases search time</i>)</span>
```
mossroy, line 317: Maybe we should rephrase 'increases search time' to 'default is 25, higher values increase search time'?

www/js/app.js
```diff
@@ -111,11 +117,15 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
     // Define behavior of HTML elements
     var searchArticlesFocused = false;
     $('#searchArticles').on('click', function() {
+        var prefix = document.getElementById('prefix').value;
+        // Do not initiate the same search if it is already in progress
+        if (globalstate.search.prefix === prefix && !/cancelled|complete/.test(globalstate.search.status)) return;
```
mossroy, line 122: It's minor, but the regexp should try to match the whole string. As it is currently, it can also match a substring, which would be wrong (even if it can not happen as the values are all known and set by us), and is probably slightly slower too.

www/js/lib/zimArchive.js
```diff
@@ -142,31 +142,74 @@ define(['zimfile', 'zimDirEntry', 'util', 'utf8'],
...
+            if (!noInterim) callback(dirEntries, localPrefix);
             var prefix = prefixVariants[0];
             prefixVariants = prefixVariants.slice(1);
-            that.findDirEntriesWithPrefixCaseSensitive(prefix, resultSize - dirEntries.length, function (newDirEntries) {
-                dirEntries.push.apply(dirEntries, newDirEntries);
-                searchNextVariant();
-            });
+            that.findDirEntriesWithPrefixCaseSensitive(prefix, resultSize - dirEntries.length, localPrefix, 
+                function (newDirEntries, interim) {
+                    if (globalstate.search.status === 'cancelled' || globalstate.search.prefix !== localPrefix)
```
mossroy, line 205: Using globalstate variable here breaks the software layers we have.
Using it in app.js is no problem (if it's where the variable is declared), but it should not be used in other files (except if passed as a parameter, of course).

The ZIM backend should not depend on external global variables. A drawback can be seen in the unit tests : you should be able to test each call independently, without being forced to declare the global variable (which might have side effects if the unit tests are ran in parallel, for example)

Jaifroid:
OK, @mossroy I understand, and hadn't realized that I was breaking the architecture by introducing a global variable into the backend. I think this is completely solvable, but isn't a "quick fix". Therefore, can I just check your timeline? If we are going to try to release this weekend, I'll revert the last two commits, and then work properly on this in #637. If you don't aim to release till next week, I *could* try to work on it more here. But I don't want to rush things, and I'm quite anxious (as you may have noticed) to get a release out.

www/index.html
Jaifroid: Yes, I'll rephrase.

www/js/app.js
Jaifroid: OK, no problem.

mossroy:
I see you just reverted your last 2 commits. I can review/test a bit more this version, right now.

Jaifroid:
I'm just adding the other two things you mentioned above (change description and regex). Then I'll be done.

Jaifroid:
OK, I've reverted and have done the two minor points in last review. Let me know if there's anything else for a "quick fix".

www/js/init.js
```diff
@@ -30,6 +30,13 @@
  */
 var params = {};
 
+/**
+ * A global object for storing app state
+ * 
+ * @type Object
+ */
+var globalstate = {};
```
mossroy, line 38: This variable can probably be moved to app.js, but it's minor and can be done later

mossroy:
You can merge and I might work on a 2.8 release in the next hour

www/js/init.js
Jaifroid: Thanks @mossroy . It would feel safer to me to move this as part of #637 when I'll have time to test a unified solution to cancelling running binary searches (and possibly also cancelling fetching of assets on pages we have navigated away from). So I'll move a summary of issues to be dealt with to #637 after squashing and merging this PR.

mossroy: Sure, that's perfect

Issue #614, opened 2020-04-14
Use const declarations instead of var wherever constants are declared


Jaifroid:
We have established with #613 that `const` works with all the browsers we support, and it improves readability as well as (presumably) improving browser/app memory management (likely a very tiny effect).

yonis9:
Hi, can I take it?

Jaifroid:
@yonis9 Yes, this would be an easy first PR. Just make sure that it is only constants that you define with `const`, i.e. nothing that can change values. Generally we indicate constants with capital letters. For example, you'll see three at the beginning of `app.js`.

ritikkuril:
i new to this so just asking i just need to convert all var type in const type in all .js scripts

Jaifroid:
@ritikkuril No, as is written above, only values that are actually constants should be converted.

fabcoder547:
@ritikkuril  have you done setup on your localpc? can you guide me in that ? how you did that?

ritikkuril:
@fabcoder547  i am also new bro

PR #625, opened 2020-06-25
Refactored the code (fixes issue #614)

TaniaMalhotra:
Changed var declarations for constants to const
Improves readability as well as browser/app memory management.


Jaifroid:
Thank you @TaniaMalhotra . I haven't tested yet, but I noticed you changed the declarations in a dependency (Boostrap-bundle.js). Dependencies shouldn't be touched because they are coded that way by the original library. Could you remove that file from the PR, please?

TaniaMalhotra:
Sir I have created a new PR for the same

Issue #590, opened 2020-02-03
Switch to GitHub Actions instead of Travis


mossroy:
We had a few issues with Travis CI in the past (like #573) and GitHub seems to push GitHub Actions to replace it.
We would have to plug GitHub Actions with Sauce Connect, which seems to be possible (see https://www.linkedin.com/pulse/getting-started-github-actions-webdriverio-sauce-labs-moataz-nabil)

Jaifroid:
There's an interesting, detailed [blog post from the author of Zettlr](https://www.zettlr.com/post/continuous-cross-platform-deployment-github-actions) (an electron-based open source markdown word processor I use) on how he implemented Continuous Deployment with GitHub actions. I found it useful to understand how actions work, though it still seems like a mammoth task. I know we don't have much deployment to do (it might be more useful in Kiwix JS Windows where there are several manual steps I have to repeat to deploy a release), but I think the principles apply to Continuous Integration as well.

mossroy:
Creating a tag on github does not seem to trigger a build on Travis any more.
So I have to manually run `./create_all_packages.sh -t -v 2.8.0` (for example) after each release (in `scripts` directory).

Krinkle:
From <https://github.com/kiwix/kiwix-js/pull/675>

> > […]  Clunky workaround! (This is one reason we want to move to GitHub actions when we have time to sort that out - #590 ).
> 
> Sorry if this is obvious, but as I understand it the tests fail for PRs from non-maintainers because the SauceLabs credentials are meant to be secret. If Travis exposed them to PRs, then anyone could submit a PR that just prints the secret to the build output. GitHub Actions or any other CI service should behave the same way. If it doesn't and it does expose it, and if that's what we want, then we shouldn't mark them as secrets but rather set them as public env variables. I would expect the same to happen for GitHub Actions.

-------
 
> I personally favour Travis a bit, as it helps the test setup to be based on open protocols and not vendor locked into this propietary Microsoft platform. Travis isn't fully open-source, but much of it is, and we could of course switch to other service providers in the future (e.g. Circle CI), or other Git hosts (e.g. GitLab).

-------

> It is of course possible that there are (or were) limitations with Travis that affect Kiwix. I only know that the specific issue of repo secrets and SauceLabs, as far as I know, is an intententional limitation that Travis and GitHub Actions both follow the same way 🙂


[...]

PR #626, opened 2020-06-25
fixes #614 

### LINKED ISSUE #614: Use const declarations instead of var wherever constants are declared

TaniaMalhotra:
Changed var declarations for constants to const
Improves readability as well as browser/app memory management.

### DIFF
--- a/service-worker.js
+++ b/service-worker.js
@@ -28,7 +28,7 @@
  * The value is defined in app.js and will be passed to Service Worker on initialization (to avoid duplication)
  * @type {String}
  */
-var CACHE_NAME;
+const CACHE_NAME;
 
 /**
  * A global Boolean that governs whether CACHE_NAME will be used
--- a/service-worker.js
+++ b/service-worker.js
@@ -28,7 +28,7 @@
  * The value is defined in app.js and will be passed to Service Worker on initialization (to avoid duplication)
  * @type {String}
  */
-const CACHE_NAME;
+var CACHE_NAME;
 
 /**
  * A global Boolean that governs whether CACHE_NAME will be used
--- a/nightwatch.js
+++ b/nightwatch.js
@@ -21,7 +21,7 @@
  * along with Kiwix (file LICENSE-GPLv3.txt).  If not, see <http://www.gnu.org/licenses/>
  */
 'use strict';
-var TRAVIS_JOB_NUMBER = process.env.TRAVIS_JOB_NUMBER;
+const TRAVIS_JOB_NUMBER = process.env.TRAVIS_JOB_NUMBER;
 
 module.exports = {
   "src_folders" : ["browser-tests"],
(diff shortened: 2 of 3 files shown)


Jaifroid:
Thanks for this PR @TaniaMalhotra . I'll test shortly.

Jaifroid:
PS Tests are failing because you're an external contributor. I'll push a minor commit the branch to check tests.

Jaifroid:
Tests now passed. @TaniaMalhotra do you see a green "Squash and merge" button below? If so, you could do the squash and merge with that button. But it may be that you don't have the right access to do it. Let me know if not.

kelson42:
@Jaifroid I took on my ton invite @TaniaMalhotra in the Team.

Regarding the CI problem with external contributions, I can say:
* During a few months we have achieved to avoid that problem on Kiwix Android with TravisCI (I don't know the details).
* Migrating to Github Actions (what we have done in most of our repos meanwhile) has solved that problem as well.

Jaifroid:
@kelson42 Thank you! We have #590 to start adding some tests using GitHub actions. I agree it would be good to move to that CI. For now I have a workaround which is to submit a PR of the same git commit (once code is ready for merging), and that triggers correct testing on the external contributor's PR.

service-worker.js
```diff
@@ -28,7 +28,7 @@
  * The value is defined in app.js and will be passed to Service Worker on initialization (to avoid duplication)
  * @type {String}
  */
-var CACHE_NAME;
+const CACHE_NAME;
```
mossroy, line 31: This can't be a constant, as this variable is assigned a value, later in this same file

Jaifroid: @mossroy You're right. I missed that when reviewing. Oddly it didn't throw an error when I tested SW mode. I guess browsers are not treating const strictly.

Jaifroid:
@TaniaMalhotra You will notice that a minor (but important) change is needed as @mossroy stated. When you make this change, could you please try to do it to this PR, rather than closing this and starting a new one? It keeps comments together in case anyone wants to refer back to PR discussions. Thank you!

TaniaMalhotra:
@Jaifroid I have made a change to this PR. I'm extremely sorry for the inconvenience last time. 😥 
@mossroy @Jaifroid Thankyou for reviewing my pull request

Jaifroid:
@TaniaMalhotra Thank you and don't be sorry - that's what review is all about!

Jaifroid:
OK @TaniaMalhotra Tests have passed, so you can squash and merge now using the button below.

Jaifroid:
@TaniaMalhotra Just a quick note that when you merge, we prefer that you squash all your commits into a single commit before merging. That's why I suggested you use the Squash and merge button rather than merge from your repo, because if you use the button it automatically squashes all your commits into one and asks you for a title. It keeps the master branch much neater, as we should only have one commit per PR. This is just friendly advice for future contributions!

TaniaMalhotra:
Sure sir
Will keep in mind from the next time

PR #635, opened 2020-07-09
Updated cookie to ensure samesite is set as Strict.

### LINKED ISSUE #627: Getting SameSite Cookie warnings in Firefox
I am getting warnings in Kiwix JS with Firefox that our Cookies have an insecure `SameSite` attribute that will soon be rejected. There is a link to https://developer.mozilla.org/es/docs/Web/HTTP/Headers/Set-Cookie/SameSite . In fact we don't set the `SameSite` attribute, but this assumes `SameSite=None` (for now) and I believe this is what is causing the warning.

For almost all use cases, we have moved from Cookies to localStorage as a result of #613 . This has been merged into master, but not yet released. However, in some rare contexts (documented in #613) the app will still use Cookies.

It seems the fix would be as simple as setting the Cookie with `SameSite=Lax` or (ideally) `SameSite=Strict`.

Abhirup-99:
Fixes #627 . I used strict as I couldn't find any reason or code to use Lax. The cookie is set only for this and doesn't need to set to any other value.

### DIFF
--- a/www/js/lib/settingsStore.js
+++ b/www/js/lib/settingsStore.js
@@ -62,10 +62,10 @@ define([], function () {
       localStorageTest = false;
     }
     // Now test for document.cookie API support
-    document.cookie = 'tempKiwixCookieTest=working;expires=Fri, 31 Dec 9999 23:59:59 GMT;samesite=Strict';
+    document.cookie = 'tempKiwixCookieTest=working;expires=Fri, 31 Dec 9999 23:59:59 GMT; SameSite=Strict';
     var kiwixCookieTest = /tempKiwixCookieTest=working/.test(document.cookie);
     // Remove test value by expiring the key
-    document.cookie = 'tempKiwixCookieTest=;expires=Thu, 01 Jan 1970 00:00:00 GMT;samesite=Strict';
+    document.cookie = 'tempKiwixCookieTest=;expires=Thu, 01 Jan 1970 00:00:00 GMT; SameSite=Strict';
     if (kiwixCookieTest) type = 'cookie';
     // Prefer localStorage if supported due to some platforms removing cookies once the session ends in some contexts
     if (localStorageTest) type = 'local_storage';
--- a/www/js/lib/settingsStore.js
+++ b/www/js/lib/settingsStore.js
@@ -62,10 +62,10 @@ define([], function () {
       localStorageTest = false;
     }
     // Now test for document.cookie API support
-    document.cookie = 'tempKiwixCookieTest=working;expires=Fri, 31 Dec 9999 23:59:59 GMT; SameSite=Strict';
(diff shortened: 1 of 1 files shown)


www/js/lib/settingsStore.js
```diff
@@ -62,10 +62,10 @@ define([], function () {
       localStorageTest = false;
     }
     // Now test for document.cookie API support
-    document.cookie = 'tempKiwixCookieTest=working;expires=Fri, 31 Dec 9999 23:59:59 GMT';
+    document.cookie = 'tempKiwixCookieTest=working;expires=Fri, 31 Dec 9999 23:59:59 GMT;samesite=Strict';
```
Jaifroid, line 65: @Abhirup-99 Please use the coding conventions / capitalization and spacing indicated in [MDN documentation](https://developer.mozilla.org/es/docs/Web/HTTP/Headers/Set-Cookie/SameSite), e.g. (this is their example):

`Set-Cookie: flavor=choco; SameSite=Secure`

So here you want a space after semicolon, and use correct capitalization. I know cookie headers are case-insensitive, but MDN is a respected source and it makes the code neater to use consistent conventions.

www/js/lib/settingsStore.js
```diff
@@ -62,10 +62,10 @@ define([], function () {
...
     var kiwixCookieTest = /tempKiwixCookieTest=working/.test(document.cookie);
     // Remove test value by expiring the key
-    document.cookie = 'tempKiwixCookieTest=;expires=Thu, 01 Jan 1970 00:00:00 GMT';
+    document.cookie = 'tempKiwixCookieTest=;expires=Thu, 01 Jan 1970 00:00:00 GMT;samesite=Strict';
```
Jaifroid, line 68: Same here.

Abhirup-99:
@Jaifroid I have updated it for samesite. Do you want it for other properties also?

Jaifroid:
If you can make that consistent, it would be great - it improves legibility of the code. this file was originally a library file, but we've modified it quite a lot, so it no longer is.

Abhirup-99:
I have updated the expires part to add the spacing. I didn't find any other issue. Take a look whenever you are free.

Jaifroid:
@Abhirup-99 This has passed automatic testing and my own manual tests. Thank you.

Do you see a green button at the bottom of this PR that looks like image below? It's possible you aren't able to use it, but if you are able to, please use this functionality to **squash the PR to one single commit** before merging. Don't try to do this from your own repo unless you know all about squashing in Git. This repo has a policy of only one commit per PR. If you are unable to use that button (because you are an external contributor), then let me know.

image

Abhirup-99:
I don't have write access to the repo, and so don't have the required permissions to do it and its not visible.

Jaifroid:
OK, no problem, I'll merge it. Thanks for letting me know, and thanks for your PR.

Issue #637, opened 2020-07-09
Optimize title search so that cancelled binary search Promises return at the earliest possible opportunity


Jaifroid:
This relates to the optimization of #617. Although searches are successfully cancelled when a user starts a new search or exits the search UI, and extraneous results from a cancelled search do not display in the UI, a certain amount of binary search processing still continues from Promises which had been set up very early on in the search process. It should be possible to return a null result, or use `throw`, from those Promises to prevent unnecessary CPU utilization. This has to be done without simply dereferencing parameters and causing potential memory leaks. It is similar to #426, where a fair amount of processing for a previous article continues after a user has navigated away to another article. Similar techniques could be used to solve both situations.

https://github.com/kiwix/kiwix-js/pull/617#issuecomment-656111971 is particularly relevant to understand the issue here.

Jaifroid:
So I don't forget, here is a quick summary of what needs to be done following on from #617:

* Move `globalstate` object from init.js to app.js; possibly rename `appstate`;
* Pass the `appstate.search` object to zimArchive.js;
* When a search is cancelled, set the `appstate.search.status` property to 'cancelled';
* When a new search is undertaken, also set `appstate.search.status` to 'cancelled': this will be reflected in the search object passed to zimArchive.js because they both still point to the same underlying object;
* Instantiate a new search object, and set `appstate.search` to this new object; the one passed to zimArchive.js will still point to the old object according to [Javascript Pointer Reference craziness](https://stackoverflow.com/questions/8318357/javascript-pointer-reference-craziness-can-someone-explain-this);
* Insert an early cancellation at the very start of the binary search loop: if a search status is cancelled, return 0 instead of -1 or +1 -- this causes the loop to exit early;
* Once all Promises have returned, the old search object is no longer referenced, and should be garbage collected automatically in due course.

Jaifroid:
If this works, extend the model to #426, which is a very similar case: we should be able to use the same appstate object to pass an `appstate.article` object, which would have a property `appstate.article.status`. In Service Worker mode, if we are allowed to emit a signal from the message channel that the user has clicked on a new article link, then this solution should be viable.

PR #642, opened 2020-07-13
Enable quick and complete cancellation of running searches

Jaifroid:
This is proof-of-concept code for the first part of the PR title (cancellation of running searches). It implements #637 and is in draft, for testing.

Because I believe the techniques and solution are the same, I intend to address #426 in this PR as well.

The code currently contains console log reports to prove that it cancels title search completely. They will be removed at a later stage.


Jaifroid:
Here is the output of console log running the above on two title searches in succession. The first is for "unesco world heritage site"; after a short time this is quickly replaced with "unesco world heritage". The former search is complex with many variations searched, and it is replaced by a simpler search.

Asterisks indicate where the first search was cancelled, and where the cancellation was registered - note that it is after the global search object has been changed, which proves that the running search object in `zimArchive.js` updates its `status` property to "cancelled" when it is modified in the global search object. When the global search object is completely replaced, the `zimArchive` search continues to point to the old search object that it was launched with (which has a modified `status` of "cancelled").

Hence, we have an effective signalling mechanism without recourse to "super-global" objects, and it is sufficient to pass all the needed information to zimArchive in a single `appstate.search` object.

image

Jaifroid:
Having looked back over this, I now believe it is too ambitious to address #426 in this PR. However, it implements #637 and could be merged as an easy PR (after removal of console.log entries).

Jaifroid:
I'm listing here the stages I mentioned in #637. I believe they have all been implemented, but I'll tick them off as I check through them before removing Draft status for this PR:

* [ ] Move globalstate object from init.js to app.js; possibly rename appstate;
* [ ] Pass the appstate.search object to zimArchive.js;
* [ ] When a search is cancelled, set the appstate.search.status property to 'cancelled';
* [ ] When a new search is undertaken, also set appstate.search.status to 'cancelled': this will be reflected in the search object passed to zimArchive.js because they both still point to the same underlying object;
* [ ] Instantiate a new search object, and set appstate.search to this new object; the one passed to zimArchive.js will still point to the old object according to Javascript Pointer Reference craziness;
* [ ] Insert an early cancellation at the very start of the binary search loop: if a search status is cancelled, return 0 instead of -1 or +1 -- this causes the loop to exit early;
* [ ] Once all Promises have returned, the old search object is no longer referenced, and should be garbage collected automatically in due course.

Jaifroid:
@mossroy I have checked over the code (see checkboxes above) and added a short technical explanation of how the object assignment works between the different scopes, as it may be useful to my future self or other future devs encountering the code. But otherwise, this is ready for review. It effectively completes #617 (which was merged despite the known issue) by implementing #637.

NB I have left in (for now) `console.log` output that verifies precisely what is happening and that the old search is cancelled in its own scope as soon as possible. If you wish to test, follow this procedure: open a large ZIM like Wikipedia English all, and search for something that will take a fairly long time to search for in all its permutations, e.g. "unesco world heritage site". (use all lowercase) After a few seconds quickly backspace in the search field to delete "site" so that a new search is launched. Observe console.log entries as per screenshot above.

I will delete diagnostic console logging before squash/merge once the code is approved.

mossroy:
I'm not forgetting this PR, and will review it as soon as I can.
But it's hard for me to find some time

Jaifroid:
No rush, I sympathize -- things going crazy here too.

www/js/app.js
```diff
@@ -974,10 +980,15 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
      */
     function searchDirEntriesFromPrefix(prefix) {
         if (selectedArchive !== null && selectedArchive.isReady()) {
-            // Store the new search term in the globalstate.search object and initialize
-            globalstate.search = {'prefix': prefix, 'status': 'init', 'type': ''};
+            // Cancel the old search (zimArchive search object will receive this change)
+            appstate.search.status = 'cancelled';
+            // Initiate a new search object and point appstate.search to it (the zimAcrhive search object will continue to point to the old object)
```
mossroy, line 985: Minor typo here: zimArchive instead of zimAcrhive

www/js/app.js
```diff
@@ -991,23 +1002,27 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
...
-     * @param {Object} reportingSearchPrefix The prefix of the reporting search
+     * @param {Object} reportingSearch The reporting search object
      */
-    function populateListOfArticles(dirEntryArray, reportingSearchPrefix) {
-        // Do not allow cancelled or changed searches to report
-        if (globalstate.search.status === 'cancelled' || globalstate.search.prefix !== reportingSearchPrefix) return;
-        var stillSearching = globalstate.search.status === 'interim';
+    function populateListOfArticles(dirEntryArray, reportingSearch) {
+        // Do not allow cancelled searches to report
+        if (!/cancelled/.test(reportingSearch.status)) console.log("Reporting search is: " + reportingSearch.prefix + " [" + reportingSearch.status + "]");
```
mossroy, line 1009: Why are you using a regular expression here, instead of comparing the string? If you prefer to keep it, it would be safer (and faster) to match the whole string instead of a subtring : add `^` at the beginning and `$` at the end)

www/js/init.js
```diff
@@ -30,13 +30,6 @@
  */
 var params = {};
 
-/**
- * A global object for storing app state
- * 
- * @type Object
- */
-var globalstate = {};
```
mossroy, line 38: It's good that this global object is removed from here

mossroy (approved):
It seems to be working well on my FirefoxOS device.
I added a few minor remarks

www/js/app.js
Jaifroid: Because I love regexes ;-) ... However, I think I can just remove this, as it's a diagnostic `console.log` we don't want in the final code.

Issue #611, opened 2020-03-31
Add support for zstd decompression


kelson42:
We just about to add support of zstd compression in the libzim:
https://github.com/openzim/libzim/issues/84

The reason is that zst can decompress around x6 time faster than xz (all the other aspects are similar to xz).

This will be shipped in the next libzim release within a week.

I guess all libzim based ports (basically everything except Kiwix JS) will have it included before summer.

As a consequence I expect to start to create ZIM file using zstd during the summer.

This is why I create this ticket to ask if zstd could be supported soon by Kiwix JS?

Jaifroid:
Hmm. This depends entirely on whether a decompressor has been written in JavaScript, or whether it can be compiled to webassembly. If not, and if we continue to be unable to compile libzim to webassembly, this would be one big breaking change that could make Kiwix JS obsolete overnight!

kelson42:
@Jaifroid https://www.npmjs.com/package/zstd-codec

Jaifroid:
@kelson42 That looks like very good news! They specifically say it is an Emscripten version "for Node and the Web", which sounds like what we need.

mossroy:
Maybe, but it might be not so easy.
I don't see in this library an API identical to the one we use for XZ format (see https://github.com/kiwix/kiwix-js/blob/master/www/js/lib/xzdec_wrapper.js#L59). But maybe it can be adapted.

I was supposed to work on #514 at the hackathon, that would supersede this issue. The fact that the hackathon has been canceled (for good covid-19 reasons) changed the plans.
Will the hackathon be re-scheduled later this year? Or simply canceled, and the next one would be in 2021? Maybe it's too early to say...
In any case, I'd prefer to work on #514 (at next hackathon, with the help of the other mates), instead of manually adding zstd support.

kelson42:
I know it does not help, but I'm quite sorry to inform you so lately, but this is something I had in the back of my mind for quite a bit of time and the move is ad the end relatively easy for the C++ backed software... But, it is true, for JS, this is challenging.

kelson42:
Here is a demo file with zstd coimpression http://tmp.kiwix.org/foo-zstd.zim

Jaifroid:
[...]

PR #648, opened 2020-08-31
Add zstd decompression support

Jaifroid:
Implements #611.

I'm creating this PR for testing, as the code is now working well on two test ZIMs: `wikipedia_bm_all_2020-07.zstd.zim` and `foo-zstd.zim`. Please note that I have not yet compiled zstdec.js with optimizations and minification, but it is very fast even as is.

I have tested this on Edge Chromium 85.0, Firefox 80.0, Edge Legacy, and IE11 - all working. I have not tested it on Firefox OS simulator yet, but I have tested on Firefox ESR which is the same codebase as the simulator, and it works in Firefox ESR. It might need to be memory optimized before it will work on lower-powered mobile devices, and at the least the binary should be compiled with -O2 optimizations to make it smaller.

However, I would appreciate any testing at this stage, even while the PR is in draft form. I'll add links to the test ZIMs in a separate post on this PR.

@mossroy I've put you as a reviewer, but don't review the code yet, just do some tests if you have time.


Jaifroid:
Links to test ZIMs:

https://github.com/kiwix/kiwix-js/raw/Add-zstd-decompression-support/tests/wikipedia_bm_all_2020-07.zstd.zim

https://github.com/kiwix/kiwix-js/raw/Add-zstd-decompression-support/tests/foo-zstd.zim

With this last one, you have to type a number in the search box to load an article. They are very short text-based articles.

Jaifroid:
I have recompiled `zstdec.js` with [O1 optimizations](https://emscripten.org/docs/tools_reference/emcc.html#arguments). This reduces the asm file size from 954KB to 255KB. I tried compiling with O2 and O3 optimizations, but these both produce compile-time errors `clang-12: error: clang frontend command failed due to signal (use -v to see invocation)`. This could be due to C++ name mangling? However, it is very fast with O1 optimizations, and seems to me to be "fast enough". So long as we don't run into memory issues on low-end mobile devices, O1 may be sufficient. And even if we do run into memory issues, it would be more important to try changing the value of `zd._outBuffer.size` first and recompile with lower `TOTAL_MEMORY`, as this would be the more important reduction in memory usage.

I have kept the original unoptimized `zstdec.js` in the emscripten/zstandard directory as repository stores may ask to see human-readable code (even though O1 does not minify).

Jaifroid:
Compiling with `--closure 1` (closure compiler) reduces size even further to 201KB.

Jaifroid:
Working on FFOS 2.2 simulator.

image

Jaifroid:
@kelson42 Do we have any larger ZSTD ZIM files now available that I can test this implementation with? I want to be sure that long articles do not cause memory issues, since most of the articles in the bm_all_2020-07 file are very short.

www/js/lib/zstdec_wrapper.js
```diff
@@ -0,0 +1,284 @@
...
+    var busy = false;
+    
+    /**
+     * @typedef Decompressor
+     * @property {Integer} _chunkSize The amount to feed to the decompressor in any one read loop
+     * @property {FileReader} _reader The filereader to use (uses plain blob reader defined in zimfile.js)
+     * @property {Integer} _inStreamPos The current known position in the steam of compressed bytes 
+     * @property {Integer} _inStreamChunkedPos The position once the currently loaded chunk will have been consumed
+     * @property {Integer} _outStreamPos The position in the decoded byte stream (offset from start of cluster)
+     * @property {Array} _outDataBuf The buffer that stores decoded bytes (it is set to the requested blob's lenght, and when full, the data are returned)
```
Jaifroid, line 59: Typo: "length"

www/js/lib/zstdec_wrapper.js
```diff
@@ -0,0 +1,284 @@
...
+    
+    /**
+     * Is the decompressor already working?
+     * @type Boolean
+     */
+    var busy = false;
+    
+    /**
+     * @typedef Decompressor
+     * @property {Integer} _chunkSize The amount to feed to the decompressor in any one read loop
```
Jaifroid, line 54: Change "amount" to "number of compressed bytes"

www/js/lib/zstdec_wrapper.js
```diff
@@ -0,0 +1,284 @@
...
+     * @property {FileReader} _reader The filereader to use (uses plain blob reader defined in zimfile.js)
+     * @property {Integer} _inStreamPos The current known position in the steam of compressed bytes 
+     * @property {Integer} _inStreamChunkedPos The position once the currently loaded chunk will have been consumed
+     * @property {Integer} _outStreamPos The position in the decoded byte stream (offset from start of cluster)
+     * @property {Array} _outDataBuf The buffer that stores decoded bytes (it is set to the requested blob's lenght, and when full, the data are returned)
+     * @property {Integer} _outDataBufPos The number of bytes of the requested blob decoded so far
+     * @property {Object} _inBuffer A JS copy of the inBuffer structure to be set in decompressor memory (malloc)
```
Jaifroid, line 61: Change "decompressor memory" to "asm memory"

www/js/lib/zstdec_wrapper.js
```diff
@@ -0,0 +1,284 @@
...
+    
+    /**
+     * @constructor
+     * @param {FileReader} reader
+     * @param {Integer} chunkSize
+     * @returns {Decompressor}
+     */
+    function Decompressor(reader, chunkSize) {
+        this._chunkSize = chunkSize || 5 * 1024;
+        // this._chunkSize = chunkSize || zd._ZSTD_DStreamInSize();
```
Jaifroid, line 73: Add comment to explain commented-out line

www/js/lib/zstdec_wrapper.js
```diff
@@ -0,0 +1,284 @@
...
+        this._reader = reader;
+    }
+    /**
+     * Read length bytes, offset into the decompressed stream. Consecutive calls may only
```
Jaifroid, line 77: This comment is opaque. Add clearer explanation.

www/js/lib/zstdec_wrapper.js
```diff
@@ -0,0 +1,284 @@
...
+        this._inBuffer = {
+            ptr: null,      /* pointer to this inBuffer structure in w/asm memory */
+            src: null,      /* void* src   < start of input buffer */
+            size: length,   /* size_t size < size of input buffer */
+            pos: 0          /* size_t pos; < position where reading stopped. Will be updated. Necessarily 0 <= pos <= size */
+        };
+        // Reserve w/asm memory for the outBuffer structure
+        this._inBuffer.ptr = mallocOrDie(3 << 2); // 3 x 32bit bytes
+        // DEV: Size of outBuffer is currently set as recommended by zd._ZSTD_DStreamOutSize() below; if you are running into
+        // memory issues, it may be possible to reduce memory consumption by setting asmaller outBuffer size here and
```
Jaifroid, line 101: typo: "a smaller"

mossroy (commented):
I tested this branch on my Firefox OS device (ZTE Open C with 512 MB of RAM), with foo-zstd.zim and wikipedia_bm_all_2020-07.zstd.zim. Both are working with no issue. And I can see the debug logs of zstddec-wrapper.js, like on desktop browsers.
I can test larger ZIM files on this device when we'll have some.

I'm truly impressed by your work on this : I know it's complicated and low-level code, and a technology (emscripten) that you had not tackled so far. All my congratulations for this achievement!
And sorry for not being available these times

www/js/lib/zstdec_wrapper.js
```diff
@@ -0,0 +1,284 @@
...
+            // zd._ZSTD_freeDStream(zd._decHandle);
+            busy = false;
+            console.log("Freed all data structures.");
+            return data;
+        });
+    };
+
+    /**
+     * Reads stream of data from file offset for length of bytes to send to the decompresor
+     * This function ensures that only one decompression runs at a time
```
Jaifroid, line 136: It would be clearer to specify that the file offset is the beginning of a compressed cluster...

www/js/lib/zstdec_wrapper.js
```diff
@@ -0,0 +1,284 @@
...
+     * @param {Integer} offset The file offset at which to begin reading compressed data
+     * @param {Integer} length The amount of data to read
+     * @returns {Promise} A Promise for the read data
```
Jaifroid, line 139: This should be `{Promise<Int8Array>}`

kelson42:
We will try to make bigger files with zstd within a week.

www/js/lib/zstdec_wrapper.js
```diff
@@ -0,0 +1,284 @@
...
+            }
+
+            // Get updated inbuffer values for processing on the JS sice
+            // NB the zd.Decoder will read these values from its own buffers
+            var ibx32ptr = that._inBuffer.ptr >> 2;
+            that._inBuffer.pos = zd.HEAP32[ibx32ptr + 2];
+            
+            // Get updated outbuffer values
+            var obx32ptr = that._outBuffer.ptr >> 2;
+            // that._outBuffer.size = zd.HEAP32[obx32ptr + 1];
```
Jaifroid, line 189: Redundant code: outBuffer.size never changes, so delete.

www/js/lib/zstdec_wrapper.js
```diff
@@ -0,0 +1,284 @@
...
+            } else {
+                return that._readLoop(offset, length);
+            }
+        });
+    };
+    
+    /**
+     * Fills in the instream buffer if needed
+     * @param {Integer} currOffset The current read offset
+     * @param {Integer} len The decompressed length of data requested
```
Jaifroid, line 227: add ... "in the decompressed stream"

www/js/lib/zstdec_wrapper.js
```diff
@@ -0,0 +1,284 @@
...
+            
+            // Transfer the (new) data to be read to the inBuffer
+            zd.HEAP8.set(data, that._inBuffer.src);
+            that._inStreamChunkedPos += data.length;
+            return 0;
+        });
+    };
+
+    /**
+     * Provision asm/wasm data block and get a pointer to the assigned location
```
Jaifroid, line 267: Add acknowledgement of useful tutorial where you go this code from

zstandard/dist/EncoderAndDecoderNodeJS.src.js
```diff
@@ -0,0 +1,134 @@
+﻿(function(global){
```
Jaifroid, line 1: This file is from another distribution I considered using, and should be deleted now that we have compiled our own (much smaller) version.

Jaifroid:
> I know it's complicated and low-level code, and a technology (emscripten) that you had not tackled so far. All my congratulations for this achievement!

Thank you @mossroy. Your words mean a lot to me! I learnt a lot doing this! Of course I relied heavily on the existing xzdec wrapper, and a lot of trial and error until I understood how all the loops worked and when to feed new data, when to extract decoded data from the asm memory buffers.

Thank you for the test: it's a relief that it works on older physical devices. It's working on UWP for desktop, I just now need to test on a couple of Windows mobiles I have with different memory sizes, and also on the low-memory mobile simulator in Visual Studio. Not that the main target is mobile any more, but it's a good test for latent memory issues I think.

A file with larger articles will be an important test too. Reading physically large ZIM files is no issue because the new code only affects decompression of strings. But a very large page will require looping through a lot of data and will be important to test for memory leaks, and to see if compiled TOTAL_MEMORY is large enough.

Jaifroid:
The Kiwix JS Windows port of this implementation, with `wikipedia_bm_all_2020-07.zstd.zim` loaded, runs fine on a 1GB Windows Mobile VM.

However, on Mobile Emulator `10.0.15063.0 WVGA 4 inch 512MB` (a very low-resourced device), although the app can load a couple of pages, it crashes frequently on cycling through ZIM articles, which is clearly due to the memory requirements of the decoder. That VM struggles also with XZ-encoded ZIMs, so it's not a surprise and not a priority. However, it may be indicative of the kind of issue that older hardware can run into, especially as the NWJS version of the app can run on Windows XP.

I have some ideas for reducing the memory requirement. As mentioned above, this involves a smaller `outBuffer.size` and recompiling with smaller memory footprint. Also commented in the code is an idea for re-using all allocated memory structures, so that the malloc allocations are not cleared after each blob is retrieved as at present. This would be possible as @mossroy wrote code to ensure only one blob is decoded at a time, so we do not need independently allocated memory space. The buffers are always the same size and are emptied as soon as they are filled, so allocating and freeing them for each blob is potentially a source of memory leaks.

Relevant comment from C++ source code is this (NB we already re-use `ZSTD_DCtx`, the decompression stream control object):

```
/*= Decompression context
 *  When decompressing many times,
 *  it is recommended to allocate a context only once,
 *  and re-use it for each successive compression operation.
 *  This will make workload friendlier for system's memory.
 *  Use one context per thread for parallel execution. */
```

Jaifroid:
With the latest memory optimizations, I can now read `wikipedia_bm_all_2020-07.zstd.zim` on the  512MB Windows Mobile VM without crashes. This is better than XZ decompression, which crashes on that VM every third article load. So I believe we now have a memory improvement as well as overall faster decode speed (compared to XZ). Reducing size of outBuffer does lead to more readLoop cycles, hence very slight decrease in decode speed compared with using larger buffers, but I don't think it's appreciable.

I implemented idea 2 above, i.e. re-using all allocated memory structures for consecutive blob reads. I also fixed an error in the read loop which incorrectly set the size of the inBuffer larger than was required by the decoder (this may also have been wasting memory, but more seriously produced a "corrrupted stream" message under certain conditions with small chunkSize). I wasn't able to reduce the TOTAL_MEMORY constant, but this doesn't seem to be the choke point, since stability in low-memory environments seems no longer to be an issue.

One thing bugs me a bit: on https://github.com/facebook/zstd/tree/dev/contrib/single_file_libs it is claimed that "The decompression library is small, adding, for example, 26kB to an Emscripten compiled WebAssembly project". However, the smallest compile I could make is 201kB. Possibly this is because I've compiled to asm, not wasm, for compatibility with older systems. Or it may be simply because I have been unable to compile with optimizations higher than -O1 (as mentioned above).

Jaifroid:
@kelson42 It's finished bar testing, and possibly optimization if testing throws anything up. We only have that bm test ZIM so far (and foo-zstd.zim), and I need to be sure that it's correctly decompressing large pages that require many iterations of the streaming loop, hence keeping it as draft until we have larger ZIMs with more complex pages. E.g. a WikiMed or a Wikivoyage ZIM would be great.

kelson42:
@Jaifroid I have created https://github.com/openzim/mwoffliner/issues/1255 to allow us to create zstd based ZIM files via the Zimfarm.

kelson42:
@Jaifroid @mossroy Here is a slightly bigger Zstandard encoded ZIM file http://tmp.kiwix.org/wikipedia_en_medicine_2020-09_zstd.zim

Jaifroid:
@kelson42 Thanks, downloading it now!

Jaifroid:
OK, most long articles work, but I've found one that reproducibly fails: "Down syndrome" (it's under Children's Health). This aborts with "Failed to decompress data stream: corrupted block detected". I have checked with Kiwix Desktop, and the same article loads fine from the same ZIM. So there's a bug. 😢 (I noticed it with a few other articles too.)

However, the console log counters show some interesting negative numbers which suggest an offset calculation problem, whcih I think will help me to locate the issue.

Jaifroid:
If I use the maximum outBuffer size that is recommended by the function `_ZSTD_DStreamOutSize()` (on my machine/browser, this is 131,072 bytes), then the Down Syndrome article decodes correctly. This is guaranteed to be the maximum amount of memory that will be used to decode a single complete frame. However, the code should be able to handle partial frame decodes and a smaller buffer, so I'll try to do a bit more investigation before settling for a larger buffer. By modern standards, 131,072 bytes is not large, but it could be an issue with older hardware.

image

kelson42:
@Jaifroid Good new, I hardly see how a few 131KB of memory could be a problem, even on an old device.

Jaifroid:
@kelson42 Agreed, it shouldn't be too much of an issue, and there are other optimizations we can do (not initialize xzdec and zstdec together on first load of Kiwix JS, but only on detecting the decompression type of the ZIM).

In any case, my tests this morning suggest that ZSTD is now stable in this PR.

Do you have a timeline for publishing ZSTD-encoded ZIMs? There's no obstacle any more from our end.

(A QUnit test is failing on Chrome and Edge, but it's unlikely to be anything to do with this PR; there seems to be a race condition in search which occasionally causes tests to fall on master, and I hope this is addressed by #642.)

kelson42:
@Jaifroid Everything is ready to start make public ZIM files with zstd. I'm only waiting the Kiwix JS based releases including your PR.

Jaifroid:
OK, in that case let's just go with this larger outBuffer and prioritize this PR.

Jaifroid:
@mossroy I've marked this ready for review. There are several diagnostic console log statements I shall remove after review, but if you'd prefer me to remove them before review, just let me know. I know you have very limited time, so I suggest this PR is more important than #642,

mossroy:
I did not look at the code, but tested this PR on my old Firefox OS device (ZTE Open C), with http://tmp.kiwix.org/wikipedia_en_medicine_2020-09_zstd.zim (I had to split it in 2GB files)

It works perfectly on it. Great work!

It's difficult to compare speed, as I don't have the same file compressed with xz.
I'll look at the code later. Keep in mind that we need to leave enough information in the repo to easily recompile with emscripten. It's required at least by Mozilla

kelson42:
@mossroy Thx for the testing, here the XZ encoded version of the file https://download.kiwix.org/zim/wikipedia/wikipedia_en_medicine_2020-09.zim

Jaifroid:
Thanks @mossroy - that's good news! 
I think there is enough in the repo to reproduce. See the readme at:
https://github.com/kiwix/kiwix-js/tree/Add-zstd-decompression-support/emscripten/zstandard (in this branch):

# Source code and compilations of zstandard decompression library

The file `zstddeclib.c` is the single-file source code for the zstandard decompressor (compressor is not included).
It has been combined by following the instructions at https://github.com/facebook/zstd/tree/dev/contrib/single_file_libs. It can be used as the source file of an Emscripten webassembly or asm compilation.

The source code can be compiled with Emscripten by setting up the Emscripten SDK and running the Linux script `./compile.sh` in this directory. This will produce `zstdec.js`. Further optimization and minification of this file may be possible by adjusting the commandline options in `compile.sh`.

For a fuller Emscripten distribution of the full zstandard library for `node.js`, see https://github.com/yoshihitoh/zstd-codec/tree/develop/js/example.

mossroy:
@Jaifroid : the README to re-compile zstd with emscripten looks great.

On my ZTE Open C, the zstd version is slower than the xz version. It depends on the article : between 5 to 30%.
I thought it was due to the debug logs, but I see the same after commenting them.
On the WikiMed ZIM files, it's especially noticeable on first access to the main page, when caching the assets : it takes around twice more time to cache these assets on the zstd version

mossroy:
In any case, the performance should not be a blocker of this PR.
What's important is to be able to read these newer ZIM files in a reliable way, which is the case.

kelson42:
@mossroy This is extremly strange as it should be at least 5x faster.

Jaifroid:
With the African ZIM, I was very impressed with the decompression speed in the desktop browser: it seemed almost instant. But first impressions can be very deceptive.

On my main Windows Mobile physical device (Lumia 950XL with 3GB RAM), it is on a par with xz, not noticeably faster, not noticeably slower, but we tuned the xz performance a lot and it has not been tuned on zstdec.js, in particular it has not been compiled with the most aggressive options (it refused to compile with higher than -O1 optimizations on my WSL V2 Linux, which leaves the decoder unminified, for example and larger than necessary). Maybe in a different compile environment it will be possible to use better optimizations.

Stability, however, is greatly improved: on a VM of a bottom-of-the-line 4.5" 500MB memory device which crashes every second or third article on xz, the zstdec African ZIM is completely stable, which suggests better memory usage.

My hunch is that on an old, slow devices, the overhead of an unminified, less-optimized Emmscripten outweighs the decompressions speed gains. This is an asm compile (like xzdec) for maximum compatibility, but the current Emscripten SDK is geared towards wasm, and I have throttled it back in options to produce asm only and support IE11.

We could experiment further with optimizations in a diferent PR. I think a lot could be done with asm, without even considering wasm...

Jaifroid:
If we believe the claims made in the [source documentation](https://github.com/facebook/zstd/tree/dev/contrib/single_file_libs) (screenshot below), we should be able to achieve a wasm binary of only 26 KB, compared with my asm compile of 201 KB (nine times larger). But wasm won't support any of these old systems (IE11, FFOS, Windows Mobile). 

image

Jaifroid:
There's some news: I updated to the latest EMSDK, and recompiled with -O2 optimizations successfully (it had refused before). This cuts the file size down to 122KB (pushed in latest commit). I will try to match the xzdec compilation level in subsequent commits.

Jaifroid:
Now down to 120KB with -O3 optimizations, which are supposed to provide performance gain too. I had to add an explicit export of _malloc and _free functions, as they've been removed from default compiles since EMSDK 2.4.

@mossroy Would you like me to remove console logging now?

Further optimizations might be achieved by following https://floooh.github.io/2016/08/27/asmjs-diet.html (but checking for the latest versions of the commands that dev uses, as API has changed a lot since 2016). It would be fun to try WASM, but asm always seems "fast enough" at least in browser context.

mossroy:
I think you can remove the debug logs

Jaifroid:
Debug logs removed.

I experimented with compiling for wasm. It produces a wasm binary (`zstdec.wasm`) of just 46 KB. However, even using the options WASM = 2, which is supposed to emit both a wasm binary and a JavaScript aleternative, IE11 fails to load at all with this build. The documentation is rather terse and just says "IE11 is end-of-life and doesn't support wasm", which of course we knew, but I was expecting it would have some built-in logic to use ASM if WASM is not available, rather than just crashing with "'WebAssembly' is undefined". I didn't attempt to test on old mobiles: I'm pretty sure they just wouldn't work.

Testing a wasm build in Chrome, I couldn't really see any noticeable performance improvement over asm. It's not as instant as, say, the Kiwix Android app. Certainly no performance gain worth dropping support for old platforms for.

We *could* do our own filtering logic. We include a wasm and an asm build, and load the appropriate one with a test. It wouldn't be hard I think, but I'm not sure it's worth it.

If you want to test the wasm yourself for performance, I've pushed a wasm folder you'll see in the latest commit. Just take the two files in there and drop them into `www/js/lib` , overwriting the existing `zstdec.js` (if you keep existing `zstdec.js` by renaming it, remove the `.js` extension so that requirejs doesn't try to run it alongside the wasm version).

(I managed to shave an extra 2KB off the asm version with a setting explicitly setting the expected runtime environment.)

mossroy:
Sorry, but the performance decrease still looks similar, with latest commits of this PR, on my ZTE Open C.
Never mind, it was worth trying.

Enhancing the performance might be done later.
The optimization options of emscripten are probably interesting. Anyway, the size of the generated file is not important in our use-case, as it's not transferred over a network.

As you said, when we tested performance of wasm against asm.js on xzdec, the difference was not very important, so we decided it was not worth it (as the fallback from wasm to asm.js was not working by default either). But the result might be different for zstd.

Jaifroid:
@mossroy I'm looking at another issue that *might* affect performance. There may be an inefficiency in the looping and interaction with `fillInBufferIfNeeded` causing more file reads than strictly necessary. Am investigating. (I definitely need to edit the code there because of some callee parameters are not consumed by the function -- accidentally left in from a previous implementation I experimented with.) 

Regarding Emscripten optimizations, yes I noticed the performance was the basically the same after the optimizations. I have included a wasm build (add it manually to the js/lib directory as mentioned if you want to test it), but clearly it only works on wasm-enabled browsers.

Jaifroid:
I've pushed some interim cleanup code which is still fully functional. However, I'm still not 100% happy that I haven't yet been able to detect in the most efficient way when the buffer should be fillled in. The test for it is currently nonsense (or at least I don't understand why it works). I suspect it works inefficiently as a result (too many file reads).

Jaifroid:
Testing more efficient use of read loop...

Jaifroid:
Well, I've been round various possibilities which I've left in the commit log, but they all bug out at some point due to having to refill the buffer before it is completely consumed. And I'm running out of time to do anything else...

Jaifroid:
I've reverted back to a working version from last night.

kelson42:
@mossroy @Jaifroid What stops usnow concretly to merge? Still a bit of code review?

Jaifroid:
It can be merged (after review). I was trying to address the performance concerns, as I suspect this code is doing more file reads than necessary. However, if it is thoroughly tested on the ZIMs we have, then a pragmatic merge can be done (as I already did with Kiwix JS Windows). Making the code more efficient could be done in a separate PR.

www/js/lib/zstdec_wrapper.js
```diff
@@ -0,0 +1,265 @@
...
+    };
+
+    /**
+     * This function ensures that only one decompression runs at a time, launching readSlice() only when
+     * the decompressor is no longer busy
+     * @param {Integer} offset The cluster offset (in decompressed stream) at which the requested blob resides
+     * @param {Integer} length The number of decompressed bytes to read
+     * @returns {Promise} A Promise for the readSlice() function
+     */
+    Decompressor.prototype.readSliceSingleThread = function (offset, length) {
```
mossroy, line 151: This Single Thread feature was necessary with xz decoder to keep memory consumption reasonable (#389). Is it also necessary for zstd? (it might be considered as an enhancement, and tested later)

www/js/lib/zstdec_wrapper.js
```diff
@@ -0,0 +1,260 @@
...
+'use strict';
+define(['q', 'zstdec'], function(Q) {
+    // DEV: zstdec.js has been compiled with `-s EXPORT_NAME="ZD" -s MODULARIZE=1` to avoid a clash with xzdec which uses "Module" as its exported object
+    // Note that we include zstdec above in requireJS definition, but we cannot change the name in the function list
+    // There is no longer any need to load it in index.html
+    // For explanation of loading method below to avoid conflicts, see https://github.com/emscripten-core/emscripten/blob/master/src/settings.js
+    
+    /**
+     * The ZSTD Decoder instance
+     * @constructor Constructs the zd object representing a ZSTD decoder Emscripten instance
```
mossroy, line 31: I don't see which constructor this corresponds to

Jaifroid:
OK, I've had a couple of ideas. I've changed `fillInBufferIfNecessary` to `fillInBuffer` on the basis that the decompressor always consumes the full inBuffer, because we know the stream cannot end until we have found our Blob. If we find the requested blob in the middle of a chunk, we copy the bytes we need and abandon the rest of the decompression (i.e. we don't continue to decompress to the end of the frame).

And the second idea was to increase our chunk size the the maximum recommended by the decompressor. This is a much larger chunk size than we use with xz (from which I derived the warapper code), and I think it significantly speeds up the operation.

The theoretical downside, as yet untested, is whether the large chunk size will cause memory issues on low-memory machines. However, I think we might be safe because everything is happening within the virtual C++ environment of Emscripten, and TOTAL_MEMORY is fixed. Emscripten is *not* saying it needs more memory in order to cope with the larger chunk size.

If @mossroy can run this on the ZTE, then that would be a good test.

mossroy (changes requested):
I've made a few remarks, that should be easy to handle.
I have to admit that I did not go deeply into zstdec_wrapper.js : not enough time (and will) for that

The main missing thing is a few unit tests. Maybe using foo-zstd.zim?

www/index.html
```diff
@@ -399,13 +399,13 @@ <h3>Expert settings</h3>
             </footer>
         </section>
 
-        <script type="text/javascript" src="js/lib/xzdec.js"></script>
```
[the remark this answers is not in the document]
Jaifroid: Thanks!

emscripten/zstandard/compile.sh
```diff
@@ -0,0 +1 @@
+emcc --memory-init-file 0 -O3 --closure 1 -s ENVIRONMENT="web" -s WASM=0 -s MALLOC="emmalloc" -s TOTAL_MEMORY=140247040 -s FILESYSTEM=0 -s DOUBLE_MODE=0 -s DYNAMIC_EXECUTION=0 -s MIN_IE_VERSION=11 -s EXPORT_NAME="ZD" -s MODULARIZE=1 -s EXPORTED_FUNCTIONS="['_malloc', '_free', '_ZSTD_createDStream', '_ZSTD_initDStream', '_ZSTD_decompressStream', '_ZSTD_isError', '_ZSTD_getErrorName', '_ZSTD_freeDStream', '_ZSTD_DStreamInSize', '_ZSTD_DStreamOutSize']" -s EXPORTED_RUNTIME_METHODS="['cwrap']" *.c -o zstdec.js
```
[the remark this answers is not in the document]
Jaifroid: Yes, I did think of that, but after  I had used zstdec everywhere. However, a search and replace can be done easily.

mossroy:
Just tested your latest commits with my Firefox OS device : I do not see a speed difference (but no regression either)

Jaifroid:
> Just tested your latest commits with my Firefox OS device : I do not see a speed difference (but no regression either)

:-(  OK. Well it was worth a try! Since it runs on your device, and in FFOS simulator, and it "seems faster" than previous commits on recent Chromium, Firefox and Edge Legacy, I'll keep this code. It clearly involves fewer loops, so saves some CPU.

emscripten/zstandard/compile.sh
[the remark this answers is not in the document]
Jaifroid: Done (will be committed with other revisions).

www/js/lib/zimfile.js
```diff
@@ -186,16 +186,30 @@ define(['xzdec_wrapper', 'util', 'utf8', 'q', 'zimDirEntry'], function(xz, util,
         {
             var clusterOffset = readInt(clusterOffsets, 0, 8);
             var nextCluster = readInt(clusterOffsets, 8, 8);
+            // This method of calculating cluster size is not safe: see https://github.com/openzim/libzim/issues/84#issuecomment-612962250
+            // var thisClusterLength = nextCluster - clusterOffset - 1;
```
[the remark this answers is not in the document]
Jaifroid: This is a note to a dev (mysefl) to remind me  that we should not use the apparently simple method of feeding a whole compressed cluster to the decompressor. There are some reasons why it is not safe to calculate cluster length by substracting the next pointer from the previous pointer. In any case, streaming means we can abandon a decompression if the blob we want is near the beginning of the cluster. The only advantage in decompressing a full cluster (which is what libzim does) would be if we had a cluster cache like the one written by peter-x. Then there might be performance gains if one cluster contains several blobs that we want belonging to an article. We would only decompress the cluster once, instead of starting again from the beginning for each blob. Swings and roundabouts...

www/js/lib/zimfile.js
```diff
@@ -186,16 +186,30 @@ define(['xzdec_wrapper', 'util', 'utf8', 'q', 'zimDirEntry'], function(xz, util,
...
                 var decompressor;
                 var plainBlobReader = function(offset, size) {
-                    return that._readSlice(clusterOffset + 1 + offset, size);
+                    // DEV: old algorithm merely returned requested data size
+                    // but I believe we need to check that we are not reading beyond the end of the cluster
+                    // Is this an oversight in original code or an unnecessary protection?
+                    // return that._readSlice(clusterOffset + 1 + offset, size);
+                    var offsetStart = clusterOffset + 1 + offset;
+                    if ( offsetStart < nextCluster) {
+                        size = offsetStart + size <= nextCluster ? size : nextCluster - offsetStart;
```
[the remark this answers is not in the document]
Jaifroid: VSCode complains about "gratuitous parentheses" if I add them.... But happy to add them for readibility.
The question is not about the code per se, but whether we should be careful not to read bits of a following cluster which could, theoretically, be after the end of the ZIM file. This is nothing to do with ZSTD, just about whether we should add this check (which we never had before).

www/js/lib/zstdec_wrapper.js
Jaifroid: It was intended to be the then function below, which instantiates the `zd` instance (object) and adds a whole load of needed properties (deocumented here) to the the object.  But I don't know technical definitions, so maybe this doesn't qualify as an object constructor?

www/js/lib/zstdec_wrapper.js
Jaifroid: It is necessary because we are not setting up new memory structures, but are explicitly re-using the same ones in the Emscripten HEAP as suggested in the C++ source code. This saves a lot of memory. I did actually try to eliminate this, but just as with XZ, the issue is that if we want to have two or more decompressions happening simultaneously, they each need their own memory structures and demand double/triple/quadruple the memory. We could potentially have dozens running. Of course none of them really are multi-threaded in JS, unless we launch a different WORKER for each decompression, but we don't have that code. Promises are only the illusion of multi-threading, after all...

Jaifroid:
I've done changes from review except for those not marked as resolved above, where there are questions. In the first case the issue is really whether adding this code is a good idea, even though we never had such a check before. It seems safer.  The second is whether a function that constructs an object can be described as a constructor, or whether that is incorrect technical usage.

The final issue is about Unit Tests. zstd-foo.zim is too basic to do any useful tests (no html, only numbers for article titles, text too uniform for each article "This is article 1"), and I think we should wait till we have a Ray Charles in ZSTD or some other, e.g. Wikipedia 100 mini. Then we can at least write similar QUnit tests with that ZIM. But given #580, other than QUnit tests maybe the effort should go into writing new tests using GitHub Actions. The rationale for this is that we will never get an exact ZSTD replica of the old Ray Charles we use for testing, and so many of the Unit Tests (such as checking for an article split across two parts of a split ZIM) will be a significant effort to write. So, could I add a new issue "Develop new Unit Tests for ZSTD"?

Jaifroid:
EDIT: Linked to wrong issue in comment above (now corrected) - I meant #590 "Switch to GitHub Actions instead of Travis".

kelson42:
@Jaifroid Which zimfarm recipe would you like to have with zstd to secure testing?

Jaifroid:
@kelson42 I guess for parity, maybe Ray Charles. Though I still would prefer to do tests as a separate issue because of lack of time right now, and it would delay this PR for quite a long time if it has to be part of this PR.

kelson42:
@Jaifroid OK, I have put the Ray Charles recipe on Zstd, from October it will make ZIMs in zstd.

www/js/lib/zimfile.js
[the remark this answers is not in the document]
mossroy: All right, you can leave it this way

www/js/lib/zstdec_wrapper.js
mossroy: I suppose the JSDoc tags should document the code of the current file.
Here you're documenting the `instance` parameter and/or the zd variable.
Maybe you could define a zd type with `@typedef`, even if this type is not used as a parameter (and thus I suppose the IDE won't use it for autocompletion), and add a comment line to explain this describes the structure of zd variable

www/js/lib/zstdec_wrapper.js
mossroy: I understand. So we have the same issue as with xz and need to keep it.
Thanks a lot for the explanation

mossroy (approved):
I agree with you that, considering the urgency and the lack of a suitable ZIM file, implementing unit test can be done later in another issue. Could you create it? As we need to keep compatibility with both xz and zstd compressions, I think we should keep the current QUnit and UI test, and simply add some QUnit tests on a zstd file. I don't think #590 has an impact on QUnit tests.

The unused files should be removed from this PR before merging (and some may be moved to another branch) : zstddec.wasm, zstddec_*.js, test ZIM files?

Jaifroid:
OK, thanks @mossroy. Will do. In view of the fact that Mozilla or other stores may want to see the unminified `zstddec.js`, I'll just keep one copy of that in the `emscripten/zstandard` compile folder, together with the final minified/optimized version we're actually using. I'm happy to add a wasm branch (which would be wasm-only) that can be used for performane comparison.

www/js/lib/zstddec_wrapper.js
```diff
@@ -0,0 +1,263 @@
...
+     * @property {Integer} _decHandle The decoder stream context object in asm memory (to be re-used for each decoder operation)
+     * @property {Object} _inBuffer A JS copy of the inBuffer structure to be set in asm memory (malloc)
+     * @property {Object} _outBuffer A JS copy of the outBuffer structure to be set in asm memory (malloc)
+     * @property {Integer} _chunkSize The number of compressed bytes to feed to the decompressor in any one read loop
+     */
+
+    /**
+     * The ZSTD Decoder instance
+     * @type EMSInstanceExt
+     */
```
Jaifroid, line 40: @mossroy This is a typedef I came up with. `{EMSInstanceExt}` stands for Emscripten Instance Extension.
IIt seems to work in the IDE: when I hover over `zd` anywhere in the code, its type pops up.
Thanks for the idea.

Jaifroid: NB it is commented above the visible code block here.

Jaifroid:
Wasm branch (https://github.com/kiwix/kiwix-js/tree/Test-wasm-version-of-zstandard) and issue (#652) now added and wasm removed from this PR.

Jaifroid:
All test files removed except: `foo-zstd.zim`, the full unoptimized build of zstddec, and the final optimized build. Source file and compile script remain.

This should be ready to squash and merge after more testing. I have ported the new code to Kiwix JS Windows to test with UWP and on mobile before merging here.

Jaifroid:
Tested, working and stable on:

* A Windows Mobile VM of a 4 inch 500MB low-end device
* On a Lumia 950XL physical device

Both were tested by porting the latest code from Kiwix JS to Kiwix JS Windows and building the UWP app. (Back-end is identical in Kiwix JS Windows.)

Everything else I've tested on has been stable and I have not had any corrupted pages.

So tomorrow I will squash and merge this, unless anyone else says they want to do more testing.

mossroy:
I don't think we should keep the unoptimized build of ztddec.
We did not have such file for xzdec and Mozilla did not complain about it.

Jaifroid:
Ok, will remove.

Jaifroid:
I've added the zstandard licences to the licences section of index.html. There was also inconsistency of spelling locally in that area between "Licence" and "License". Strictly, "licence" = noun; "license" = verb (like advice/advise), but US spelling tends to ignore the difference. I haven't attempted to apply this throughout the repo, however, as it would mean multiple changes unrelated to this PR, and it's not very important. I just thought I should explain the changes locally to that file for consistency of what the reader sees in that area.

Issue #619, opened 2020-05-10
Consider adding dirEntry and block cache


Jaifroid:
@mossroy As you know, @peter-x worked on (and completed) a [dirEntry and block cache](https://github.com/kiwix/kiwix-js/pull/183) feature for Kiwix JS.

When I was trying to optimize Kiwix JS for Android browsers (#581), in desperation I tried it out in Kiwix JS Windows, and found it to make a huge difference. Coincidentally for #617, where peter-x's code makes a really noticeable difference is in title search in slow binary search contexts (IE11, FFOS, Android and, believe it or not, Chrome [even on a PC] which is significantly slower at binary search than Firefox, probably due to its file-handling).

I integrated peter-x's code into Kiwix JS Windows, and you can do a simple test of the rather dramatic search speed gain by going to https://kiwix.github.io/kiwix-js-windows/ in IE11 and loading a big ZIM like full English Wikipedia. Then search for "the second world war", and you will see how much faster the search is even than Chrome running kiwix-js.

It's almost no work to integrate the cache into Kiwix JS. I'd be happy to do it, but I understand that it was not merged at the time because it was not felt to make enough of a performance gain. With the heavier binary searches that near-case-insensitivity introduces, it now does make a significant gain IMHO. I leave it to you to decide whether you wish to have this for Kiwix JS.

PS If you visit the Kiwix JS Windows implementation in Chrome or Firefox, and you have visited it previously, you may need to delete the service worker cache to be sure you are immediately running the latest code: Dev Tools -> Application -> Cache Storage -> right-click `sw-precache` -> Delete, then reload browser with Ctrl-Shift-R followed by Ctrl-R.

Issue #595, opened 2020-02-28
Can Kiwix-JS deal with an online ZIM file available through HTTP?


kelson42:
I know we have talked about that in the past... but it the outcome of the discussion is unclear to me.

Jaifroid:
Kiwix JS can load a ZIM file into memory over HTTP(S) and use it that way, which is not much good because you might as well be downloading it in that case. Sharun-s had developed a technique using XML Range Requests to access part of a ZIM file over http(s). It was slow.

Jaifroid:
I might add that the slowest element here would be binary search of the title list and url list, as we would be opening potentially hundreds of XMLHttpRequests to the server for very small bits of data. There is a caching module (developed by peter-x) that helps in that situation, but given #581 this could still be pretty slow on large files.

kelson42:
@Jaifroid I mean without downloading the whole file. Just dealing like with a local filesystem file, just downloading the cluster and unpacking them?

Jaifroid:
@kelson42 Sorry, did not see your further query till now. Kiwix JS cannot currently download clusters over http, but sharun-s developed a prototype of that capability in his branch. He used XML range requests to read clusters from a file on localhost, but there was no reason why the exact same technique could not be used for a remote server. XML range requests ask the server only for a small part of a file, so it fits what you are asking about.

The problem is not integrating sharun-s's technique (that is not particularly difficult). The problem is that it would inevitably be slow: binary search in particular is the main problem, because you would generate hundreds of requests to the server for each article. We do have block cache code that would greatly reduce the number of requests by caching dirEntries and clusters that are already requested, and that helps a lot.

So, it may be feasible for smallish ZIM files, but probably excruciatingly slow for anything with a very large url pointer list that needs to be searched.

What's your use case?

Jaifroid:
PS the server would need to accept range requests: see https://developer.mozilla.org/en-US/docs/Web/HTTP/Range_requests. I think most do (those that allow resumable download, or skipping through a streaming media file).

kelson42:
@Jaifroid Thank you for your answer. My feelings about this approach are mitigated, I don't really believe in this for the future (but I might be wrong). But I wanted to get this documented properly for my own knowlege and for a few other reasons as well.

[...]

Issue #581, opened 2020-01-01
Extremely slow binary search in Chromium for Anroid


Jaifroid:
Although this is not an official target of Kiwix JS, I think this might uncover a problem in the coding of binary search. If you open https://kiwix.github.io/kiwix-js/www/index.html on an Android phone in Chrome or Edge-Chromium with a reasonably large ZIM (e.g. WikiMed) on a microSD card, the binary search is **_ridiculously_ slow** and bugs out when backspacing in the search field (a race condition, I believe).

The system is generally pretty slow, and I believe it is because ASM is not optimized in the Android browser. However, slow binary search seems to be another issue because binary search does not decompress the title index (and therefore should not use `xzdec`), it merely accesses it on-disk repeatedly using `FileReader()`. The same ZIM accessed with the official Android Kiwix client does not have this issue.

I think this could be allieviated by reading more of the title index into memory (up to a predefined limit) in order to reduce storage/disk access, but I'm guessing wildly.

I note that it is claimed that WASM is supported on mobile Chrome, and therefore it might be worth experimenting with this to speed up access in mobile contexts generally, though it wouldn't make a difference to binary search.

NB the aim is not to challenge the official Kiwix Android app, but to optimize support for a PWA version of our app. An installable PWA of Kiwix JS Windows can be experimented with by going to https://kiwix.github.io/kiwix-js-windows/www/ in a browser that can install PWAs (Chrome or Edge-Chromium, on PC or Android). It has the limitation that the user must pick the ZIM each time s/he launches the app. However, this can be solved on Windows 10, because WinRT APIs can be accessed by a PWA if it is packaged for the Store or sideloaded (but not if it is installed from the Web).

mossroy:
I confirm that issue on my device. It's very slow on Firefox Mobile too.
On Firefox Preview, it looks much faster (except the first access, right after choosing the ZIM file, where the screen remains black for a very long time).
It appears to be Android-specific.
I suspect it might come from the way the file is "given" to the browser and/or how it is accessed by the browser itself.
On Android, these browsers don't actually browse the filesystem by themselves. They ask for an application to provide the file(s), certainly via an "intent". I tested by choosing the File Browser application of LineageOS after clicking on the file input of our app.

Jaifroid:
> On Firefox Preview, it looks much faster (except the first access, right after choosing the ZIM file, where the screen remains black for a very long time).

[...]

PR #658, opened 2020-10-17
Add a low level block cache

Jaifroid:
As discussed in #595, this PR implements #619 and revives #183. It is almost entirely @peter-x's code. I have simply tuned the cache size based on my experience running this in KJSW, and have added a faster method of reading an arrayBuffer from the updated Blob API that elminates the need to use FileReader (for those browsers that support the new method).

Apart from Edge (Chromium), I have tested very quickly on Firefox OS simulator (working) and on IE11 (working).

I have left in a console log output that shows the block cache hits and misses, as well as the percentage hit rate. Below is from IE11 console log when running the full 96GB English Wikipedia and searching for "the second world war". You can clearly see here the number of file reads that are avoided and an impressive 100% hit rate for some 6,000 dirEntry reads, and above 74% for the initial 3,400 reads.

image

### DIFF
--- a/www/js/lib/filecache.js
+++ b/www/js/lib/filecache.js
@@ -63,6 +63,7 @@ define(['q'], function(Q) {
         this.moveToTop(entry);
         return entry.value;
     };
+    
     /**
      * Stores a value in the cache by id and prunes the least recently used entry if the cache is larger than MAX_CACHE_SIZE
      * @param {String} id The key under which to store the value (consists of filename + file number)
--- a/www/js/lib/filecache.js
+++ b/www/js/lib/filecache.js
@@ -51,7 +51,9 @@ define(['q'], function(Q) {
 
     /**
      * Tries to retrieve an element by its namespace and id. If it is not present in the cache,
-     * returns undefined.
+     * returns undefined
+     * @param {Integer} id The block cache entry id
+     * @returns {Uint8Array|undefined} The requested cache data or undefined 
      */
     LRUCache.prototype.get = function(id) {
         var entry = this._entries[id]; 
@@ -114,7 +116,10 @@ define(['q'], function(Q) {
     /**
      * Read a certain byte range in the given file, breaking the range into chunks that go through the cache
      * If a read of more than blocksize (bytes) is requested, do not use the cache
-     * @return {Promise} promise that resolves to the correctly concatenated data.
+     * @param {Object} file The requested file to read from
+     * @param {Integer} begin The byte from which to start reading
+     * @param {Integer} end The last byte to read
+     * @return {Promise<Uint8Array>} A Promise that resolves to the correctly concatenated data
      */
     var read = function(file, begin, end) {
         // Read large chunks bypassing the block cache because we would have to
--- a/www/js/lib/filecache.js
+++ b/www/js/lib/filecache.js
@@ -161,7 +161,9 @@ define(['q'], function(Q) {
     };
     var readInternal = function (file, begin, end) {
         if ('arrayBuffer' in Blob.prototype) {
-            // DEV: This method uses the Native FS handle if available
+            // DEV: This method uses the native arrayBuffer method of Blob, if available, as it eliminates
+            // the need to use FileReader and set up event listeners; it also uses the method's native Promise
+            // rather than setting up potentially hundreds of new Q promises for small byte range reads
             return file.slice(begin, end).arrayBuffer().then(function (buffer) {
                 return new Uint8Array(buffer);
             });
--- a/www/js/lib/filecache.js
+++ b/www/js/lib/filecache.js
@@ -20,7 +20,7 @@
  * along with Kiwix JS (file LICENSE).  If not, see <http://www.gnu.org/licenses/>
  */
 'use strict';
-define(['q'], function(Q) {
+define(['q', 'util'], function(Q, util) {
     /**
      * Set maximum number of cache blocks of BLOCK_SIZE bytes each
      * Maximum size of cache in bytes = MAX_CACHE_SIZE * BLOCK_SIZE
@@ -116,15 +116,15 @@ define(['q'], function(Q) {
     /**
      * Read a certain byte range in the given file, breaking the range into chunks that go through the cache
      * If a read of more than blocksize (bytes) is requested, do not use the cache
-     * @param {Object} file The requested file to read from
+     * @param {Object} file The requested ZIM archive to read from
      * @param {Integer} begin The byte from which to start reading
-     * @param {Integer} end The last byte to read
-     * @return {Promise<Uint8Array>} A Promise that resolves to the correctly concatenated data
+     * @param {Integer} end The byte at which to stop reading (end will not be read)
+     * @return {Promise<Uint8Array>} A Promise that resolves to the correctly concatenated data from the split ZIM file set
      */
     var read = function(file, begin, end) {
         // Read large chunks bypassing the block cache because we would have to
         // stitch together too many blocks and would clog the cache
-        if (end - begin > BLOCK_SIZE * 2) return readInternal(file, begin, end);
+        if (end - begin > BLOCK_SIZE * 2) return util.readFileSlice(file, begin, end);
         var readRequests = [];
         var blocks = {};
         for (var i = Math.floor(begin / BLOCK_SIZE) * BLOCK_SIZE; i < end; i += BLOCK_SIZE) {
@@ -160,22 +160,35 @@ define(['q'], function(Q) {
         });
     };
     var readInternal = function (file, begin, end) {
-        if ('arrayBuffer' in Blob.prototype) {
-            // DEV: This method uses the native arrayBuffer method of Blob, if available, as it eliminates
-            // the need to use FileReader and set up event listeners; it also uses the method's native Promise
-            // rather than setting up potentially hundreds of new Q promises for small byte range reads
-            return file.slice(begin, end).arrayBuffer().then(function (buffer) {
-                return new Uint8Array(buffer);
-            });
+        var readRequests = [];
+        var currentOffset = 0;
+        for (var i = 0; i < file._files.length; currentOffset += file._files[i].size, ++i) {
+            var currentSize = file._files[i].size;
+            if (begin < currentOffset + currentSize && currentOffset < end) {
+                var readStart = Math.max(0, begin - currentOffset);
+                var readSize = Math.min(currentSize, end - currentOffset - readStart);
+                readRequests.push(util.readFileSlice(file._files[i], readStart, readStart + readSize));
+            }
+        }
+        if (readRequests.length === 0) {
+            return Q(new Uint8Array(0).buffer);
+        } else if (readRequests.length === 1) {
+            return readRequests[0];
         } else {
-            return Q.Promise(function (resolve, reject) {
-                var reader = new FileReader();
-                reader.readAsArrayBuffer(file.slice(begin, end));
-                reader.addEventListener('load', function (e) {
-                    resolve(new Uint8Array(e.target.result));
+            // Wait until all are resolved and concatenate.
+            console.log("CONCAT");
+            return Q.all(readRequests).then(function(arrays) {
+                var length = 0;
+                arrays.forEach(function (item) {
+                    length += item.byteLength;
                 });
-                reader.addEventListener('error', reject);
-                reader.addEventListener('abort', reject);
+                var concatenated = new Uint8Array(length);
+                var offset = 0;
+                arrays.forEach(function (item) {
+                    concatenated.set(new Uint8Array(item), offset);
+                    offset += item.byteLength;
+                });
+                return concatenated;
             });
         }
     };
--- a/www/js/lib/filecache.js
+++ b/www/js/lib/filecache.js
@@ -124,7 +124,7 @@ define(['q', 'util'], function(Q, util) {
     var read = function(file, begin, end) {
         // Read large chunks bypassing the block cache because we would have to
         // stitch together too many blocks and would clog the cache
-        if (end - begin > BLOCK_SIZE * 2) return directRead(file, begin, end);
+        if (end - begin > BLOCK_SIZE * 2) return file._readSplitSlice(begin, end);
         var readRequests = [];
         var blocks = {};
         for (var i = Math.floor(begin / BLOCK_SIZE) * BLOCK_SIZE; i < end; i += BLOCK_SIZE) {
@@ -132,7 +132,7 @@ define(['q', 'util'], function(Q, util) {
             if (block === undefined) {
                 misses++;
                 readRequests.push(function(offset) {
-                    return directRead(file, offset, offset + BLOCK_SIZE).then(function(result) {
+                    return file._readSplitSlice(offset, offset + BLOCK_SIZE).then(function(result) {
                         cache.store(file.name + offset, result);
                         blocks[offset] = result;
                     });
@@ -159,44 +159,8 @@ define(['q', 'util'], function(Q, util) {
             return result;
         });
     };
-    var directRead = function (file, begin, end) {
-        var readRequests = [];
-        var currentOffset = 0;
-        for (var i = 0; i < file._files.length; currentOffset += file._files[i].size, ++i) {
-            var currentSize = file._files[i].size;
-            if (begin < currentOffset + currentSize && currentOffset < end) {
-                // DEV: Math.max is used below because we could be reading the last part of a blob split across two files,
-                // in which case (begin - currentOffset) could be negative!
-                var readStart = Math.max(0, begin - currentOffset);
-                var readEnd = Math.min(currentSize, end - currentOffset);
-                readRequests.push(util.readFileSlice(file._files[i], readStart, readEnd));
-            }
-        }
-        if (readRequests.length === 0) {
-            return Q(new Uint8Array(0).buffer);
-        } else if (readRequests.length === 1) {
-            return readRequests[0];
-        } else {
-            // Wait until all are resolved and concatenate.
-            console.log("CONCAT");
-            return Q.all(readRequests).then(function(arrays) {
-                var length = 0;
-                arrays.forEach(function (item) {
-                    length += item.byteLength;
-                });
-                var concatenated = new Uint8Array(length);
-                var offset = 0;
-                arrays.forEach(function (item) {
-                    concatenated.set(new Uint8Array(item), offset);
-                    offset += item.byteLength;
-                });
-                return concatenated;
-            });
-        }
-    };
 
     return {
-        read: read,
-        directRead: directRead
+        read: read
     };
 });
--- a/www/js/lib/filecache.js
+++ b/www/js/lib/filecache.js
@@ -50,9 +50,9 @@ define(['q'], function(Q) {
     }
 
     /**
-     * Tries to retrieve an element by its id. If it is not present in the cache,
-     * returns undefined
-     * @param {Integer} id The block cache entry id
+     * Tries to retrieve an element by its id. If it is not present in the cache, returns undefined; if it is present,
+     * then the value is returned and the entry is moved to the top of the cache
+     * @param {String} id The block cache entry key
      * @returns {Uint8Array|undefined} The requested cache data or undefined 
      */
     LRUCache.prototype.get = function(id) {
@@ -63,6 +63,11 @@ define(['q'], function(Q) {
         this.moveToTop(entry);
         return entry.value;
     };
+    /**
+     * Stores a value in the cache by id and prunes the least recently used entry if the cache is larger than MAX_CACHE_SIZE
+     * @param {String} id The key under which to store the value (consists of filename + file number)
+     * @param {Uint16Array} value The value to store in the cache 
+     */
     LRUCache.prototype.store = function(id, value) {
         var entry = this._entries[id];
         if (entry === undefined) {
@@ -80,6 +85,11 @@ define(['q'], function(Q) {
             this.moveToTop(entry);
         }
     };
+
+    /**
+     * Delete a cache entry
+     * @param {String} entry The entry to delete 
+     */
     LRUCache.prototype.unlink = function(entry) {
         if (entry.next === null) {
             this._last = entry.prev;
@@ -92,6 +102,11 @@ define(['q'], function(Q) {
             entry.prev.next = entry.next;
         }
     };
+
+    /**
+     * Insert a cache entry at the top of the cache
+     * @param {String} entry The entry to insert 
+     */
     LRUCache.prototype.insertAtTop = function(entry) {
         if (this._first === null) {
             this._first = this._last = entry;
@@ -101,6 +116,11 @@ define(['q'], function(Q) {
             this._first = entry;
         }
     };
+
+    /**
+     * Move a cache entry to the top of the cache
+     * @param {String} entry The entry to move 
+     */
     LRUCache.prototype.moveToTop = function(entry) {
         this.unlink(entry);
         this.insertAtTop(entry);
--- a/www/js/lib/filecache.js
+++ b/www/js/lib/filecache.js
@@ -124,7 +124,7 @@ define(['q', 'util'], function(Q, util) {
     var read = function(file, begin, end) {
         // Read large chunks bypassing the block cache because we would have to
         // stitch together too many blocks and would clog the cache
-        if (end - begin > BLOCK_SIZE * 2) return readInternal(file, begin, end);
+        if (end - begin > BLOCK_SIZE * 2) return directRead(file, begin, end);
         var readRequests = [];
         var blocks = {};
         for (var i = Math.floor(begin / BLOCK_SIZE) * BLOCK_SIZE; i < end; i += BLOCK_SIZE) {
@@ -132,7 +132,7 @@ define(['q', 'util'], function(Q, util) {
             if (block === undefined) {
                 misses++;
                 readRequests.push(function(offset) {
-                    return readInternal(file, offset, offset + BLOCK_SIZE).then(function(result) {
+                    return directRead(file, offset, offset + BLOCK_SIZE).then(function(result) {
                         cache.store(file.name + offset, result);
                         blocks[offset] = result;
                     });
@@ -159,15 +159,17 @@ define(['q', 'util'], function(Q, util) {
             return result;
         });
     };
-    var readInternal = function (file, begin, end) {
+    var directRead = function (file, begin, end) {
         var readRequests = [];
         var currentOffset = 0;
         for (var i = 0; i < file._files.length; currentOffset += file._files[i].size, ++i) {
             var currentSize = file._files[i].size;
             if (begin < currentOffset + currentSize && currentOffset < end) {
+                // DEV: Math.max is used below because we could be reading the last part of a blob split across two files,
+                // in which case (begin - currentOffset) could be negative!
                 var readStart = Math.max(0, begin - currentOffset);
-                var readSize = Math.min(currentSize, end - currentOffset - readStart);
-                readRequests.push(util.readFileSlice(file._files[i], readStart, readStart + readSize));
+                var readEnd = Math.min(currentSize, end - currentOffset);
+                readRequests.push(util.readFileSlice(file._files[i], readStart, readEnd));
             }
         }
         if (readRequests.length === 0) {
@@ -194,6 +196,7 @@ define(['q', 'util'], function(Q, util) {
     };
 
     return {
-        read: read
+        read: read,
+        directRead: directRead
     };
 });
--- a/www/js/lib/filecache.js
+++ b/www/js/lib/filecache.js
@@ -153,7 +153,7 @@ define(['q', 'util'], function(Q, util) {
             for (var i = Math.floor(begin / BLOCK_SIZE) * BLOCK_SIZE; i < end; i += BLOCK_SIZE) {
                 var b = Math.max(i, begin) - i;
                 var e = Math.min(end, i + BLOCK_SIZE) - i;
-                result.set(blocks[i].subarray(b, e), pos);
+                if (blocks[i].subarray) result.set(blocks[i].subarray(b, e), pos);
                 pos += e - b;
             }
             return result;
--- a/www/js/lib/filecache.js
+++ b/www/js/lib/filecache.js
@@ -124,7 +124,7 @@ define(['q', 'util'], function(Q, util) {
     var read = function(file, begin, end) {
         // Read large chunks bypassing the block cache because we would have to
         // stitch together too many blocks and would clog the cache
-        if (end - begin > BLOCK_SIZE * 2) return util.readFileSlice(file, begin, end);
+        if (end - begin > BLOCK_SIZE * 2) return readInternal(file, begin, end);
         var readRequests = [];
         var blocks = {};
         for (var i = Math.floor(begin / BLOCK_SIZE) * BLOCK_SIZE; i < end; i += BLOCK_SIZE) {
--- a/www/js/lib/filecache.js
+++ b/www/js/lib/filecache.js
@@ -20,7 +20,7 @@
  * along with Kiwix JS (file LICENSE).  If not, see <http://www.gnu.org/licenses/>
  */
 'use strict';
-define(['q', 'util'], function(Q, util) {
+define(['q'], function(Q) {
     /**
      * Set maximum number of cache blocks of BLOCK_SIZE bytes each
      * Maximum size of cache in bytes = MAX_CACHE_SIZE * BLOCK_SIZE
--- a/www/js/lib/filecache.js
+++ b/www/js/lib/filecache.js
@@ -26,13 +26,13 @@ define(['q'], function(Q) {
      * Maximum size of cache in bytes = MAX_CACHE_SIZE * BLOCK_SIZE
      * @type {Integer}
      */
-    var MAX_CACHE_SIZE = 4000;
+    const MAX_CACHE_SIZE = 4000;
 
     /**
      * The maximum blocksize to read or store via the block cache (bytes)
      * @type {Integer}
     */
-    var BLOCK_SIZE = 4096;
+    const BLOCK_SIZE = 4096;
 
     /**
      * Creates a new cache with max size limit
--- a/www/js/lib/filecache.js
+++ b/www/js/lib/filecache.js
@@ -1,0 +1,179 @@
+﻿/**
+ * filecache.js: Generic least-recently-used-cache used for reading file chunks.
+ *
+ * Copyright 2020 Mossroy, peter-x, jaifroid and contributors
+ * License GPL v3:
+ *
+ * This file is part of Kiwix.
+ *
+ * Kiwix JS is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * Kiwix JS is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Kiwix JS (file LICENSE).  If not, see <http://www.gnu.org/licenses/>
+ */
+'use strict';
+define(['q'], function(Q) {
+    /**
+     * Set maximum number of cache blocks of BLOCK_SIZE bytes each
+     * Maximum size of cache in bytes = MAX_CACHE_SIZE * BLOCK_SIZE
+     * @type {Integer}
+     */
+    var MAX_CACHE_SIZE = 4000;
+
+    /**
+     * The maximum blocksize to read or store via the block cache (bytes)
+     * @type {Integer}
+    */
+    var BLOCK_SIZE = 4096;
+
+    /**
+     * Creates a new cache with max size limit
+     * @param {Integer} limit The maximum number of 2048-byte blocks to be cached
+     */
+    function LRUCache(limit) {
+        console.log("Creating cache of size " + limit);
+        this._limit = limit;
+        this._size = 0;
+        // Mapping from id to {value: , prev: , next: }
+        this._entries = {};
+        // linked list of entries
+        this._first = null;
+        this._last = null;
+    }
+
+    /**
+     * Tries to retrieve an element by its namespace and id. If it is not present in the cache,
+     * returns undefined.
+     */
+    LRUCache.prototype.get = function(id) {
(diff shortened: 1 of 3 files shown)


Jaifroid:
Hmm. QUnit tests are failing this version on several browsers. I revert to draft while I investigate. It works in browsers outside the test environment.

Jaifroid:
@javgh Feel free to comment on this PR since you have also integrated peter-x's block cache into your version. If there are any optimizations to the cache you have implemented that could be useful here, please discuss.

Jaifroid:
It's failing on reading an article split across two different split ZIM file parts. Interesting, because I use split ZIMs with this code quite extensively, and have noticed that failure before. Possibly because the split part fragments on Ray Charles are so small. But good that the tests found that error.

Jaifroid:
OK, I'm pretty sure I've discovered the issue with split files. The block cache code doesn't have logic to detect when it is reading beyond the end of a split ZIM chunk, which is probably a vanishingly small evenet with large split ZIM archives, but because our Ray Charles test is plit into tiny 96K parts, it happens with some regularity. Hence, I've never detected this issue before.

IMO, the cache code is patching file slice reads at the wrong level. Instead of patching `util.readFileSlice` (which is the level at which individual reads of split parts of a ZIM happen, to be concatenated later), it shoudl patch `zimFile.readSlice`, and the logic to split reads across different ZIM parts should be included in the cache code.

I think this is relatively easy to do. It will also simplify the cache, as it will store relative offsets from the start of the ZIM file set instead of absolute offsets from the start of each split ZIM part. It would then be indifferent to how small split ZIM chunks are.

Jaifroid:
Although it's now passing tests, I want to do some cleanup of the location of some of the code.

Jaifroid:
I have implemented the suggestions in my previous post, except that I left the logic for dealing with split ZIM files as a dedicated function in zimfile that the cache can use, rather than putting it wholly in the cache code.

I have cleaned up and documented the code (I documented the cache code, but also added missing documentation to zimfile), and have tested the last commit on Edge Chromium, Firefox OS Simulator, IE11 and Edge Legacy.

The console log entry that measures the block cache hit rate displays only once every 2,000 cache requests. For searches the hit rate is very high, often near 100%. For randomly cycling through a ZIM, the hit rate is 60 - 75%, higher for smaller ZIM archives.

I would suggest leaving that console log entry in the code as it is a useful metric that could help future devs to tune the block cache to modern devices. However, I can remove it or comment it out if that is preferred.

The code is ready for review.

kelson42:
@Jaifroid Congratulation for the super fast step forward. What is the memory consumption of this cache? How is that different from @javgh's cache system?

Jaifroid:
@kelson42 It's the same underlying cache system, originally developed by peter-x but never merged. I have corrected the logic for dealing with split archives, and have documented the code fully, and tuned the size of the cache based on my experience with Kiwix JS Windows. I believe @javgh has tuned their code to deal with network requests, so the cache is probably much larger. However, there are constants at the head of the cache file that can easily change the size of the cache and of the blocks read, and these could easily be turned into parameters that a dev (or a user) could set based on usage requirements.

As configured in this PR, the maximum cache size would be 4000 * 4096 bytes = 16,384,000 bytes. I use these values in KJSW, and the code can run on a low-end VM of Windows Mobile with 512MB RAM. If the cache is too big for @mossroy's physical FFOS device (though I doubt it will be an issue, based on the FFOS simulator), then we can try 2000 * 4096 or 2000 * 2048.

www/js/lib/zimfile.js
```diff
@@ -48,89 +47,98 @@ define(['xzdec_wrapper', 'zstddec_wrapper', 'util', 'utf8', 'q', 'zimDirEntry'],
...
         } else {
             // Wait until all are resolved and concatenate.
             console.log("CONCAT");
-            return Q.all(readRequests).then(function(arrays) {
-                var concatenated = new Uint8Array(size);
-                var sizeSum = 0;
-                for (var i = 0; i < arrays.length; ++i) {
-                    concatenated.set(new Uint8Array(arrays[i]), sizeSum);
-                    sizeSum += arrays[i].byteLength;
-                }
```
Jaifroid, line 107: Consider reverting code below to above (largely). Although code below works fine, it was part of an early attempt to correct the split ZIM archive issue. Original code would be marginally faster and the first arrays.forEach is no longer needed.

www/js/lib/zimfile.js
```diff
@@ -203,15 +205,17 @@ define(['xzdec_wrapper', 'zstddec_wrapper', 'util', 'utf8', 'q', 'zimDirEntry'],
                 };
                 if (compressionType[0] === 0 || compressionType[0] === 1) {
                     // uncompressed
-                    decompressor = { readSliceSingleThread: plainBlobReader };
```
Jaifroid, line 206: Revert to above - over-zeleous linter.

www/js/lib/zimfile.js
```diff
@@ -282,6 +285,7 @@ define(['xzdec_wrapper', 'zstddec_wrapper', 'util', 'utf8', 'q', 'zimDirEntry'],
                 var urlPtrPos = readInt(header, 32, 8);
                 return readMimetypeMap(fileArray[0], mimeListPos, urlPtrPos).then(function (data) {
                     var zf = new ZIMFile(fileArray);
+                    zf.name = fileArray[0].name.replace(/(\.zim)[a-z]{2}$/i, '$1');
```
Jaifroid, line 288: Add comment above to explain what this line does: provides an abstract filename, turning e.g. `en_wikipedia.zimaa` into `en_wikipedia.zim` so that the cache code doesn't have to deal with individual split ZIM files, but just sees one abstracted ZIM archive and uses absolute offsets from the start of the abstracted archive.

(This line will leave non-split filenames alone.)

www/js/lib/filecache.js
```diff
@@ -0,0 +1,186 @@
...
+     * @returns {Uint8Array|undefined} The requested cache data or undefined 
+     */
+    LRUCache.prototype.get = function(id) {
+        var entry = this._entries[id]; 
+        if (entry === undefined) {
+            return entry;
+        }
+        this.moveToTop(entry);
+        return entry.value;
+    };
```
Jaifroid, line 65: Add space to separate this function from next.

www/js/lib/zimfile.js
Jaifroid: Done.

mossroy:
I tested on my Firefox OS device : it works with no problem. So the cache size looks compatible with such old hardware.
The debug logs usually give very good hit/miss ratios.

But, surprisingly, I don't measure any performance difference on this device, at least on the time to first appearance of a wikipedia article. I tested on big articles like "Paris" on wikipedia, and on the home page : it takes the same time on this branch and on master (although I see the good hit/miss ratio in the browser log, when using the branch).

Anyway, if you measure significant improvements on other devices, we should not bother about Firefox OS devices performance any more.
I'll have a look at the code, which looks good at first sight

Jaifroid:
@mossroy This cache does not make much difference with a single article load time. The cache is designed to speed up binary search for dirEntries where we can easily get around 10,000 file reads for a single search (see https://github.com/openzim/libzim/pull/430#issuecomment-713721164). Without this PR, each of those is a file read; with this PR, about 80% are taken from the cache, and 100% for repeated searching.

So, where you will see a really big difference is if you do search on a device with slow disk access, or, as in #595 and #659, accessing file slices over a network. Favourite searches I use to test this in full English Wikipedia are: "the second world war", or "unesco world heritage site" (the latter requires a lot of searching till it discovers the all caps "UNESCO" with other caps combinations.

Binary search of course also affects loading an article with a lot of assets, because a search has to be undertaken for each asset.

It should be noted that the solution here is a version of what is being worked on in https://github.com/openzim/libzim/pull/430 and also https://github.com/openzim/libzim/issues/424 . As explained there, it will become even more important once ZIMs without namespaces are generally available, becuase then binary search will not be easy to limit by the namespace.

I see a really big increase in speed on any device with slow file access. IE11 search is massively sped up. Running the app on Android makes it usable in Chrome/Edge even with full English Wikipedia (it is completely unusable without this cache). Even Chrome on a PC has relatively slow file access. However, the ONE browser that has very fast file access is.... Firefox (we already noted that on Android in #581). As a result, you won't notice much speedup in Firefox, since file access is not a bottleneck in that browser (and maybe it does its own caching to achieve that: it certainly attempts to do that on Android, which then fails with large ZIMs).

Jaifroid:
Just a further point: you will see in the code that the cache is tuned **not** to cache file reads that are larger than twice the block size, to avoid clogging the cache. So it will not cache the contents of an article, or even of an image for the most part (and we have a specialized assets cache for that purpose anwyay). What this cache specializes in is dirEntries that are repeatedly and uselessly read from file in the binary search process (thousands and thousands of them, and for the initial hops always the same ones).

mossroy:
You're right : the search speed is significantly improved on my Firefox OS device with this branch.
I now can barely see the results progressively displaying in the list : they appear almost always all at the same type, and much faster.

www/js/lib/filecache.js
```diff
@@ -0,0 +1,187 @@
+﻿/**
+ * filecache.js: Generic least-recently-used-cache used for reading file chunks.
```
Jaifroid, line 2: Maybe say here: "used for fast Directory Entry lookup" (instead of "used for reading file chunks").

Jaifroid: Or better: "Generic cache for small, frequently read file slices. It discards cached blocks according to a least-recently-used algorithm. It is used primarily for fast Directory Entry lookup, speeding up binary search."

www/js/lib/filecache.js
```diff
@@ -0,0 +1,187 @@
...
+    var hits = 0;
+    var misses = 0;
+
+    /**
+     * Read a certain byte range in the given file, breaking the range into chunks that go through the cache
+     * If a read of more than blocksize (bytes) is requested, do not use the cache
+     * @param {Object} file The requested ZIM archive to read from
+     * @param {Integer} begin The byte from which to start reading
+     * @param {Integer} end The byte at which to stop reading (end will not be read)
+     * @return {Promise<Uint8Array>} A Promise that resolves to the correctly concatenated data from the split ZIM file set
```
Jaifroid, line 143: Change description to "A Promise that resolves to the correctly concatenated data from the cache or from the ZIM archive". (This function is now indifferent as to whether the ZIM is split or not, as that is handled in `zimfile.js`.)

www/js/lib/filecache.js
```diff
@@ -0,0 +1,187 @@
...
+     */
+    var read = function (file, begin, end) {
+        // Read large chunks bypassing the block cache because we would have to
+        // stitch together too many blocks and would clog the cache
+        if (end - begin > BLOCK_SIZE * 2) return file._readSplitSlice(begin, end);
+        var readRequests = [];
+        var blocks = {};
+        for (var i = Math.floor(begin / BLOCK_SIZE) * BLOCK_SIZE; i < end; i += BLOCK_SIZE) {
+            var block = cache.get(file.name + i);
```
mossroy, line 152: A bug might occur here in a **very** rare case : If I have 2 ZIM files named foo and foo1, and open them one after the other one (in the same session), there might be a mix if their cache blocks, for example between block 2 of foo1 and block 12 of foo : both would have the same id foo12.
I see at least 2 options to avoid that : 

- use a separator between the file name and the block number. But it needs to be a character that can never be used in a file name (on any OS)
- use a javascript object with 2 different properties in it, as the key. It's certainly safer and cleaner. It should be supported by using a https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map instead of a javascript Object. Based on this documentation from Mozilla, it also might be faster. And it seems to be supported by all our supported browsers (with some restrictions on IE11, but we probably don't use the unsupported functions)

www/js/lib/util.js
```diff
@@ -196,22 +196,31 @@ define(['q'], function(Q) {
...
-                resolve(new Uint8Array(e.target.result));
-            };
-            reader.onerror = reader.onabort = reject;
-            reader.readAsArrayBuffer(file.slice(begin, begin + size));
-        });
+    function readFileSlice(file, begin, end) {
+        if ('arrayBuffer' in Blob.prototype) {
+            // DEV: This method uses the native arrayBuffer method of Blob, if available, as it eliminates
+            // the need to use FileReader and set up event listeners; it also uses the method's native Promise
+            // rather than setting up potentially hundreds of new Q promises for small byte range reads
```
mossroy, line 209: Nice! It might improve performance and recent browsers

www/js/lib/zimfile.js
```diff
@@ -48,89 +47,94 @@ define(['xzdec_wrapper', 'zstddec_wrapper', 'util', 'utf8', 'q', 'zimDirEntry'],
...
 
     /**
-     * 
-     * @param {Integer} offset
-     * @param {Integer} size
-     * @returns {Integer}
+     * Read and decode an integer value from the ZIM archive
+     * @param {Integer} offset The offset at which the integer is found
+     * @param {Integer} size The size of data to read
+     * @returns {Promise<Integer>} A Promise for the returned value
```
mossroy, line 64: Many thanks for these jsdocs additions

www/js/lib/zimfile.js
```diff
@@ -48,89 +47,94 @@ define(['xzdec_wrapper', 'zstddec_wrapper', 'util', 'utf8', 'q', 'zimDirEntry'],
...
             return readInt(data, 0, size);
         });
     };
 
     /**
-     * 
-     * @param {Integer} offset
-     * @param {Integer} size
-     * @returns {Promise}
+     * Read a slice from the ZIM set starting at offset for size of bytes
```
mossroy, line 73: It might be added in this description that it also makes use of the cache

mossroy (changes requested):
Here are a few remarks and suggestions.
The main one is to try to use a Map instead of an Object to store blocks of the cache, to solve a rare bug case, and (maybe) give better performance

Other than that, it's minor remarks and the code looks good!

www/js/lib/zimfile.js
```diff
@@ -48,89 +47,94 @@ define(['xzdec_wrapper', 'zstddec_wrapper', 'util', 'utf8', 'q', 'zimDirEntry'],
...
         }
-        if (readRequests.length == 0) {
+        if (readRequests.length === 0) {
             return Q(new Uint8Array(0).buffer);
-        } else if (readRequests.length == 1) {
+        } else if (readRequests.length === 1) {
             return readRequests[0];
         } else {
             // Wait until all are resolved and concatenate.
             console.log("CONCAT");
```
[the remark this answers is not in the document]
Jaifroid: Happy to remove it, it doesn't serve any purpose except to show that a concatenation ocurred, which will pretty much always occur with a split ZIM.

www/js/lib/filecache.js
```diff
@@ -0,0 +1,187 @@
...
+    // Create a new cache
+    var cache = new LRUCache(MAX_CACHE_SIZE);
+
+    // Counters for reporting only
+    var hits = 0;
+    var misses = 0;
+
+    /**
+     * Read a certain byte range in the given file, breaking the range into chunks that go through the cache
+     * If a read of more than blocksize (bytes) is requested, do not use the cache
```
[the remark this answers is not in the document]
Jaifroid: Thanks, I missed that -- it was part of "tuning" to try out caching larger or smaller reads, and I settled on twice the block size as a maximum for the cache.

www/js/lib/filecache.js
```diff
@@ -0,0 +1,187 @@
...
+     * @param {Integer} end The byte at which to stop reading (end will not be read)
+     * @return {Promise<Uint8Array>} A Promise that resolves to the correctly concatenated data from the split ZIM file set
+     */
+    var read = function (file, begin, end) {
+        // Read large chunks bypassing the block cache because we would have to
+        // stitch together too many blocks and would clog the cache
+        if (end - begin > BLOCK_SIZE * 2) return file._readSplitSlice(begin, end);
+        var readRequests = [];
+        var blocks = {};
+        for (var i = Math.floor(begin / BLOCK_SIZE) * BLOCK_SIZE; i < end; i += BLOCK_SIZE) {
```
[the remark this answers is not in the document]
Jaifroid: OK

www/js/lib/filecache.js
Jaifroid: OK, thanks for the suggestion.

www/js/lib/zimfile.js
Jaifroid: Will do.

Jaifroid:
I've rebased this on master to do more accurate performance testing.

www/js/lib/filecache.js
Jaifroid: @mossroy It seems that a new Map object is not easily extensible - https://stackoverflow.com/questions/52015305/how-to-extend-a-map-in-js . I could extend the prototype of the native Map object, but then every new Map (in this app) will have the defined functions -- would that be an issue? We currently use one other Map (a cache in app.js).

Alternatively, we could try to use principles of functional programming (#363), and not attach the functions to the prototype of the Cache itself. But I think then we would have some hybrid here of OOP and functional that is probably not ideal!

Any advice?

mossroy: I don't understand why you need to extend a Map.
What I was suggesting was to replace (line 48) `this._entries = {};` by `this._entries = new Map();`
and replace subsequent `_entries[id]` by `_entries.set(id, value)`or `_entries.get(id)` or `_entities.delete(id)`

Jaifroid: Ah, OK, thanks! I had understood that the entire FileCache object should be a Map with extended get, store and unlink functions in its prototype. What you say makes a lot more sense now. (I struggle with OOP.)

Jaifroid: I've managed to convert the code in `filecahce.js` to use a Map to store the file slices. However, the code is currently non-functional due to an issue with using a composite key that is an Array or an Object (I am using an array `key = [filename, id]`). While this stores correctly, to retrieve the data corresponding to this Array key, I would need to use the exact same array (in memory) as the retrieval key. It is not enough to make a `newkey = [filename, id]`, as even if the array key looks exactly the same, it is in fact a different array (i.e. a different location in memory), and `Map.get(newkey)` returns undefined, even though I can see that there is an entry in the Map using a key `[filename, id]`. This is the same reason that in console.log checking `{} === {}` returns `false`. The two Objects here are two different memory locations, so are not condered identical.

To use an Array (or Object) as a key, we would need to save a reference to a non-primitive instance of the Array, which is not viable for us. For discussion, see https://stackoverflow.com/questions/32660188/using-array-objects-as-key-for-es6-map

I still think using a Map is a good idea, but it looks like we are stuck with a composite key that is a String with a separator that would not be used in a ZIM filename. Rather than looking for exotic characters, I suggest we just use "@", which is most unlikely ever to appear at the end of a ZIM filename (especially since the filename part includes the `.zim` extension). Adding this would add approx 4K to the memory requirement (1 character * 4000 entries). I think we can be pretty sure no ZIM file will ever be called `wikipedia_en_all_maxi_2020-06.zim@1`, can't we?

An alternative would be to ensure the Cache is emptied on new ZIM load. I haven't checked whether in fact this is the case already.

Jaifroid: PS If we did empty the cache, we would no longer need to use the filename at all in the key. Since I modified peter-x's code to use absolute offsets from the start of the archive, not offsets relative to each ZIM file part, it's kind of redundant to use the filename in any case, unless we cannot empty the cache between ZIM loads for whatever reason. I'll investigate.

mossroy: I agree that it could be a good workaround
We don't really need to keep blocks of several ZIM files at the same time

Jaifroid: @mossroy This is now complete. To summarize: I've converted the storage to a Map, and have stripped out the use of the filename as part of the key. The key is now numeric (it is the absolute archive offset of the block to be read/stored). Since Maps have a size property, we no longer need to keep track of the Cache size with a counter. The Cache is reset in `zimFile.js` when a new ZIM is loaded. There is no need for explicit reset commands in `app.js`. I may be able to strip out the added generic filename in `zimFile.js` as well if it is no longer needed anywhere (will add a comment in code for that). But this is now passing tests and ready.

www/js/lib/zimfile.js
```diff
@@ -282,6 +278,9 @@ define(['xzdec_wrapper', 'zstddec_wrapper', 'util', 'utf8', 'q', 'zimDirEntry'],
                 var urlPtrPos = readInt(header, 32, 8);
                 return readMimetypeMap(fileArray[0], mimeListPos, urlPtrPos).then(function (data) {
                     var zf = new ZIMFile(fileArray);
+                    // Line below provides an abstracted filename in case the ZIM file is split into multiple parts;
+                    // it greatly simplifies coding of the block cache, as it can store and respond to offsets from the start of the file set
+                    zf.name = fileArray[0].name.replace(/(\.zim)\w\w$/i, '$1');
```
Jaifroid, line 283: See if above lines can be removed now.

www/js/lib/zimfile.js
```diff
@@ -48,89 +47,93 @@ define(['xzdec_wrapper', 'zstddec_wrapper', 'util', 'utf8', 'q', 'zimDirEntry'],
...
                 return concatenated;
             });
         }
     };
 
     /**
-     * 
-     * @param {Integer} offset
-     * @returns {DirEntry} DirEntry
+     * Read and parse a a Directory Entry at the given archive offset
```
Jaifroid, line 121: Typo "a a Directory Entry".

mossroy (changes requested):
I was about to approve the PR, but I might have a found a bug.
On my slow Firefox OS device, if you open a big article (with many images in it), then quickly switch to another ZIM file, the welcome page of the new ZIM file can be corrupt, with error messages in the browser console, like : 
```
could not find DirEntry for image:I/m/Emojione_1F5FA.svg.png TypeError: dirEntry is null
Trace de la pile :
ZIMArchive.prototype.readBinaryFile@app://kiwix.org/www/js/lib/zimArchive.js:293:9
displayArticleContentInIframe/loadImagesJQuery/</<@app://kiwix.org/www/js/app.js:1386:21
_fulfilled@app://kiwix.org/www/js/lib/q.js:854:54
Promise.prototype.then/</<@app://kiwix.org/www/js/lib/q.js:883:30
Promise/promise.promiseDispatch@app://kiwix.org/www/js/lib/q.js:816:13
defer/become/</<@app://kiwix.org/www/js/lib/q.js:624:17
runSingle@app://kiwix.org/www/js/lib/q.js:137:13
flush@app://kiwix.org/www/js/lib/q.js:125:13
```
It's certainly because the browser is still reading images in the background, even after we call the `init` method of the cache. So I suppose the cache gets populated by blocks coming from both ZIM files.

So I was wrong when I said keeping blocks from several ZIM files was not a need. It is not a need in functional terms, but it's still a technical need for this kind of edge cases.

So we probably need to keep the ZIM file name in the cache key in some way, in the end.

mossroy:
It's probably possible to concatenate the strings (filename and offset) with a non-printable character between them, that would be even more unlikely than @.

Based on https://en.wikipedia.org/wiki/Filename#Comparison_of_filename_limitations, the NULL character (ASCII code 0) would be the only one forbidden on both Windows and Linux filesystems. It's allowed on Mac OS, but is certainly the most unlikely one

Jaifroid:
Good catch, @mossroy. It is quite an edge case, but clearly we shouldn't let such a potential bug stand when it can be fixed fairly easily. I would be tempted to say that the real bug here is not this code, but the fact that we have not managed to cancel resource extraction fully once we navigate away from an article. However, fixing that is more complicated than adding back an identifier to this code.

Jaifroid:
@mossroy Commit https://github.com/kiwix/kiwix-js/pull/658/commits/38cb3e7f8675a9ca32cc80019a9482a6511438c4 appears to solve this issue.

To avoid wasting an extra ~132 KB of memory by storing the full filename repeatedly 4,000 times in the cache, I have instead implemented a temporary numeric id for each loaded ZIM file. The ID is a simple incrementing integer, that will never be the same after each ZIM load, and that is voided between app sessions (obviously). It allows the construction of a simple string ID, with format `fileId:offset` e.g. "1:03145684". This only uses two extra bytes per entry (the id and a ":" separator), i.e. 8 KB. Because we are not using a filename, we don't need to find an exotic separator.

I have tested in IE11 and Edge (latter in jQuery and SW modes) with console.log open, by loading the full English Wikipedia "Paris" article, and then quickly loading a different ZIM while the "Paris" assets are being read. I have not encountered any read errors.

Jaifroid:
I have determined that the ':' separator is redundant in the cache key (humans will not read the cache entries, and the Map retrieval functions do not need it). The new format of the cache key is therefore `offset + '' + fileId`. (The `''` is merely used to convert the numbers into strings so they can be concatenated.)

I have also added some type definitions, to complete the documentation.

www/js/lib/filecache.js
```diff
@@ -36,21 +36,36 @@ define(['q'], function (Q) {
...
-        // linked list of entries
-        this._first = null;
-        this._last = null;
     }
 
     /**
      * Tries to retrieve an element by its id. If it is not present in the cache, returns undefined; if it is present,
      * then the value is returned and the entry is moved to the top of the cache
-     * @param {String} key The block cache entry key (file.id + ':' + byte offset)
+     * @param {String} key The block cache entry key (byte offset + '' + file.id)
```
mossroy, line 68: The file id allows for a more simple separator : that's a good idea!
But I think the separator should be kept : even if it's unlikely, there could be a mix between 1+''+11 and 11+''+1 (for example). I think it's better to have a 100% safe key (with a separator like `:`) than to save only 4kB (without a separator)

Jaifroid: He he, you never miss anything, @mossroy! I agree there is a theoretical ambiguity in the key format. In practice, the ambiguity should never occur, because the smallest ambiguous number is (I think) 110, which could be 11:0 or 1:10. However, we void `cache._entries` (the Map) on each ZIM change, so the only entries from a previous ZIM that can remain in the cache would be assets still resolving from the previous ZIM load (which woudl be numbered 9 in the case of 1:10).  Nevertheless, it is a small cost to remove any possibility of ambiguity, as you say, and I have restored the separator. Thank you for your eagle eye!

mossroy:
It's working fine on my Firefox OS device.

In fact, I was wrong when I said I saw the bug of mixed content in the cache : it was in fact not a regression but a faulty ZIM file (wikivoyage_fr_all_novid_2019-06.zim). Sorry about that, I deleted it on my devices.

Anyway, I think it's still safer to add the ZIM file id in the cache key because it could happen. I would find it more logical to have this id as a prefix than as a suffix, but it's not important.
In any case, with this id in the key, it's not longer necessary to reinit the cache after each file change. Its content will be "naturally" replaced with the time. Should we remove this reinit code? As you wish : keeping it might slightly improve performance on a ZIM file change. If you prefer to keep it, it would be good to add comments in the code to explain why

www/js/lib/filecache.js
```diff
@@ -0,0 +1,221 @@
...
+                        blocks[offset] = result;
+                    });
+                }(id));
+            } else {
+                hits++;
+                blocks[id] = block;
+            }
+        }
+        if (misses + hits > 2000) {
+            console.log("** Block cache hit rate: " + Math.round(hits / (hits + misses) * 1000) / 10 + "% [ hits:" + hits + " / misses:" + misses + " ]");
```
mossroy: I agree

Jaifroid, line 198: @mossroy I shall remove all console logging from this file before merging. However, I think it is woth keeping this line commented out, and the associated counters (`hits` and `misses`, also commented out) because future DEVs may wish to tune the Cache according to changing software and hardware environments.

Jaifroid:
> Anyway, I think it's still safer to add the ZIM file id in the cache key because it could happen.

Actaully, I checked, and I *did* find (in rare circumstances) keys with a previous ZIM ID in the cache, so the accident with the faulty ZIM file was fortuitous. The rare circumstance is the one you identified by accident, i.e. that even after voiding and re-initializing the Map, completing asset Promises can re-populate it erroneously with data from the previously loaded ZIM. In this case, the Promises do not keep their own copy of the Cache (Map) object, because it is referenced outside the Promise closure, so they can still populate the newly initialized Cache.

You are right that previous Cache Keys will be purged over time. However, they are also completely useless, because unless I maintain a separate Map of ZIM filenames to ZIM IDs, they can never be re-used. To do that might be "over-coding": unless we think it is likely that someone will go repeatedly back and forward between two ZIMs in the same app session, there is no advantage, and possibly some performance disadvantage, in not re-initializing the Cache Map. Let me know if you think we should address this scenario (it would need a little bit of code in `zimfile.js`.

Otherwise, I'll add some comments and switch the `offset:id` order (I was thinking that it was a least-significant bit of information, only there to cover a sort of bug in our code, i.e. #426, but I guess that's arguable).

Jaifroid:
Hmm I've found a bug: redundant Cache entries are not being deleted over time and the cache is growing larger than the set limit. The Cache is coded with a very clever recursive logic (not my logic!) where all entries are linked to other entries by containing a prev field and a next field that actually contain the liniked Cache Entry. I can see that the code to "unlink" an entry is called, but the actual `Map.delete(entry.id)` command is returning `false`, which indicates an unsuccessful deletion attempt, and indeed the Cache keeps growing larger and larger.

Sorry I didn't notice this earlier. I need to investigate, as this is a serious issue that could crash the app over time. I'm converting this back to draft while I investigate.

Jaifroid:
Because Maps retain the insertion order (see excerpt from MDN below), it should be possible to simplify peter-x's code. The catch is that it's necessary to move an entry to the "most-recent" position (actually the last entry in the Map, rather than the first) if it is read from the Cache, ensuring only unused older entries are removed.

> The keys in Map are ordered in a simple, straightforward way: A Map object iterates entries, keys, and values in the order of entry insertion.

www/js/lib/filecache.js
```diff
@@ -0,0 +1,228 @@
...
+     * @property {Uint8Array} value The cached data
+     */
+
+    /**
+     * A Block Cache employing a Least Recently Used caching strategy
+     * @typedef BlockCache
+     * @property {Integer} _limit The maximum number of entries in the cache 
+     * @property {Map} _entries A map to store the cache keys and data
+     * @property {CacheEntry} _first The most recently used entry in the cache
+     * @property {CacheEntry} _last The least recedntly used entry in the cache
```
mossroy, line 54: Minor typo here : recently instead of recedntly

mossroy:
Don't bother more on the "not re-initializing cache" topic.

Regarding the cache implementation, you also could have a look at existing ones (and even use them instead of our current) : https://gist.github.com/dashsaurabh/bf7a02e85d325743076ba9943a286e9f , https://blog.lmerza.com/2019/01/24/lfu-and-lru-caching-in-javascript/ , https://github.com/monsur/jscache . They're not using a `Map`, though

Jaifroid:
Thanks for those examples and theory of LRU and LFU caches. I followed the example from https://gist.github.com/dashsaurabh/bf7a02e85d325743076ba9943a286e9f and the Cache is now deleting the least frequently used entry correctly. It seems peter-x's implementation hadn't quite completed the logic of unlinking and deleting the lfu entry.

Still in draft, because I need to trace each step through carefully to be sure it's right. The linked lists are a bit mind-bending, and I can't help thinking that an easier implementation should be possible with a Map, because a Map preserves the order of insertion of entries, meaning we wouldn't have to maintain the linking lists and reset the links. However, I don't feel I understand the theory thoroughly enough to re-do it myself in Map format.

Jaifroid:
Having said that, I find someone has explained how to do it with JS Map - and it's along the lines I was thinking. Apparently it's a common interview question... I'll review this later this week, because it has the potential to simplify quite radically:

https://markmurray.co/blog/lru-cache/

Jaifroid:
Quick update: latest commit refactors to use a simple Map-based implementation. It works very well (and fast) except in IE11, where the `Map.keys()` iterator is not supported (used to find the first = oldest key quickly in order to delete it once the cache is full). Therefore we will either have to find another way to iterate just part of the Map to get the first (oldest) value -- `forEach` is supported but we wouldn't want to loop over the full 4000 Map entries, and there is no way of breaking a forEach loop -- or else we'll have to keep track of the first entry, which means keeping the linking code. I tried `for ... of` but IE doesn't support it either, and `for ... in` doesn't work with Maps. There must be a low-level way (in IE11) to get entries out of a Map when we don't know the key.

Jaifroid:
OK, I've added IE11 support. The only way getting the first key from the Map in IE11 is to do a forEach loop over all 4,000 map entries. However, I've pared down the processing to a minimum, so that each loop iteration simply returns instantly once the first (=oldest) value has been found, and I have tested with performance.now(). The full iteration of 4000 entries takes an average of 1 millisecond (sometimes 0.6 milliseconds) in IE11. So, to a human, this will not be noticeable.

There is an alternative: once the cache is full we could empty it instantly in IE11 only. It takes a *lot* of full Wikipedia queries and page loads to fill the 4,000 cache entries, so the cache would be being emptied rarely.

See what you think. The current loop can be seen in commit https://github.com/kiwix/kiwix-js/pull/658/commits/f22aaa68f97cfdf985807047557920b9e249edcc.

Jaifroid:
I've thought of a solution for IE11: instead of iterating the Map 4,000 times for each Cache.store operation in order to delete one entry, and instead of clearing the entire Map, we can simply clear the oldest 10% of the entries. This means that we don't hit the Cache too hard, and we don't use excessive CPU once the Cache is full. I have run some tests in IE11, and below is the output showing how many entries were cleared and when in the read cycle. As can be seen, there's a slight losss of hit / miss ratio immediately after a delete operation, but it doesn't then run again for several article loads (unless they are huge articles), and the loss seems pretty acceptable.

image

(Tests done with full English Wikipedia)

mossroy:
This `Map` version works fine on my Firefox OS device.

Regarding the IE workaround, I don't see why https://github.com/kiwix/kiwix-js/pull/658/commits/f22aaa68f97cfdf985807047557920b9e249edcc is so bad : it will exit after the first loop, so should not iterate over all the cache entries? Is it because calling `forEach` is slow?
In any case, it's not a big issue if IE11 is slower. What is important is that this PR does not reduce its performance. If this PR does not enhance its performance as much as other browsers, don't bother too much on optimizing it

Jaifroid:
> I don't see why [f22aaa6](https://github.com/kiwix/kiwix-js/commit/f22aaa68f97cfdf985807047557920b9e249edcc) is so bad : it will exit after the first loop, so should not iterate over all the cache entries? Is it because calling `forEach` is slow?

Unfortunately, it will still run through all 4,000 entries, even if it only exexutes `if (c > q) return;` 3,999 times and does useful work only on the very first entry. There is no `break` for `forEach` loops (there is a hacky way to break the loop by causing an exception within a `try-catch` block, but it's "very bad" to use exceptions for programme flow, so I read!). It kind of offends my sensibilities to write something so inefficient, even if the slowdown is trivial on a modern CPU. Remember that this 4,000 churn will execute for every read of a small amount of data not already in the cache, and that can be thousands of reads.

I thnk deleting 25% of the Cache in one go gets round this churn without impacting IE11 performance., The loop will only ever run once, and then wait till the Cache is filled with another 1,000 entries before running agiain. This is implemented in the latest commits.

This is ready for final testing / review. I never thought it would be a complex PR, but JS is never easy!

mossroy:
OK I understand.
I found a polyfill for IE and the Map feature : https://github.com/zloirock/core-js#map. But it might be overkill for the need

Jaifroid:
> I found a polyfill for IE and the Map feature : https://github.com/zloirock/core-js#map. But it might be overkill for the need

I think there's a good argument for using `babel/core-js`, which would allow us to write ES6 (or its superset, Typescript) and compile to compatible code without worrying so much about workarounds for specific browsers (mostly IE, but not only).

But I think this needs to be done as part of #554 (which presupposes #367, and also removing Q, because core-js handles Promise support as well, including modern features such as `async`).

As I understand it (but I don't have any direct experience), automatic code analysis is done by babel, which then attaches the core-js polyfills required by the code based on the language features used.

mossroy (approved):
It's working fine on my Firefox OS device, and also on Firefox 82 and Chromium 86.

I think it was a really good choice to use a `Map` and its ability to keep the order of keys : it makes the code more simple (and probably faster).
It might have been a bit faster on IE11 to use an Object (instead of the `Map`, like it was done in earlier versions of this PR), because of its lack or .keys(). But it would be too much code to maintain for a very deprecated browser : I prefer your workaround.

Jaifroid:
> It might have been a bit faster on IE11 to use an Object (instead of the `Map`

The issue is that, because Objects do not guarantee the order of entries (at least in older browsers), it is necessary to maintain complex lists linking the entries in order of insertion (which is what the versions of the LRU Cache that use Objects implement). And as you say that would have been a lot of extra code to maintain for one broswer.

www/js/lib/filecache.js
```diff
@@ -0,0 +1,178 @@
...
+     * Stores a value in the cache by id and prunes the least recently used entry if the cache is larger than MAX_CACHE_SIZE
+     * @param {String} key The key under which to store the value (file.id + ':' + byte offset from start of ZIM archive)
+     * @param {Uint8Array} value The value to store in the cache 
+     */
+    LRUCache.prototype.store = function (key, value) {
+        var entry = this.cache.get(key);
+        // If the key already exists, delete it so that it will be added
+        // to the bottom of the Map (bottom = most recent)
+        if (entry) this.cache.delete(key);
+        else entry = value;
```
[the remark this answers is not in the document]
Jaifroid: I decided to keep the cached value for memory management purposes. Since the Map stores a *reference* to an object stored in memory (rather than storing a new object), then if we just overwrite the key with a new object (even with identical data), then we are not re-using memory optimally, and have to rely on the garbage collector to delete unlinked objects. I read somewhere that Maps hold on strongly to an object (as opposed to a [WeakMap](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakMap)).

So this explains why I decided to get the object reference from the map, and only store a new value if it returns `undefined`.

In practice, I think we will only very rarely look up in the ZIM a file slice that was already in the Cache: only where there is a race between two lookups requesting the same data before any one has completed and stored the data in the Cache. Once the Cache is substantially populated, this will be very rare.

But if you feel any of this is unsafe (or my reasoning flawed), I'll be happy to modify.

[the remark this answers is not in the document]
mossroy: No, it's fine, you are right.
But you should add these explanations in a comment, because this code looks weird at first sight

Jaifroid:
Tested on IE11 and FFOS Simulator in jQuery mode, and on Fiefox 81.0.2, Edge Legacy, Edge Chromium 86.0, in jQuery and SW modes. I still need to test on Windows Mobile.

Jaifroid:
@mossroy I have completed my testing. I'll squash and merge tomorrow (Thursday) afternoon unless I hear otherwise.

Issue #639, opened 2020-07-11
Improve our CodeFactor mark


mossroy:
CodeFactor does not give us a very good mark for now : C-
especially for app.js (F) : https://www.codefactor.io/repository/github/kiwix/kiwix-js/file/master/www/js/app.js

There are many minor issues that could easily be fixed.
And we also might configure it to ignore the external libraries.

Jaifroid:
I agree. Codefactor even offers to make PRs for simple fixes. Might be worth experimenting with that, if the proposed fixes are sensible.

ram690514:
@kelson42  i would like to work on this issue

kelson42:
@ram690514 You are welcome.

mossroy:
@ram690514 : I'd prefer that you finalize work on #660 PR first, before starting work on another issue.

ram690514:
@mossroy :I am working on it . I will fix it surely as fast as possible.

ram690514:
@kelson42 : I have pushed a pull request . please check it

ram690514:
@mossroy : I have pushed a pull request. can you please check it

Jaifroid:
@ram690514 Thank you, but it's impossible to see what you have actually fixed from that PR. Could you please close it and re-do it using with only the actual code changes, or perhaps with one commit per change, Please do not change formatting and indenting, only focus on the actual code changes. If the code change isn't clearly shown in the diff, we cannot see what you have done. and so cannot check your work!

ram690514:
@Jaifroid  I will surely do it

ram690514:
I have created a pull request regarding this issue..please check @kelson42

ram690514:
@Jaifroid : I have pushed a pull request.please check it.

Jaifroid:
I've reviewed it and suggested changes. As a matter of etiquette, when you create a PR, please give it an informative title ("Changes" means nothing and could apply to any PR), and also provide an overview description of what you have done. You don't need to provide a description now, as we know what you have done, but it is good practice for a PR. You should change the PR title, however.

ram690514:
Thanks for your your suggestions, I will surely do it

ram690514:
@Jaifroid : I have updated the pull request ,please check it

ram690514:
@kelson42 : I have updated the pull request,please check it

ram690514:
I had restored the element message channel and updated the pull request

Jaifroid:
[...]

PR #662, opened 2020-10-20
Issue #639 fixed

ram690514:


mossroy (changes requested):
This is unusable.
You did not say what you did, and the diff https://github.com/kiwix/kiwix-js/pull/662/files shows that almost all lines are changed.
How can you expect us to review?

If you want to have chances that a PR is reviewed (and merged), take the time to test your changes, explain what you did, check the diff is understandable etc.
And I would advise to focus on one issue/PR instead of doing several in parallel. It's not a race, and quality should be preferred over quantity.

Keep in mind that we are all volunteers on this project, with limited personal time to spend on it

ram690514:
@mossroy,
I am really sorry  , I  am a beginner in the field of open
source contribution that's why there came so many errors while fixing an
issue .
I actually don't know the fact that for each and every change i should do a
commit .
Thank you , for showing me the fault that i did.

mossroy:
I did not say that (even if it's indeed advised to do a commit for each different change).
What I say is to make sure the PR is readable and understandable. A global reformatting should not be mixed in the same PR as other changes, and usually creates conflicts. So only do other changes, and create one PR with all your changes, and explain what you did

PR #668, opened 2020-11-05
Exempt articleId assignment from Active Content Warning

### LINKED ISSUE #667: Latest ZSTD WikiMed shows our custom "Unable to display active content" warning
It's very nice to see the first mainstream ZSTD WikiMed (the November English version). It's 230MB smaller than last month's, and noticeably snappier, at least on my PC!

There is a minor issue with the landing page (affecting only Kiwix JS and KJSW in jQuery mode): for some reason, the landing page is throwing up our "Unable to display active content" warning (see screenshot). However, there is no active content on the landing page. There must be some new embedded script that we're detecting as active content.

It should be a quick PR to exempt that script.

image

Jaifroid:
Fixes #667 for latest WikiMed ZIMs. The warning is triggered because a new inline script block has been added to the landing page that looks like this:

`<script>let articleId = 'Wikipedia:WikiProject_Medicine/Open_Textbook_of_Medicine2'</script>`

The fix is to add another exemption to our `regexpActiveContent`. The full amended regex is:

`var regexpActiveContent = /<script\b(?:(?![^>]+src\b)|(?=[^>]+src\b=["'][^"']+?app\.js))(?!>[^<]+(?:importScript\(\)|toggleOpenSection|articleId\s?=\s?['"]))(?![^>]+type\s*=\s*["'](?:math\/|[^"']*?math))/i;`

And the added code is: `|articleId\s?=\s?['"]`. This expands the negative lookahead so that any inline script that contains `articleId = '` will **not** match, and so will not trigger the Active Content Warning. I have made it as generic as possible (no spaces around the `=`, or different quotation style will also not match).

The Regex has been tested in RegexBuddy and the PR has been tested quickly on the new WikiMed ZIM. It works as intended.

### DIFF
--- a/www/js/app.js
+++ b/www/js/app.js
@@ -1227,12 +1227,11 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
     var regexpTagsWithZimUrl = /(<(?:img|script|link|track)\b[^>]*?\s)(?:src|href)(\s*=\s*["'])(?:\.\.\/|\/)+(?=[-IJ]\/)/ig;
     // Regex below tests the html of an article for active content [kiwix-js #466]
     // It inspects every <script> block in the html and matches in the following cases: 1) the script loads a UI application called app.js;
-    // 2) the script block has inline content that does not contain "importScript()" or "toggleOpenSection" (these strings are used widely
-    // in our fully supported wikimedia ZIMs, so they are excluded); 3) the script block is not of type "math" (these are MathJax markup
-    // scripts used extensively in Stackexchange ZIMs). Note that the regex will match ReactJS <script type="text/html"> markup, which is
-    // common in unsupported packaged UIs, e.g. PhET ZIMs.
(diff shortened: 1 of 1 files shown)


mossroy (approved):
I did not test, but it looks good.

Of course, this whole regexp is not very reliable.
But, well, the consequence of a false-positive or false-negative is minor, and I don't have a better idea for now

Jaifroid:
I agree it's become a bit of a monster regex, and I had to read my comment on it carefully in order to maintain it.

The only alternative I can think of is a cosmetic one to aid maintainability: we could run a short regex that recognizes a simple `<script ...>...</script>` tag pair, and then runs a function over each pair. The function would then have a series of short, documented tests on the string. The tests would be essentially the same as those already in this regex. It would be a more readable/maintainable regex function, but would do the same thing as this, just a tiny a bit slower.

Maybe if we need to expand this regex any further in the future, we should do that. I can add a comment reminding myself of this.

mossroy:
My remark was not about the fact that the regexp is complicated (the comment is meaningful), but on the fact that it _could_ match articles that don't have active content, and not match articles that have active content. It's making assumptions on the ZIM content, which can become wrong (this PR shows it).
But it's "only" a warning message (and only in jQuery mode), and I don't see how to easily do that better : let's leave that this way

PR #670, opened 2020-11-17
Add WebP Polyfills in jQuery mode

Jaifroid:
@mossroy I 'solved' the issue with the polyfill clashing with require.js `define` by loading the polyfills as a shim module in the RequireJS config. So, the code in this PR nearly works, i.e. this will decode _most_ images correctly in IE11, and also in Firefox OS simulator.

**BUT**, while most images decode and display just fine, on both IE11 and FFOS simulator, some images do not load at all when there are multiple images on a page. They seem to get lost somewhere in the WebpMachine, and never make it back to the `then` funciton of the Promise (see code in `uiUtil.js`). This does not happen in KJSW, where I am currently loading the module in `index.html` using some conditional logic.

The main WebpMachine is an Emscripten module. My only guess at the moment is that RequireJS is killing the module between repeated calls to `uiUtil.js`.

I think you know RequreJS better than I do. Can you see anything wrong in the way I'm requiring and defining (look at `init.js` and `uiUtil.js`)? I remember you had an issue with `xz` originally which could not be loaded with RequireJS, but I managed to solve that for both `xz` and `zstddec`, so in principle Emscripten modules can be loaded with RequireJS..

EDIT: Is there a way to get RequireJS to persist a module in memory? A crude way might be define the `webpMachine` as a global const in init.js, but I don't know if this would somehow defeat the purpose of RequireJS:


Jaifroid:
I've tested defining the WebpMachine in init.js, and it doesn't improve the situation.

So, my next guess is that the WebpMachine rejects processing an image if it is busy, which is a similar situation to the one we have with `xzdec` and `zstdddec`. This would explain why on a page with lots of images, we see alternate gaps in the processing. I'll test this hypothesis.

mossroy:
It's the hypothesis I was thinking of, too

www/js/lib/uiUtil.js
```diff
@@ -399,6 +409,20 @@ define([], function() {
...
+    var webpMachine;
+    var testWebP = function (callback) {
+        var webP = new Image();
+        webP.onload = webP.onerror = function () {
+            callback(webP.height === 2);
+        };
+        webP.src = 'data:image/webp;base64,UklGRjoAAABXRUJQVlA4IC4AAACyAgCdASoCAAIALmk0mk0iIiIiIgBoSygABc6WWgAA/veff/0PP8bA//LwYAAA';
+    };
+    testWebP(function (support) {
+        if (!support) webpMachine = new webpHero.WebpMachine();
```
mossroy, line 422: It's minor, but, when webp is not supported, the webp support detection is probably done twice here : manually by the testWebP function above, and also by the webpHero library.
It could be avoided by using passing a dummy webpSupport option to the WebpMachine constructor, that always return false (because it has already been detected)

mossroy:
Regarding the issue with several webp images, did you check you use the latest version of webpHero (currentlt v0.0.0-dev.27)?

By the way, it would be good to be able to see this version in our code.
I suggest to add this version as a suffix of the javascript file(s), and to use requireJS feature to point to this file (at the beginning of init.js, just like it is done for jQuery)

www/js/lib/uiUtil.js
Jaifroid: Yes, originally I attached the scripts inside this `testWebP` block (in index.html) to avoid loading a memory-heavy script for the majority of browsers that don't need it. I'm still thinking of conditionally including the polyfills in the RequireJS configuration object, to address https://github.com/chase-moskal/webp-hero/issues/39:

> * conditional loading of the whole library — the currently recommended strategy in the readme is alarmingly wasteful, because it will load the entire 1MB bundle, even on new browsers which already support webp. i think we should create some kind of wrapper which avoids loading the whole library when webp is supported

Jaifroid:
> Regarding the issue with several webp images, did you check you use the latest version of webpHero (currently v0.0.0-dev.27)?

I used the distribution loaded by the demo, but you're right, the Dev may not have updated the demo with the latest dist. I'll check, though I don't think the API has changed, and in any case it's best to use the official `unpkg` repo (since we're not using npm).

> I suggest to add this version as a suffix of the javascript file(s), and to use requireJS feature to point to this file (at the beginning of init.js, just like it is done for jQuery)

Agreed.

mossroy:
They should have the same problem in webpHero when they parse the DOM and decode every webp image.
Reading https://github.com/chase-moskal/webp-hero/blob/master/source/webp-machine.ts, it seems to me that they indeed do not decode several images in parallel : they use `await` to make sure one image is decoded before decoding the next one.

`await` is unfortunately not available on IE11 and some old browsers, but we might do a similar sequential decoding of webp images

NB : for performance reasons, we also could override the `detectWebpImage` function (by passing it as constructor option, that always returns true), because we already call it only with webp images

Jaifroid:
Thanks for the ideas @mossroy and for checking the source code. I've added a simple queue to the WebpMachine, and it seems to solve the issue (poorman's `awat`!). It is not passing tests, but it looks like the test issue is unrelated: there is an error connecting to Sauce Labs before any code is run.

The queue is event driven rather than using `setTimeout`: i.e. the loop is called on the next queued item by the the `.then` (and .`catch`) clasues of the WebpMachine Promise. I can probably streamline this code a bit, but as proof of concept, it works on IE11 (untested on anything else).

I've only tested this ZIM in jQuery mode so far -- I haven't yet downloaded a browser that supports Service Worker but does not support WebP.

Jaifroid:
> for performance reasons, we also could override the detectWebpImage function (by passing it as constructor option, that always returns true), because we already call it only with webp images

Good idea.

Jaifroid:
Also working on Firefox OS simulator.

Jaifroid:
@mossroy Regarding your comment on overriding the `detectWebpImage` function, I tried the code below, but it is never actually executed. Have I done it right? It may be the case that the `WebpMachine.decode()` function (the only one we use) never actually does any detection on the image, and that this function is only used as part of the `polyfillDocument()` function.

```
    testWebP(function (support) {
        if (!support) webpMachine = new webpHero.WebpMachine({
            detectWebpImage: function () {
                return true;
            }
        });
    });

```

Jaifroid:
Regarding the "alarmingly wasteful" strategy of loading the entire polyfill even on browsers that will never use it, I cannot think of a way of conditionally loading it via RequireJS. I know how to load it conditionally in `index.html`, but it's not as elegant as using require, and presumably require has some performance or memory management advantages.

Actually I did think of one way: if WebP is natively supported, maybe we could redefine the 'path' property of the polyfills (in the RequireJS configuration block) to a tiny dummy .js file that just returns true. Is it worth doing something like this in your opinion? It seem a bit hacky!

www/js/lib/uiUtil.js
mossroy: It might be possible to load WebpHero only if needed, using requirejs : https://stackoverflow.com/a/8457158/6431035 (I did not test) : the idea would be to not require the WebpHero dependency on the uiUtil.js module, but only in a function that creates the webpMachine variable (hold in the uiUtil.js)

But if it does not work or is too complicated, this could be left as it is. We'll see later if it can be optimized

mossroy:
About `detectWebpImage`, I think you do it right.
The `decode` function does not seem to call `detectWebpImage`, but `polyfillImage` does : I thought you were using this one (sorry, my bad) : https://github.com/chase-moskal/webp-hero/blob/5d671f81fdd3b30f68bfb5ea4bd9ae1e74478af8/source/webp-machine.ts#L66
Should we use `polyfillImage` instead of `decode`? `polyfillImage` handles some cases that are not relevant for us, and adds a cache. But it's not usable here as it fetches the data from the `src` attribute of images.
So you used the right API.

Should we add some cache to avoid decoding the same image several times? I'm not so sure.
If we did, it should also avoid reading the image in the ZIM file, so we should do that at the ZIM backend level.
It's true that there are sometimes the same small images on many article contents. On the other hand, browsers that do not support webp are usually on old devices, that might not have a lot of memory : it might be worse to keep some images in a memory cache.

So I'd leave the code as it is, in the end

www/js/lib/uiUtil.js
Jaifroid: @mossroy OK, thanks - that confirms an idea I was working on. I've pushed a **prototype** of conditional loading. I think it can be improved -- on reflection, the test doesn't need to happen in init.js and can be moved back to uiUtil.js as suggested in the example above. But the latest commit enables conditional loading.

Jaifroid:
OK. I agree. It's quite rare that any one page will use the same image twice, except flags as you say (and they are SVG, not sure if they are converted to WebP or not).

By the way, tests are failiing because my latest commit did not update the init.js used by tests. I'll refactor first, as I don't think it's necessary.

Jaifroid:
I couldn't relocate the WebP test to `uiUtil.js` because the test is async, and I cannot put the entire uiUtil inside the callee of an async function (it won't export the functions needed by app.js). The test needs to be ready before we call `define([], funciton()...)` if we are to load the WebP polyfill conditionally. So I've kept it in `init.js`, which is the only script that runs early enough.

Conditional loading is working and tested in IE11 and Edge Chromium so far.

Unless you can think of a better way to do this, this is now ready for review.

www/js/lib/uiUtil.js
```diff
@@ -20,27 +20,65 @@
...
-        var url = URL.createObjectURL(blob);
-        jQueryNode.on('load', function () {
-            URL.revokeObjectURL(url);
-        });
-        jQueryNode.attr(nodeAttribute, url);
+    function feedNodeWithBlob(node, nodeAttribute, content, mimeType) {
+        // Decode WebP data if the browser does not support WebP and the mimeType is webp
+        if (webpMachine && /image\/webp/i.test(mimeType)) {
+            // Queue WebP images to be decoded (the WebP polyfill is single-threaded and will reject a job if it is busy)
+            webpQueue.push({ 'node': node, 'nodeAttribute': nodeAttribute, 'content': content });
```
Jaifroid, line 56: We should consider the possibility of applying an image queue in `app.js` to the whole function `loadImagesJQuery()` instead of doing it just for WebP images. The same logic as here could be transferred there very easily. The main advantage would be that images would be loaded from the top of the page, sequentially, rather than in a somewhat random order as at present, at least in jQuery mode (and some browsers do this for SW mode automatically). This would be a better user experience, as images above the fold would appear fairly instantly.

Jaifroid:
I changed the title of this PR to "Add WebP Polyfills in jQuery mode" to make clear that I have not attempted to address SW mode at all, as it should "just work". We do however need to test that, in order to feed back to mwoffliner if something is wrong with the built-in polyfill loading. We need to find a portable version of a browser that supports SW but not WebP.

www/js/lib/uiUtil.js
```diff
@@ -20,27 +20,65 @@
...
-define([], function() {
 
+// DEV: Put your RequireJS definition in the rqDef array below, and any function exports in the function parenthesis of the define statement
+// (This method is needed in order to load WebP polyfills conditionally)
+var rqDef = [];
+
+// Add WebP polyfill only if webpHero was loaded in init.js
+if (webpMachine) {
+    rqDef.push('webpHeroBundle');
+}
```
Jaifroid, line 31: Do we need to detect whether the app is in SW mode, and avoid adding the polyfills if that is the case? We certainly need to test whether loading polyfills via RequireJS conflicts with them being loaded in the iframe by the ZIM in SW mode. There *shouldn't* be a conflict because the iframe `contentWindow` is isolated from the `window` object. If we did disable polyfills in SW mode, it would imply that a user switching to jQuery mode might need to exit the app and reload it. (We can detect SW mode easily by scanning document.cookie.)

www/js/lib/uiUtil.js
mossroy: If you want.
It's true that it would probably improve the user experience in jQuery mode.
But maybe it should be a separate issue/PR?

www/js/lib/uiUtil.js
```diff
@@ -20,27 +20,65 @@
  * along with Kiwix (file LICENSE-GPLv3.txt).  If not, see <http://www.gnu.org/licenses/>
  */
 'use strict';
-define([], function() {
 
+// DEV: Put your RequireJS definition in the rqDef array below, and any function exports in the function parenthesis of the define statement
+// (This method is needed in order to load WebP polyfills conditionally)
```
mossroy, line 25: The comment should explain why we load it conditionally : it's not necessary, but an optimization : we want to avoid loading a useless library for the most common use-case (where webp is supported)

mossroy (approved):
It's working fine on my Firefox OS device with the test ZIM file : nice job!
I did not test on any other browser

tests/init.js
```diff
@@ -24,11 +24,19 @@
 
 // Define global params needed for tests to run on existing app code
 var params = {};
+var webpMachine = true;
```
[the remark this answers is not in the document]
Jaifroid: I agree. I'll have a go.

www/js/lib/uiUtil.js
Jaifroid: OK, I agree we need to test in an old Firefox before merging.

www/js/lib/uiUtil.js
Jaifroid: I'll add it as a separate issue.

Jaifroid:
I've now downloaded Firefox Portable 64.0, and have confirmed that it requires the Polyfill and that Service Worker mode functions.

My tests are *generally* positive. This is what I have discovered:

* If the app is started in Service Worker mode, then most images load and display very briefly as broken images before being converted to a readable format;
* However, on some pages, images fail to be converted. This is random, and revisiting the page after loading another page can often cause the images to show;
* If the app starts in jQuery mode and the user subsequently switches to Service Worker mode, then the image on the home page shows as a broken image *the first time it is loaded after switching*. Thereafter the app works normally, except that on some pages, randomly, the images fail to be converted.

My hypothesis: there is a race condition affecting load order, and on some pages the polyfill is not loaded in time. This is merely a guess: I have done no investigation as yet.

@mossroy We could patch this behaviour to use the polyfill already loaded in the main `window` context instead of the polyfill in the `iframe.contentWindow`, for a consistent UX, but of course it would be better to diagnose the issue properly and get it fixed at source.

@kelson42 Have you heard reports of such behaviour (inconsistent loading of WebP images) in any other Kiwix apps that use the polyfill - I guess that's probably only the iOS app?

Jaifroid:
Just to add that the above tests were done with `wikipedia_en_medicine-app_maxi_2020-11.zim` downloaded from http://download.kiwix.org/zim/.hidden/custom_apps/ .

www/js/lib/uiUtil.js
Jaifroid: Done

kelson42:
@Jaifroid Not so far, but does not mean much...

Jaifroid:
Thanks, @kelson42 . I have done some tests by disabling our custom Polyfill, to be sure that there is no interference between the top-level version and the version loaded from the ZIM in the iframe. I can confirm that the issue persists: some pages randomly fail to convert images in SW mode, but visiting another page and returning to the page that failed often causes images to load correctly. This strongly suggests a timing issue to me.

kelson42:
@Jaifroid I'm really not an expert, but we should probably then have a look in detail to the ordering of events. Check as well if there is no incompatibility beetween img lazy loading and the polyfill.

Jaifroid:
@kelson42 I'll add a new issue for investigating and dealing with issues caused by the Polyfill in the ZIM, as it's a bit more complicated than can be dealt with in this PR, I think, and should be linked to an issue on mwoffliner (which I can create once I've done a bit more investigation).

tests/init.js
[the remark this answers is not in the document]
Jaifroid: Given the refactoring happening in #672, I think it would be best to add this test after the refactor has been accepted, so I'll add a new issue for this.

www/js/lib/uiUtil.js
Jaifroid: This has been tested, and an issue (possible race condition) unrelated to this PR has emerged, which will be dealt with in #673. I tested that the issue is unrelated to this PR by running Firefox 64 Potable against master (i.e. without this Polyfill) in SW mode, and randomly cycling though pages until the error occurs. It may take several page loads, but eventually images fail on a page.

PR #675, opened 2020-11-22
tests: Remove redundant xzdec.js include

Krinkle:
Follows-up 0c530101573cf95b, which removed this from www/index.html, but it was still present in the test suite.

This is now loaded automatically by RequireJS.

### DIFF
--- a/tests.html
+++ b/tests.html
@@ -31,7 +31,6 @@
              js files. This loads "main.js", which in turn can load other
              files, all handled by require.js:
              http://requirejs.org/docs/api.html#jsfiles -->
-        <script type="text/javascript" src="www/js/lib/xzdec.js"></script>
         <script type="text/javascript"
                 data-main="tests/init.js"
         src="www/js/lib/require.js"></script>


Jaifroid:
Thanks @Krinkle . I think tests have failed because of the external repo issue. @kelson42 can you give @Krinkle contributor access to Kiwix JS?

kelson42:
@Jaifroid Done

Krinkle:
Should I resubmit this as a local pull request?

Jaifroid:
@Krinkle Yes, I think it is the only way tests will run currently!

Jaifroid:
@Krinkle When you say "local", I should clarify that t's sufficient to do it from a clone of the repo on your machine now you are a contributor. Maybe I should have suggested you try pushing an inconsequential commit first to see if the tests run that way...

Krinkle:
OK. I can try that first :)

Jaifroid:
@Krinkle I'm not sure! I've tried restarting the build, but no luck. The way I've had to get builds running for external contributors is to push a clone of the branch to a test branch under my account (from my clone of the Repo), and *then* add a further commit (e.g. a comment that can be removed after). This usually gets the tests running, then I can set the contributors branch to the same commit as the test branch. Clunky workaround! (This is one reason we want to move to GitHub actions when we have time to sort that out - #590 ).

Krinkle:
> […]  Clunky workaround! (This is one reason we want to move to GitHub actions when we have time to sort that out - #590 ).

Sorry if this is obvious, but as I understand it the tests fail for PRs from non-maintainers because the SauceLabs credentials are meant to be secret. If Travis exposed them to PRs, then anyone could submit a PR that just prints the secret to the build output. GitHub Actions or any other CI service should behave the same way. If it doesn't and it does expose it, and if that's what we want, then we shouldn't mark them as secrets but rather set them as public env variables. I would expect the same to happen for GitHub Actions.

Jaifroid:
@Krinkle OK! I don't have any particular reason to favour one over the other, it was kelson who suggested that other repos had solved this issue with GitHub Actions. If we can do the same with Travis, that would be a relief (saves a lot of work).

Krinkle:
It is of course possible that there are (or were) limitations with Travis that affect Kiwix. I only know that the specific issue of repo secrets and SauceLabs, as far as I know, is an intententional limitation that Travis and GitHub Actions both follow the same way 🙂 

I'll ask for more details over at https://github.com/kiwix/kiwix-js/issues/590 to see if I can help with anything.

Issue #674, opened 2020-11-22
Add QUnit test for conversion of WebP images


Jaifroid:
Following on from https://github.com/kiwix/kiwix-js/pull/670#discussion_r527896572, once #672 is complete, we should add a QUnit test for WebP image reading:

> It would be good to add a QUnit test on feedNodeWithBlob function : it could create a DOM image element, call this function with a webp content (like the sample image with 2 pixels height), and the check its height afterwards.
This might probably need to refactor this function a bit : add a callback parameter or make it return a Promise, so that the unit test can know when the image has been decoded.

kelson42:
The value of auch a unit test decreases rapidly with time because les and less people uses a Browser without native webp support.

mossroy:
This should be superseded by #748 that should have a UI test that checks that an image is displayed (and check its dimensions).
If we keep running this scenario on browsers that do not support webp (IE11, Firefox 45), it will check that conversion of webp images work.
I agree with you that it would be enough.

Issue #671, opened 2020-11-21
Extract images sequentially in jQuery mode


Jaifroid:
Following on from #670 , it would be possible to transfer the sequential WebP image extraction logic in `uiUtil.js` to the whole `loadImagesJQuery()` function in `app.js`. This would mean that all images would be extracted one by one starting at the top of the page, hence images above the fold would load first. This would improve the UX for jQuery users (who also often have slow browsers). For users of SW mode, some browsers are already including lazy loading, and/or load from the top of the DOM.

Issue #505, opened 2019-05-28
Enable progressive extraction of images


Jaifroid:
If #504 is merged, it would be very easy to add (at least for jQuery mode, but potentially also for SW mode) progressive extraction of images based on processing only images visible in the viewport. The principle is that all images in viewport are extracted on page load (we already have a function for testing whether an element is in the viewport). Then, the process is run again after the user scrolls and the scroll has ended (on "scrollstop").

PR #677, opened 2020-11-23
Extract images sequentially in jQuery mode

Jaifroid:
This implements #671 and #505 in jQuery mode. I have removed the WebP image queue from `uiUtil.js` and have transferred the same self-invoking recursive function to the `loadImagesJQuery()` function in `app.js`.

This greatly improves UX on long articles with lots of images, since it ensures that images are loaded from the top of the page, sequentially. The images extracted first are those in the visible area ("above the fold").

Additionally, this PR partially implements #426 for jQuery mode image assets, cancelling extraciton of images if a user navigates away from the page. This is a side-effect of a need to avoid an error if the `images` array is not redefined cleanly. Cancellation does not occur at a very low level, but is still relatively effective (tested with a console log, removed in the latest commit).

I have also implemented a callback (necessary for this PR) which will help in implementing #674 once #672 is complete.

### DIFF
--- a/www/js/app.js
+++ b/www/js/app.js
@@ -1379,11 +1379,13 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
         function loadImagesJQuery() {
             // Make an array from the images that need to be processed
             var images = Array.prototype.slice.call(iframeArticleContent.contentDocument.querySelectorAll('img[data-kiwixurl]'));
+            // This ensures cancellation of image extraction if the user navigates away from the page before extraction has finished
+            images.owner = expectedArticleURLToBeDisplayed;
             // DEV: This self-invoking function is recursive, calling itself only when an image has been fully processed into a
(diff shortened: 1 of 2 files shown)


Jaifroid:
I have now tested this on IE11, Firefox OS simulator, Edge Legacy and Chromium. All working as expected.

A good page to test this in the English WikiMed ZIM is "COVID-19 pandemic by country and territory" (using the WebP ZIM). This page has hundreds of little flags on it, all coded in WebP, and you can clearly see the sequential nature of extraction if you scroll down to where the flags are being decoded. Note that some of the boxes with tables are scrollable, so if the decoding appears to stop for a period, it is in fact decoding images that are below the visible area in one of the scroll boxes.

I'm marking this as ready for review.

mossroy (approved):
It appears to work fine on my Firefox OS device.
It makes a real UX difference on big articles in JQuery mode : thanks for this improvement!

Issue #678, opened 2020-11-24
"Open in New tab" does not work with local links (in Jquery)


kelson42:
Nothing is loaded, not even the Kiwix extension.

I'm not sure if this is possible at all.

But if not possible we should probably not open a new tab for local links

Jaifroid:
@kelson42 We don't have any logic to decode an article and all assets in a shadow DOM and then transfer the HTML to a new tab. I implemented a "breakout link" in KJSW which transfers a copy of the currently loaded article (including images loaded as data URIs) into a new tab (with no Kiwix chrome) for bookmarking an article or side-by-side comparison, or for better printing. But the extracted HTML and images are not navegable, it is a static copy of the article DOM.

Going further than that would require the ability to track and communicate between tabs, and quite a lot of extra logic to manage tabs.

Are you suggesting we try to disable the built-in browser behaviours for ctrl-click or right-click "open in new tab"?

kelson42:
I primarly report a behaviour which is not really nice for user (actually this is a report from a user).

I suspect that, if we can open a ZIM file without going through the "open file" dialog (which is #656 if I'm right), then we should be in position to open a new Kiwix-JS an open the right ZIM file at the right place. There is as well an obvious "problem" around the URLs to manage.

If that would really not work we might think about catching the click event via JQuery or SW. But again, would be sad to have to follow that path.

Jaifroid:
OK, I'll investigate.

Jaifroid:
@kelson42 I've investigated a bit, and it is technically possible to open a new tab using something like:

```
var newWindow = window.open("article.html", "KiwixTab");
newWindow.onload = function {
    var newWindowDoc = newWindow.document.documentElement;
    newWindowDoc.innerHTML= htmlArticle;
    parseAnchorsJQuery();
    loadImagesJQuery();
    loadCSSJQuery();
    ...
}
```

The newWindow can be manipulated from the main window in exactly the same way as we manipulate the iframe. It is possible to navigate in `newWindow` (by clicking on links), and main window will supply the new document, images and CSS, or from main window by using Random button.

I pushed a demo branch of this working pretty well in Chromium and Firefox (jQuery mode) (for Firefox, the user must explicitly allow the "popup" tab). It's very hacky (it closes tab and re-opens it for each new article, just as a kludge to make it work). It does not (currently) work at all in IE11, and works poorly in Edge Legacy (styles and images are not loaded).

[...]

PR #680, opened 2020-11-29
Send article to new tab in jQuery mode

Jaifroid:
This is currently a rough draft of a possible way to manage tabbed browsing in Kiwix JS in jQuery mode, addressing #678. This PR is not needed for Service Worker mode because the browser handles tabbed browsing natively.

With this PR, a user can launch a new tab or a new window (depending on OS/browser) in one of two ways:

1. Ctrl-click or command-click on a hyperlink (if user has a keyboard and mouse, or keyboard and tap);
2. Long press on a hyperlink for more than 600ms (only on browsers that support the [`touchstart` event](https://caniuse.com/?search=touchstart) - i.e. FFOS and Windows Mobile are not supported, but I don't think anyone really wants to open a new window on these mobile apps).

The second option is an alternative to patching the right-click context menu. This menu has useful things a user may wish to do, so we should not override it. As a result, this PR does *not* support opening a new tab from the context menu (it is natively supported in SW mode). There is no reliable way to patch the "open in new tab" entry in the context menu without completely replacing the menu. 

The tabs opened by this PR are independently browsable, i.e. the user can click on ZIM hyperlinks and they will open in the same window, and also ctrl-click on links to open yet another window/tab).

The new tabs do not have the Kiwix JS UI (no search bar, no random button or navigation buttons). I see this as an advantage for the user who would like to read a page fullscreen or print a page using the browser's native print function.

There is currently a small bug with setting the title of each tab on some browsers: it will intermittently show "Localhost" instead of the page title - this is probably a minor race condition with setting the title at the right moment.

### DIFF
--- a/browser-tests/nightwatch_runner.js
+++ b/browser-tests/nightwatch_runner.js
@@ -107,6 +107,8 @@ module.exports = {
             .frame('articleContent')
             // Check the text in the article "Ray Charles"
             .useXpath()
+            // Let's pause for a second
+            .pause(1000)
             .waitForElementPresent("//div[@id='content']/div[@id='mw-content-text']/h2[@id='mweQ']", 40000)
             .assert.containsText("//div[@id='content']/div[@id='mw-content-text']/h2[@id='mweQ']", 'Life and career')
             // Wait for a particular image to be visible and check its size
--- a/browser-tests/nightwatch_runner.js
+++ b/browser-tests/nightwatch_runner.js
@@ -74,9 +74,9 @@ module.exports = {
         browser.click("//div[@id='articleList']/a[text()='Ray Charles']")
             .frame('articleContent')
             // Check the text in the article "Ray Charles"
-            .useXpath()
-            .waitForElementPresent("//div[@id='content']/div[@id='mw-content-text']/h2[@id='mweQ']", 40000)
-            .assert.containsText("//div[@id='content']/div[@id='mw-content-text']/h2[@id='mweQ']", 'Life and career')
+            // .useXpath()
+            // .waitForElementPresent("//div[@id='content']/div[@id='mw-content-text']/h2[@id='mweQ']", 40000)
+            // .assert.containsText("//div[@id='content']/div[@id='mw-content-text']/h2[@id='mweQ']", 'Life and career')
             // Wait for a particular image to be visible and check its size
             .useXpath()
             .waitForElementVisible("//td[@id='mwCA']/p/span/img", 20000)
--- a/browser-tests/nightwatch_runner.js
+++ b/browser-tests/nightwatch_runner.js
@@ -74,9 +74,9 @@ module.exports = {
         browser.click("//div[@id='articleList']/a[text()='Ray Charles']")
             .frame('articleContent')
             // Check the text in the article "Ray Charles"
-            // .useXpath()
-            // .waitForElementPresent("//div[@id='content']/div[@id='mw-content-text']/h2[@id='mweQ']", 40000)
-            // .assert.containsText("//div[@id='content']/div[@id='mw-content-text']/h2[@id='mweQ']", 'Life and career')
+            .useXpath()
+            .waitForElementPresent("//div[@id='content']/div[@id='mw-content-text']/h2[@id='mweQ']", 20000)
+            .assert.containsText("//div[@id='content']/div[@id='mw-content-text']/h2[@id='mweQ']", 'Life and career')
             // Wait for a particular image to be visible and check its size
             .useXpath()
             .waitForElementVisible("//td[@id='mwCA']/p/span/img", 20000)
--- a/nightwatch.js
+++ b/nightwatch.js
@@ -43,7 +43,13 @@ module.exports = {
         "enabled" : false
       },
       "globals": {
-        "waitForConditionTimeout": 600
+        "waitForConditionTimeout": 10000
+      },
+      // Configure when a request to the Selenium server should time out and optionally define the number of retries for a timed-out request
+      // See https://github.com/nightwatchjs/nightwatch/issues/1936
+      "request_timeout_options": {
+        "timeout": 100000,
+        "retry_attempts": 3
       }
     },
     "firefox45" : {
@@ -60,7 +66,8 @@ module.exports = {
         "browserName": "firefox",
         "javascriptEnabled": true,
         "acceptSslCerts": true,
-        "build": build
+        "build": build,
+        "extendedDebugging": true
       }
     },
     "chrome58" : {
--- a/www/article.html
+++ b/www/article.html
@@ -2,7 +2,7 @@
 <html>
     <head>
         <meta charset="utf-8">
-        <title>Placeholder for injecting an article into the iframe</title>
+        <meta name="description" content="Placeholder for injecting an article into the iframe or window">
     </head>
     <body></body>
 </html>
--- a/www/js/app.js
+++ b/www/js/app.js
@@ -85,12 +85,18 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
     params['useHomeKeyToFocusSearchBar'] = settingsStore.getItem('useHomeKeyToFocusSearchBar') === 'true';
     document.getElementById('useHomeKeyToFocusSearchBarCheck').checked = params.useHomeKeyToFocusSearchBar;
     switchHomeKeyToFocusSearchBar();
+    // A setting that determines whether right-click/long-press of a ZIM link opens a new window/tab
+    params['rightClickOpensTab'] = settingsStore.getItem('rightClickOpensTab') === 'true';
+    document.getElementById('rightClickOpensTabCheck').checked = params.rightClickOpensTab;
+
     // An object to hold the current search and its state (allows cancellation of search across modules)
     appstate['search'] = {
         'prefix': '', // A field to hold the original search string
         'status': '',  // The status of the search: ''|'init'|'interim'|'cancelled'|'complete'
         'type': ''    // The type of the search: 'basic'|'full' (set automatically in search algorithm)
     };
+    // The target for article loads (this should always be 'iframe' initially, and will only be changed as a result of user action)
+    appstate['target'] = 'iframe';
     
     // Define globalDropZone (universal drop area) and configDropZone (highlighting area on Config page)
     var globalDropZone = document.getElementById('search-article');
@@ -132,6 +138,8 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
         $("#welcomeText").hide();
         $('.alert').hide();
         $("#searchingArticles").show();
+        // Ensure selected search item is displayed in the iframe, not a new window or tab
+        appstate.target = 'iframe';
         pushBrowserHistoryState(null, prefix);
         // Initiate the search
         searchDirEntriesFromPrefix(prefix);
@@ -224,6 +232,8 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
         }
     });
     $("#btnRandomArticle").on("click", function() {
+        // In jQuery mode, only load random content in iframe (not tab or window)
+        appstate.target = 'iframe';
         $('#prefix').val("");
         goToRandomArticle();
         $("#welcomeText").hide();
@@ -254,6 +264,8 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
     });
     // Top menu :
     $('#btnHome').on('click', function() {
+        // In jQuery mode, only load landing page in iframe (not tab or window)
+        appstate.target = 'iframe';
         // Highlight the selected section in the navbar
         $('#liHomeNav').attr("class","active");
         $('#liConfigureNav').attr("class","");
@@ -354,6 +366,10 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
         params.useHomeKeyToFocusSearchBar = this.checked ? true : false;
         settingsStore.setItem('useHomeKeyToFocusSearchBar', params.useHomeKeyToFocusSearchBar, Infinity);
         switchHomeKeyToFocusSearchBar();
+    });
+    document.getElementById('rightClickOpensTabCheck').addEventListener('change', function (e) {
+        params.rightClickOpensTab = e.target.checked;
+        settingsStore.setItem('rightClickOpensTab', params.rightClickOpensTab, Infinity);
     });
     document.getElementById('appThemeSelect').addEventListener('change', function (e) {
         params.appTheme = e.target.value;
@@ -745,11 +761,14 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
 
 
     // Display the article when the user goes back in the browser history
-    window.onpopstate = function(event) {
+    var historyPop = function(event) {
         if (event.state) {
             var title = event.state.title;
             var titleSearch = event.state.titleSearch;
-            
+            appstate.target = event.target.kiwixType;
+            // Select the correct window to which to write the popped history in case the user
+            // siwtches to a tab and navigates history without first clicking on a link
+            if (appstate.target === 'window') articleContainer = event.target;
             $('#prefix').val("");
             $("#welcomeText").hide();
             $("#searchingArticles").hide();
@@ -771,6 +790,8 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
             }
         }
     };
+
+    window.onpopstate = historyPop;
     
     /**
      * Populate the drop-down list of archives with the given list
@@ -855,6 +876,9 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
             resetCssCache();
             selectedArchive = zimArchiveLoader.loadArchiveFromDeviceStorage(selectedStorage, archiveDirectory, function () {
                 settingsStore.setItem("lastSelectedArchive", archiveDirectory, Infinity);
+                // Ensure that the new ZIM output is initially sent to the iframe (e.g. if the last article was loaded in a window)
+                // (this only affects jQuery mode)
+                appstate.target = 'iframe';
                 // The archive is set : go back to home page to start searching
                 $("#btnHome").click();
             });
@@ -942,6 +966,9 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
         }
         resetCssCache();
         selectedArchive = zimArchiveLoader.loadArchiveFromFiles(files, function () {
+            // Ensure that the new ZIM output is initially sent to the iframe (e.g. if the last article was loaded in a window)
+            // (this only affects jQuery mode)
+            appstate.target = 'iframe';
             // The archive is set : go back to home page to start searching
             $("#btnHome").click();
             document.getElementById('downloadInstruction').style.display = 'none';
@@ -1214,7 +1241,7 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
                 // Line below was inserted to prevent the spinner being hidden, possibly by an async function, when pressing the Random button in quick succession
                 // TODO: Investigate whether it is really an async issue or whether there is a rogue .hide() statement in the chain
                 $("#searchingArticles").show();
-                selectedArchive.readUtf8File(dirEntry, displayArticleContentInIframe);
+                selectedArchive.readUtf8File(dirEntry, displayArticleContentInContainer);
             }
         }
     }
@@ -1290,15 +1317,17 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
     // the link. This is currently the case for epub and pdf files in Project Gutenberg ZIMs -- add any further types you need
     // to support to this regex. The "zip" has been added here as an example of how to support further filetypes
     var regexpDownloadLinks = /^.*?\.epub($|\?)|^.*?\.pdf($|\?)|^.*?\.zip($|\?)/i;
+    // Placeholders for the article container and the documentElement of the article
+    var articleContainer, articleDocument;
     
     /**
      * Display the the given HTML article in the web page,
      * and convert links to javascript calls
      * NB : in some error cases, the given title can be null, and the htmlArticle contains the error message
-     * @param {DirEntry} dirEntry
-     * @param {String} htmlArticle
+     * @param {DirEntry} dirEntry The Directory Entry of the article
+     * @param {String} htmlArticle The decoded HTML of the article
      */
-    function displayArticleContentInIframe(dirEntry, htmlArticle) {
+    function displayArticleContentInContainer(dirEntry, htmlArticle) {
         if(! isDirEntryExpectedToBeDisplayed(dirEntry)){
             return;
         }		
@@ -1325,23 +1354,18 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
         var htmlCSS = htmlArticle.match(/<html[^>]*class\s*=\s*["']\s*([^"']+)/i);
         htmlCSS = htmlCSS ? htmlCSS[1] : '';
         
-        // Tell jQuery we're removing the iframe document: clears jQuery cache and prevents memory leaks [kiwix-js #361]
-        $('#articleContent').contents().remove();
-
         // Hide any alert box that was activated in uiUtil.displayFileDownloadAlert function
         $('#downloadAlert').hide();
 
-        var iframeArticleContent = document.getElementById('articleContent');
-        
-        iframeArticleContent.onload = function() {
-            iframeArticleContent.onload = function(){};
+        var windowLoaded = function() {
+            articleDocument = articleContainer.document.documentElement;
+            
             $("#articleList").empty();
             $('#articleListHeaderMessage').empty();
             $('#articleListWithHeader').hide();
             $("#prefix").val("");
             
-            var iframeContentDocument = iframeArticleContent.contentDocument;
-            if (!iframeContentDocument && window.location.protocol === 'file:') {
+            if (!articleDocument && window.location.protocol === 'file:') {
                 alert("You seem to be opening kiwix-js with the file:// protocol, which is blocked by your browser for security reasons."
                         + "\nThe easiest way to run it is to download and run it as a browser extension (from the vendor store)."
                         + "\nElse you can open it through a web server : either through a local one (http://localhost/...) or through a remote one (but you need SSL : https://webserver/...)"
@@ -1349,25 +1373,22 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
                 return;
             }
             
-            // Inject the new article's HTML into the iframe
-            var articleContent = iframeContentDocument.documentElement;
-            articleContent.innerHTML = htmlArticle;
-            
-            var docBody = articleContent.getElementsByTagName('body');
-            docBody = docBody ? docBody[0] : null;
-            if (docBody) {
-                // Add any missing classes stripped from the <html> tag
-                if (htmlCSS) docBody.classList.add(htmlCSS);
-                // Deflect drag-and-drop of ZIM file on the iframe to Config
-                docBody.addEventListener('dragover', handleIframeDragover);
-                docBody.addEventListener('drop', handleIframeDrop);
+            if (articleContainer.kiwixType === 'iframe') {
+                var docBody = articleDocument.querySelector('body');
+                if (docBody) {
+                    // Add any missing classes stripped from the <html> tag
+                    if (htmlCSS) docBody.classList.add(htmlCSS);
+                    // Deflect drag-and-drop of ZIM file on the iframe to Config
+                    docBody.addEventListener('dragover', handleIframeDragover);
+                    docBody.addEventListener('drop', handleIframeDrop);
+                }
+                // Set the requested appTheme
+                uiUtil.applyAppTheme(params.appTheme);
             }
-
-            // Set the requested appTheme
-            uiUtil.applyAppTheme(params.appTheme);
             // Allow back/forward in browser history
             pushBrowserHistoryState(dirEntry.namespace + "/" + dirEntry.url);
 
+            loadCSSJQuery();
             parseAnchorsJQuery();
             loadImagesJQuery();
             // JavaScript is currently disabled, so we need to make the browser interpret noscript tags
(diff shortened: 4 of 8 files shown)


Jaifroid:
Here is a demo of what the PR does (jQuery mode in Firefox, using Ctrl-click):

tabsGalore

Jaifroid:
This is nearly ready, with the following mostly minor issues to resolve:

* Tests are failing consistently, even though QUnit tests are passing locally (I have yet to debug this);
* History is stored manually by the app in jQuery mode, and so to go back and forward in a tab, it currently has to be done from the main app tab (where the controls are): maybe allowing native history in tabs could work, as the pages are not in an iframe;
* Tabbed browsing in IE11 works fine with non-WebP ZIMs, but with WebP, the images in a tab will not load unless the user switches back to the main (app) window and then back to the tab (the switching can be done a long time page has "finished" loading, and the images will then all show at once).

The last issue is curious: it is as if the Emscripten engine for WebP refuses to run "in the background" (only in IE11), and has to be in the foreground to complete a Promise or a timeout. I cannot find an obvious cause, and I cannot find a way to give focus to the main tab programmatically (blur and focus methods on tabs were blocked a long time ago due to abuse by spammers). Possibly the way image queueing is done inside a closure has this side-effect? I don't think it's a big deal if this feature remains a bit buggy in IE11.

Jaifroid:
Tested and working as a Chromium Add-in and a Firefox Extension.
Firefox erroneously claims it has blocked popup windows (they are not actually blocked since they are initiated by a user action, which is allowed, and I didn't notice any blocked functionality). Dismissing the message about blocking (which only appears in the main app tab) makes it go away.

Chromium, Edge Legacy and IE11 do not flag the new tabs as popups.

Jaifroid:
Hmm. I wonder why Travis tests are not running any more (only codefactor).
Last step for this PR is to be sure it is passing tests, but with the tests having gone down that's a bit tricky.

Jaifroid:
I@kelson42 @mossroy, I see the error below in Travis about our having run out of credit. Do you know what can be done about this?

image

kelson42:
@Jaifroid No, I don't have changeg anything. Actually we have moved two additional repos from Travis to Actions recently.

Jaifroid:
@kelson42 The message says "Builds have been temporarily disabled for public repositories due to a negative credit balance. Please go to the Plan page to replenish your credit balance or alter your Consume paid credits for OSS setting." When I try to go to the Plan Page - https://travis-ci.com/organizations/kiwix/plan - I just get "There was an error, please try again". Is this something you can check, or @Popolechien (becuase it seems to be related to the Kiwix Organiziation Plan)?

mossroy:
I discover that Travis CI announced last month that it stops providing free unlimited CI for OSS : https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing
We're not the only affected OSS repo : https://www.jeffgeerling.com/blog/2020/travis-cis-new-pricing-plan-threw-wrench-my-open-source-works

Shit, I would have liked to be informed earlier, and have more time to think about our options, which seem to be : 

- switch to a paid option : I doubt we'll do that
- beg Travis to give us some OSS credits : it would be a welcome temporary measure, to give us some time
- switch to another CI/CD provider, like GitHub Actions (see #590) : it's probably what we'll need to do in the mid-term

And this will make #672 more complicated to merge (as we won't be able to test it in the short-term...)

Jaifroid:
Hmm, so they're asking us either to request more credits by saying "please", or to pay USD $15 for 25,000 credits. @kelson42 what would you advise? I don't know how long 25,000 credits would last, but might be a short-term inexpensive solution while we work on migrating to github actions (#590). Of course I completely understand that there may be an issue of principle at stake rather than just the amount of money.

kelson42:
@Jaifroid I would advise to move to Github actions.

Jaifroid:
I guess this PR needs to wait until we have fixed Unit Tests. For my part, pending tests, it is ready.

www/js/app.js
```diff
@@ -1350,35 +1392,94 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
...
+            // However, we cannot rely on the download attribute having been set, so we also need to test for known download file types
+            var isDownloadableLink = a.hasAttribute('download') || regexpDownloadLinks.test(href);
+            if (isDownloadableLink) {
+                downloadAttrValue = a.getAttribute('download');
+                // Normalize the value to a true Boolean or a filename string or true if there is no download attribute
+                downloadAttrValue = /^(download|true|\s*)$/i.test(downloadAttrValue) || downloadAttrValue || true;
+                contentType = a.getAttribute('type');
+            }
+            // DEV: We need to use the '#' location trick here for cross-browser compatibility with opening a new tab/window
+            a.setAttribute('href', '#');// Store the current values, as they may be changed if user switches to another tab before returning to this one
```
Jaifroid, line 1425: Move comment to next line.

www/js/app.js
```diff
@@ -1350,35 +1392,94 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
...
+                appstate.target = kiwixTarget;
+                articleContainer = thisWindow;
+                // This detects Ctrl-click, Command-click, the long-press event, and middle-click
+                if (e.ctrlKey || e.metaKey || touched || e.which === 2 || e.button === 4) {
+                    // We open the new window immediately so that it is a direct result of user action (click)
+                    // and we'll populate it later - this avoids popup blockers
+                    articleContainer = window.open('article.html', '_blank');
+                    appstate.target = 'window';
+                    articleContainer.kiwixType = appstate.target;
+                } else if (a.tagName !== 'A') {
```
Jaifroid, line 1471: Add 'AREA' tagname here too.

kelson42:
I guess we can carry on with this PR now? Maybe rebase and ask @mossroy to review?

Jaifroid:
Yes, thanks @kelson42 . I need to rebase and get the new testing system to run on this PR (that was the previous blocker).

Jaifroid:
OK tests have run, and it's the same as under Travis: tests are failing on Firefox (only). I'm guessing there is a race condition due to Firefox not running `document.write()` sunchronously. The spec says it should be synchronous, but I'm pretty sure we've encountered this issue before when using `document.write()` with Firefox. Unfortunately, neither Edge Legacy nor IE11 expose the onload event for secondary tabs, so I have to insert a clunky workaround.

Jaifroid:
I've restored asynchronous loading for browsers that support it. However, the Firefox error persists. I'm a bit stumped by this one. In the logs we have this identified as the issue:

image

And element `<hd id="mweQ" ...>` corresponds to the heading 'Live and career' (of Ray Charles). Checking the video on Sauce Labs for the problematic Firefox 84, I can clearly see that the corresponding text is shown as soon as the article is displayed:

image

The only thing I can think of is that there are two elements marked with `id="mweQ"` (one of them is a td element, the other is this h2 element). This clearly breaks the spec, since no two elements should have the same ID, but that would not be a reason for this particular PR to be failing while master passes.

I'm trying with the local Karma versions of the tests.

mossroy:
I don't think it comes from the ZIM content. Because we look for the exact same element earlier in the test scenario, and it's found by all browsers.
I notice that the following error message appears in the log : 

    Error ECONNRESET: socket hang up

Could it be related to https://github.com/nightwatchjs/nightwatch/issues/1936 ?
Maybe you could test to downgrade to version "=0.9.20" of nightwatch instead of "^1.5.1" (in package.json), to see if it helps

Krinkle:
The issue is unrelated to this patch indeed. It happens when the concurrency limit with SauceLabs is reached, which was happening because the same job was running twice at the same time (once as `push` and once as `pull_request`). The secrets worked differently at Travis where this issue did not happen. We didn't catch it before because the PR for adding these tests in CI ran more slowly with the jobs not overlapping (lucky).

I have fixed it at https://github.com/kiwix/kiwix-js/pull/699. When you rebase this branch, the issue should go away automatically :)

www/js/app.js
```diff
@@ -1261,65 +1289,88 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
         var htmlCSS = htmlArticle.match(/<html[^>]*class\s*=\s*["']\s*([^"']+)/i);
         htmlCSS = htmlCSS ? htmlCSS[1] : '';
         
-        // Tell jQuery we're removing the iframe document: clears jQuery cache and prevents memory leaks [kiwix-js #361]
-        $('#articleContent').contents().remove();
-
```
Jaifroid, line 1266: Above code has been moved to line 1340.

Jaifroid:
Hmm. I've rebased, but still getting some socket hangups, even after re-running tests. Some of them seem to occur in random places, but the error with `<hd id="mweQ" ...>` mentioned above seems to crop up in every run. Local tests are running fine. Very puzzling.

Jaifroid:
I could try the nightwatch suggestion (when I get a chance).

Jaifroid:
I finally got around to downgrading nightwatch to 0.9.20. It fixes the ECONNRESET errors, or at least the reporting of those errors (i.e. they don't appear in the log), and it fixes any errors in Chrome or Edge that were caused by that socket reset. However, with Firefox 85 the system still reports that it cannot find the header `#mweQ`, even though it's clearly visible on-screen (screenshots). No more tests run after this error. I have tested manually in Firefox 85 imitating the exact same steps, and cannot see any problem. I suspect there is a socket disconnect that is not reported in this old version of nightwatch, becuase the behaviour is exactly the same: the instance on sauce labs never receives further instructions after that test, waits for another 3 minutes+, then it times out and deletes the environment. On the nightwatch side, it waits for around 40 seconds for the test result and then reports "not found" for that element. 

My only current clue is this warning when I ran `npm install` to update package-lock.json: 

`npm WARN deprecated socks@1.1.10: If using 2.x branch, please upgrade to at least 2.1.6 to avoid a serious bug with socket data flow and an import issue introduced in 2.1.0
`

However, this is not a package we specifiy in package.json, it appears to be a dependency of another package.

image

image

Jaifroid:
I've reverted the downgrade, because it merely suppresses the error *display* in the log, but the underlying socket hangup still occurs only in Firefox 85.

Jaifroid:
Rebased on master.

Jaifroid:
Rebased on master.

Jaifroid:
I solved the problem with the tests using some "forensic" patience and observation. It turned out to be our old friend: the fact that Firefox asynchronously writes html to the `innerHtml` property of an element. It's anomalous in this compared to the other browsers, and the loading method I'd used for iframes and windows/tabs was causing the test failure, even though it could not be reproduced in real-world conditions. The test report was misleading in its description of the failure, an why this caused a socket disconnect on using `history.back()`.

Jaifroid:
OK, based on real-world testing with Kiwix JS Windows, where this is in production, and numerous tests here, I think this PR is ready for review.

I realize it looks big and there are some architectural changes. However, many of the changes are about removing the hard-coded dependency on the iframe. The app is now highly flexible about where it can place content read from the backend. We can place it in the iframe, in a new tab, or in a new window, in most environments that support these options (this excludes most mobile environments, though the code does work on Android, albeit slowly). We can even target a `<div>`, which opens up interesting future possibilities such as Wikipedia's standard "Page Preview" feature (#719).

Most of this PR affects jQuery mode only. Service Worker mode supports opening tabs and windows natively. This does introduce some differences. Currently, the "theme" (dark/light) is only applied to a jQuery window or tab, because I would need to put hooks into the DOM in SW mode to achieve this. This is because the Service Worker cannot tell us where it is getting its requests from. There are two ways to get a handle on the opened window: intercept clicks on links (possibly on the iframe itself, rather than each link), or ask the Service Worker to inject a small script into the frame/window (which means adding the script to the raw html before sending it to the window, which will send the handle back to the parent. There may be other ways of scanning for child windows, but I haven't researched that much (it's probably blocked for security reasons). Maybe it's not a priority.

I realize some of this PR looks big. If you don't want if for Kiwix JS, @mossroy, don't worry. I've made it a feature of Kiwix JS Windows, so no work has been lost developing it. I do think it's a nice feature, though, and it was requested by a user of the Firefox Extension (who cannot use SW mode).

Jaifroid:
Tested and working in:

* Firefox extension
* Chromium (Edge) extension
* IE11 - but the user must set the target as "window", not as "tab" ("tab" fails to open a tab, but opens a popup window instead which is not scrollable or resizeable, whereas "window" loads a scrollable and resizable popup wiindow)
* Tested in Edge HTML via Kiwix JS Windows (works only in Service Worker mode).

It does not work in Firefox OS Simulator: the system attempts to open a new browser window, but this window cannot be controlled by the app, so never receives any content. It is possible to close this window and go back to the app and continue browsing. This is expected behaviour on mobile (though it does work, sluggishly, on Android with Chromium).

www/js/app.js
```diff
@@ -1325,83 +1399,117 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
...
         // Hide any alert box that was activated in uiUtil.displayFileDownloadAlert function
         $('#downloadAlert').hide();
 
-        var iframeArticleContent = document.getElementById('articleContent');
-        
-        iframeArticleContent.onload = function() {
-            iframeArticleContent.onload = function(){};
+        // Code below will run after we have written the new article to the articleContainer
```
Jaifroid, line 1405: Comment is incorrect now. Should read:

`// Code below will run after we have loaded the dummy article into the articleContainer`

www/js/app.js
```diff
@@ -1325,83 +1399,117 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
...
-            
-            var iframeContentDocument = iframeArticleContent.contentDocument;
-            if (!iframeContentDocument && window.location.protocol === 'file:') {
-                alert("You seem to be opening kiwix-js with the file:// protocol, which is blocked by your browser for security reasons."
-                        + "\nThe easiest way to run it is to download and run it as a browser extension (from the vendor store)."
-                        + "\nElse you can open it through a web server : either through a local one (http://localhost/...) or through a remote one (but you need SSL : https://webserver/...)"
-                        + "\nAnother option is to force your browser to accept that (but you'll open a security breach) : on Chrome, you can start it with --allow-file-access-from-files command-line argument; on Firefox, you can set privacy.file_unique_origin to false in about:config");
-                return;
+            if (appstate.target === 'iframe' && !articleContainer.contentDocument && window.location.protocol === 'file:') {
+                if (!appstate.fileAccessWarning) {
```
Jaifroid, line 1412: Because this warning about file:// access is longer now, with more user-friendly explanation, I have formatted it better and have used a simple DOM-based (Bootstrap-styled), non-blocking `systemAlert()` function. Arguably we should make all system alerts use this function.

Also, rather than bombing out at this point, I found that in Chrome and Firefox allow us to inject the HTML into a new window and show it fully styed. Chrome/Edge launches this window immediately and shows it, but Firefox blocks it by default as a popup. After the user allows the popup, they can try again and it will show just like in Chromium. I thought this was fun (and potentially useful)!

www/js/app.js
```diff
@@ -1325,83 +1399,117 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
...
+                    uiUtil.systemAlert("<p>You seem to be opening kiwix-js with the file:// protocol, which blocks access to the app's iframe. "
+                    + "We have tried to open your article in a separate window. You may be able to use it with limited functionality.</p>"
```
Jaifroid, line 1414: Add "You may need to allow a blocked popup and try again to see your content."

www/js/app.js
```diff
@@ -1325,83 +1399,117 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
...
+            // (to avoid popup blockers). Firefox loads windows asynchronously, so we need to wait for onclick load to be fully
+            // cleared, or else Firefox overwrites the window immediately after we load the html content into it.
+            setTimeout(windowLoaded, 250);
+        }
+        
         function parseAnchorsJQuery() {
-            var currentProtocol = location.protocol;
-            var currentHost = location.host;
+            var currentProtocol = articleWindow.location.protocol;
+            currentProtocol === 'about:' ? currentProtocol = ':' : currentProtocol;
```
Jaifroid, line 1505: This is needed to parse anchors in a window opened as `about:blank`.

www/js/app.js
```diff
@@ -1325,83 +1399,117 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
...
+                    + "We have tried to open your article in a separate window. You may be able to use it with limited functionality.</p>"
+                    + "<p>The easiest way to run this app fully is to download and run it as a browser extension (from the vendor store). "
+                    + "Alternatively, you can open it through a web server: either use a local one (http://localhost/...) "
+                    + "or a remote one. For example, you can try your ZIM out right now with our online version of the app: "
+                    + "<a href='https://kiwix.github.io/kiwix-js/'>https://kiwix.github.io/kiwix-js/</a>.</p>"
+                    + "<p>Another option is to force your browser to accept file access (a potential security breach): "
+                    + "on Chrome, you can start it with the <code>--allow-file-access-from-files</code> command-line argument; on Firefox, "
+                    + "you can set <code>privacy.file_unique_origin</code> to <code>false</code> in about:config.</p>");
+                }
+                articleContainer = window.open('', dirEntry.title, 'toolbar=0,location=0,menubar=0,width=800,height=600,resizable=1,scrollbars=1');
```
Jaifroid, line 1423: This opens as about:blank, which is not considered cross-origin, and so is not blocked.

www/js/app.js
```diff
@@ -1422,35 +1530,109 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
...
-                        e.preventDefault();
-                    });
+                    addListenersToLink(anchor, href);
                 }
             });
+            // Add event listeners to the main document so user can open current document in new tab or window
+            addListenersToLink(articleDocument, encodeURIComponent(dirEntry.url.replace(/[^/]+\//g, '')));
         }
-        
+
```
Jaifroid, line 1539: Move funciton below to `uiUtil.js` to de-clutter `app.js`.

www/js/lib/uiUtil.js
```diff
@@ -375,18 +394,18 @@ define(rqDef, function() {
         // Show any specific help for selected contentTheme
         var help = document.getElementById(theme + '-help');
         if (help) help.style.display = 'block';
-        
+        var frame = container.classList ? container : container.document.documentElement;
```
Jaifroid, line 397: If articleContainer is an iframe, it will already have a `classList`. If it is a Window, we have to apply theme classes to the documentElement (i.e. the `<html>` node).

browser-tests/nightwatch_runner.js
```diff
@@ -75,7 +75,7 @@ module.exports = {
             .frame('articleContent')
             // Check the text in the article "Ray Charles"
             .useXpath()
-            .waitForElementPresent("//div[@id='content']/div[@id='mw-content-text']/h2[@id='mweQ']", 40000)
+            .waitForElementPresent("//div[@id='content']/div[@id='mw-content-text']/h2[@id='mweQ']", 20000)
```
Jaifroid, line 78: Revert these changes, which were done in an attempt to fix the tests, and see if reverting them destabilizes currently working tests.

browser-tests/nightwatch_runner.js
```diff
@@ -107,6 +107,8 @@ module.exports = {
             .frame('articleContent')
             // Check the text in the article "Ray Charles"
             .useXpath()
+            // Let's pause for a second
+            .pause(1000)
```
Jaifroid, line 111: Change possibly no longer neceesry, but maybe it does no harm?

nightwatch.js
```diff
@@ -43,7 +43,13 @@ module.exports = {
         "enabled" : false
       },
       "globals": {
-        "waitForConditionTimeout": 600
+        "waitForConditionTimeout": 10000
```
Jaifroid, line 46: Try reverting this change.

nightwatch.js
```diff
@@ -43,7 +43,13 @@ module.exports = {
...
+      },
+      // Configure when a request to the Selenium server should time out and optionally define the number of retries for a timed-out request
+      // See https://github.com/nightwatchjs/nightwatch/issues/1936
+      "request_timeout_options": {
+        "timeout": 100000,
+        "retry_attempts": 3
```
Jaifroid, line 52: May be worth keeping these ones?

www/article.html
```diff
@@ -2,7 +2,7 @@
 <html>
     <head>
         <meta charset="utf-8">
-        <title>Placeholder for injecting an article into the iframe</title>
+        <meta name="description" content="Placeholder for injecting an article into the iframe or window">
```
Jaifroid, line 5: We need to avoid setting a title in the dummy article, because it sometimes sticks and doesn't get replaced by the article title when loading a new tab or window. However, we still need a string in the html that allows us to identify that the dummy article is loaded (see in code below where we search for this string).

www/js/app.js
Jaifroid: Actually, having tried this (a bit) it turns out to be quite tricky, because the function needs access to a fair amount of state that is stored in app.js, and needs to set crucial variables such as `articleContainer` and `articleWindow`. It also needs to access the function `goToArticle()` in `app.js`, and for architectural reasons RequireJS won't allow `app` to be a dependency of `uiUtil`. While it would be possible to pass variables and functions needed, and we could call back to app.js with the values to set in global variables, it becomes a bit unwieldy, and maybe exceeds the definition of UI 'utility'.

Jaifroid:
Reverted changes to tests necessitated a force-push.

browser-tests/nightwatch_runner.js
Jaifroid: Reverting causes test failure.

nightwatch.js
Jaifroid: Reverting causes test failure.

mossroy:
I don't forget this PR : I will have a look when you think it's ready

mossroy:
When starting testing this on Chromium and Firefox (on desktop, both in jQuery mode), I faced some bugs.
My habit is to use the middle-click of my mouse to open a new tab so I configured kiwix-js for tabs, but I could not make it work.
On both browsers, it opens an empty new tab on article.html, and the iframe in the original tab is emptied and shows a spinner. It also triggers the popup blocker of both browsers : if I unblock the popup, it opens the same empty tab.

So I tested to use ctrl-click instead (still configured to use tabs). It works much better : the new tab is opened with the right content, and the original tab only displays the spinner while the new tab is rendered. But afterwards I faced other issues : if I continue browsing the content in the new tab, it works fine. But if I quickly use the back/forward buttons (of the browser, or of my mouse), I end up with an empty tab (and the original tab displays a spinner).

If you don't manage to reproduce these issues, please let me know. It's the same behavior with any kind of ZIM file.

Jaifroid:
Thanks, it sounds very much as if the middle-click isn't being trapped. If it were being trapped, it would run the same way as ctrl-click or right-click (or long-press if it is a touch-screen). I am slightly hampered by the fact that my everyday computer (which I also use for developing) has no middle mouse button. However, I have an old Lenovo Thinkpad with a middle mouse button I can test on (I did test on it in the early days of this PR, but the code may have drifted since then).

Regarding history manipulation, there is clearly a timing or race condition with moving fast backwards and forwards.

Thanks for these reports, I'll look at them at the weekend.

mossroy:
In terms of UI in the Configuration section, I think we miss a way to indicate that this new setting only affects jQuery mode (before the user tries to click on it) and that it works out-of-the-box in SW mode.
The need is similar for the "active content warning" : maybe both could be moved at the end of the "display options", with an indication like "The following settings are specific to jQuery mode to attempt to catch-up the features of ServiceWorker mode"

The triple state of the setting works fine and the "indeterminate" state of the checkbox is supported by all our supported browsers : https://caniuse.com/indeterminate-checkbox. But I'm not a big fan in terms of UI. Using tabs is not "indeterminate", this state is more commonly used in hierarchical options.
Regarding tabs vs windows, it seems to me that tabs is the most usual and convenient way to open a content somewhere else. I even wonder if it's useful to keep the option to open in a new window. I understand it could be for IE11, but I think it's not a problem if this new feature does not work on very deprecated platforms like IE11 of FirefoxOS.

I did not have a deep look at the code but, yes, it's a very huge PR for a feature that does not seem crucial to me. But I also understand it was the occasion for you to clean up some parts of the code

Jaifroid:
Personally, having used it quite a lot, I really like the option to open in a new window compared to a new tab, as I can re-arrange material better that way and can have two windows visible together with OneNote or another note-taking app (for example). I should point out that in Electron, NWJS, PWA and UWP frameworks, the *only* option available is to open in a new window (there is no possibility of a tabbed interface in those frameworks).

Regarding UI, I agree that the tri-state checkbox has some downsides, mostly discoverability. I considered three radio buttons, but this clutters the interface quite a lot. I guess you could have two check-boxes, with the second hidden and appearing only after the feature is switched on. So you would have:

[] Open ZIM link in new tab
[] Open in new window instead <-- (hidden until first is activated)

We could even hide the first option if Service Worker mode is activated.

Happy to add wording about the feature only affecting jQuery, and supported natively in SW mode.

Jaifroid:
I was looking into the wording to be displayed when SW mode is activated. You should see the following wording displayed under the option to open a new tab or window if you switch to SW mode (also see screenshot):

> This setting has no effect in ServiceWorker mode because opening new tabs or windows (if supported by the context) is handled natively with right-clcik or ctrl-click. Turn this setting off to hide this message.

The message is not shown if the user has not activated one of the options. Is this message not clear enough? Is there anything you'd like to change, @mossroy?

image

Jaifroid:
I'm having trouble reproducing the reported navigation bug. In Firefox and Chrome (Edge Chromium) on Windows 10, I can open a new tab, navigate extensively, and use back and forward buttons as fast as I like without the reported issue. What I notice is that communication across tabs is a lot slower than communication with a newly opened window. The latter is very fast, almost instantaneous, whereas the browser is clearly using a slower protocol to send data between tabs.

@mossroy Could you please specifcy which Operating System or Linux flavour you are experiencing those issues with? Could you also please try opening a new window, and let me know if the same navigation bug occurs there? This information might help me reproduce and understand the issue.

www/js/app.js
```diff
@@ -355,10 +365,48 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
...
+    });
+    function setWindowOpenerUI(setCheckbox) {
+        var woState = document.getElementById('windowOpenerState');
+        var woHelp = document.getElementById('windowOpenerHelp');
+        if (params.windowOpener) {
+            woState.innerHTML = params.windowOpener;
+            woHelp.hidden = false;
+            if (contentInjectionMode === 'serviceworker') {
+                woHelp.innerHTML = 'This setting has no effect in ServiceWorker mode because opening new tabs or windows (if supported by the context) ' +
+                    'is handled natively with right-clcik or ctrl-click. Turn this setting off to hide this message.';
```
Jaifroid, line 384: NB typo `clcik` -> `click`

mossroy:
>  I guess you could have two check-boxes, with the second hidden and appearing only after the feature is switched on

I would indeed prefer this UI.
About hiding checkboxes depending on the mode (jQuery/SW), my opinion is that we should not do that if the checkboxes are not next to the choice of the mode (because of discoverability).
Your wording looks OK to me but I would remove the "(if supported by the context)"

I had the navigation issues on Ubuntu 20.04, with latest versions of Firefox (88.0.1) and Chromium (90.0.4430.212), from standard ubuntu repos

Jaifroid:
I've found an old Lenovo Bluetooth keyboard with trackpoint and middle-click mouse button, so I can now test on a modern system. I confirm the issue with the middle click opening one tab that contains article.html, unprocessed. It also (for me) opens another tab that does work correctly. I think I know how to fix.

mossroy:
Regarding the back/forward, I use some buttons of my mouse to do that quickly. If you don't have such buttons, you probably can do it with keyboard shortcuts (alt-left arrow and alt-right arrow should work)

Jaifroid:
> Regarding the back/forward, I use some buttons of my mouse to do that quickly. If you don't have such buttons, you probably can do it with keyboard shortcuts (alt-left arrow and alt-right arrow should work)

I did use alt-left and alt-right in the video I recorded and posted in [the comment above](https://github.com/kiwix/kiwix-js/pull/680#issuecomment-846380938) to illustrate intended experience with fast back and forward, and I can't reproduce on Windows. But I'll try on an Ubuntu VM I have set up.

mossroy:
I have the same issue with new windows or new tabs. It's probably when I use back/forward before the article is fully loaded

Jaifroid:
What I'm seeing with out-of-the-box Firefox (75.0) on Ubuntu 20.4 (VM) is that navigation with alt-left often produces a double or even triple navigation backwards, seeming to pass very briefly through the previous article back to the one before, or the one before that, without giving the intermediate ones time to load. Forward navigation seems to be lost once I go back in a tab as well. Oddly, neither of these things happen in the same browser on Windows. I don't see any kind of bugging out, but I can imagine that a timing issue could emerge in certain combinations of hardware, OS, software.

I propose to try blocking further navigation until the previous page has loaded (i.e. until the point where the spinner would stop). The alternative is to try to find a way to abandon all running Promises on a navigation that has commenced, and as we know that is very hard to achieve, requiring multiple low-level interventions in the back end.

mossroy:
I never saw the double or triple navigation backwards you mention : it always goes to the next or previous page (without skipping any).

The behavior of my issue is not as consistent as I thought, I'll have to test more.

On Chromium, I could not reproduce my issue while using the back/forward buttons of the browser : it only happens when I use the back/forward buttons of my mouse. Here is an example (I use the mouse buttons at the end, when the pointer is in a blank area. I first use "back" several times, and when use "forward", the tab becomes empty) :

https://user-images.githubusercontent.com/890923/119228522-a01d9c80-bb13-11eb-88ba-2bec38afc6ea.mp4

Still with the code of this PR, if the setting is set to "tab" and I don't open another tab, I also have an issue when using the "forward" button of my mouse : it opens a new tab. I'll try to make a screen record.

With the same Chromium, I don't have any issue navigating with my mouse buttons in Kiwix 3.1.0 (as extension), even very quickly, and in jQuery mode

Jaifroid:
Oh, OK, that's very helpful. It looks like it's navigating all the way back to the blank `article.html`, the very first document that is opened in a new tab/window before we populate the document. That would explain the spinner still being shown, because none of the loading code is run and therefore we never get to the point where the spinner is hidden. So I need to prevent navigation back beyond an actual ZIM article, or (better), prevent an empty article.html being recorded in the history.

Jaifroid:
With commit https://github.com/kiwix/kiwix-js/pull/680/commits/6580907b958aee211547766687654ee2818e638a, middle-click now works reliably for me with my bluetooth keyboard on Windows. On Ubuntu it is working reliably for me if i click on a normal ZIM link. However, on one of the graphical landing pages (in the Geography ZIM), I am still getting multiple tabs opening with a single middle click on one of the graphical tiles.

Jaifroid:
Unfortunately, middle-click is still misbehaving, opening both a tab and a window on some occasions. Investigating...

Jaifroid:
Hmm, this is interesting. Seems like recent browsers define an extra "auxclick" event. See https://developer.mozilla.org/en-US/docs/Web/API/Element/auxclick_event for middle-click, and I may need to handle that separately.

Jaifroid:
I think I've fixed the problem with middle-click both on Windows and Ubuntu. Ctrl-click should also work correctly (if set to open a new window, it would open a new window AND a new empty tab). There's a fair amount of diagnostic logging left in the click code, but I'll leave it there for a while during testing.

Jaifroid:
Having said that, I think I see a way to prevent popup blockers: there is an `isTrusted` property on the click event, and it is set to true if the event is the direct result of a click, and to false if it's called from another event (like mousedown) even if the latter is the direct results of a click. It would be best, therefore, not to break that trust chain, and in some cases I am breaking it. Javascript is never easy!!

mossroy:
Middle-click now works on my ubuntu computer (like with ctrl-click), on both Firefox and Chromium. Nice job.

But when I click on another link in the new tab, the spinner appears on the main tab : it's not visible. So the user does not have the usual feedback that the click is being processed. It's not a big issue but is not consistent with the main tab.
A new middle-click in the new tab does not do anything (which might be an acceptable limitation if it's too complicated to implement), but also triggers the popup blocker in the main tab.

Regarding the "empty tab" symptom (of my video above) when using back/forward, I only see it on Chromium, not on Firefox. And only when I use the "forward" button of my mouse (no problem when using the buttons of the browser). I have the same behavior with 2 different brands of mouse (that both have back/forward buttons)

Jaifroid:
> Middle-click now works on my ubuntu computer (like with ctrl-click), on both Firefox and Chromium. Nice job.

Good to hear.
 
> But when I click on another link in the new tab, the spinner appears on the main tab : it's not visible. So the user does not have the usual feedback that the click is being processed. It's not a big issue but is not consistent with the main tab.

The spinner belongs to the chrome in the main window, and I have not attempted to replicate it in a secondary tab or window, as the window/tab is effectively a transparent container that only contains the HTML of the loaded article, much like the iframe. I could of course add the spinner to the secondary container in jQuery mode, but it would then be inconsistent with SW mode, and I believe I should not attempt to modify the DOM in SW mode. My "solution" was to show the spinner only in the main window. Please let me know if you think we could/should reconsider that.

> A new middle-click in the new tab does not do anything (which might be an acceptable limitation if it's too complicated to implement), but also triggers the popup blocker in the main tab.

In jQuery mode, we parse the anchors in the usual way, so a middle-click should be caught and treated in the same way as a middle-click in the iframe, and should open a new tab or window. If you have permanently allowed popups in the main window for the domain in question, there should be no further popups, as tabs are in the same domain (localhost, usually). In SW mode, the middle-click should be supported natively as it is in any other web page (we don't parse the DOM at all for SW mode).

> Regarding the "empty tab" symptom (of my video above) when using back/forward, I only see it on Chromium, not on Firefox. And only when I use the "forward" button of my mouse (no problem when using the buttons of the browser). I have the same behavior with 2 different brands of mouse (that both have back/forward buttons)

OK, thanks. I'm not quite sure how to debug this, or why it is happening. I put in code that prevents navigating while the article is loading, but this is clearly not the issue. Possibly, because it is the mouse that is doing the navigation, it could count as an auxilliary click of some kind, so we are getting an aux mouse click as well as a navigation event... The best I can offer right now is to write more information in console.log about which event(s) have been triggered, which might give us a clue.

mossroy:
I have more information to debug the white page issue :
When the problem happens, I have the following lines in the console of Chromium : 

```
Uncaught TypeError: Cannot set property 'kiwixType' of null
    at onDetectedClick (app.js:1611)
    at HTMLBodyElement.<anonymous> (app.js:1661)
onDetectedClick @ app.js:1611
(anonymous) @ app.js:1661
```

`auxclick`

```
Uncaught TypeError: Cannot read property 'contentDocument' of null
    at Object.applyAppTheme (uiUtil.js:386)
    at windowLoaded (app.js:1467)
```


When using back/forward quickly, I also sometimes get : 

```
Uncaught TypeError: Cannot read property 'title' of null
    at historyPop (app.js:812)
```

Jaifroid:
Thank you @mossroy . There seem to be a few things here. First, a click has been detected, but it doesn't seem to be associated with a Window object, hence attempting to store the window's type (which we do for history manipulation) fails, since `articleContainer` is null. I had never seen that situation, since a click should only be trapped on elements in a loaded document, hence by definition the document and its Window are available. Possibly the window is in the process of being unloaded (wild guess).

I'm confused by why a back or forward button should produce a click at all on any DOM element. Could there be a proprietary mouse driver that reads a type of mouseclick and then programmatically does `history.back()` or `history.forward()`, possibly with a timeout? A clue might be in that `auxclick` event, which I'd hypothesized above.

The final error could be caused by going back in history to the initial load of the blank `article.html` which doesn't have a title associated with it.

I'll thiink more about this and see if I can reproduce.

Jaifroid:
Looking into this, your mouse almost certainly comes with software that remaps buttons to custom actions, or else you have set up a Linux mapping/driver to do the same. On Windows there is this: https://www.addictivetips.com/windows-tips/remap-mouse-buttons-on-windows-10/ which allows various remappings for people with generic mice that don't come with a proprietary driver, so I'll see if I can simulate what's happening with that.

I guess we would not normally try to work around proprietary hardware, but in this case it could be revealing flaws in the implementation, so a good proxy for a testing for race conditions, etc.

mossroy:
I did not have to install or configure anything for the mouse. The driver (generic or specific) must be open-source and bundled in the linux kernel. And the mapping of buttons is the default one
It's the same mapping as in https://medium.com/@Aenon/bind-mouse-buttons-to-keys-or-scripts-under-linux-with-xbindkeys-and-xvkbd-7e6e6fcf4cba : back button is button 8, and forward button is button 9. I checked with `xev` command-line
They are normally mapped with XF86Back and XF86Forward events

But I don't manage to reproduce the issue with the keyboard, even when mapping a real key (like pause) to XF86Forward with `xmodmap -e "keycode 127=XF86Forward"`

Jaifroid:
OK, so I gues it's the combination of a mouseclck being registerd and a navigation event both happening (roughly) at the same time, since the underlying driver must see a click in order to convert it into the appropriate browser command, or map it to a keyboard equivalent. I think we're getting closer, thanks, and I can work with that!

mossroy:
I think that we're spending too much time and code on this feature.
It's more complicated than initially expected in #678 and there will be limitations (see comments above).

I tend to think that we should not invest that time to improve the jQuery mode in kiwix-js. I'll open a separate issue to discuss that more broadly.

But this PR also contains some refactoring that might be kept (probably in a new PR) if it might ease to keep this feature in kiwix-js-windows

Jaifroid:
That's fine, @mossroy . I think the feature works quite well when targeted at sepecific ecosystems where you can control the variables, but targeted across multiple OS's and a range of potential hardware inputs, it gets complicated to iron out all the bugs, as we've seen, particularly managing history and so on. And I'm still not 100% satisfied that the tests are working as expected: with every other push, I have to restart the tests (they do complete correctly on the second go), but that seems indicative of a race condition I haven't been able to get to the bottom of.

So I can close this if you like, but keep the branch to decide what from here could be kept in kiwix-js. On the UWP app, the feature seems to be working well. Admittedly I haven't tried a mouse with various buttons (other than right-, left- and middle-click, which work) but UWP is more touch-and-click-centric, and that seems to be covered. Electron/NWJS may be less controlled as they explicitly target various operating systems, but the usage is very small compared to the UWP app, and only really for those who can't run Kiwix Desktop (i.e., for 32-bit).

PR #683, opened 2020-12-08
Create all packages on release

Jaifroid:
@mossroy I thought we might use this release to test a GitHub Action that could run `create_all_packages.sh` on publishing a source code release (and creating a tag), once we've added the secret in the approved way.

However, I've just realized that no doubt this script will run into the same problems I had in terms of access to commands to build the .click package, and maybe that's not easy to solve.

The information on adding encrypted secret files is here: https://docs.github.com/en/free-pro-team@latest/actions/reference/encrypted-secrets .

Is it worth trying to do this for this release, or is it a future project?


Jaifroid:
I've copied the Android setup for the ssh secret, but it seems we will need to add the Mozilla secret as well, and I don't have the Repo permissions to add a secret at Repo level. @mossroy Shall we leave this setup for later development, and not delay the current release?

mossroy:
If you're in the mood of trying to use Github Actions for that, it would be a great starting point for #590 
I just added MOZILLA_API_KEY and MOZILLA_API_SECRET as new repository secrets.
I also configured your github user as admin of kiwix-js : you should have the right to manage secrets now

Regarding .click packages, it was easier on Travis as it used an old ubuntu version, where the packages were available in standard repos.
As GitHub Actions probably uses a more recent version of ubuntu, you'll probably have to install them manually, as you did on your PC. It should not be very complicated : `sudo apt-get install ./*.deb` (if the deb files are in the current directory)

Regarding the command-line to run, have a look at the one of Travis : last lines of https://github.com/kiwix/kiwix-js/blob/master/.travis.yml. You probably miss the `-v` command-line parameter to pass the tag name, and maybe the `DISPLAY=:99.0` to let Chrome run (to sign the package) without crashing because it does not find a display (maybe there's now a better way to tell Chrome to run in full headless mode)

Jaifroid:
@mossroy I would like to merge this to master for testing. The way I shall test is to use the -d dryrun switch, which I have set in this branch, and I will manually run the action to debug it, i.e. without creating a release.

If you are able to look over the code and offer any advice about the linux commands, that would be helpful -- and in particular to check that the code will in fact run as a "dry run" (i.e. I haven't misunderstood the syntax).

However, if you haven't got time to look at the code, then just let me know if I can merge this to master for testing (i.e. approve it).

Jaifroid:
And please note that if this is merged to master, you will need to rename your `travisci_builder_id_key` to `ssh_key` if you want to run the build script locally on your own machine.

mossroy:
Sorry, I won't have time to review the code this evening or tomorrow.
But you can merge it if it helps testing

mossroy (approved):

Jaifroid:
Reopening this branch because I think I can see a way to debug the action from a branch rather than from master.

Jaifroid:
Even when explicitly choosing to run the action from this branch (Actions tab -> run action -> select branch), it still runs the code on master. Strange.

So @mossroy it seems that to debug properly I will have to push all my debugging changes to master, which could be a few commits that will pollute the commit log (because I only know what to commit after a failure, and some commits will be trial-and-error).

I could fix that afterwards by resetting master to commit https://github.com/kiwix/kiwix-js/commit/4a8efd48edf42e5cd7e6eb2e0eb8ad2b3cbe3926 and then force pushing a single commit with all changes once I have clean code. I know force pushing to "rewrite history" is usually frowned upon, but with only two active contributors here, maybe it's OK in order to have a clean commit log? Clearly you'd have to hard-reset your master branch after that.

If you know a better way to do it, let me know!

mossroy:
You can go ahead, just be careful when your force-push on master

Jaifroid:
@mossroy FYI I found out how to run the script from a branch (actually it was working first time, I just thought it had failed for the wrong reason), so I've cleaned up master, and am now experimenting on the branch that this PR was made from (but commits don't show up in this PR, because already merged). It's really tricky to get the commands right, with the interaction between YAML and bash, especially as I don't know syntax for either well...

PR #685, opened 2020-12-10
Action build all packages

Jaifroid:
I'm opening this new PR to get help on the code, as I'm currently a bit stuck and nothing I've tried works. I'm trying to echo the SSH_KEY secret to an `ssh_key` file in the scripts directory, so that our bash script can use it as an input, but I keep getting the error:

`echo "$SSH_KEY" > scripts/ssh_key: No such file or directory`

I've tried writing it as `echo "$SSH_KEY" > ./scripts/ssh_key` and `echo "$SSH_KEY" > $GHWS/scripts/ssh_key` (where `$GHWS = /home/runner/work/kiwix-js/kiwix-js` which is `pwd`). I included an `ls` command to show the files and directories, and `scripts` is clearly there (see screenshot). There is probably some real noob error here (on my part).

image


Jaifroid:
More infor on the reason for using single quotes round that line in YAML here: https://stackoverflow.com/questions/59481933/github-actions-how-can-i-save-secret-to-file 

Also, I tried to follow the syntax used in the Android app's file, but a) they use an old checkout API (checkout v1) which does not have the GitHub Workspace; and b) I don't think our secret is base64 encoded (they pipe it through base64). In any case, that syntax gives the same not found error.

Jaifroid:
OK, so I've temporarily changed the `ssh_key` file to `~/ssh_key`, which seems to get over the write error, so maybe it's not possible to write into a checked out directory. This means we will need to change the logic in the `create_all_packages` script, but for now I've just edited it to use `~/ssh_key` as its identity file.

I now have another problem with the .click installation. @mossroy would you be able to look over the log file from the attempted installation which was not successful? What is missing? The log file can be seen here:

https://github.com/kiwix/kiwix-js/runs/1532393802?check_suite_focus=true

The last few lines summarize the issue -- it possibly is a dependency problem:
```
Errors were encountered while processing:
 /tmp/apt-dpkg-install-4CamcF/03-libclick-0.4-0_0.4.43+16.04.20170613-0ubuntu1_amd64.deb
 /tmp/apt-dpkg-install-4CamcF/05-python3-click-package_0.4.43+16.04.20170613-0ubuntu1_amd64.deb
E: Sub-process /usr/bin/dpkg returned an error code (1)
```

mossroy:
At least the libclick file seems to be corrupted in the repo. It probably explains the issues when installing them. My file has a size of exactly 58300 octets on my PC, it has a size of 58299 octets on github

Jaifroid:
Oh, how strange -- I literally took them from the directory where I had saved them. I'll check, thanks.

Jaifroid:
Yes of course -- please do!
(Also, I took the files just now directly from your email and saved them to the ubuntu-touch folder (commit above), and re-ran the tests, but it's still failing with a corrupted error:

https://github.com/kiwix/kiwix-js/runs/1533511918?check_suite_focus=true

mossroy:
The corrupted files might come from the fact that you are using windows, that does not use the same line-ending as on unix. If your git client considers these .deb files as text files (and not binary), it will try to change their line-endings and corrupt them.
Try adding the following line at the end of .gitattributes file, so that it considers them as binary : 

    *.deb   binary

Jaifroid:
Ah, OK, will try that.

mossroy:
I modified the script to write to ./scripts/ssh_key and it seems to work ok : I don't understand what was wrong

mossroy:
Your git client might not see these files as changed and might not allow you to commit them. You might need to delete them, commit, then add them again.
At worst, you can upload the files through the github web interface : in https://github.com/kiwix/kiwix-js/tree/Action-build-all-packages/ubuntu_touch : Add file -> upload files
But it would be better to fix this issue (your .gitattributes looks correct)

Jaifroid:
Thank you -- probably a noob error on my part then. I've pushed gitattributes to both master and to this branch, and am going to delete the .deb packages, commit, and then re-upload them.

Jaifroid:
Thank you, thank you! That's now got us past the environment setup. The next error to resolve is this (the headless display syntax):

image

mossroy:
No problem!
For the latest issue, try to remove the "sh " at the beginning

Jaifroid:
> For the latest issue, try to remove the "sh " at the beginning

I believe the reason to use `sh` was to be able to run the script without making it executable (from a Stackexchange answer). But it's easy enough to do that with a chmod on the `create_all_packages` script. Will try this tomorrow.

mossroy:
> > For the latest issue, try to remove the "sh " at the beginning
> 
> I believe the reason to use `sh` was to be able to run the script without making it executable (from a Stackexchange answer). But it's easy enough to do that with a chmod on the `create_all_packages` script. Will try this tomorrow.

It should not be necessary as git keeps the executable bit : 

     -rwxr-xr-x  1 runner docker 4020 Dec 10 21:53 create_all_packages.sh

Jaifroid:
Nearly there! It seems to have built all the packages (well, not sure about the .click one), but failed on an upload step (but I'm glad it failed, because it was supposed to be a dry run and shouldn't have attempted an upload...). I think I must have got the dry run switch wrong because it seems to have incorporated `-d` into the click package's filename (or I'm seeing things because it's getting late). I'll look at this properly in the morning:

image

Jaifroid:
Everything seems to be building reliably now (with some warnings about the .click build, see below), using a test TAG_NAME of '0.0.0'. I included this only to be sure that I was passing TAG_NAME correctly (and it's only available on publishing a release). I'll remove that logic before rebasing/merging to master because I know the script produces its own build tag if one is not passed.

image

Remaining warnings - are these significant? It does have FAIL in capital letters! :
```
Errors
------
 - lint:framework
	'ubuntu-sdk-16.04' is not a supported framework
	http://askubuntu.com/questions/460512/what-framework-should-i-use-in-my-manifest-file
 - security:policy_version_matches_framework:kiwix.apparmor
	Invalid framework 'ubuntu-sdk-16.04'
./kiwix_0.0.0_all.click: FAIL
Successfully built package in './kiwix_0.0.0_all.click'.
Skipping uploading the files, because it's a dryrun test
```

Jaifroid:
The final test before merging would be to remove the dryrun and let it build and upload a version 0.0.0, to see if the secrets are being used correctly. We did have a failure earlier, with this error (mistake with dryrun meant it attempted to upload builds):
```
Uploading the files on https://download.kiwix.org/nightly/2020-12-10/
ssh_askpass: exec(/usr/bin/ssh-askpass): No such file or directory
Host key verification failed.
ssh_askpass: exec(/usr/bin/ssh-askpass): No such file or directory
Host key verification failed.
lost connection
```
This looks a bit serious "Host key verification failed" - it doesn't know whether to trust the host or not.

mossroy:
I'm pretty sure we already had this error message (and I reproduce the same on my PC) : the .click file is generated, but it's the "click-review" that fails.
We might avoid this by adding "--no-validate" to the "click build" command line

There's a minor issue, though, in the click package : it includes the .deb files inside it, because they are in the "ubuntu_touch" directory it uses. You need to find a way to exclude them. Maybe by copying only the necessary files in create_all_apackages.sh instead of "cp -f ubuntu_touch/* tmp/"

Jaifroid:
Thanks for spotting that. I could put the *.deb files in the scripts directory, or maybe in `scripts/ubuntu_touch` to keep things neat?

mossroy:
Regarding the ssh/scp error messages, the "Host key verification failed" might be because it does not trust the SSH key of download.kiwix.org.
In Travis, there was a line to do that : 

    ssh_known_hosts: download.kiwix.org

I don't find the equivalent on GitHub Actions.
On kiwix-android, they used `-o StrictHostKeyChecking=no` command-line parameter to scp to avoid that.
This parameter is usually considered as a security risk : it's true, but what Travis did was the same IMHO : it trusted the SSH key that it read a few lines above

Jaifroid:
Hooray!

image

Jaifroid:
I'm going to rename `main.yml` to `release.yml` (using the names adopted in Android app), and presumably we copy this, change the event and save it as a `nightly.yml` as well (though possibly we can add more than one launch event, so it can serve in both scenarios -- I'll check). I'll rename the action name accordingly, if it can be used for both events, or only for one event. Then I'll rebase this branch.

Jaifroid:
@mossroy I've renamed, added a nightly cron event, and have rebased on master.
Because I've renamed it, we can no longer run it from the Actions tab until this PR has been merged to master (because no equivalent yaml file exists in master). It also won't run nightly until merged (I believe).

.github/workflows/release.yml
```diff
@@ -1,14 +1,17 @@
 # This sets the environment for building Kiwix JS store packages and runs the build script
 
-name: Build all packages on release publish
+name: Build all packages nightly or on publish new release
```
mossroy, line 3: As it covers also the nightly builds, maybe the YAML file should not be named "release.yml"

.github/workflows/release.yml
```diff
@@ -24,24 +27,27 @@ jobs:
       - uses: actions/checkout@v2
```
mossroy, line 27: This comment is about line : 

    runs-on: ubuntu latest

It might be better to choose the version of ubuntu here, so that we decide when to switch to a newer ubuntu version instead of having to follow the rythm chosen by Github

.github/workflows/release.yml
```diff
@@ -24,24 +27,27 @@ jobs:
...
-          # only exists if the release event has fired
-          TAG_NAME: ${{ github.event.release.tag_name } || {'0.0.0'}}
+          TAG_NAME: ${{ github.event.release.tag_name }}
+        shell: bash
         run: |
-          echo "$SSH_KEY" | base64 -d > ./$GHWS/scripts/ssh_key
-          chmod 600 ./$GHWS/scripts/ssh_key
+          echo "$SSH_KEY" > ./scripts/ssh_key
+          chmod 600 ./scripts/ssh_key
+          ls -al ./scripts/
```
mossroy, line 40: This line was meant for debug, I suppose, and could be removed now

.github/workflows/release.yml
```diff
@@ -24,24 +27,27 @@ jobs:
...
+          sudo apt-get update
+          sudo apt-get --yes install ./scripts/ubuntu_touch_packages/*.deb
 
       # Runs the build scripts
       - name: Run the build scripts
-        # Switch -t indicates a tag release; -d indicates dry run (for testing)
-        run: sh DISPLAY=:99.0 ./$GHWS/scripts/create_all_packages.sh -t -v $TAG_NAME -d
+        shell: bash
+        # Switch -t indicates a tag release; add -d for dry run (for testing)
+        run: DISPLAY=:99.0 ./scripts/create_all_packages.sh -t -v "$TAG_NAME"
```
mossroy, line 53: The parameters must be different when it's a nightly build : see .travis.yml

mossroy (changes requested):
Congratulations for this work! It will probably work for releases.
For nightly builds, you'll have to adjust some command-line parameters

.github/workflows/release.yml
Jaifroid: OK, done.

.github/workflows/release.yml
Jaifroid: OK, I could rename it `nightly_or_release.yml`,

.github/workflows/release.yml
Jaifroid: Interestingly, if `$TAG_NAME` is empty, then the script builds a nightly build and uploads it to nightly with the correct naming convention (see screenshot below from my test earlier today, taken from http://download.kiwix.org/nightly/2020-12-11/ ). In the light of this, do you still think I should add logic to remove the parameters if it is a nightly build?

image

.github/workflows/release.yml
Jaifroid: I've checked the labels here: https://github.com/actions/virtual-environments . We could choose `ubuntu-18.04` (which is what we've tested on), or else we could choose upcoming `ubuntu-20.04`, which is in preview. As we are still testing, we could test on this, and adopt it if it works.

.github/workflows/release.yml
mossroy: OK

.github/workflows/release.yml
mossroy: I agree, we could test ubuntu-20.04.
It will probably work as we're doing basic things, and would be a more recent baseline than 18.04

Jaifroid: Done.

.github/workflows/release.yml
mossroy: The parameters need to be removed on a nightly build, at least for the firefox extension : depending on the presence of `-t` parameter, it will create a "unsigned listed" or "signed unlisted" extension. They have different ids, and different purposes : 

- "unlisted" extensions are for testing purpose, automatically signed, and not available on Mozilla store
- "listed" extensions are for deployment purpose. They are manually uploaded, checked by Mozilla, and signed by Mozilla if accepted, and afterwards available on their store

Jaifroid: Ah, OK. I think I can simply do:
```
if [ "x$TAG_NAME" = "x" ]; then
   DISPLAY=:99.0 ./scripts/create_all_packages.sh -t -v "$TAG_NAME"
else
   DISPLAY=:99.0 ./scripts/create_all_packages.sh
fi
```
Because the script will only ever be run with a TAG_NAME (in which case it's a release), or without one (in which case it's a nightly build or else an Action dispatch build). I'll try this but won't be able to test properly till merged to master.

Jaifroid: Oops, I think the if and else blocks need to be reversed.

Jaifroid: Done (with the then-else blocks the reverse of the sample above).

Jaifroid:
So I think all requested changes have been done, but testing will have to wait till after approval and merge to master, as the Actions tab doesn't know about the existence of this script yet (due to name change).

Jaifroid:
Ooh, pushing a commit seems to have caused a run (strange), with this result (it says there is an error in my YAML on line 56):

image

.github/workflows/nightly_or_release.yml
```diff
@@ -0,0 +1,57 @@
...
+      # Runs the build scripts
+      - name: Run the build scripts
+        shell: bash
+        # Switch -t indicates a tag release; add -d for dry run (for testing)
+        run: |
+          if [ "x$TAG_NAME" = "x" ]; then
+            DISPLAY=:99.0 ./scripts/create_all_packages.sh
+          else
+            DISPLAY=:99.0 ./scripts/create_all_packages.sh -t -v "$TAG_NAME"
+         fi
```
mossroy, line 57: It's probably an indentation issue : the `fi` should be shifted one space on the right

Jaifroid: Thanks, I've corrected that. I didn't know indentation was that significant in YAML files.

Jaifroid:
Should I add a Changelog item?

> * UPDATE: Switched from Travis to GitHub Actions for nightly and release builds

I'm not sure whether the Changelog should register meta items, or only items that represent a public-facing change to the app itself.

I can also add the release date of today for version 3.1 (to the Changelog), as I think we should be able to build release today. Or I can do that as a separate commit to master.

mossroy (approved):
Looks good!
When I have a look at the changelog on kiwix-android, I only see changes relevant for end-users, so maybe we should do the same.
You can add the date and do the release

Jaifroid:
Merged. So far so good: manually dispatching the Action from master produces the expected packages in Nightly:

image

I'll now do a release to see if it triggers a release build as expected.

Jaifroid:
Well, the Action did run on making the release, but it looks to me as if the tag was not passed to the script correctly, and it has built another Nightly rather than a public release. Moreover, there appear to be a bunch of unexpected npm errors (which nevertheless did not prevent the script uploading packages) in the log. See:

https://github.com/kiwix/kiwix-js/runs/1541780265?check_suite_focus=true 

and open "Run the build scripts". I'll check to find out why the event did not appear to contain the correct tag, and what the correct way to pass that is.

Jaifroid:
I can see an obvious error in the logs:

`Warning: Identity file ../scripts/travisci_builder_id_key not accessible: No such file or directory.
`
Clearly one of the scripts still has a reference to that, which I was not aware of.  Will correct.

mossroy:
It's also trying to use chromium-browser and it's not installed. We could install it first, but Chrome is installed and could be used instead. We could modify the script to use chromium-browser if it exists, and else us Chrome. I can help on that

Jaifroid:
OK. I'll quickly make a new branch/PR to include the changes I've just made.

Jaifroid:
We should now switch to #686.

PR #686, opened 2020-12-12
Corrections to nightly_or_release GitHub Action

Jaifroid:
This PR is to make corrections as a result of testing of #685.

### DIFF
--- a/.github/workflows/nightly_or_release.yml
+++ b/.github/workflows/nightly_or_release.yml
@@ -54,7 +54,7 @@ jobs:
 
       # Runs the build scripts
       - name: Run the build scripts
-        env: 
+        env:
           MOZILLA_API_SECRET: ${{ secrets.MOZILLA_API_SECRET }}
           MOZILLA_API_KEY: ${{ secrets.MOZILLA_API_KEY }}
         shell: bash
--- a/.github/workflows/nightly_or_release.yml
+++ b/.github/workflows/nightly_or_release.yml
@@ -32,12 +32,11 @@ jobs:
         shell: bash
         run: echo ::set-output name=SOURCE_TAG::${GITHUB_REF#refs/tags/}
 
-      # Decrypt the ssh secret
+      # Set up secret files from encrypted secrets
       - name: Set up secret files
         env:
           SSH_KEY: ${{ secrets.SSH_KEY }}
           CHROME_EXTENSION_KEY: ${{ secrets.CHROME_EXTENSION_KEY }}
-          TAG_NAME: ${{ steps.source_tag.outputs.SOURCE_TAG }}
         shell: bash
         run: |
           echo "$SSH_KEY" > ./scripts/ssh_key
@@ -57,6 +56,7 @@ jobs:
         env:
           MOZILLA_API_SECRET: ${{ secrets.MOZILLA_API_SECRET }}
           MOZILLA_API_KEY: ${{ secrets.MOZILLA_API_KEY }}
+          TAG_NAME: ${{ steps.source_tag.outputs.SOURCE_TAG }}
         shell: bash
         # Switch -t indicates a tag release; add -d for dry run (for testing)
         run: |
--- a/.github/workflows/nightly_or_release.yml
+++ b/.github/workflows/nightly_or_release.yml
@@ -65,8 +65,11 @@ jobs:
           else
             echo "MOZILLA_API_SECRET is available"
           fi
+          # Run xvfb to make Chrome/Chromium believe it has a display. Else it fails signing the package
+          xvfb -ac :99 -screen 0 1280x1024x16 &
+          export DISPLAY=:99
           if [ "x$TAG_NAME" = "x" ]; then
-            DISPLAY=:99.0 ./scripts/create_all_packages.sh
+            ./scripts/create_all_packages.sh
           else
-            DISPLAY=:99.0 ./scripts/create_all_packages.sh -t -v "$TAG_NAME"
+            ./scripts/create_all_packages.sh -t -v "$TAG_NAME"
           fi
--- a/.github/workflows/nightly_or_release.yml
+++ b/.github/workflows/nightly_or_release.yml
@@ -60,10 +60,15 @@ jobs:
         shell: bash
         # Switch -t indicates a tag release; add -d for dry run (for testing)
         run: |
+          if [ "x$MOZILLA_API_KEY" = "x" ]; then
+            echo "WARNING: MOZILLA_API_KEY is empty!"
+          else
+            echo "MOZILLA_API_KEY is available and is of size ${#MOZILLA_API_KEY}"
+          fi
           if [ "x$MOZILLA_API_SECRET" = "x" ]; then
             echo "WARNING: MOZILLA_API_SECRET is empty!"
           else
-            echo "MOZILLA_API_SECRET is available"
+            echo "MOZILLA_API_SECRET is available and is of size ${#MOZILLA_API_SECRET}"
           fi
           # Run xvfb to make Chrome/Chromium believe it has a display. Else it fails signing the package
           xvfb -ac :99 -screen 0 1280x1024x16 &
--- a/.github/workflows/nightly_or_release.yml
+++ b/.github/workflows/nightly_or_release.yml
@@ -48,6 +48,8 @@ jobs:
           chmod 600 ./scripts/kiwix-html5.pem
           if [ "x$MOZILLA_API_SECRET" = "x" ]; then
             echo "WARNING: MOZILLA_API_SECRET is empty!"
+          else
+            echo "MOZILLA_API_SECRET is available"
           fi
 
       # Set up the environment
--- a/.github/workflows/nightly_or_release.yml
+++ b/.github/workflows/nightly_or_release.yml
@@ -71,7 +71,7 @@ jobs:
             echo "MOZILLA_API_SECRET is available and is of size ${#MOZILLA_API_SECRET}"
           fi
           # Run xvfb to make Chrome/Chromium believe it has a display. Else it fails signing the package
-          xvfb -ac :99 -screen 0 1280x1024x16 &
+          Xvfb -ac :99 -screen 0 1280x1024x16 &
           export DISPLAY=:99
           if [ "x$TAG_NAME" = "x" ]; then
             ./scripts/create_all_packages.sh
--- a/.github/workflows/nightly_or_release.yml
+++ b/.github/workflows/nightly_or_release.yml
@@ -46,6 +46,9 @@ jobs:
           chmod 600 ./scripts/ssh_key
           echo "$CHROME_EXTENSION_KEY" > ./scripts/kiwix-html5.pem
           chmod 600 ./scripts/kiwix-html5.pem
+          if [ "x$MOZILLA_API_SECRET" = "x" ]; then
+            echo "WARNING: MOZILLA_API_SECRET is empty!"
+          fi
 
       # Set up the environment
       - name: Setup environment
--- a/.github/workflows/nightly_or_release.yml
+++ b/.github/workflows/nightly_or_release.yml
@@ -26,13 +26,19 @@ jobs:
       # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
       - uses: actions/checkout@v2
 
+      # Get the source tag
+      - name: Source tag
+        id: source_tag
+        shell: bash
+        run: echo ::set-output name=SOURCE_TAG::${GITHUB_REF#refs/tags/}
+
       # Decrypt the ssh secret
       - name: Decrypt files and set constants
         env:
           SSH_KEY: ${{ secrets.SSH_KEY }}
           MOZILLA_API_SECRET: ${{ secrets.MOZILLA_API_SECRET }}
           MOZILLA_API_KEY: ${{ secrets.MOZILLA_API_KEY }}
-          TAG_NAME: ${{ github.event.release.tag_name }}
+          TAG_NAME: ${{ steps.source_tag.outputs.SOURCE_TAG }}
         shell: bash
         run: |
           echo "$SSH_KEY" > ./scripts/ssh_key
--- a/.github/workflows/nightly_or_release.yml
+++ b/.github/workflows/nightly_or_release.yml
@@ -60,16 +60,6 @@ jobs:
         shell: bash
         # Switch -t indicates a tag release; add -d for dry run (for testing)
         run: |
-          if [ "x$MOZILLA_API_KEY" = "x" ]; then
-            echo "WARNING: MOZILLA_API_KEY is empty!"
-          else
-            echo "MOZILLA_API_KEY is available and is of size ${#MOZILLA_API_KEY}"
-          fi
-          if [ "x$MOZILLA_API_SECRET" = "x" ]; then
-            echo "WARNING: MOZILLA_API_SECRET is empty!"
-          else
-            echo "MOZILLA_API_SECRET is available and is of size ${#MOZILLA_API_SECRET}"
-          fi
           # Run xvfb to make Chrome/Chromium believe it has a display. Else it fails signing the package
           Xvfb -ac :99 -screen 0 1280x1024x16 &
           export DISPLAY=:99
--- a/.github/workflows/nightly_or_release.yml
+++ b/.github/workflows/nightly_or_release.yml
@@ -44,6 +44,8 @@ jobs:
         run: |
           echo "$SSH_KEY" > ./scripts/ssh_key
           chmod 600 ./scripts/ssh_key
+          echo "$CHROME_EXTENSION_KEY" > ./scripts/kiwix-html5.pem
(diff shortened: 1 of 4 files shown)


scripts/package_firefox_extension.sh
```diff
@@ -40,9 +40,9 @@ if [ "${TAG}zz" == "zz" ]; then
     else
             echo "Extension not signed by Mozilla. It might be because this commit id has already been signed : let's look for it in a previous nightly build"
             FOUND=0
-            for FILE in $(ssh -i ../scripts/travisci_builder_id_key ci@download.kiwix.org "find /data/download/nightly -name \"kiwix-firefox-signed-extension-$VERSION.xpi\""); do
+            for FILE in $(ssh -o StrictHostKeyChecking=no -i ./scripts/ssh_key ci@download.kiwix.org "find /data/download/nightly -name \"kiwix-firefox-signed-extension-$VERSION.xpi\""); do
```
mossroy, line 43: It's probably ../scripts/ssh_key instead of ./scripts/ssh_key

Jaifroid: I did wonder. Please feel free to make any changes you notice, as you have a much better grasp of how these scripts are working (and of bash syntax) than I do.

mossroy:
To use Chromium or Chrome, you could run something like : 

    if hash chromium-browser 2>/dev/null
    then
        echo "Chromium is found : use Chromium"
        CHROME_BIN=chromium-browser
    else
        echo "Chromium is not found : use Chrome"
        CHROME_BIN=chrome
    fi
    $CHROME_BIN (followed by the needed parameters)

Jaifroid:
OK, thanks, I'll try.

.github/workflows/nightly_or_release.yml
```diff
@@ -26,13 +26,19 @@ jobs:
       # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
       - uses: actions/checkout@v2
 
+      # Get the source tag
+      - name: Source tag
+        id: source_tag
+        shell: bash
+        run: echo ::set-output name=SOURCE_TAG::${GITHUB_REF#refs/tags/}
```
Jaifroid, line 33: @mossroy I have copied code from https://github.community/t/how-to-get-just-the-tag-name/16241/11 to output the source tag. The output from this step is then read into line 41 below to set the TAG_NAME. Untested so far.

scripts/package_firefox_extension.sh
Jaifroid: I'll make these changes now.

scripts/package_chrome_extension.sh
```diff
@@ -16,9 +16,16 @@ cd tmp
...
+    # Package the extension with Chrome or Chromium, if we're not packaging a public version
+    if hash chromium-browser 2>/dev/null then
+        echo "Chromium is available"
+        CHROME_BIN=chromium-browser
+    else
+        echo "Chrome is available"
+        CHROME_BIN=chrome
+    fi
+    echo "Signing the extension for $CHROME_BIN, version $VERSION"
+    $CHROME_BIN --no-sandbox --pack-extension=tmp --pack-extension-key=./scripts/kiwix-html5.pem
```
Jaifroid, line 28: @mossroy Please check that the above code looks correct.

I'm also a bit worried about the single `.` before `/scripts` in the line above: `--pack-extension-key=./scripts/kiwix-html5.pem`. By analogy with the firefox packaging file, it would seem this might need to be `../scripts` (but it was originally like this, so I'm probably missing something).

Jaifroid: Ah, also, kiwix-html5.pem is a secret file not committed to the repo, right? In which case we'll have to make it too and upload it as a secret. Correct?

Jaifroid: I've added it as CHROME_EXTENSION_KEY and have uploaded the secret. I need to find out if --pack-extension-key can accept the constant, or if we need to make a file.

Jaifroid: Seems it only takes a file. Will make one.

.github/workflows/nightly_or_release.yml
mossroy: Needs to be tested (including on nightly), but looks good

scripts/package_chrome_extension.sh
Jaifroid: I've made the file from the secret. My remaining doubt is about whether to use `../scripts` or `./scripts` here. Is the location read from the current working directory (in which case I think we need `../scripts`) or from the perspective of the chrome executable (in which case we may need something more elaborate that includes the $GITHUB_WORKSPACE directory structure)?

mossroy: The `./scripts` is probably OK because there is a `cd ..` above.
These scripts would deserve a good refactoring : the bash coding style could really be improved

You're right to create this new secret for Chrome. You'll probably have to serialize it to a file, like for the ssh key

Jaifroid: Thanks. I've added code to make the file already. I can run a tests from this branch. It will upload to nightly, but should allow us to see if things are working in the "nightly" mode.

Jaifroid:
I've run a test, output: https://github.com/kiwix/kiwix-js/runs/1542212680?check_suite_focus=true 

It didn't like my if ... then ... else syntax in bash 😥:
```
scripts/package_chrome_extension.sh: line 23: syntax error near unexpected token `else'
scripts/package_chrome_extension.sh: line 23: `    else'
```
Will correct.

There is also another error relating to the Firefox extension -- it looks like we haven't set or referenced one of the secret constants correctly:
```
Error: required argument was empty: apiSecret
    at reportEmpty (/home/runner/work/kiwix-js/kiwix-js/node_modules/sign-addon/dist/webpack:/src/index.js:66:11)
    at v (/home/runner/work/kiwix-js/kiwix-js/node_modules/sign-addon/dist/webpack:/src/index.js:78:5)
    at /home/runner/work/kiwix-js/kiwix-js/node_modules/web-ext/dist/web-ext.js:1:70282
    at processTicksAndRejections (internal/process/task_queues.js:93:5)
    at A.execute (/home/runner/work/kiwix-js/kiwix-js/node_modules/web-ext/dist/web-ext.js:1:34732)
```

Jaifroid:
Now we're getting:
```
Chrome is available
Signing the extension for chrome, version 3.1commit-9b48447
scripts/package_chrome_extension.sh: line 29: chrome: command not found
```
Maybe we need some path information to run the chrome executable?

mossroy:
I probably fixed the command-line for Chrome.
Regarding the Mozilla API secret, I've set it again in https://github.com/kiwix/kiwix-js/settings/secrets/actions , in case I did it wrong

Jaifroid:
Thanks, nearly there now...
The Chrome command is fixed, it now runs. There is an error reported in moving a file that it produces (second line below):
```
Signing the extension for google-chrome-stable, version 3.1commit-4f42e2a
mv: cannot stat 'tmp.crx': No such file or directory
```
For Mozilla, we're still getting (last line):
```
Signing the extension for Firefox with Mozilla API, version 3.1commit-4f42e2a
Building web extension from /home/runner/work/kiwix-js/kiwix-js/tmp
Error: required argument was empty: apiSecret
```
I am guessing this is something to do with the option `--api-secret=${MOZILLA_API_SECRET}`. Maybe $MOZILLA_API_SECRET is an empty variable, in which case the error is in how we set it up in the YAML file (I doubt it's an issue with the way you uploaded it: simply copy and paste the private key into the secret's value field).

I can push some test code to see if it's empty. Apparently we shouldn't try to echo it to the log even though it should be redacted (see https://docs.github.com/en/free-pro-team@latest/actions/reference/encrypted-secrets):

image

Jaifroid:
Interesting. I pushed test code in two places, one in the YAML file, and one in the mozilla script. In the first MOZILLA_API_SECRET is clearly populated. In the second, when we try to access it, it is void:
```
Signing the extension for Firefox with Mozilla API, version 3.1commit-a77b56b
WARNING: MOZILLA_API_SECRET is empty!
```
I think I know what can be done about this (access the variable from env.MOZILLA_API_SECRET instead).

Jaifroid:
I've fixed the access to the variables. Despite what it explicitly says in https://docs.github.com/en/free-pro-team@latest/actions/reference/environment-variables ("You can define environment variables for a step, job, or entire workflow using the jobs.<job_id>.steps.env, jobs.<job_id>.env, and env keywords"), in fact it seems you have to define the environment variable in the relevant step for it to be available to a script.

But now it's trying to sign, we're getting another error:
```
Signing the extension for Firefox with Mozilla API, version 3.1commit-c21f72a
MOZILLA_API_SECRET is available
Building web extension from /home/runner/work/kiwix-js/kiwix-js/tmp

Error: Received bad response from the server while requesting https://addons.mozilla.org/api/v3/addons/kiwix-html5-unlisted%40kiwix.org/versions/3.1commitcCBfHCa/

status: 401
response: {"detail":"Unknown JWT iss (issuer)."}
```
There's a lot more detail in the latest build log https://github.com/kiwix/kiwix-js/runs/1542604400?check_suite_focus=true .

Any thoughts on what this is?

.github/workflows/nightly_or_release.yml
```diff
@@ -47,9 +54,17 @@ jobs:
...
+          MOZILLA_API_SECRET: ${{ secrets.MOZILLA_API_SECRET }}
+          MOZILLA_API_KEY: ${{ secrets.MOZILLA_API_KEY }}
         shell: bash
         # Switch -t indicates a tag release; add -d for dry run (for testing)
         run: |
+          if [ "x$MOZILLA_API_SECRET" = "x" ]; then
+            echo "WARNING: MOZILLA_API_SECRET is empty!"
+          else
+            echo "MOZILLA_API_SECRET is available"
+          fi
```
Jaifroid, line 67: Above lines should be removed before merge (testing only).

scripts/package_firefox_extension.sh
```diff
@@ -24,6 +24,11 @@ if [ "${TAG}zz" == "zz" ]; then
     if [ "${DRYRUN}zz" == "zz" ]; then
         # Sign the extension with the Mozilla API through web-ext, if we're not packaging a public version
         echo "Signing the extension for Firefox with Mozilla API, version $VERSION"
+        if [ "x$MOZILLA_API_SECRET" = "x" ]; then
+            echo "WARNING: MOZILLA_API_SECRET is empty!"
+        else
+            echo "MOZILLA_API_SECRET is available"
+        fi
```
Jaifroid, line 31: Above lines should be removed before merge.

mossroy:
Signing of the extension with local Chrome is now working. I did not manage to make it work with --headless, so I used a similar trick as with Travis : run xvfb to make it pretend there is a display
But I still don't understand the 401 error from Mozilla API : the API key and secrets seem to be there

mossroy:
Finally! The Mozilla credentials were simply wrong...
I took them from Mozilla website and updated them on GitHub secrets, and the API managed to sign the firefox addon

mossroy (approved):
I think this can be merged (after removing the debug lines on MOZILLA_*) to test a new release

Jaifroid:
OK, great, I'll do the cleanup, squash/merge, and then see if I can test a public release by deleting the current release and then recreating it. If that doesn't work, possibly I might need to delete the tag, but I don't think so because the trigger event is a release event, not a tag-creation event.

Jaifroid:
The build release ran, but the code to get the tag did not work, so it ran as a nightly release, although several previous errors seem to have crept back in. See https://github.com/kiwix/kiwix-js/runs/1542952446?check_suite_focus=true .

I then ran a standard "nightly" build, and it failed with several directory not found errors, very strange!
https://github.com/kiwix/kiwix-js/runs/1542952446?check_suite_focus=true 

Grrr. The second one could be a transitory error. Will retry. For the first I need to try another way to get the tag name.

Jaifroid:
Edit, second failed build is: https://github.com/kiwix/kiwix-js/runs/1542970929?check_suite_focus=true (gave wrong link earlier)

mossroy:
It seems to me that the version passed as a parameter to create_all_packages.sh is wrong : instead of a version number (the tag name), it's "refs/heads/master" which is not numeric : it's probably why the sed commands fail

mossroy:
It's probably because the source_tag step uses GITHUB_REF all the time, even when it's not on a tag. And it probably keeps the original git ref if it does not matches a tag.
This step should be fixed to feed SOURCE_TAG only when it's a release

Jaifroid:
OK, I have some test code with another way of getting the tag name, actually reverting to the original method but making it available in the appropriate step.

The tag name is supposed to be stored in `${{ github.event.release.tag_name }}` if the event has fired, and it should be empty if not. But previously I had used that in an earlier step, and our experiments showed that it has to be set in the same step as the code that will use it. I have added debugging code too.

This can only be tested on master, because to test it I need to delete the release and re-create it.

If this turns out to be a lot of debugging, then I guess I can always reset master to its current state to clean the commit history.

Jaifroid:
Ha! I've worked out what is happening by looking at the raw logs. Because the tag 3.1.0 already exists, when I create a new release based on it, the script checks out the ref of 3.1.0, instead of master, which is of course old code. I probably need to delete the tag in the remote repo, if that's possible.
```
2020-12-12T18:02:10.9799154Z Note: switching to 'refs/tags/3.1.0'.
2020-12-12T18:02:10.9801790Z 
2020-12-12T18:02:10.9804228Z You are in 'detached HEAD' state. You can look around, make experimental
2020-12-12T18:02:10.9806216Z changes and commit them, and you can discard any commits you make in this
2020-12-12T18:02:10.9808125Z state without impacting any branches by switching back to a branch.
2020-12-12T18:02:10.9809898Z 
2020-12-12T18:02:10.9811335Z If you want to create a new branch to retain commits you create, you may
2020-12-12T18:02:10.9813273Z do so (now or later) by using -c with the switch command. Example:
```

Jaifroid:
Yes, that was it!

image

Build log: https://github.com/kiwix/kiwix-js/runs/1543107017?check_suite_focus=true
```
Creating a release with TAG_NAME: 3.1.0
BASEDIR is /home/runner/work/kiwix-js/kiwix-js/scripts/..
Packaging version 3.1.0 because it has been passed as an argument
This version is a tag : we're releasing a public version
```

Jaifroid:
I *think* we have our release now, but will check tomorrow -- I have to go now. Wow, that was much more work than I anticipated, but I'm glad we've done it. I'll also clean up the debugging code tomorrow. We also need to test a nightly build.

mossroy:
Congratulations!
I'll do a few checks on the content of these released files.

mossroy:
The content of released files look good to me
A nightly build should run tonight : we'll see how it works

Jaifroid:
Nightly seems to have run perfectly well, and everything was signed without errors. The .crx files and other signed versions have been uploaded, so I think this is success! Thank you @mossroy for all your help and guidance with this.

I've cleaned up the debugging code in master. How should I proceed? All the commits associated with this switch to GitHub Actions are shown in screenshot below. Three alternatives:

* If you would like a single commit for this switch to GitHub Actions, I could reset master to [Changelog for v3.1](https://github.com/kiwix/kiwix-js/commit/21a3a3b218fee84a5198acfd6afadbe0ecd6e35e) and then commit the diff as a single commit "Implement continuous Integration using GitHub Actions (#683, #685, #686)". (Don't worry - I've done this many times before on KJSW!) You would need to resync your copy of master after that.
* Alternatively, we could keep three commits for each of those PRs (#683, #685, #686), and I could incorporate the last two commits into #686. This would be a less drastic rewrite of history.
* Or we can leave things as they are to keep a record of the somewhat bumpy history of making this switch.

image

Jaifroid:
I think we may need to remove the following lines from `.travis.yml`:
```
deploy:
    - provider: script
      # Nightly builds (launched by cron)
      script: sudo apt-get update -q && sudo apt-get --yes install click click-reviewers-tools && ./scripts/setup_travis_env.sh && DISPLAY=:99.0 ./scripts/create_all_packages.sh
      on:
          condition: ( "$TRAVIS_EVENT_TYPE" = "cron" )
    - provider: script
      # Generation of packages for public releases (launched by a tag)
      script: sudo apt-get update -q && sudo apt-get --yes install click click-reviewers-tools && ./scripts/setup_travis_env.sh && DISPLAY=:99.0 ./scripts/create_all_packages.sh -t -v ${TRAVIS_TAG}
      on:
          tags: true
```
If that is correct, I can include this as part of any cleanup above.

mossroy:
As with previous release (3.0), you almost did all the job by yourself @Jaifroid :-)
Regarding the git history, it's not very important because there are not that much commits. I'd vote for option 2 but feel free to switch to another one if necessary

About .travis.yml, yes you can remove this part, as it's now covered by Github Actions

Jaifroid:
OK, I've done option 2, including in the last commit the change to `travis.yml`. You now need to reset your master branch. I'll rebase any affected PRs.

PR #695, opened 2021-01-05
Document npm dependency for unit tests

Krinkle:
Follows-up https://github.com/kiwix/kiwix-js/pull/672.

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,6 +61,8 @@ The source code can be found at https://github.com/kiwix/kiwix-js
 
 Unit tests can be run by opening `tests/index.html` file in Firefox, Edge, or Chromium/Chrome.
 
+Before running the tests, a one-time set up is needed to fetch development dependencies from the npm registry. Run `npm ci --ignore-scripts` to fetch the same versions as we use in CI.
+
 ## Public releases and nightly builds
 
 The browser extensions are distributed through the stores of each vendor (see links above). But the packages are also saved in https://download.kiwix.org/release/browsers/ if necessary.
--- a/tests/karma.conf.local.js
+++ b/tests/karma.conf.local.js
@@ -4,7 +4,13 @@ module.exports = function (config) {
     // https://karma-runner.github.io/5.2/config/browsers.html
     browsers: [
       'FirefoxHeadless',
-      'ChromeHeadless'
+      // During local development, consider Chrome and Chromium to be similar enough
+      // and pick whichever the developer is most likely to have.
+      // In general, Linux distros provide and update Chromium only,
+      // whereas Windows and macOS users tend to have auto-updating Google Chrome.
+      //
+      // See package.json for commands to run tests in a single browser only.
+      process.platform === 'linux' ? 'ChromiumHeadless' : 'ChromeHeadless'
     ],
     frameworks: ['qunit'],
     client: {
--- a/package.json
+++ b/package.json
@@ -3,6 +3,9 @@
   "scripts": {
     "test": "npm run test-unit-local",
     "test-unit-local": "karma start tests/karma.conf.local.js",
+    "test-unit-firefox": "karma start tests/karma.conf.local.js --browsers FirefoxHeadless",
+    "test-unit-chromium": "karma start tests/karma.conf.local.js --browsers ChromiumHeadless",
+    "test-unit-chrome": "karma start tests/karma.conf.local.js --browsers ChromeHeadless",
     "test-unit-saucelabs": "karma start tests/karma.conf.saucelabs.js"
   },
   "engines": {


tests/karma.conf.local.js
```diff
@@ -4,7 +4,16 @@ module.exports = function (config) {
...
-      'ChromeHeadless'
+      // In general, Linux distros provides and updates Chromium only,
+      // whereas Windows and macOS users tend to have auto-updating Google Chrome.
+      // For developer convenience, treat these as the same as pick the are most likely
+      // to have. The exception for CHROME_BIN is to accomodate GitHub Actions (CI)
+      // which oddly enough runs Ubuntu but with google-chrome instead of chromium-browser.
+      // It exposes CHROME_BIN to indicate this.
+      // Ref https://github.com/actions/virtual-environments/issues/2388.
+      //
+      // Se package.json for sub commands that run tests in specific browser only.
```
Jaifroid, line 15: Typo: See

tests/karma.conf.local.js
```diff
@@ -4,7 +4,16 @@ module.exports = function (config) {
     // https://karma-runner.github.io/5.2/config/browsers.html
     browsers: [
       'FirefoxHeadless',
-      'ChromeHeadless'
+      // In general, Linux distros provides and updates Chromium only,
```
Jaifroid, line 7: Should read: "distros provide and update" (plural)

tests/karma.conf.local.js
```diff
@@ -4,7 +4,16 @@ module.exports = function (config) {
...
+      // whereas Windows and macOS users tend to have auto-updating Google Chrome.
+      // For developer convenience, treat these as the same as pick the are most likely
```
Jaifroid, line 9: Some typos have garbled the syntax here: "treat these as the same as pick the are" ??

README.md
```diff
@@ -61,6 +61,8 @@ The source code can be found at https://github.com/kiwix/kiwix-js
 
 Unit tests can be run by opening `tests/index.html` file in Firefox, Edge, or Chromium/Chrome.
 
+Before running the tests, a one-time set up is needed to fetch development dependencies from the npm registry. Run `npm ci --ignore-scripts` to fetch the same versions as we use in CI.
```
Jaifroid, line 64: I presume the dev also needs to run `npm install` before `npm ci --ignore-scripts`? If that's correct, then for the sake of newbie devs (and we do get quite a lot of undergraduate compsci students wanting to get PR experience), maybe add that to this instruction?

Jaifroid (approved):
Just some typos in comments to attend to.

Jaifroid:
Interestingly, this branch is failing with the same error in Firefox 84 as in my PR, so it seems the error may not be related to my PR. Any thoughts on what is going on here, @mossroy? Could it have anything to do with searching on an `id` that appears to be erroneously applied to two different elements in the ZIM article? But we should have seen this regularly before in that case.

Jaifroid:
@Krinkle This might be a separate PR, but as the version of Chromium that comes "pre-installed" on all Windows 10 now is Edge Chromium, it might make sense to add support for this. There is this npm package for Karma that looks to be kept regularaly up-to-date (last changed 4 days ago):

https://github.com/ChiragRupani/karma-chromiumedge-launcher 

Would it be easy to add? I'd be very happy to test.

Jaifroid:
Sorry for storm of comments, but following a suggestion from @mossroy, I've discovered that if I add:
```
if (process.platform === 'win32') process.env.CHROME_BIN =
  "C:\\Program Files (x86)\\Microsoft\\Edge\\Application\\msedge.exe";
```
At the top of `karma.conf.local.js` then the tests run fine on Windows 10 using the ChromeHeadless configuration. So maybe the plugin is overkill.

image

README.md
mossroy: No, it's better to keep only `npm ci` (provided that npm version is >=6). See https://medium.com/better-programming/npm-ci-vs-npm-install-which-should-you-use-in-your-node-js-projects-51e07cb71e26

mossroy (approved):
Apart from the typos, it's fine for me

Krinkle:
> @Krinkle This might be a separate PR, but as the version of Chromium that comes "pre-installed" on all Windows 10 now is Edge […]

Chrome, Chromium, Opera, and Edge are all different "browsers" for end-users. They indeed use the same source code for the rendering engine, but the versions of that engine, the compiler flags, feature flags, platform bindings, and higher-level APIs differ. They also differ in accessibility and developer tools.

I don't think they should be swapped out interchangeable in that way, and setting CHROME_BIN to Edge will likely result in failures at some point due to contract failure as they are not interchangeable. If they were, Karma would not consider them as separate "browsers".

Having said that, for the unspecified case of just running `npm test` locally, I propose we either 1) just run in Firefox only, or 2) run in Firefox + one Chromium-like browser that is most likely installed by the developer, which could be Edge on Windows, Chrome on macOS, and Chromium on Linux?

Jaifroid:
@Krinkle Yes, I only meant this for local tests. I completely agree about the browsers being different. The console will tell the Dev against which browser the local tests have run. So I would support your second model (2).

Jaifroid:
In the other PR, @mossroy has made the suggestion below about the test failure on Firefox we are currently seeing. @Krinkle is it feasible to downgrade nightwatch to 0.9.20 to see if it fixes the test failures on this branch? I can try it out also on the other PR, but not till the weekend.

> I don't think it comes from the ZIM content. Because we look for the exact same element earlier in the test scenario, and it's found by all browsers.
> I notice that the following error message appears in the log :
> 
> ```
> Error ECONNRESET: socket hang up
> ```
> 
> Could it be related to [nightwatchjs/nightwatch#1936](https://github.com/nightwatchjs/nightwatch/issues/1936) ?
> Maybe you could test to downgrade to version "=0.9.20" of nightwatch instead of "^1.5.1" (in package.json), to see if it helps

kelson42:
@Krinkle Any chance you could help to do the little left to be able to finally merge this PR?

Jaifroid:
Recent test, e.g. on #698 seem to be running correctly. Maybe this one can be re-run to see if the strange failures are fixed now?

README.md
Krinkle: Indeed, as @mossroy says.

`npm ci` is a stricter variant of `npm install` that fully adheres to the recursive package-lock file and prefers local caches, instead of freshly computing and calculating direct and indirect dependency version ranges.

It is an unfortunate turth that `npm ci` does what what one would reasonably expect `npm install` to do, and `npm install` does effectively what one would expect `npm update` to do.

The only difference to me is that `npm update` also saves the implied updates to `package.json`. In other words, when package.json specifies `foo: ^1.2.3`, npm install could install 1.5.5 which is compatible, whereas  `npm update` would install 1.5.5 and then also update package.json to `^1.5.5`.

Krinkle:
Fixed typos, confirmed re-run tests are passing.

Jaifroid:
Thank you @Krinkle for the further commits.

Do you have any idea why three of the the tests have failed again? Master is now also failing. 😒

Krinkle:
That is strange, they were passing a minute ago prior to merging. I'm investigating now.

Jaifroid:
Could it just be a connection issue to Sauce? I see some "socket hangup" errors.

Krinkle:
Yeah, that error goes away when clicking "Re-run jobs". But I do notice another issue as well.  Ref https://github.com/kiwix/kiwix-js/pull/699.

Krinkle:
@Jaifroid I figured out the socket issue as well. It's because the job is running twice at the same time (once as push, once as pull_request), and thus the Nightwatch tests fail when the concurrency limit is reached.

The secrets worked differently at Travis where naturally only one of them ran the sauce variant. I've reproduced the same behaviour in a fix as part of https://github.com/kiwix/kiwix-js/pull/699.

Issue #597, opened 2020-03-04
Support reading ZIM archives with no namespace


Jaifroid:
See comment here: https://github.com/kiwix/kiwix-js/issues/230#issuecomment-301271039 (and that issue more broadly) and the issue for this on Libzim: https://github.com/openzim/libzim/issues/15. If we haven't achieved #514 by the time this becomes a reality, we will need to do a fair amount of adjustment to the back end. It should not be difficult, just a bit tedious... :-)

Rajat379:
I am keen on working on this issue, should I work straight away?

Jaifroid:
@Rajat379 This is a rather specialized ssue: unless you are very experienced with JS development and with the format of ZIM archives (and the changes happening to that format) I suggest you should choose an easier issue to tackle as a first contribution.

Jaifroid:
@kelson42  wrote in #684:

> First test ZIM files have been created and are available at:
> http://tmp.kiwix.org/nons_zims/
> 
> Almost all Kiwix ports work with them so far, but not Kiwix-JS unfortunately.

Jaifroid:
@kelson42 What is the new ZIM format specification we should work with here? I've checked https://wiki.openzim.org/wiki/ZIM_file_format#Namespaces and it doesn't have any updated information. Looking at the Ray Charles sample you provided, it seems the landing page (at least) is in Namespace `-` which is why it is not showing in Kiwix JS: we have code that checks if the landing page is an article, and rejects it if it is not in namespace `A`. While I can change that to accept Namespace `-` as well, it would be useful to know whether `-` is fixed as part of a stable specification, and any other details of the new specification that are relevant for adapting the code.

kelson42:
@Jaifroid @mgautierfr Couldyou please talk together on thisad updatethe spec wherever is needed?

kelson42:
@Jaifroid The curret roadma looks like this:
* until end of december, polishing last details + doc
* mid of january, release of the new libzim
* mid of february, first scrapers are adapted
* march, first official ZIM releases without namespaces

Hope in the meantime we will be able to fix all related bugs in Kiwix JS and other ports.

Jaifroid:
Thanks, that's good to know. I don't anticipate big difficulties adapting the code. The main issue, from previous discussions, is likely to be slower binary search if we have to search through one huge namespace instead of one focused on articles. But we now have a block cache merged in 3.1 which should help with that.

mgautierfr:
[...]

PR #698, opened 2021-01-25
Support ZIM archives with no namespace

Jaifroid:
This is preliminary work on #597.

I've done some hacks to the code in jQuery mode (for now) as a proof-of-concept.

This supports reading and showing articles in the sample ZIM file `wikipedia_en_ray_charles_2018-09_nons.zim`. It is also backwardly compatible with older ZIM files. It supports searching for a title, with the proviso that results are not yet filtered for an article MIME type.

One confusing thing is that the landing article of this ZIM is in namespace '-' instead of 'W' or 'C'. This could simply be an earlier version of the spec.

Service Worker mode is not currently working.

### DIFF
--- a/www/js/app.js
+++ b/www/js/app.js
@@ -1624,6 +1624,9 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
                 $("#searchingArticles").hide();
                 alert("Error finding random article.");
             } else {
+                // We fall back to the old A namespace to support old ZIM files without a text/html MIME type for articles
+                // DEV: This will need to be changed if we search titlePtrList version 1
+                // in a future PR, as that list contains only articles
                 if (dirEntry.getMimetype() === 'text/html' || dirEntry.namespace === 'A') {
                     params.isLandingPage = false;
                     $('#activeContent').hide();
@@ -1646,7 +1649,9 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
                 $("#searchingArticles").hide();
                 $("#welcomeText").show();
             } else {
-                if (dirEntry.redirect || dirEntry.getMimetype() === 'text/html') {
+                // DEV: see comment above under goToRandomArticle()
+                if (dirEntry.redirect || dirEntry.getMimetype() === 'text/html' || 
+                dirEntry.namespace === 'A') {
                     params.isLandingPage = true;
                     readArticle(dirEntry);
                 } else {
--- a/www/js/app.js
+++ b/www/js/app.js
@@ -1614,7 +1614,7 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
                 $("#searchingArticles").hide();
                 alert("Error finding random article.");
             } else {
-                if (dirEntry.namespace === 'A') {
+                if (dirEntry.getMimetype() === 'text/html') {
                     params.isLandingPage = false;
                     $('#activeContent').hide();
                     $('#searchingArticles').show();
@@ -1636,7 +1636,7 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
                 $("#searchingArticles").hide();
                 $("#welcomeText").show();
             } else {
-                if (/[ACW-]/.test(dirEntry.namespace)) {
+                if (dirEntry.redirect || dirEntry.getMimetype() === 'text/html') {
                     params.isLandingPage = true;
                     readArticle(dirEntry);
                 } else {
--- a/www/js/app.js
+++ b/www/js/app.js
@@ -1224,7 +1224,7 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
     // URLs that begin 'http' (i.e. non-relative URLs). It then captures the whole of the URL up until either the opening delimiter
     // (" or ', which is capture group \3) or a querystring or hash character (? or #). When the regex is used below, it will be further
     // processed to calculate the ZIM URL from the relative path. This regex can cope with legitimate single quote marks (') in the URL.
-    var regexpTagsWithZimUrl = /(<(?:img|script|link|track)\b[^>]*?\s)(?:src|href)(\s*=\s*(["']))(?!http)((?:.(?!\3|\?|#))+.)/ig;
+    var regexpTagsWithZimUrl = /(<(?:img|script|link|track)\b[^>]*?\s)(?:src|href)(\s*=\s*(["']))(?!http)(.+?)(?=\3|\?|#)/ig;
     // Regex below tests the html of an article for active content [kiwix-js #466]
     // It inspects every <script> block in the html and matches in the following cases: 1) the script loads a UI application called app.js;
     // 2) the script block has inline content that does not contain "importScript()", "toggleOpenSection" or an "articleId" assignment
--- a/www/js/app.js
+++ b/www/js/app.js
@@ -1224,7 +1224,7 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
     // matches ZIM URLs with namespaces [-IJ] ('-' = metadata or 'I'/'J' = image). When the regex is used below, it will also
     // remove any relative or absolute path from ZIM-style URLs.
     // DEV: If you want to support more namespaces, add them to the END of the character set [-IJ] (not to the beginning) 
-    var regexpTagsWithZimUrl = /(<(?:img|script|link|track)\b[^>]*?\s)(?:src|href)(\s*=\s*["'])(?:\.\.\/|\/)+(?=[-IJ]\/)/ig;
+    var regexpTagsWithZimUrl = /(<(?:img|script|link|track)\b[^>]*?\s)(?:src|href)(\s*=\s*["'])(?!http)(?:\.\.\/|\/)*([^"']+)/ig;
     // Regex below tests the html of an article for active content [kiwix-js #466]
     // It inspects every <script> block in the html and matches in the following cases: 1) the script loads a UI application called app.js;
     // 2) the script block has inline content that does not contain "importScript()", "toggleOpenSection" or an "articleId" assignment
@@ -1255,7 +1255,11 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
 
         // Replaces ZIM-style URLs of img, script, link and media tags with a data-kiwixurl to prevent 404 errors [kiwix-js #272 #376]
         // This replacement also processes the URL to remove the path so that the URL is ready for subsequent jQuery functions
-        htmlArticle = htmlArticle.replace(regexpTagsWithZimUrl, '$1data-kiwixurl$2');
+        htmlArticle = htmlArticle.replace(regexpTagsWithZimUrl, function(m0, m1, m2, url) {
+            // If it's a new ZIM URL without namespace, add the C namespace (=Content)
+            url = /^[-IJ]/.test(url) ? url : 'C/' + url;
+            return m1 + 'data-kiwixurl' + m2 + url;
+        });
 
         // Extract any css classes from the html tag (they will be stripped when injected in iframe with .innerHTML)
         var htmlCSS = htmlArticle.match(/<html[^>]*class\s*=\s*["']\s*([^"']+)/i);
@@ -1632,7 +1636,7 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
                 $("#searchingArticles").hide();
                 $("#welcomeText").show();
             } else {
-                if (dirEntry.namespace === 'A') {
+                if (/[AC-]/.test(dirEntry.namespace)) {
                     params.isLandingPage = true;
                     readArticle(dirEntry);
                 } else {
--- a/www/js/app.js
+++ b/www/js/app.js
@@ -1266,7 +1266,7 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
                 console.log('ZIM URL: ' + dirEntry.namespace + '/' + dirEntry.url);
                 console.log('Full URL: ' + relAssetUrl);
                 console.log('Base URL: ' + baseUrl);
-                console.log('Calculated asset URL: ' + decAssetZIMUrl);
+                console.log('Calculated asset URL: ' + assetZIMUrl);
             **/
             // DEV: Note that deriveZimUrlFromRelativeUrl produces a *decoded* URL (and incidentally would remove any URI component
             // if we had captured it). We therefore re-encode the URI with encodeURI (which does not encode forward slashes) instead
--- a/www/js/app.js
+++ b/www/js/app.js
@@ -1261,16 +1261,9 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
         // with the correct namespace (this works for old-style -,I,J namespaces and for new-style C namespace)
         htmlArticle = htmlArticle.replace(regexpTagsWithZimUrl, function(match, blockStart, equals, quote, relAssetUrl) {
             var assetZIMUrl = uiUtil.deriveZimUrlFromRelativeUrl(relAssetUrl, baseUrl);
-            // Uncomment logging below to test the calculation of relative URLs if you are having issues 
-            /**
-                console.log('ZIM URL: ' + dirEntry.namespace + '/' + dirEntry.url);
-                console.log('Full URL: ' + relAssetUrl);
-                console.log('Base URL: ' + baseUrl);
-                console.log('Calculated asset URL: ' + assetZIMUrl);
-            **/
             // DEV: Note that deriveZimUrlFromRelativeUrl produces a *decoded* URL (and incidentally would remove any URI component
             // if we had captured it). We therefore re-encode the URI with encodeURI (which does not encode forward slashes) instead
-            // of encodeURIComponent. 
+            // of encodeURIComponent.
             return blockStart + 'data-kiwixurl' + equals + encodeURI(assetZIMUrl);
         });
 
--- a/www/js/app.js
+++ b/www/js/app.js
@@ -1624,7 +1624,7 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
                 $("#searchingArticles").hide();
                 alert("Error finding random article.");
             } else {
-                if (dirEntry.getMimetype() === 'text/html') {
+                if (dirEntry.getMimetype() === 'text/html' || dirEntry.namespace === 'A') {
                     params.isLandingPage = false;
                     $('#activeContent').hide();
                     $('#searchingArticles').show();
--- a/www/js/app.js
+++ b/www/js/app.js
@@ -1268,8 +1268,9 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
                 console.log('Base URL: ' + baseUrl);
                 console.log('Calculated asset URL: ' + decAssetZIMUrl);
             **/
-            // DEV: Note that deriveZimUrlFromRelativeUrl produces a *decoded* URL (and removes any URI component)
-            // so we re-encode it with encodeURI (this does not encode forward slashes) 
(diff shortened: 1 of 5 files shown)


Jaifroid:
Service Worker mode is now supported, but has only been tested on the Ray Charles sample ZIM.

www/js/app.js
```diff
@@ -1255,7 +1254,12 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
 
         // Replaces ZIM-style URLs of img, script, link and media tags with a data-kiwixurl to prevent 404 errors [kiwix-js #272 #376]
         // This replacement also processes the URL to remove the path so that the URL is ready for subsequent jQuery functions
-        htmlArticle = htmlArticle.replace(regexpTagsWithZimUrl, '$1data-kiwixurl$2');
+        // If we are in a new-style ZIM archive with a single namespace for user content, then that namespace is added to the URL
+        htmlArticle = htmlArticle.replace(regexpTagsWithZimUrl, function(m0, m1, m2, url) {
+            // If it's a new ZIM URL without namespace, add the C namespace (=Content)
+            url = /^[-IJ]/.test(url) ? url : 'C/' + url;
```
Jaifroid, line 1260: Do we need to calculate the relative path to the asset before adding the `C` namespace? For hyperlinks, we do ned to calculate relative paths, but for assets, it depends whether they can now exisst in subdirectories of an article. Until recently, all assets would be in a separate namespace `-`, `I` or `J`, therefore the relative path to that namespace was irrelevant and could be discarded. Without knowing the structure of the ZIM, i.e. whether it can place assets such as stylesheets as subdirectories of articles, we don't know whether we need to do a relative path calculation here. @kelson42 do you have any idea about this? (NB in Service Worker mode this will "just work", but not in jQuery mode.)

Jaifroid: For example, if we have an article on China, let's say `../China`, could we potentially have assets appearing like `../China/the_great_wall.webp`? Or with the Wikivoyage `../Singapore/Chinatown`, could an asset appear as `Tooth_Relic_Temple.png.webp`, so we have to calculate that the URL to the asset is `C/Singapore/Chinatown/Tooth_Relic_Temple.png.webp`?

Jaifroid: I've now written some "universal" code that calculates ZIM asset URLs relative to the page's ZIM URL. This appears to work indifferently on old-style URLs with namespace and on new-style URLs without namespace. (So long as the relative URLs are written correctly - see  https://github.com/kiwix/kiwix-js/issues/597#issuecomment-767828723 for a (no-namespace) case where the URLs appear to be have written incorrectly.

www/js/app.js
```diff
@@ -1253,9 +1252,21 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
...
-        htmlArticle = htmlArticle.replace(regexpTagsWithZimUrl, '$1data-kiwixurl$2');
+        // This replacement also processes the URL relative to the page's ZIM URL so that we can find the ZIM URL of the asset
+        // with the correct namespace (this works for old-style -,I,J namespaces and for new-style C namespace)
+        htmlArticle = htmlArticle.replace(regexpTagsWithZimUrl, function(m0, m1, m2, fullUrl) {
+            var url = uiUtil.deriveZimUrlFromRelativeUrl(fullUrl, baseUrl);
+            // Uncomment logging below to test the calculation of relative URLs if you are having issues 
+            /** console.log('ZIM URL: ' + dirEntry.namespace + '/' + dirEntry.url);
+            console.log('Full URL: ' + fullUrl);
+            console.log('Base URL: ' + baseUrl);
+            console.log('Calculated URL: ' + url); **/
```
Jaifroid, line 1267: Remove above console logging before merge.

www/js/app.js
```diff
@@ -1220,11 +1220,10 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
     var regexpZIMUrlWithNamespace = /^[./]*([-ABIJMUVWX]\/.+)$/;
     // Regex below finds images, scripts, stylesheets and tracks with ZIM-type metadata and image namespaces [kiwix-js #378]
     // It first searches for <img, <script, <link, etc., then scans forward to find, on a word boundary, either src=["']
-    // or href=["'] (ignoring any extra whitespace), and it then tests the path of the URL with a non-capturing lookahead that
-    // matches ZIM URLs with namespaces [-IJ] ('-' = metadata or 'I'/'J' = image). When the regex is used below, it will also
-    // remove any relative or absolute path from ZIM-style URLs.
-    // DEV: If you want to support more namespaces, add them to the END of the character set [-IJ] (not to the beginning) 
-    var regexpTagsWithZimUrl = /(<(?:img|script|link|track)\b[^>]*?\s)(?:src|href)(\s*=\s*["'])(?:\.\.\/|\/)+(?=[-IJ]\/)/ig;
+    // or href=["'] (ignoring any extra whitespace), and it then tests the path of the URL with a non-capturing megative lookahead
```
Jaifroid, line 1223: typo: negative

www/js/app.js
```diff
@@ -1253,9 +1252,21 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
...
-        htmlArticle = htmlArticle.replace(regexpTagsWithZimUrl, '$1data-kiwixurl$2');
+        // This replacement also processes the URL relative to the page's ZIM URL so that we can find the ZIM URL of the asset
+        // with the correct namespace (this works for old-style -,I,J namespaces and for new-style C namespace)
+        htmlArticle = htmlArticle.replace(regexpTagsWithZimUrl, function(m0, m1, m2, relAssetUrl) {
+            var decAssetZIMUrl = uiUtil.deriveZimUrlFromRelativeUrl(relAssetUrl, baseUrl);
+            // Uncomment logging below to test the calculation of relative URLs if you are having issues 
+            /** console.log('ZIM URL: ' + dirEntry.namespace + '/' + dirEntry.url);
+            console.log('Full URL: ' + relAssetUrl);
+            console.log('Base URL: ' + baseUrl);
+            console.log('Calculated asset URL: ' + decAssetZIMUrl); **/
```
Jaifroid, line 1267: Remove console logging before merge.

www/js/lib/zimArchive.js
```diff
@@ -237,7 +238,7 @@ define(['zimfile', 'zimDirEntry', 'util', 'utf8'],
                 return that._file.dirEntryByTitleIndex(index).then(function(dirEntry) {
                     var title = dirEntry.getTitleOrUrl();
                     // Only return dirEntries with titles that actually begin with prefix
-                    if (dirEntry.namespace === 'A' && title.indexOf(prefix) === 0) {
+                    if (/[AC]/.test(dirEntry.namespace) && title.indexOf(prefix) === 0) {
```
Jaifroid, line 241: Should the namespace test be replaced with a MIME type test? Probably not, becuase it involves a lookup and for old-style ZIMs the `A` namespace test is sufficient and much faster. And for new-style ZIMs, we should use the v2 `titlePtrList`, which only contains articles, so a test is redundant.

Jaifroid:
Code tested on Edge Classic and IE11, as well as Edge Chromium, in both jQuery and SW mode (for browsers that support it). I thought the URI calculation of asset ZIM URIsTh would be slow on IE11, but actually it seems as fast as before.

This PR completes the first of the two PRs I suggested in https://github.com/kiwix/kiwix-js/issues/597#issuecomment-767006966. What remains to be done, with a separate PR, is to switch to using the v1 `titlePtrList` if it is available at `X/listing/titleOrdered/v1`. This will allow us to display only article titles in the title search.

Jaifroid:
@mossroy Would you rather we incorporate the use of the dedicated article list (`X/listing/titleOrdered/v1`) for searches, instead of the `titlePtrList`, into this PR, or would you rather review this (smallish) PR first? I was thinking it might be easier to follow the changes if we do it in two stages, test and merge this one and then work on the second part.

mossroy:
It would be better with 2 PRs, if possible. I should find some time to review on Friday
I would suggest to add some unit tests on this new version of the format, with a sample ZIM file

Jaifroid:
That's fine, thanks!

I agree with adding tests, but necessarily it must be a new ZIM, so we'd be testing an old and a new ZIM. I think this is quite a lot of work and can't be completed until these conditions are met:

1. A stable new ZIM produced by the final version of the ZIM writer, not by a script that converts an old ZIM to new format for test purposes;
2. Completion of the second PR, so that we can test searching for an article in the new-format ZIM using the new API.

In sum, I think adding tests that proprerly test the new-format ZIMs might be a third PR?

Jaifroid:
I have revised the code to deal with media sources (video/audio) in the new format (jQuery mode), but I have to stress this is untested and a little speculative, until I can find a video in one of these ZIMs. I noted we have a double decodeUriComponent happening in the media decode block, which I've corrected. It hasn't caused us issues up till now, but evidently could in theory in some languages.

mossroy:
> In sum, I think adding tests that proprerly test the new-format ZIMs might be a third PR?

All right. Hope we'll have one suitable ZIM file soon.
It's true that it's a bit of work, but it would be a lot safer to have both formats automatically tested on our supported browsers (now and with all future commits)

www/js/app.js
```diff
@@ -1253,9 +1252,25 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
...
+            // Uncomment logging below to test the calculation of relative URLs if you are having issues 
+            /**
+                console.log('ZIM URL: ' + dirEntry.namespace + '/' + dirEntry.url);
+                console.log('Full URL: ' + relAssetUrl);
+                console.log('Base URL: ' + baseUrl);
+                console.log('Calculated asset URL: ' + decAssetZIMUrl);
+            **/
+            // DEV: Note that deriveZimUrlFromRelativeUrl produces a *decoded* URL (and removes any URI component)
+            // so we re-encode it with encodeURI (this does not encode forward slashes) 
+            return m1 + 'data-kiwixurl' + m2 + encodeURI(assetZIMUrl);
```
Jaifroid, line 1272: Query: As noted in the comment, `deriveZimUrlFromRelativeUrl` has the side effect of discarding any URI component. Is this an issue for assets? We are only here processing links for tags `img|script|link|track` in jQuery mode. We don't support script in any case (where query string could be important). I think it's not relevant for css. It would only be relevant for images if some processing were being done on the image, which we don't do in jQuery mode. We could save the query string in this function and add it back here, but is it necessary?

Jaifroid:
I haven't yet found any video, but there is audio on the Ray Charles article "Side by Side (1927 song)". This plays in jQuery mode, but not in SW mode (at least in Chromium). However, I think the problem with SW mode is that some of the JavaScript (and some of the CSS) is missing from this converted ZIM (there are errors in console log, but I'm pretty sure these errors are missing assets in the ZIM, not errors in constructing the ZIM URL).

www/js/app.js
Jaifroid: Actually, I've now resolved this simply by not capturing the query string or hash part of the URL before processing and adding back into the document. This is a better solution.

mossroy:
I probably found a regression : if you open foo-zstd.zim (the one inside test directory) and start a random search among articles. It seems to trigger an infinite loop (at least on Chromium, both in jQuery and SW modes).

Jaifroid:
> I probably found a regression : if you open foo-zstd.zim (the one inside test directory) and start a random search among articles. It seems to trigger an infinite loop (at least on Chromium, both in jQuery and SW modes).

Oh dear! I hadn't seen that with other test ZIMs, but I'll check it out.

Jaifroid:
I confirm the regression. I guess the reason is because the articles in foo-zstd.zim do not appear to have a MIME type of text/html, and I have introduced a test to ensure that only articles are displayed by the random function.

I guess this ZIM is a pretty extreme case in having that MIME type missing, but on the other hand infinite loops are always bad. I'll re-introduce the A namespace test as a fallback.

Jaifroid:
I've re-introduced the A namespace fallback. However, I could modify the test (in the random article function) to accept 'text/plain' as well as 'text/html'. I'm just slightly wary about not knowing what future MIME types a ZIM could hold as "articles". For old ZIMs, clearly that 'A' namespace test is definitive. For newer ZIMs going forwards, we may only be searching the v1 titlePtrList in any case, which shold only have articles in it.

Do you have any advice, @mossroy , on the best way to deal with this generically? The code in question is [here](https://github.com/kiwix/kiwix-js/pull/698/files#diff-484c9e8aa09e09b8eb1fafaf55e8eadcd0eaab230e38d02bb9440841bfa090a4R1627).

www/js/app.js
```diff
@@ -1632,7 +1646,7 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
                 $("#searchingArticles").hide();
                 $("#welcomeText").show();
             } else {
-                if (dirEntry.namespace === 'A') {
+                if (dirEntry.redirect || dirEntry.getMimetype() === 'text/html') {
```
Jaifroid, line 1649: @mossroy  Here also I could add a fallback `|| dirEntry.namespace === 'A'`.

www/js/app.js
```diff
@@ -1610,7 +1624,7 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
                 $("#searchingArticles").hide();
                 alert("Error finding random article.");
             } else {
-                if (dirEntry.namespace === 'A') {
+                if (dirEntry.getMimetype() === 'text/html' || dirEntry.namespace === 'A') {
```
Jaifroid, line 1627: Alternative here would be to test only `^text`, i.e.

`if (/^text\b/i.test(dirEntry.getMimetype() || dirEntry.namespace === 'A')`.

mossroy:
My first reflex would be to keep the fallback on namespace 'A', to be sure we don't break compatibility with strange old ZIM files (a comment would be welcome to explain that above).
But, in fact, we should do the same as in libzim, whatever it is. If they have the same issue with foo-zstd.zim, we should report it on libzim to discuss it

Jaifroid:
> But, in fact, we should do the same as in libzim, whatever it is. If they have the same issue with foo-zstd.zim, we should report it on libzim to discuss it

OK, will check.

mossroy (commented):
Except on foo-zstd.zim file, I did not see any regression after quick tests on Chromium (jQuery and SW modes) and Firefox OS device

www/js/lib/zimArchive.js
```diff
@@ -223,9 +223,10 @@ define(['zimfile', 'zimDirEntry', 'util', 'utf8'],
         util.binarySearch(0, this._file.articleCount, function(i) {
             return that._file.dirEntryByTitleIndex(i).then(function(dirEntry) {
                 if (search.status === 'cancelled') return 0;
-                if (dirEntry.namespace < 'A') return 1;
-                if (dirEntry.namespace > 'A') return -1;
-                // We should now be in namespace A
+                var ns = dirEntry.namespace;
+                if (ns < 'A') return 1;
+                if (ns === 'B' || ns > 'C') return -1;
```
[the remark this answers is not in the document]
Jaifroid: I compared the two specs. The old-format spec for namespaces is here: https://wiki.openzim.org/wiki/ZIM_file_format_old_namespace#Namespaces , and the new one is here: https://wiki.openzim.org/wiki/ZIM_file_format#Namespaces .

The new format can only have namespaces C,M,W,X. The old format has namespaces -,A,B,I,J,M,U,V,W,X. I guess C was chosen precisely because there is no ambiguity with earlier content namespaces.

I'll check libzim if I can find the corresponding code (I'm struggling to find the random function anywhere in libzim!). But this is the most economical logical test that works universally for both ZIM types. If a future spec is introduced, we'll need to change this code anyway.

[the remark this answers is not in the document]
Jaifroid: PS I will certainly add a comment.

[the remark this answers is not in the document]
Jaifroid: @mossroy My lack of cpp knowledge is hampering me. I found a long explanation here:

https://github.com/openzim/libzim/blob/master/src/archive.cpp#L217 

And there is a test function `if (m_impl->hasNewNamespaceScheme())` being called here (and in lots of places in this script):

https://github.com/openzim/libzim/blob/master/src/archive.cpp#L217

But I can't find any function actually corresponding to that.

And `search.cpp` seems to be full of Xapian-specific code.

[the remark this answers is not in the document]
mossroy: They seem to read the "minor version" in the header of the ZIM file to know if the ZIM file is using the new namespace sheme or not. See https://github.com/openzim/libzim/blob/93464a63019619fa9c9ef71401d2491df75bd189/src/fileimpl.cpp#L200 and https://openzim.org/wiki/ZIM_file_format#Header

It looks like a good and reliable idea. It should be easy to add that in our zimfile.js, so that we can use this information to know if we have to use the old way or new way to parse the ZIM file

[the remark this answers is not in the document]
Jaifroid: OK I think I can implement that. Should be much easier than adding the MIME type table was.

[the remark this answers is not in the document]
Jaifroid: Done.

www/js/app.js
Jaifroid: Done

www/js/app.js
Jaifroid: This doesn't work because it catches text/css as well. We should keep the fallback and revise when we suppport titlePtrList v1.

www/js/app.js
```diff
@@ -1253,9 +1253,26 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
...
 
+        // Calculate the current article's ZIM baseUrl to use when processing relative links
+        var baseUrl = dirEntry.namespace + '/' + dirEntry.url.replace(/[^/]+$/, '');
+
         // Replaces ZIM-style URLs of img, script, link and media tags with a data-kiwixurl to prevent 404 errors [kiwix-js #272 #376]
-        // This replacement also processes the URL to remove the path so that the URL is ready for subsequent jQuery functions
-        htmlArticle = htmlArticle.replace(regexpTagsWithZimUrl, '$1data-kiwixurl$2');
+        // This replacement also processes the URL relative to the page's ZIM URL so that we can find the ZIM URL of the asset
+        // with the correct namespace (this works for old-style -,I,J namespaces and for new-style C namespace)
+        htmlArticle = htmlArticle.replace(regexpTagsWithZimUrl, function(m0, m1, m2, m3, relAssetUrl) {
```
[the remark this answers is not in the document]
Jaifroid: It's a bit difficult because they map onto non-discrete units of the tag, but I've had a go at providing more meaningful parameter names. Note that some are throwaway (e.g. the third capture group is only used internally within the regex, not in the function).

www/js/lib/zimArchive.js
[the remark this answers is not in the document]
Jaifroid: Test replaced with actual namespace determined from minorVersion.

www/js/app.js
```diff
@@ -1253,9 +1253,26 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
...
+        // with the correct namespace (this works for old-style -,I,J namespaces and for new-style C namespace)
+        htmlArticle = htmlArticle.replace(regexpTagsWithZimUrl, function(match, blockStart, equals, quote, relAssetUrl) {
+            var assetZIMUrl = uiUtil.deriveZimUrlFromRelativeUrl(relAssetUrl, baseUrl);
+            // Uncomment logging below to test the calculation of relative URLs if you are having issues 
+            /**
+                console.log('ZIM URL: ' + dirEntry.namespace + '/' + dirEntry.url);
+                console.log('Full URL: ' + relAssetUrl);
+                console.log('Base URL: ' + baseUrl);
+                console.log('Calculated asset URL: ' + assetZIMUrl);
+            **/
```
Jaifroid, line 1270: Reminder to self, remove above testing code before merge.

Jaifroid:
@mossroy I think this first-stage PR is ready for final review. There are some tests to determine whether a dirEntry represents an article or not that could have been coded more thoroughly (I currenlty only test for `text/html` or fallback to namespace A), but considering that the MIME type test should become redundant with the next PR (where we should be doing title searches on the Version 1 `titlePtrList` which only contains articles), it doesn't seem worth encoding more elaborately.

mossroy (approved):
Looks good. It's much cleaner and safer to use the metadata like this.
I quickly tested again on my Firefox OS device and saw no regression.
The bug on foo-zstd.zim is solved

Jaifroid:
Tested and working on IE11 and Edge Legacy (already tested on Firefox and Edge Chromium).

PR #699, opened 2021-01-27
tests: Use Chrome for 'npm test' in GitHub CI

Krinkle:
Follows-up 55186c8. I had this check originally, but then removed it after GitHub merged the upstream improvement.
It looks like some CI runners might not yet have it applied, so keep it a little longer for now.


Krinkle:
Bolding landing so as to unbreak CI. Feel free to revert if it causes issues :)

Jaifroid:
Thank you, @Krinkle!

Issue #708, opened 2021-02-27
Support version 1 of titlePtrList accessed from X/listing/titleOrdered/v1


Jaifroid:
Having completed #698 I'm looking into how to support the new `titePtrList`. I note that libzim has implemented a `getTitleAccessor()` function:

https://github.com/openzim/libzim/blob/master/src/fileimpl.cpp#L158

There are two versions of the `titlePtrList`, one of which is `X/listing/titleOrdered/v0`, which is the traditional one that is also accessed via the `titlePtrPos` byte contained in the [ZIM file header](https://wiki.openzim.org/wiki/ZIM_file_format#Header). The other is `X/listing/titleOrdered/v1`, which only contains article titles, and so this is the one we need to use for title search and random articles.

I'm having trouble getting my head around how we can "bootstrap" the app to the point where it can read a title from the X namespace (in order to retireve these new `titlePtrList`s without first accessing the traditional `titlePtrPos` that is in the archive header, but which is now considered obsolete, even though it points to the same offset represented by `X/listing/titleOrdered/v0`.

Can I use the `urlPtrPos` to binary search the `urlPtrList` instead? I can't find code corresponding to this in libZim (but the fact that I can't find it means nothing: I don't know C++ well enough).

Jaifroid:
@mgautierfr Is it safe for our code to make the following assumptions in order to find `X/listing/titleOrdered/v0` and `X/listing/titleOrdered/v1` while loading a new archive, without using the obsolete `titlePtrPos` contained in the archive header? -

* There will never be a namespace below X in a ZIM (in practical terms, assuming ASCII, we will not have a namespace Z)
* Therefore the dirEntries for `X/listing/titleOrdered/...` will always be the last few dirEntries (currently the last three, if there is a Xapian index)
* It would be safe to iterate through the last 10 (say) entries in `urlPtrList` to find the two entries for `X/listing/titleOrdered/...`

If the above assumptions are not safe ones to make, how do you recommend we find these listings without using `titlePtrPos`? Must we initiate a binary search through the `urlPtrList`?

[...]

Issue #552, opened 2019-08-18
Incorrect rendering of landing page of recent ZIMs in jQuery mode


Jaifroid:
At least in `wikipedia_en_history_maxi_2019-08.zim`, the landing page in jQuery mode looks like the top image (and no further images are displayed), whereas it should look something like the bottom image (and have many more images as you scroll down). It works fine in SW mode. So it's a stylesheet injection issue, I think. Tested in Edge 44 and Edge 78 (Chromium).

![image](https://user-images.githubusercontent.com/4304337/63226868-c80dfc80-c1d7-11e9-9e55-b1a501b7b709.png)

![image](https://user-images.githubusercontent.com/4304337/63226844-78c7cc00-c1d7-11e9-8b81-8dbaf457a85d.png)

kelson42:
I confirm the bug is still there.

kelson42:
@Jaifroid Considering that no new dev is foreseen fornthe JQuery mode, would that not be more appropriate to close this ticket?

Jaifroid:
Agreed, and fixing this would require intervening in the HTML of the landing page at reader level, which is against the philosophy of this Repo.

PR #709, opened 2021-03-04
Support new v1 directory listings

Jaifroid:
This is a draft of support for the new directory listings, corresponding to issue #708.

In line with a suggestion from @mgautierfr , I've implemented it in an extensible way: a developer can add further (future) listings in the `ZIMArchive` function of `zimArchive.js` by adding more known listings to the array of Listing objects. Obviously the only two listed there are currently as in code block below. This is very much just-in-case, as there are no plans to extend the listings.

Just note that, for testing with the ZIMs we have a available, the code is currently rigged to fill out the `articlePtrPos` and `articleCount` metadata from `X/listing/titleOrdered/v0` because we don't have a ZIM (at least not the ones I've tested) with a non-empty `X/listing/titleOrdered/v1`. The rigged code will be changed (a one-letter change) once we have a valid ZIM for testing. It is clearly commented where I've rigged it in `zimArchive.js`.

I've only done preliminary testing on two `nons` ZIMs (Ray Charles and Basketball), and on one legacy ZIM (Ray Charles, plus it is passing push tests with a split Ray Charles).

```
[{
    path: 'X/listing/titleOrdered/v0',
    ptrName: 'titlePtrPos',
    countName: 'entryCount'
},
{
    path: 'X/listing/titleOrdered/v1',
    ptrName: 'articlePtrPos',
    countName: 'articleCount'
}]
```

### DIFF
--- a/www/js/lib/zimArchive.js
+++ b/www/js/lib/zimArchive.js
@@ -54,9 +54,30 @@ define(['zimfile', 'zimDirEntry', 'util', 'utf8'],
         var that = this;
         that._file = null;
         that._language = ""; //@TODO
-        var createZimfile = function(fileArray) {
-            zimfile.fromFileArray(fileArray).then(function(file) {
+        var createZimfile = function (fileArray) {
+            zimfile.fromFileArray(fileArray).then(function (file) {
                 that._file = file;
+                // File has been created, but we need to add any listings which extend the archive metadata
+                that._file.setListings([
+                    // Provide here any listings for which we need to extract metadata as key:value obects to be added to the file
+                    // 'ptrName' and 'countName' contain the key names to be set in the archive file object
+                    {
+                        path: 'X/listing/titleOrdered/v0',
+                        ptrName: 'titlePtrPos',
+                        countName: 'entryCount'
+                    },
+                    {
+                        // !!! CHANGE THIS PATH TO .../v1 BEFORE MERGING !!!
+                        // It is set to v0 for testing because we do not have any archive with a valid v1 index yet
+                        path: 'X/listing/titleOrdered/v0',
+                        ptrName: 'articlePtrPos',
+                        countName: 'articleCount'
+                    }
+                ]);
+                // DEV: Currently, extended listings are only used for title (=article) listings when the user searches
+                // for an article or uses the Random button, by which time the listings will have been extracted.
+                // If, in the future, listings are used in a more time-critical manner, consider forcing a wait before
+                // declaring the archive to be ready, by chaining the following callback in a .then() function of setListings.
                 callbackReady(that);
             });
         };
@@ -245,7 +266,9 @@ define(['zimfile', 'zimDirEntry', 'util', 'utf8'],
     ZIMArchive.prototype.findDirEntriesWithPrefixCaseSensitive = function(prefix, resultSize, search, callback) {
         var that = this;
         var cns = this.getContentNamespace();
-        util.binarySearch(0, this._file.articleCount, function(i) {
+        // Search v1 article listing if available, otherwise fallback to v0
+        var articleCount = this._file.articleCount || this._file.entryCount;
+        util.binarySearch(0, articleCount, function(i) {
             return that._file.dirEntryByTitleIndex(i).then(function(dirEntry) {
                 if (search.status === 'cancelled') return 0;
                 var ns = dirEntry.namespace;
@@ -328,7 +351,7 @@ define(['zimfile', 'zimDirEntry', 'util', 'utf8'],
      */
     ZIMArchive.prototype.getDirEntryByTitle = function(title) {
         var that = this;
-        return util.binarySearch(0, this._file.articleCount, function(i) {
+        return util.binarySearch(0, this._file.entryCount, function(i) {
             return that._file.dirEntryByUrlIndex(i).then(function(dirEntry) {
                 var url = dirEntry.namespace + "/" + dirEntry.url;
                 if (title < url)
--- a/www/js/lib/zimArchive.js
+++ b/www/js/lib/zimArchive.js
@@ -280,7 +280,7 @@ define(['zimfile', 'zimDirEntry', 'util', 'utf8'],
         }, true).then(function(firstIndex) {
             var dirEntries = [];
             var addDirEntries = function(index) {
-                if (search.status === 'cancelled' || index >= firstIndex + resultSize || index >= that._file.articleCount) {
+                if (search.status === 'cancelled' || index >= firstIndex + resultSize || index >= articleCount) {
                     return dirEntries;
                 }
                 return that._file.dirEntryByTitleIndex(index).then(function(dirEntry) {
@@ -374,7 +374,8 @@ define(['zimfile', 'zimDirEntry', 'util', 'utf8'],
      * @param {callbackDirEntry} callback
      */
     ZIMArchive.prototype.getRandomDirEntry = function(callback) {
-        var index = Math.floor(Math.random() * this._file.articleCount);
+        var articleCount = this._file.articleCount || this._file.entryCount;
+        var index = Math.floor(Math.random() * articleCount);
         this._file.dirEntryByUrlIndex(index).then(callback);
     };
     
--- a/www/js/lib/zimfile.js
+++ b/www/js/lib/zimfile.js
@@ -54,14 +54,17 @@ define(['xzdec_wrapper', 'zstddec_wrapper', 'util', 'utf8', 'q', 'zimDirEntry', 
      * @property {Array<File>} _files Array of ZIM files
      * @property {String} name Abstract archive name for file set
      * @property {Integer} id Arbitrary numeric ZIM id used to track the currently loaded archive
-     * @property {Integer} articleCount Total number of articles
+     * @property {Integer} entryCount Total number of entries in the URL pointerlist
+     * @property {Integer} articleCount Total number of articles in the v1 article-only pointerlist (async calculated entry)
      * @property {Integer} clusterCount Total number of clusters
      * @property {Integer} urlPtrPos Position of the directory pointerlist ordered by URL
-     * @property {Integer} titlePtrPos Position of the directory pointerlist ordered by title
+     * @property {Integer} titlePtrPos Position of the legacy v0 pointerlist ordered by title
+     * @property {Integer} articlePtrPos Position of the v1 article-only pointerlist ordered by title (async calculated entry)
      * @property {Integer} clusterPtrPos Position of the cluster pointer list
      * @property {Integer} mimeListPos Position of the MIME type list (also header size)
      * @property {Integer} mainPage Main page or 0xffffffff if no main page
      * @property {Integer} layoutPage Layout page or 0xffffffffff if no layout page
+     * @property {Map} mimeTypes The ZIM file's MIME type table rendered as a Map (calculated entry)
      */
     
     /**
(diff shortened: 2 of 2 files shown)


www/js/lib/zimArchive.js
```diff
@@ -328,7 +351,7 @@ define(['zimfile', 'zimDirEntry', 'util', 'utf8'],
      */
     ZIMArchive.prototype.getDirEntryByTitle = function(title) {
```
Jaifroid, line 352: I think this funciton is now misnamed. There is now a tangible difference between "titles" and "urls", yet this funciton is actually used for getting dirEntires by "url", not by title. I could re-name it to `getDirEntryByUrl` (which would require renaming code throughout the app where it is called - 16 results in 3 files).

Jaifroid: I think `getDirEntryByPath` might be a more accurate name (it searches on namespace + filename, e.g. `A/path/to/interesting_file` or `C/user_assets/mobile_display.css`). However, I'm torn on whether to do it as part of this PR or as a separate one.

**In favour of doing it in this PR**: this PR clearly separates "titles" (with their own pointer list) from "urls". Articles have titles, URLs (which include assets) may have titles, but more often than not don't. The fact that this function searches the url pointer list, but is named `getDirEntryByTitle`, is now very confusing . I nearly made a mistake, and was half-way to forcing it to use the new title/article pointer list, before remembering that it seaches on namespace + filename. So this PR sort of requires us to rename this function.

**In favour of doing it as a separate PR**: there are lots and lots of references to this function, and a number of variable names would need changing in disparate functions to rename this properly. We use "title" liberally throughout the code to refer to a url or a path. It would introduce lots of changes only tangentially relevant to this PR (including changing names of properites in the messages passed between Service Worker and `app.js`).

mossroy: I agree that there is some confusion in the code namings around these concepts. Some cleanup would be indeed very welcome.
About doing that in a separate PR or not, I understand the pros and cons. You can choose what seems better in your opinion.

Jaifroid: Thanks @mossroy. I've renamed the function in the latest commit, and have cleaned up as much as was viable to do in the scope of this PR in the functions that directly use this one. I've opened #710 for a more thorough cleanup of the many other instances of 'title' being used where 'url' or 'path[name]' would be clearer and avoid ambiguity.

www/js/app.js
```diff
@@ -1627,9 +1627,9 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
                 alert("Error finding random article.");
             } else {
                 // We fall back to the old A namespace to support old ZIM files without a text/html MIME type for articles
-                // DEV: This will need to be changed if we search titlePtrList version 1
-                // in a future PR, as that list contains only articles
-                if (dirEntry.getMimetype() === 'text/html' || dirEntry.namespace === 'A') {
+                // DEV: If articlePtrPos is defined in zimFile, then we are using a v1 article-only title listing. By definition,
+                // all dirEntries in an article-only listing must be articles.  
+                if (selectedArchive._file.articlePtrPos || dirEntry.getMimetype() === 'text/html' || dirEntry.namespace === 'A') {
```
Jaifroid, line 1632: NB With the currently available test ZIMs that have empty v1 title pointer lists, `articlePtrPos` will point to a v0 title pointer list instead of a v1. This is so we can test. However, the side effect is that this random function will incorrectly show non-article dirEntries when used on one of these test ZIMs. As soon as we have a valid test ZIM, this should start to work correctly.

www/js/lib/zimArchive.js
```diff
@@ -46,17 +46,38 @@ define(['zimfile', 'zimDirEntry', 'util', 'utf8'],
...
+                    // 'ptrName' and 'countName' contain the key names to be set in the archive file object
+                    {
+                        path: 'X/listing/titleOrdered/v0',
+                        ptrName: 'titlePtrPos',
+                        countName: 'entryCount'
+                    },
+                    {
+                        // !!! CHANGE THIS PATH TO .../v1 BEFORE MERGING !!!
+                        // It is set to v0 for testing because we do not have any archive with a valid v1 index yet
+                        path: 'X/listing/titleOrdered/v0',
```
Jaifroid, line 72: This is where the code is rigged. `v0` needs to be changed to `v1` as soon as we have a ZIM with a valid, non-empty v1 title pointer list.

Jaifroid:
@mossroy Just to let you know I've got as far as I can with this PR, until we have a test ZIM with a valid (non-empty) v1 listing. (see https://github.com/kiwix/kiwix-js/issues/708#issuecomment-792311126). Bear in mind that any test of this PR using one of the test ZIMs in http://tmp.kiwix.org/nons_zims/ will currently act as if it's using a v1 listing, but in fact it is using a v0 listing, as confirmed by the diagnostic console.log entry that is written shortly after loading a new ZIM. The random button will occasionally show non-article assets with such a ZIM.

Jaifroid:
I've rebased this branch on master, have removed the rigged code, and have introduced a test for an empty title listing (which will be rejected gracefully), as suggested by mgautier. The code will fall back to the next best listing (which is usually the full `entryPtrList`).

Jaifroid:
Quickly tested and working with the NEW `wikipedia_fr_basketball_maxi_2019-08_nons.zim` on Edge Chromium, IE11 and Firefox ESR (52.5.2). I haven't tested on FFOS.

I think this is ready for review.

mossroy (approved):
Tested on Firefox OS, with several old ZIM files and wikipedia_fr_basketball_maxi_2019-08_nons.zim
I did not see any issue, except the main page that can not be scrolled on my device (but I doubt it is a regression of this PR)
Congratulations!

I did not have a deep look in the code, but did not see any problem

Jaifroid:
OK, thanks. I need to test on more of the supplied ZIMs.

The Basektball landing page has an issue common to many themed Wkipedia ZIMs in jQuery mode whereby only some of the tiles are loaded. We have had an open issue #552 for this (since August 2019). As you say this is unrelated to this PR. It is to do with the dependency on the ['masonry' JavaScript library](https://masonry.desandro.com/) to produce these tiled views. It's possible to work around it by preventing some of the CSS from loading, or else we could maybe inject the required scripts into jQuery mode, but that would be a separate support option for Wikipedia-based archives.

mossroy:
Forgot to add that, as soon as we have a "no-namespace" reference ZIM file (like an updated version of Ray Charles archive, that would also cover zstd compression?), we should add unit tests on it (while keeping at least a few unit tests on the previous format, to cover compatibility)

Jaifroid:
Agreed.

Jaifroid:
I have found one issue with reading the v1 title index of the new Ray Charles no-namespace ZIM. Opening this ZIM and searching for "r" only displays two title entries with this PR: "Ray Charles" and "Ray Charles discography". However, the landing page shows there should be 15 articles beginning with "r" (see screen shot below). I need to check whether this is a problem with this PR or a problem with the coding of this sample ZIM.

@mgautierfr Would you happne to know whether this ZIM only contains a sample subset of the articles in the ZIM?

image

www/js/lib/zimfile.js
```diff
@@ -223,15 +244,85 @@ define(['xzdec_wrapper', 'zstddec_wrapper', 'util', 'utf8', 'q', 'zimDirEntry',
...
+     * Read the metadata (archive offset pointer, and number of entiries) of one or more ZIM directory Listings.
+     * This supports reading a subset of user content that might be ordered differently from the main URL pointerlist.
+     * In particular, it supports v1 title pointerlists, which contain articles sorted by title, superseding the article
+     * namespace ('A') in legazy ZIM archives.  
+     * @param {Array<DirListing>} listings An array of DirListing objects (see zimArchive.js for examples)  
+     */
+    ZIMFile.prototype.setListings = function(listings) {
+        // If we are in a legacy ZIM archive, there is nothing further to look up
+        if (this.minorVersion === 0) {
+            console.log('ZIM DirListing version: legacy');
```
Jaifroid, line 274: @mossroy Are you OK with our keeping this `console.log` entry (and the one below line 282) when merging? These will only ever print out a single message when a user first loads a ZIM, and I think it might be quite a useful diagnostic to have in case we notice any problem with production no-namespace ZIMs going forward.

mossroy: OK for me, but use debug level : `console.debug` instead of `console.log`

Jaifroid: OK, thanks. I wasn't aware of `console.debug()`. Looks useful.

Jaifroid:
So I've found the issue that is preventing search from finding some of the title listings. The issue is that the title field of the Directory Entries contains html markup in some cases, e.g. `<i>Title</i> (album)` (see screenshot of sample dirEntry below, and notice the italic tags in the title field). Of course this means a) that these titles are sorted in the v1 listing under `<` instead of under the letter of the first semantic word in the title; and b) that if we search for a title using the first word of the title, it won't be found. The missing titles can in fact be searched for by typing `<` in the search field!

@mgautierfr I assume this is a format error in the sample Ray Charles ZIM, and that you are not intending to implement rich text formatting for directory titles?

image

Result of searching for `<` in search field:

image

kelson42:
@Jaifroid Looks like a bug in MWoffliner, similar to https://github.com/openzim/mwoffliner/issues/605. Could you please open a ticket in openzim/mwoffliner and give a recent file where this problem occurs?

Jaifroid:
@kelson42 To be clear, this PR is for reading the NEW sample no-namespace ZIM archives with v1 title listings that @mgautierfr has uploaded to http://tmp.kiwix.org/nons_zims/ for us to test this PR on. I **believe** these have not been produced with mwoffliner, but with a conversion script. However, if you think they have been produced my mwoffliner I'd be happy to open a ticket. I am only seeing this with the sample Ray Charles ZIM provided (I haven't tested others).

kelson42:
@Jaifroid I understand that these zim files have been recreated with zim tool "zimrecreate". But at the origin, there is a ZIM file created with MWoffliner and zimrecreate has for sure not "invented" the DOM node in the title. I believe you probably have the same bug already with the original zim file and this would not be a regression then. This should be verified.

Jaifroid:
@kelson42 I've just tested, and this problem is **not** in the latest Ray Charles (March 2021). I think the sample Ray Charles no-namespace ZIM has been made with an old Ray Charles that may well have had the old https://github.com/openzim/mwoffliner/issues/605 issue (or similar). Since it's not in the current Ray Charles, then I think we can assume it's fixed.

kelson42:
@mgautierfr Would be nice if you could just confirm this (and redo the test ZIM file based on a recent version).

mgautierfr:
New http://tmp.kiwix.org/nons_zims/wikipedia_en_ray_charles_maxi_2021-03_nons.zim uploaded.

Jaifroid:
@mgautierfr Thank you! That was the missing piece in the puzzle and confirms that this PR is working correctly. There are no titles in the article pointer list beginning with an html tag, and a search for 'r' yields the expected articles. For information, a readout of the ZIM heade after reading the v1 title index is in the screenshot below.

image

Jaifroid:
In addition to previous tests, I've now tested on FFOS simulator and IE11 with the newest no-namespace Ray Charles ZIM. Everything seems to be working as expected, so I'll go ahead and merge now.

@mossroy I don't consider the latest Ray Charles to be a "reference version", because it was created with ZIMRecreate rather than with mwoffliner. Therefore I don't think we're ready yet to add tests specific to a new Ray Charles ZIM. As it may represent some work, it would be better to add tests on the first valid no-namespace Ray Charles produced by MWOffliner with updated libzim code.

mossroy:
All right, I agree

Issue #704, opened 2021-02-17
Make the "home" keyboard key focus the search form


mossroy:
It's a suggestion done via e-mail by RMS.

The "home" key of the keyboard scrolls the current page (or frame) to its top (in most browsers).
RMS would find it useful that it also puts the cursor in the search form, so that the user can immediately type another search (without using the mouse)

KaranKaira:
hey , @mossroy I am interested in this ,  I am new to open source  and want to contribute . I will be grateful if you guide me.

fabcoder547:
hey @mossroy  I am using version 2.8 of this extension in firefox . I am not seeing such a scroll on clicking the home key.
can you help me to find out this bug?

fabcoder547:
hey, @mossroy I got this. can you elaborate on what we have to do here? Do we have to fix this scrolling bug? and what do we have to do about the cursor on the search bar?

Jaifroid:
@fabcoder547 The description at the top of this issue is very clear. When pressing the home key (not button) the browser scrolls the page to the top. One user would like the same action to set the focus on the search field.

KaranKaira:
I tried this . (code is not formatted properly)



    document.addEventListener('keydown', function (event) {
      if (event.key === 'Home') {
        $('#btnHomeBottom').click() 
        // console.log('done')
        return false
      }
    })
 



@mossroy  @Jaifroid 


On the 'Configure' page and 'About' page when home button(keyboard) is pressed then it takes us to 'Home' Page and search form is focused.
But if we are already on the 'Home' page then pressing home button only scroll to the top without focusing on the search form .
How can we solve this?

Please have a look .
Thank You.

mossroy:
@KaranKaira 
Using `$('#btnHomeBottom').click()` is not a good idea : it indeed does what is needed, but it also switches to the welcome page of the ZIM file (which is not what is expected). You should simply put the focus in the prefix field.
You probably should not `return false` because we want to keep the default browser behavior too (which is usually to scroll to the top)
Please create a Pull Request with your following code changes, it will be easier to understand & discuss

KaranKaira:
@mossroy Sure , Today I am little busy with my college work , I will make these changes in this weekend.
Thanks

KaranKaira:
@mossroy  
Did this. This is should work .
```
  document.addEventListener('keydown', function (event) {

    if (event.key === 'Home') {
      const searchBox = document.getElementById('prefix')
      // console.log(searchBox)
[...]

PR #711, opened 2021-03-06
Work on #704

KaranKaira:
@mossroy 
Changed www/js/app.js file .

Added this code 

```
 document.addEventListener('keydown',(event)=>{
  if(event.code==="Home"){

  const prefix=document.getElementById("prefix");
  prefix.focus();
  prefix.select();
  }   
  
  })
```


mossroy:
Thanks for this PR.
But, as you can see in the "Files changed" tab, almost all the content of app.js is considered as modified.
It's probably because you formatted all the content of this file : don't do that because we can't check what has really been changed.
For this PR, please restore the original file content (without formatting it), and then add your modifications again and commit/push. Or any other way to have only the relevant changes in the "Files changed" tab.

Did you manage to find what was not working?

PR #712, opened 2021-03-07
Issue 704

KaranKaira:
Changed www\index.html and www\js\app.js .
In www\index.html
```
 <input type="search" id="prefix" placeholder="Search..." class="form-control" onclick = "this.focus()" />

```

In www\js\app.js : 

```
document.addEventListener('keydown' , (event)=>{
        if(event.code === 'Home'){
            const prefix = document.getElementById("prefix");
            prefix.click();
        }
    })
```

[Reference](https://stackoverflow.com/questions/3724749/html-input-fields-does-not-get-focus-when-clicked#:~:text=I%20used%20the%20disableSelection(),not%20get%20focused%20on%20clicking.&text=The%20solution%20was%20to%20remove,method%20for%20the%20parent%20DIV)

Behaviour is inconsistent . Sometimes working properly sometimes not.
I think there is a issue with bootstrap .
Check out the above link for various cases of what might be going wrong.
Thanks

### DIFF
--- a/www/index.html
+++ b/www/index.html
@@ -79,7 +79,7 @@
                         <div class="row">
                             <span class="col-10">
                                 <span class="input-group">
-                                    <input type="search" id="prefix" placeholder="Search..." class="form-control" onClick = "this.focus()" />
+                                    <input type="search" id="prefix" placeholder="Search..." class="form-control" onclick = "this.focus()" />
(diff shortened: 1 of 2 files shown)


KaranKaira:
Ignore

PR #713, opened 2021-03-07
Make the "home" keyboard key focus the search form #704

KaranKaira:
Changed www\index.html and www\js\app.js .
In www\index.html

```
 <input type="search" id="prefix" placeholder="Search..." class="form-control" onclick = "this.focus()" />
```
In www\js\app.js :
```
document.addEventListener('keydown' , (event)=>{
        if(event.code === 'Home'){
            const prefix = document.getElementById("prefix");
            prefix.click();
        }
    })
```
[Reference](https://stackoverflow.com/questions/3724749/html-input-fields-does-not-get-focus-when-clicked#:~:text=I%20used%20the%20disableSelection(),not%20get%20focused%20on%20clicking.&text=The%20solution%20was%20to%20remove,method%20for%20the%20parent%20DIV)

Behaviour is inconsistent . Sometimes working properly sometimes not.
I think there is a issue with bootstrap .
Check out the above link for various cases of what might be going wrong.
Thanks

### DIFF
--- a/www/index.html
+++ b/www/index.html
@@ -79,7 +79,7 @@
                         <div class="row">
                             <span class="col-10">
                                 <span class="input-group">
-                                    <input type="search" id="prefix" placeholder="Search..." class="form-control"  />
+                                    <input type="search" id="prefix" placeholder="Search..." class="form-control" />
                                     <button type="button" class="btn btn-outline-secondary" id="searchArticles"><i class="fas fa-search"></i></button>
                                 </span>
                             </span>
--- a/www/index.html
+++ b/www/index.html
@@ -79,7 +79,7 @@
                         <div class="row">
                             <span class="col-10">
                                 <span class="input-group">
-                                    <input type="search" id="prefix" placeholder="Search..." class="form-control" onClick = "this.focus()" />
+                                    <input type="search" id="prefix" placeholder="Search..." class="form-control" onclick = "this.focus()" />
                                     <button type="button" class="btn btn-outline-secondary" id="searchArticles"><i class="fas fa-search"></i></button>
                                 </span>
                             </span>
--- a/www/index.html
+++ b/www/index.html
@@ -79,7 +79,7 @@
                         <div class="row">
                             <span class="col-10">
                                 <span class="input-group">
-                                    <input type="search" id="prefix" placeholder="Search..." class="form-control" />
+                                    <input type="search" id="prefix" placeholder="Search..." class="form-control" onClick = "this.focus()" />
                                     <button type="button" class="btn btn-outline-secondary" id="searchArticles"><i class="fas fa-search"></i></button>
                                 </span>
                             </span>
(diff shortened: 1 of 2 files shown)


www/index.html
```diff
@@ -79,7 +79,7 @@
                         <div class="row">
                             <span class="col-10">
                                 <span class="input-group">
-                                    <input type="search" id="prefix" placeholder="Search..." class="form-control" />
+                                    <input type="search" id="prefix" placeholder="Search..." class="form-control" onclick = "this.focus()" />
```
Jaifroid, line 82: @KaranKaira Due to restrictive Content Security Policies in some contexts, we tyr to avoid using inline `on`-style events in HTML. Instead, you should add a click event listener to this input in `app.js`. I can't remember, but there may be such an event already defined in the code, in which case you can use it. On the other hand, I'm not sure why you want to focus the prefix on click, when a click should focus it anyway, without any additional code?

KaranKaira: hi @Jaifroid ,  Thanks for your review.
<blockquote> 
On the other hand, I'm not sure why you want to focus the prefix on click, when a click should focus it anyway, without any additional code?
</blockquote>
I did this to be 100%sure.
I will remove it now . 

Kindly see now.

mossroy:
If the behavior is not consistent, you'll need to debug to understand.
For example by adding breakpoints in your browser, or by adding logs. In order to know if/when your code is used or not, if it depends on where the focus is when the event is triggered etc
Why do you think it could be related to Bootstrap?

KaranKaira:
Please  verify if my logic in app.js is correct or not.
Thank You

www/js/app.js
```diff
@@ -1664,4 +1664,13 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
         });
     }
 
+    
+    
+    document.addEventListener('keydown' , (event)=>{
```
Jaifroid, line 1669: @KaranKaira Because we still support some old browsers, including IE11, you can't use the arrow notation. Please refactor this function and be sure to test that your code works on IE11 as a baseline. Note that we're not using Babel or Core JS, so code currently has to be written in ES5. Also, please ensure you use consistent spacing (look at examples of how we use spacing with if statements, and avoid adding extraneous spaces before comma separators.

KaranKaira: Sure.

KaranKaira: Done.

KaranKaira:
Please have a look now.

Jaifroid:
@KaranKaira Did you test this? I just tested on Edge Chromium and IE11, and it's not working in either of them. You need to test thoroughly on at least Chrome/Chromium, Firefox and IE11, with a ZIM file loaded and an article showring, scrolled to the bottom of the article, and the focus on the article. Test in both jQuery and Service Worker modes (see Expert Settings in Configuration). Before requesting a review, it would be very helpful if you could indicate which tests you have done and the outcome of those tests.

Jaifroid:
This PR was superseded by #714 .

PR #714, opened 2021-03-14
"Home" keyboard key now focuses search form.

Hrishi-Raj:
Issue: #704 
The changes are made in www/js/app.js

It does the intended action of focusing search form and scrolling to top on 'home' keydown.

### DIFF
--- a/www/index.html
+++ b/www/index.html
@@ -417,6 +417,13 @@
                                         <strong>Animate transition between app pages</strong>
                                     </label>
                                 </div>
+                                <div class="checkbox">
+                                    <label>
+                                        <input type="checkbox" name="useHomeKeyToFocusSearchBar"
+                                               id="useHomeKeyToFocusSearchBarCheck">
+                                        <strong>Use Home key to focus search bar</strong> (may rarely have side effects on ZIM files that handle Home key)
+                                    </label>
+                                </div>
                                 <div class="form-group">
                                     <label>
                                         <b>Select app theme</b> (content inversion is experimental):
@@ -524,25 +531,6 @@
                                 <div id="serviceWorkerStatus"></div>
                                 <div id="messageChannelStatus"></div>
                                 <div id="settingsStoreStatus"></div>
-                            </div>
-                        </div>
-                    </div>
-                    <br />
-                    <div class="container">
-                        <h3>Extra Features</h3>
-                        <div class="card" id="extraFeatureDiv">
-                            <div class="card-header">Use Home Key to focus search bar</div>
-                            <div class="card-body">
-                                <div class="checkbox">
-                                    <label>
-                                        <input type="checkbox" name="useHomeKeyToFocusSearchBar"
-                                               id="useHomeKeyToFocusSearchBarCheck">
-                                        <strong>Activate</strong>
-                                    </label>
-                                </div>
-                                <p>
-                                    (Caution!! May produce side effect for few ZIM files in ServiceWorker mode.)
-                                </p>
                             </div>
                         </div>
                     </div>
--- a/www/index.html
+++ b/www/index.html
@@ -528,6 +528,25 @@
                         </div>
                     </div>
                     <br />
+                    <div class="container">
+                        <h3>Extra Features</h3>
+                        <div class="card" id="extraFeatureDiv">
+                            <div class="card-header">Use Home Key to focus search bar</div>
+                            <div class="card-body">
+                                <div class="checkbox">
+                                    <label>
+                                        <input type="checkbox" name="useHomeKeyToFocusSearchBar"
+                                               id="useHomeKeyToFocusSearchBarCheck">
+                                        <strong>Activate</strong>
+                                    </label>
+                                </div>
+                                <p>
+                                    (Caution!! May produce side effect for few ZIM files in ServiceWorker mode.)
+                                </p>
+                            </div>
+                        </div>
+                    </div>
+                    <br />
                 </div>
                 <div id="welcomeText" class="container">
                     You can search the content of your ZIM archive by typing in the above search field.
--- a/www/js/app.js
+++ b/www/js/app.js
@@ -117,11 +117,11 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
             }, 100);
         }
     }
-    $(document).ready(function(){
+    $(document).ready(function() {
         resizeIFrame();
         //handle home key press in initial Home Page (empty iframe)
         var iframeContentWindow = document.getElementById('articleContent').contentWindow;
-        $(iframeContentWindow).on('keydown',function (e){
+        iframeContentWindow.addEventListener('keydown',function(e) {
             focusPrefixOnHomeKey(e.key);
         });
     });
@@ -153,7 +153,7 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
         return false;
     });
      //Handle Home key press inside window(outside iframe) to focus #prefix
-     $(window).on('keydown', function(e) {
+     window.addEventListener('keydown',function(e) {
          focusPrefixOnHomeKey(e.key);
      });
     // Handle keyboard events in the prefix (article search) field
@@ -398,9 +398,9 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
     });
 
     //focus #prefix if Home key is pressed
-    function focusPrefixOnHomeKey(key){
+    function focusPrefixOnHomeKey(key) {
         //check if home key is pressed
-        if(key === 'Home') {
+        if (key === 'Home') {
             //scroll to top of #search-article section if scroll bar is associated with it
             $('#search-article').scrollTop(0);
             //in case the scroll bar is associated with iframe #articleContent
@@ -1166,6 +1166,11 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
                 if (docBody) {
                     docBody.addEventListener('dragover', handleIframeDragover);
                     docBody.addEventListener('drop', handleIframeDrop);
+                    docBody.addEventListener('keydown',function(e) {
+                        // if cursor is not inside a input element only then focus #prefix
+                        if (e.target.nodeName !== 'INPUT')
+                            focusPrefixOnHomeKey(e.key);
+                    });
                 }
                 resizeIFrame();
                 // Reset UI when the article is unloaded
@@ -1339,9 +1344,9 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
                 // Deflect drag-and-drop of ZIM file on the iframe to Config
                 docBody.addEventListener('dragover', handleIframeDragover);
                 docBody.addEventListener('drop', handleIframeDrop);
-                docBody.addEventListener('keydown',function(e){
+                docBody.addEventListener('keydown',function(e) {
                     // if cursor is not inside a input element only then focus #prefix
-                    if(e.target.nodeName !== 'INPUT')
+                    if (e.target.nodeName !== 'INPUT')
                         focusPrefixOnHomeKey(e.key);
                 });
             }
--- a/www/js/app.js
+++ b/www/js/app.js
@@ -117,7 +117,14 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
             }, 100);
         }
     }
-    $(document).ready(resizeIFrame);
+    $(document).ready(function(){
+        resizeIFrame();
+        //handle home key press in initial Home Page (empty iframe)
+        var iframeContentWindow = document.getElementById('articleContent').contentWindow;
+        $(iframeContentWindow).on('keydown',function (e){
+            focusPrefixOnHomeKey(e.key);
+        });
+    });
     $(window).resize(resizeIFrame);
     
     // Define behavior of HTML elements
@@ -145,32 +152,10 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
         document.getElementById('searchArticles').click();
         return false;
     });
-     //Handle Home key press to focus #prefix
-     $(window,document).on('keydown', function(e) {
-         //if home key is pressed then
-         if(/^Home/.test(e.key)) {
-             //scroll to top of #search-article section since scroll bar is associated with it and not to window or document
-             $('#search-article').scrollTop(0);
-             $('#prefix').focus();
-         }
+     //Handle Home key press inside window(outside iframe) to focus #prefix
+     $(window).on('keydown', function(e) {
+         focusPrefixOnHomeKey(e.key);
      });
-     //Handle Home key event when focus is inside iframe #articleContent
-     var iframe= document.getElementById('articleContent');
-     //onload event is not fired in chrome for iframe if content is from local storage
-     //to compensate for onload event setTimeout can be used
-     setInterval(function (){
-         let iframewindow= iframe.contentWindow? iframe.contentWindow : iframe.contentDocument.defaultView;
-         $(iframewindow.document).on('keydown',function(e){
-             //if home key is pressed then
-             if(/^Home/.test(e.key)) {
-                 //scroll to top of #search-article section since scroll bar is associated with it and not to window or document
-                 $('#search-article').scrollTop(0);
-                 //in case scroll bar is associated with iframe #articleContent
-                 $(iframewindow.document).scrollTop(0);
-                 $('#prefix').focus();
-             }
-         });
-     },2000);
     // Handle keyboard events in the prefix (article search) field
     var keyPressHandled = false;
     $('#prefix').on('keydown', function(e) {
@@ -412,6 +397,18 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
         }, 600);
     });
 
+    //focus #prefix if Home key is pressed
+    function focusPrefixOnHomeKey(key){
+        //check if home key is pressed
+        if(key === 'Home') {
+            //scroll to top of #search-article section if scroll bar is associated with it
+            $('#search-article').scrollTop(0);
+            //in case the scroll bar is associated with iframe #articleContent
+            var iframeArticleDocument = document.getElementById('articleContent').contentWindow.document;
+            $(iframeArticleDocument).scrollTop(0);
+            $('#prefix').focus();
+        }
+     }
     /**
      * Displays or refreshes the API status shown to the user
      */
@@ -1342,6 +1339,11 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
                 // Deflect drag-and-drop of ZIM file on the iframe to Config
                 docBody.addEventListener('dragover', handleIframeDragover);
                 docBody.addEventListener('drop', handleIframeDrop);
+                docBody.addEventListener('keydown',function(e){
+                    // if cursor is not inside a input element only then focus #prefix
+                    if(e.target.nodeName !== 'INPUT')
+                        focusPrefixOnHomeKey(e.key);
+                });
             }
             // Set the requested appTheme
             uiUtil.applyAppTheme(params.appTheme);
--- a/www/js/app.js
+++ b/www/js/app.js
@@ -1187,11 +1187,10 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
                     // Configure home key press to focus #prefix only if the feature is in active state
                     if (params.useHomeKeyToFocusSearchBar)
                         iframeArticleContent.contentWindow.addEventListener('keydown', focusPrefixOnHomeKey);
-
+                    // Reset UI when the article is unloaded
                     iframeArticleContent.contentWindow.onunload = function () {
                         // remove eventListener to avoid memory leaks
                         iframeArticleContent.contentWindow.removeEventListener('keydown', focusPrefixOnHomeKey);
-                        // Reset UI when the article is unloaded
                         $("#articleList").empty();
                         $('#articleListHeaderMessage').empty();
                         $('#articleListWithHeader').hide();
--- a/www/js/app.js
+++ b/www/js/app.js
@@ -145,6 +145,32 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
         document.getElementById('searchArticles').click();
         return false;
     });
+     //Handle Home key press to focus #prefix
+     $(window,document).on('keydown', function(e) {
+         //if home key is pressed then
+         if(/^Home/.test(e.key)) {
+             //scroll to top of #search-article section since scroll bar is associated with it and not to window or document
+             $('#search-article').scrollTop(0);
+             $('#prefix').focus();
+         }
+     });
+     //Handle Home key event when focus is inside iframe #articleContent
+     var iframe= document.getElementById('articleContent');
+     //onload event is not fired in chrome for iframe if content is from local storage
+     //to compensate for onload event setTimeout can be used
+     setInterval(function (){
+         let iframewindow= iframe.contentWindow? iframe.contentWindow : iframe.contentDocument.defaultView;
+         $(iframewindow.document).on('keydown',function(e){
+             //if home key is pressed then
+             if(/^Home/.test(e.key)) {
+                 //scroll to top of #search-article section since scroll bar is associated with it and not to window or document
+                 $('#search-article').scrollTop(0);
+                 //in case scroll bar is associated with iframe #articleContent
+                 $(iframewindow.document).scrollTop(0);
+                 $('#prefix').focus();
+             }
+         });
+     },2000);
     // Handle keyboard events in the prefix (article search) field
     var keyPressHandled = false;
     $('#prefix').on('keydown', function(e) {
--- a/www/js/app.js
+++ b/www/js/app.js
@@ -81,6 +81,9 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
     params['appTheme'] = settingsStore.getItem('appTheme') || 'light'; // Currently implemented: light|dark|dark_invert|dark_mwInvert
     document.getElementById('appThemeSelect').value = params.appTheme;
     uiUtil.applyAppTheme(params.appTheme);
+    // A global parameter to turn on/off the use of Keyboard HOME Key to focus search bar
+    params['useHomeKeyToFocusSearchBar'] = settingsStore.getItem('useHomeKeyToFocusSearchBar') === 'true';
+    document.getElementById('useHomeKeyToFocusSearchBarCheck').checked = params.useHomeKeyToFocusSearchBar;
 
     // An object to hold the current search and its state (allows cancellation of search across modules)
     appstate['search'] = {
@@ -119,11 +122,7 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
     }
     $(document).ready(function() {
         resizeIFrame();
-        //handle home key press in initial Home Page (empty iframe)
-        var iframeContentWindow = document.getElementById('articleContent').contentWindow;
-        iframeContentWindow.addEventListener('keydown',function(e) {
-            focusPrefixOnHomeKey(e.key);
-        });
+        switchHomeKeyToFocusSearchBar();
     });
     $(window).resize(resizeIFrame);
     
@@ -152,10 +151,6 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
         document.getElementById('searchArticles').click();
         return false;
     });
-     //Handle Home key press inside window(outside iframe) to focus #prefix
-     window.addEventListener('keydown',function(e) {
-         focusPrefixOnHomeKey(e.key);
-     });
     // Handle keyboard events in the prefix (article search) field
     var keyPressHandled = false;
     $('#prefix').on('keydown', function(e) {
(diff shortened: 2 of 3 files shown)


mossroy:
Thanks for this PR.
I did not test a lot but it seems to be working fine on my Firefox and Chromium desktop browsers.

I found one caveat that we did not think about, though : when the ZIM content already has a behavior on the Home key.
For example, in Gutenberg ZIM files (any one from https://download.kiwix.org/zim/gutenberg/), there is a search field (to search by author name, it only works in ServiceWorker mode) : if you type something in it, and use the home key, you'd expect it to simply move the cursor at the beginning of the field. In this PR, it moves the cursor in the #prefix field instead.
It would probably be easy to workaround the issue in this specific case, but there could be many other cases (like some javascript inside the ZIM file that adds an event handler on it).

I did not have a look at the implementation, yet.

www/js/app.js
```diff
@@ -145,6 +145,32 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
         document.getElementById('searchArticles').click();
         return false;
     });
+     //Handle Home key press to focus #prefix
+     $(window,document).on('keydown', function(e) {
+         //if home key is pressed then
+         if(/^Home/.test(e.key)) {
```
mossroy, line 151: Is there a specific reason to test that `e.key` **starts** with `Home`?
If not, it would be safer to test for equality : `/^Home$/`

www/js/app.js
```diff
@@ -145,6 +145,32 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
...
+             //scroll to top of #search-article section since scroll bar is associated with it and not to window or document
+             $('#search-article').scrollTop(0);
+             $('#prefix').focus();
+         }
+     });
+     //Handle Home key event when focus is inside iframe #articleContent
+     var iframe= document.getElementById('articleContent');
+     //onload event is not fired in chrome for iframe if content is from local storage
+     //to compensate for onload event setTimeout can be used
```
mossroy, line 160: We're already using the onload event of the iframe in `readArticle` function. And it seems to work well (even on Chrome).
Using `setInterval` here is unsafe : it might run too early or too late

www/js/app.js
```diff
@@ -145,6 +145,32 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
...
+     $(window,document).on('keydown', function(e) {
+         //if home key is pressed then
+         if(/^Home/.test(e.key)) {
+             //scroll to top of #search-article section since scroll bar is associated with it and not to window or document
+             $('#search-article').scrollTop(0);
+             $('#prefix').focus();
+         }
+     });
+     //Handle Home key event when focus is inside iframe #articleContent
+     var iframe= document.getElementById('articleContent');
```
mossroy, line 158: Be careful to not collide with another `iframe` variable (like the one in `loadCSSJQuery`, that is not the same).

www/js/app.js
```diff
@@ -145,6 +145,32 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
...
+     //onload event is not fired in chrome for iframe if content is from local storage
+     //to compensate for onload event setTimeout can be used
+     setInterval(function (){
+         let iframewindow= iframe.contentWindow? iframe.contentWindow : iframe.contentDocument.defaultView;
+         $(iframewindow.document).on('keydown',function(e){
+             //if home key is pressed then
+             if(/^Home/.test(e.key)) {
```
mossroy, line 165: The function started by `keydown` event is almost the same for the main window and the iframe.
I suppose we could have a single one (to avoid duplication), that would always scroll the iframe?

www/js/app.js
```diff
@@ -145,6 +145,32 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
         document.getElementById('searchArticles').click();
         return false;
     });
+     //Handle Home key press to focus #prefix
+     $(window,document).on('keydown', function(e) {
```
mossroy, line 149: Wouldn't using `$(window)` work the same as `$(window,document)`, here?

www/js/app.js
```diff
@@ -145,6 +145,32 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
...
+             $('#search-article').scrollTop(0);
+             $('#prefix').focus();
+         }
+     });
+     //Handle Home key event when focus is inside iframe #articleContent
+     var iframe= document.getElementById('articleContent');
+     //onload event is not fired in chrome for iframe if content is from local storage
+     //to compensate for onload event setTimeout can be used
+     setInterval(function (){
+         let iframewindow= iframe.contentWindow? iframe.contentWindow : iframe.contentDocument.defaultView;
```
mossroy, line 162: I suppose this is necessary to handle some old versions of IE?
We only support IE11 for kiwix-js, not earlier versions.
So it's probably enough to use directly `.contentWindow` (like it's done in `findDirEntryFromDirEntryIdAndLaunchArticleRead` function, for example)

www/js/app.js
mossroy, line 158: Very minor : when you assign a value to a variable with `=`, it's preferred to have a space both before and after it

mossroy (changes requested):
The `setInterval` usage should be removed to handle the iframe.
The other comments are less important, and are mainly to keep simple and readable/maintainable code

www/js/app.js
Hrishi-Raj: When I tried using onload event, the changes were not working as expected. Now I understand that it was not due to problem with firing of onload event. The problem was that onload event handler is redefined by `readArticle` function and `displayArticleContentInIframe` function whenever called.
Then the keydown event handler for iframe must be defined in these functions inside onload handler. 
Please do guide me if I am wrong.

Hrishi-Raj: I have already implemented as described above. Only issue remains in case of initial empty home page (or initial empty article.html loaded in iframe). In this case iframe.onload does not get triggered.
A possible solution is to add keydown handler to iframe when the document is ready.

www/js/app.js
Hrishi-Raj: changed to equality
 `if(key === 'Home')`

www/js/app.js
Hrishi-Raj: Done

www/js/app.js
Hrishi-Raj: Thanks for clarifying. New commit uses `.contentWindow` .

www/js/app.js
Hrishi-Raj: Corrected. I will follow this from now on.

Jaifroid:
I tested quickly, and it's not working in Service Worker mode at least in Edge Classic or Edge Chromium. Pressing Home key simply scrolls to the top, without focusing the search area. It works in jQuery mode in IE11 and both flavours of Edge. However, see below about why it should perhaps *not* be implemented in SW mode.

The issue raised by @mossroy about not overriding behaviour programmed in a ZIM is an important one without an easy resolution. Only two things occur to me:

* Make it an option in configuration
* Support only jQuery mode

The second point overcomes the problem because we do not support JS-in-the-ZIM in jQuery mode, so there will be nothing to override. However, it comes at the expense of splitting the code base further between jQuery and SW modes, which could be undesirable.

www/js/app.js
```diff
@@ -117,7 +117,14 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
             }, 100);
         }
     }
-    $(document).ready(resizeIFrame);
+    $(document).ready(function(){
+        resizeIFrame();
+        //handle home key press in initial Home Page (empty iframe)
+        var iframeContentWindow = document.getElementById('articleContent').contentWindow;
+        $(iframeContentWindow).on('keydown',function (e){
```
Jaifroid, line 124: Why are you using jQuery here? You have defined `iframeContentWindow` using native DOM methods, so jQuery is redundant and just adds overhead. You simply want `iframeContentWindow.addEventListener('keydown', function ...`

www/js/app.js
```diff
@@ -145,6 +152,10 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
         document.getElementById('searchArticles').click();
         return false;
     });
+     //Handle Home key press inside window(outside iframe) to focus #prefix
+     $(window).on('keydown', function(e) {
```
Jaifroid, line 156: Ditto, not need to use jQuery.

www/js/app.js
```diff
@@ -386,6 +397,18 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
         }, 600);
     });
 
+    //focus #prefix if Home key is pressed
+    function focusPrefixOnHomeKey(key){
+        //check if home key is pressed
+        if(key === 'Home') {
```
Jaifroid, line 403: Minor, but our style uses a space between `if` and parentheses.

Jaifroid (changes requested):

Hrishi-Raj:
If we use `addEventListener` to add handler and there was an existing handler for the event then both handlers are executed whenever event is triggered. Maybe this can help to prevent overriding.
Then,
for SW mode will it be sufficient to add following condition to avoid unexpected behaviour?
_(if element under focus is not input element)_
<code>
if (event.target.nodeName !== 'INPUT')
                        focusPrefixOnHomeKey(e.key);
</code>

www/js/app.js
Hrishi-Raj: Thank you for indicating. Changes made in new commit.

www/js/app.js
Hrishi-Raj: Done.

www/js/app.js
Hrishi-Raj: Done. I will follow this from now on.

www/js/app.js
```diff
@@ -1143,6 +1166,11 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
                 if (docBody) {
                     docBody.addEventListener('dragover', handleIframeDragover);
                     docBody.addEventListener('drop', handleIframeDrop);
+                    docBody.addEventListener('keydown',function(e) {
```
mossroy, line 1169: It can be confusing to put this code here : as all the lines above handle the drag-and-drop feature, one could tend to think this piece of code is also to handle drag-and-drop.
I suggest to put that outside this code block, and add a comment above to explain the purpose. Never mind if you duplicate the `if (docBody)`

www/js/app.js
```diff
@@ -1143,6 +1166,11 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
...
+                        // if cursor is not inside a input element only then focus #prefix
+                        if (e.target.nodeName !== 'INPUT')
```
mossroy, line 1171: If we put an option for this feature, I suggest to not try to cover its side effects : it's probably impossible to achieve

www/js/app.js
```diff
@@ -1316,6 +1344,11 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
                 // Deflect drag-and-drop of ZIM file on the iframe to Config
                 docBody.addEventListener('dragover', handleIframeDragover);
                 docBody.addEventListener('drop', handleIframeDrop);
+                docBody.addEventListener('keydown',function(e) {
+                    // if cursor is not inside a input element only then focus #prefix
+                    if (e.target.nodeName !== 'INPUT')
+                        focusPrefixOnHomeKey(e.key);
+                });
```
mossroy, line 1351: Same comments as for the SW mode

Jaifroid:
> I suggest to disable it by default : for safety, and because I think most people won't care about this feature (and it's useless on mobile, of course).
> When the option is enabled, I think we should _always_ enable the event listener : in both jQuery and SW modes, and without trying to disable it in input fields (it's working, but can not cover all the possible cases)

I agree with this approach. @Hrishi-Raj , if you implement this as an option, you'll need to save a reference to the event listener, so that you can disable it if the user switches the option back off. Also, we're trying to put all new Configuraiton options together in the block of options that you'll see starting round about [line 69 in app.js](https://github.com/kiwix/kiwix-js/blob/master/www/js/app.js#L69). You can copy some of the code there. It also sets the state of the configuration option from the Settings Store, and instantiates it if the user had turned the option on in a previous session.

www/js/app.js
Hrishi-Raj: Done.

www/js/app.js
Hrishi-Raj: New commit follows this comment.

www/js/lib/settingsStore.js
```diff
@@ -34,7 +34,7 @@ define([], function () {
    */
   var regexpCookieKeysToMigrate = new RegExp([
     'hideActiveContentWarning', 'showUIAnimations', 'appTheme', 'useCache',
-    'lastContentInjectionMode', 'listOfArchives', 'lastSelectedArchive'
+    'lastContentInjectionMode', 'listOfArchives', 'lastSelectedArchive', 'useHomeKeyToFocusSearchBar'
```
mossroy, line 37: It's not necessary to add your new setting here.
This was to provide a smooth migration of the settings found in old storage : this can not happen on new settings, that are always stored in the new storage

www/index.html
```diff
@@ -528,6 +528,25 @@ <h3>Expert settings</h3>
...
+                            <div class="card-body">
+                                <div class="checkbox">
+                                    <label>
+                                        <input type="checkbox" name="useHomeKeyToFocusSearchBar"
+                                               id="useHomeKeyToFocusSearchBarCheck">
+                                        <strong>Activate</strong>
+                                    </label>
+                                </div>
+                                <p>
+                                    (Caution!! May produce side effect for few ZIM files in ServiceWorker mode.)
```
mossroy, line 544: The side effects should occur rarely, and only on specific ZIM files.
We can probably reword this to something less scary like `(may rarely have side effects on ZIM files that also handle the home key)`

www/js/app.js
```diff
@@ -386,6 +397,36 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
...
+    }
+    //switch on/off the feature to use Home Key to focus search bar
+    function switchHomeKeyToFocusSearchBar() {
+        var iframeContentWindow = document.getElementById('articleContent').contentWindow;
+        // when the feature is in active state
+        if (params.useHomeKeyToFocusSearchBar) {
+            //Handle Home key press inside window(outside iframe) to focus #prefix
+            window.addEventListener('keydown', focusPrefixOnHomeKey);
+            //handle home key press inside iframe window
+            iframeContentWindow.addEventListener('keydown', focusPrefixOnHomeKey);
```
mossroy, line 420: I'm wondering if we should also turn it on/off here in the docBody.
I think the way our UI is currently setup makes it unnecessary (because the current page is reloaded after coming back from Settings section).
But I'd be in favor of adding it here, in case we change this UI behavior.
In this case, a comment should explain why we have to set the eventListener on both the iframeContentWindow and the docBody

mossroy (commented):
Nice job, it seems to work fine, thanks.
I've added a few remarks, but we're getting close to something that can be merged, for me.

Maybe "Display Settings" and "Extra features" zones (in the Settings) could be merged as something like "Behavior settings"? @Jaifroid is the master of wording, I'd like to have his opinion ;-)

www/js/app.js
Hrishi-Raj: Sorry, for not writing an explanation for this code. It will be done.

It is not required to set event listener on both `iframeContentWindow` and `docBody`. 
This part was included to handle the initial iFrame which is displayed when no zim file is loaded yet.
For loading which none of the function `readArticle` or `displayArticleContentInIframe` are called. This is to be handled when document is ready and when the Home-Key-Feature is turned on/off.
Also, Since the Zim Contents are reloaded each time we visit the HomePage. The eventlistener associated with `iFrameContentWindow` will also be removed on the reload itself hence does not remain useful (given current UI behaviour) in any case other than initial iframe.

Jaifroid:
I've found a bug in Internet Explorer 11. To reproduce:

1. Load any Wikipedia-based ZIM in IE11.
2. Navigate to any article either by clicking on one or by searching for one (ideally a long article).
3. Scroll down the article using keyboard or mouse scroll but DO NOT CLICK into the article.
4. Press the Home key.

Result -> the search field is not selected.

This bug does not exist on Edge Classic or Edge Chromium.

www/index.html
Jaifroid: I don't think we need a whole new section for this setting, and "Extra Features" is not very useful as a description of the section. This is actually a User Interface Setting, and we currently include UI settings under Display settings.

We could either rename "Display Settings" to "User Interface Settings", or else accept that "Display Settings" includes UI features as we currently do ("Animate transition between app pages" is a UI setting, not strictly a "display" setting).

Personally, I think "Display Settings" is fine for this purpose, and that we should put this option there. It shouldn't have a label "Activate" - the label should describe the setting, i.e. the label should be "Use Home Key to focus search bar".

I agree with @mossroy's wording for the warning.

www/js/app.js
```diff
@@ -386,6 +397,36 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
...
 
+    //focus search bar (#prefix) if Home key is pressed
+    function focusPrefixOnHomeKey(event) {
+        //check if home key is pressed
+        if (event.key === 'Home') {
+            //scroll to top of #search-article section if scroll bar is associated with it
+            $('#search-article').scrollTop(0);
+            //in case the scroll bar is associated with iframe #articleContent
+            var iframeArticleDocument = document.getElementById('articleContent').contentWindow.document;
+            $(iframeArticleDocument).scrollTop(0);
```
Jaifroid, line 408: @Hrishi-Raj  You're using a hybrid of native DOM methods and jQuery here. Since you select iframeArticleDocument using the native DOM method, you should also use the native DOM `iframeArticleDocument.scrollTo(0, 0);`

In my opinion this whole function doesn't need jQuery at all.

Jaifroid: Looking at this more closely, @Hrishi-Raj , I don't think you should add an event listener **every time** the Home key is pressed. This is wasteful. Why can't you just define the required listeners once in the `switchHomeKeyToFocusSearchBar()` function when the user activates this feature?

www/js/app.js
```diff
@@ -117,7 +120,10 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
             }, 100);
         }
     }
-    $(document).ready(resizeIFrame);
+    $(document).ready(function() {
+        resizeIFrame();
+        switchHomeKeyToFocusSearchBar();
```
Jaifroid, line 125: @Hrishi-Raj I don't think you need to run this in the `document.ready` function. Just launch `switchHomeKeyToFocusSearchBar()` after line 86 above.

www/js/app.js
```diff
@@ -386,6 +397,36 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
         }, 600);
     });
 
+    //focus search bar (#prefix) if Home key is pressed
+    function focusPrefixOnHomeKey(event) {
+        //check if home key is pressed
+        if (event.key === 'Home') {
+            //scroll to top of #search-article section if scroll bar is associated with it
+            $('#search-article').scrollTop(0);
```
Jaifroid, line 405: @Hrishi-Raj Could I ask why you are trying to scroll the search-article element when it is not scrollable?
jQuery swallows an error here. If you convert to native JS:

`document.getElementById('search-article').scrollTo(0, 0);`

you'll see that IE11 throws an error that the element is not scrollable.

www/js/app.js
```diff
@@ -386,6 +397,36 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
...
+            //in case the scroll bar is associated with iframe #articleContent
+            var iframeArticleDocument = document.getElementById('articleContent').contentWindow.document;
```
Jaifroid, line 407: @Hrishi-Raj  I think you can delete this line (407), and change the next line to:

`document.getElementById('articleContent').contentWindow.scrollTo(0, 0);

Even better, use smooth scrolling for modern browsers (not IE11 or Edge Legacy):

`document.getElementById('articleContent').contentWindow.scrollTo({top: 0, behavior: 'smooth'});`
(syntax can still be used for old browsers, it doesn't throw an error).      
`

www/js/app.js
Jaifroid: Maybe you instead want to scroll some element into view? For that there is the native command:

`document.getElementById('someElement').scrollIntoView({behavior: 'smooth'});`

Jaifroid:
@Hrishi-Raj I have tested use of the Home key on master (without your PR), and it already scrolls the article to the top in all the contexts that we need it to, at least in IE11, Edge Legacy, Edge Chromium and Firefox. Therefore I think you can remove the aspects of your PR that programmatically simulate a scroll to top on pressing the Home key. You simply need **not** to cancel the keydown or keyup event, so that the native action is performed. All that is needed is to focus the search bar. I don't think that focusing the search bar should interfere with the scrolling, but if it does, then the focus could run after a timeout.

Hrishi-Raj:
@Jaifroid This was the first thing I tried at the very begining which did not work.
I was un aware that use of setTimeout solves this problem. Thank You.

What should be appropriate timeout which will work for all supported browsers?

Jaifroid:
@Hrishi-Raj I just tested and confirm that setTimeout does the job.
Comment out the scrolling code, and change `$('#prefix').focus();` to:
```
setTimeout(function() {
     document.getElementById('prefix').focus();
}, 250);
```
If it doesn't work for you, add `true` to the event listeners, i.e.:
`window.addEventListener('keydown', focusPrefixOnHomeKey, true);`
And same for other event listener.
It works for me on a quick test in IE11.

Hrishi-Raj:
> setTimeout(function() {
>      document.getElementById('prefix').focus();
> }, 250);

@Jaifroid  I tested, it works on Chrome and Firefox as well (even without adding true).
Just pointing out one difference, if clicked outside iFrame it will not scroll iFrame to top. 
Is this alright?

Jaifroid:
@Hrishi-Raj Where outside the iframe? Do you mean, for example, if you click on the top bar (but not actually on a button)? I think in that case the Home key still focuses the search bar, but obviously doesn't scroll because there is no scrollable element in focus. Since this is the normal behaviour of the browser (i.e. it also wouldn't scroll in that situation without your PR), there is no problem.

The only issue I think is still the IE11 bug where if the user hasn't clicked into the iframe, the keyup and keydown events do not fire at all. But I think this is is an IE11 quirk and is not solveable with the keyup or keydown event listeners. Once the user has clicked into the document then the focusing works. I think it is reasonable not to work around this marginal case for a deprecated browser.

Hrishi-Raj:
Made a new commit following all the above comments. 
Requesting Review.

mossroy (approved):
Looks good to me.
Quickly tested on Firefox 87 and Chromium 89
I'll let @Jaifroid test on Microsoft browsers

About scrolling or not scrolling, I personally prefer the previous version proposed by @Hrishi-Raj : scrolling programmatically before setting the focus
Because the default scrolling of browsers only occurs when the focus is inside the iframe : it can be surprising that the home key does not always have the same behavior, depending on if you clicked inside the iframe (or if you put the focus in another way). The user original wish was that the Home key always scrolls and sets the focus.
It would also avoid the setTimeout that might be not very safe

mossroy:
I also agree that it's not necessary to work on marginal use-cases for IE11

Jaifroid:
> About scrolling or not scrolling, I personally prefer the previous version proposed by @Hrishi-Raj : scrolling programmatically before setting the focus
> Because the default scrolling of browsers only occurs when the focus is inside the iframe : it can be surprising that the home key does not always have the same behavior, depending on if you clicked inside the iframe (or if you put the focus in another way). The user original wish was that the Home key always scrolls and sets the focus.

It may be possible to use dispatchEvent instead of setTimeout (though I can't think of a scenario where the setTimeout is unsafe, it merely makes the scroll async). I'll test that and report back.

However, there are other problems with changing the browser's default scroll: in all browsers, the default scroll when pressing the home key is to scroll, not jump, to the top of the page. When simulating the scroll programmatically, we get a jump to the top of the page, which is startling and disorientating. While some browsers can be instructed to scroll using "smooth scrolling", this does not work in older browsers, whereas smooth scrolling *does* work in these older browsers if we just let the browser handle the scroll natively.

Furthermore, this PR works fine *without* clicking into the iframe, for example if the user searches for an article or when using the random button. It only fails to *scroll* if the user deliberately clicks somewhere in the app outside of the iframe, which would be the case anyway without this PR. Even so, this PR still focuses the search bar in that situation, so it is still functional.

I feel we should only override native behaviour on something like scroll if there is a very good reason to do so.

However, if both of you prefer a programmatic scroll, I'll obviously come along quiety! I would just ask @Hrishi-Raj to make sure it is a "smooth scroll", not a jump, if you decide to simulate the scroll, for browsers that support smooth scroll.

Jaifroid:
DispatchEvent doesn't work, because the event is in the process of being dispatched. It is only for newly created events.

However, it is possible to reduce the setTimeout from 250 ms to 0 ms. This in efect merely delays the focussing of the search bar to the next tick in the Event Loop. This should be safer than a 250ms delay, and also shows that we are not in danger of introducing a race condition here.

@Hrishi-Raj To do this, change 250 to 0 in the setTimeout argument.

mossroy:
The setTimeout parameter set to 0 is certainly a good move, and probably shows it's safe.
I understand your position, and agree to only focus (i.e. without a programmatic scroll)

Hrishi-Raj:
> 
> @Hrishi-Raj To do this, change 250 to 0 in the setTimeout argument.

@Jaifroid Changes made.

Jaifroid:
@Hrishi-Raj Looks good now! I'll just do some testing on old browsers and will report back.

Jaifroid:
@Hrishi-Raj I've tested on all browsers, but found an issue with IE11. However the good news is that it's easy to fix, and the fix works in all browsers as well.

The issue is that registering the event listener on the `docBody` of the iframe instead of on the `contentWindow` prevents IE11 from capturing the key press unless the user has explicitly clicked in the iframe. In other words, in IE11 only, after searching an article or using the random button, the user has to click somewhere in the iframe before this PR will work. However, if we instead attach the `keydown` event to the `contentWindow` of the iframe, then the PR works without clicking into the document in all browsers. This is very easy to remedy.

While doing this, I also think it would be best practice to remove the event listener in the `onunload` event of the `contentWindow` to avoid potiential memory leaks. We currently have jQuery code that removes all its listeners, but I think we should remove this listener attached using native DOM methods. There is already an onunload event int he Service Worker code and the removeEventListener can be added there. Then it's a case of copying that to the jQuery equivalent.

@Hrishi-Raj I've tested these changes and would be happy to push a commit to your PR implementing them. However, if you would rather I paste some sample code here for you to do, that's fine as well. Let me know what you would prefer.

Hrishi-Raj:
@Jaifroid 
Please let me do it. 
I am Ubuntu user, I am setting up Virtual Machine to test on IE11. This may take few hours.

www/js/app.js
```diff
@@ -1144,6 +1181,10 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
                     docBody.addEventListener('dragover', handleIframeDragover);
                     docBody.addEventListener('drop', handleIframeDrop);
                 }
+                // Configure home key press to focus #prefix only if the feature is in active state
+                if (docBody && params.useHomeKeyToFocusSearchBar)
+                    docBody.addEventListener('keydown', focusPrefixOnHomeKey);
```
Jaifroid, line 1186: @Hrishi-Raj Here is my suggestion:

Instead of `docBody` here, you can use `iframeArticleContent.contentWindow.addEventListener( ..... );` This works much better with IE11.

But then you can move this code to be inside the test on line 1190 (add the missing curly braces):

```
if (iframeArticleContent.contentWindow) {
    if (params.useHomeKeyToFocusSearchBar)
        ....
    iframeArticleContent.contentWindow.onunload = function () { 
       ....
```
You can add the `.removeEventListener` code inside this `onunload` function, before the next statements.

Then you need to copy the same idea for jQuery mode and put it round about [line 1378](https://github.com/kiwix/kiwix-js/pull/714/files#diff-484c9e8aa09e09b8eb1fafaf55e8eadcd0eaab230e38d02bb9440841bfa090a4R1378).

I hope this is helpful.

Hrishi-Raj: This was very helpful. I made the changes and also tested it on IE11, it works as you described (without clicking anywhere).

www/js/app.js
```diff
@@ -1145,14 +1182,23 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
...
+
+                if (iframeArticleContent.contentWindow) {
+                    // Configure home key press to focus #prefix only if the feature is in active state
+                    if (params.useHomeKeyToFocusSearchBar)
+                        iframeArticleContent.contentWindow.addEventListener('keydown', focusPrefixOnHomeKey);
+
+                    iframeArticleContent.contentWindow.onunload = function () {
+                        // remove eventListener to avoid memory leaks
+                        iframeArticleContent.contentWindow.removeEventListener('keydown', focusPrefixOnHomeKey);
+                        // Reset UI when the article is unloaded
```
Jaifroid, line 1194: @Hrishi-Raj I think it would be better to move this comment to the currently blank line 1190.

Jaifroid (approved):
@Hrishi-Raj Thank you this is great! I have one very minor small suggestion, to move a comment, but otherwise this is ready to squash and merge. Thank you very much for this PR and for your flexibility in implementing our changing suggestions. Once you tell me it's ready, I can ensure it is squashed and merged, or let me know if you need experience in squashing a PR to a single commit.

Hrishi-Raj:
@Jaifroid  I have made the changes. 
If you feel it is ready, please Squash and Merge these commits.

Jaifroid:
OK, I've done my last tests, so I'll squash and merge now.

PR #724, opened 2021-05-18
Calculate true number of articles in a legacy ZIM

### LINKED ISSUE #720: REGRESSION: random button no longer truly random
Due to the changes implemented to support v1 title indices, the random button appears no longer to be random. This is best observed on a small Wikipedia ZIM such as `wikipedia_en_100_maxi_2021-03`. The random buttom predominantly picks articles towards the end of the alphabet, often repeating the same three articles in succession, as if caught in a loop.

I *believe* this may be caused by picking the article from the full pool of entries in the v0 index, and then filtering for those that are articles. This will cause the random function to be selecting from a much larger pool of articles than are actually in the article index. This will tend to skew the random result probabilistically towards larger numbers, and hence will tend to pick articles weighted towards the end of the range of actual articles.

This should not be difficult to fix.

Jaifroid:
This fixes #720.

It turned out to be an interesting logical problem. For a legacy (i.e. current) ZIM format, the random function takes the full range of titles in the title pointerlist and hops through it until it finds an article. The title pointerlist is merely the entire list of dirEntries ordered by title, as opposed to url. The number of entries in the title pointerlist is thus the same as the number of entries in the url pointerlist, which is equal to the number of directory entries. Random is therefore quite inefficient, because it is trying to find a random article in a sea of other directory entries (e.g. the number of images can easily outnumber the number of articles).

However, in a v1 no-namespace ZIM, we will have an article pointerlist which is the subset of dirEntries that represent articles, and the random function will search through this list, thus being more efficient and more truly random.

However, it struck me that the v1 article pointerlist is actually equivalent to the subset of the v0 (legacy) title pointerlist that is in the `A` namespace, i.e. all articles and only articles. If we pre-calculate the start of that subset in a v0 ZIM, and find out the number of entries in that subset, then we have the equivalent of the v1 article pointerlist, i.e., we have emulated it for a legacy ZIM.

Doing this doesn't only make the random function more efficient. It also speeds up title binary search, because we are only hopping through a reduced subset of the full title list that would otherwise include images, metadata, etc. We end up saving quite a number of hops (you can see them in the debug console log during the initial pre-calculation in this PR). Effectively we do this calculation once in advance instead of doing it for every title (binary) search.

While it may seem pointless enhancing search (and random) in legacy ZIMs, when the format is about to be deprecated, I think in fact we'll be dealing with these legacy ZIMs for quite a long time, so this PR is worth it.

There is one thing to note that will always prevent the random function from being truly random (in the practical rather than the theoretical sense). This is the fact that the article list contains (at least for Wikipedia ZIMs) lots of redundant entries. For example, we may have `(USA)`, `US`, `united states`, `United States`, `United states of America`, `United States of America`, etc., etc., all pointing to the same underlying article. Thus, the random function will end up biased towards those articles with lots and lots of aliases, whereas less popular articles will often have no aliases. There is no way to overcome this, as we don't have a list of unique articles.

I need to test this PR more extensively before it is ready for review. Note that I'll clean out most of the console debug log before merging.

### DIFF
--- a/www/js/lib/zimArchive.js
+++ b/www/js/lib/zimArchive.js
@@ -273,6 +273,8 @@ define(['zimfile', 'zimDirEntry', 'util', 'utf8'],
             return that._file.dirEntryByTitleIndex(i).then(function(dirEntry) {
                 if (search.status === 'cancelled') return 0;
                 var ns = dirEntry.namespace;
+                // DEV: This search is redundant if we managed to populate articlePtrLst and articleCount, but it only takes two instructions and
+                // provides maximum compatibility with rare ZIMs where attempts to find first and last article (in zimArchive.js) may have failed
                 if (ns < cns) return 1;
                 if (ns > cns) return -1;
                 // We should now be in namespace A (old format ZIM) or C (new format ZIM)
--- a/www/js/lib/zimfile.js
+++ b/www/js/lib/zimfile.js
@@ -282,22 +282,11 @@ define(['xzdec_wrapper', 'zstddec_wrapper', 'util', 'utf8', 'q', 'zimDirEntry', 
                         var ns = dirEntry.namespace;
                         var url = ns + '/' + dirEntry.getTitleOrUrl();
                         var prefix = ordinal === 'first' ? 'A' : 'B';
-                        // DIAGNOSTIC TO BE REMOVED
-                            console.debug(ordinal + ': ' + url);
                         if (prefix < ns) return -1;
                         else if (prefix > ns) return 1;
                         return prefix < url ? -1 : 1;
                     });
                 }, true).then(function(index) {
-                    // DIAGNOSTIC CODE TO BE REMOVED //
-                        console.log('The **' + ordinal + '** dirEntry in range is index ' + index);
-                        console.log('Checking dirEntry title for index and index-1 (async)...');
-                        [index, index-1].forEach(function (idx) {
-                            that.dirEntryByTitleIndex(idx).then(function (dEntry) {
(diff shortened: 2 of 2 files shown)


Jaifroid:
This is now tested on a range of ZIMs. I've added diagnostic code (clealry marked) to be removed before merge. Look for items like this in console when you load a current ZIM (not a no-ns ZIM):
```
The **first** dirEntry in range is index 48
Checking dirEntry title for index and index-1 (async)...
--> A/"Angel of the North"
--> -/webpHeroPolyfill.js
...
The **last** dirEntry in range is index 57048
Checking dirEntry title for index and index-1 (async)...
Calculated article count is: 57000
--> I/!!_Auberge_de_Castille_at_sunrise_!!.jpg.webp
--> A/重庆
```
The two dirEntry titles shown with `-->` are the two boundary dirEntries, the first between the `-` and `A` namespaces, and the second between the `A` and the next namespace (usually `I`, but it could be `B`). This is just to prove that the code is capturing the full range of dirEntries, and produces a correct `articleCount`.

We store the values for `articlePtrList` and `articleCount`, and we use this henceforth as an article-only list against which to do binary search and random. These functions become more efficient and act in the same way as they will with a no-ns ZIM.

The code is ready for review bearing in mind that the diagnostic code will be removed before merging.

mossroy (approved):
Tested OK on Firefox OS with various ZIM files.
On my old Firefox OS device, with big ZIM files, the initial binary search of first and last articles takes a few seconds to end. But it runs in "parallel" with the load of the home page and does not introduce any significant slowdown.
The code looks good.
Thanks!

Jaifroid:
I've now tested this on IE11, Chromium and Firefox on a range of archive types.

Issue #735, opened 2021-07-14
Only polyfill Promise API if browser needs it


Jaifroid:
I realize this is an optimization that would be better served by https://github.com/kiwix/kiwix-js/issues/554, where we would include babel or whatever the latest polyfill solution is, but "in the meantime" we could do "babel-lite", by using the currently recommended Promise polyfill from https://polyfill.io/v3/url-builder/ . This would be a more intelligent version than our current Promise polyfill, as it would only load if it detects that it is needed. Modern browsers handle promises very well and fast.

Alternatively, we could include the full ES6 polyfill package, and it would only polyfill these APIs if missing. It would allow us to write more modern code:

- Array.from
- Array.of
- Array.prototype.@@iterator
- Array.prototype.copyWithin
- Array.prototype.entries
- Array.prototype.fill
- Array.prototype.find
- Array.prototype.findIndex
- Array.prototype.keys
- Array.prototype.sort
- Array.prototype.values
- Function.prototype.name
- Map
- Math.acosh
- Math.asinh
- Math.atanh
- Math.cbrt
- Math.clz32
- Math.cosh
- Math.expm1
- Math.fround
- Math.hypot
- Math.imul
- Math.log10
- Math.log1p
- Math.log2
- Math.sign
- Math.sinh
- Math.tanh
- Math.trunc
- Number.Epsilon
- Number.MAX_SAFE_INTEGER
- Number.MIN_SAFE_INTEGER
- Number.isFinite
- Number.isInteger
- Number.isNaN
- Number.isSafeInteger
- Number.parseFloat
- Number.parseInt
- Object.assign
- Object.is
- Object.seal
- Object.setPrototypeOf
- Promise
- RegExp.prototype.flags
- Set
- String.fromCodePoint
- String.prototype.@@iterator
- String.prototype.codePointAt
- String.prototype.endsWith
- String.prototype.includes
- String.prototype.repeat
- String.prototype.startsWith
- Symbol.hasInstance
- Symbol.isConcatSpreadable
- Symbol.iterator
- Symbol.match
- Symbol.replace
- Symbol.search
- Symbol.species
- Symbol.split
- Symbol.toPrimitive
- Symbol.toStringTag
- Symbol.unscopables
- Symbol
- WeakMap
- WeakSet

Jaifroid:
PS current recommended Promise polfill source is here:

https://github.com/Financial-Times/polyfill-library/tree/master/polyfills/Promise

[...]

PR #736, opened 2021-07-15
Use Promise polyfill instead of Q

Jaifroid:
This implements #735 for testing. Please note that tests will probably **fail** due to IE11. While the polyfill works perfectly in IE11, one of the tests fails (something to do with uppercase Unicode). It may be that IE11 needs another polyfill to deal with the specific Unicode issue that was (I don't know why) dealt with by Q.

Second issue: I can't currently test on Firefox OS because Firefox surreptitiously updated the old LTS version with the FFOS IDE in it (despite my turning off updates, etc.). Grrr. It may take me a while to set it up again. A test on a physical FFOS device (no rush) would be appreciated.

Overall first impressions of this PR are pretty positive. I tried it out first on Kiwix JS Windows with a deep brute-force search of around 500,000 titles, and in Edge/Chrome it went through them about 50% faster, using native Promises, than with Q, and with a lot less CPU being used (judging by quieter fan noise during intensive search -- very scientific!!).

Of those I've tested, only IE11 actually uses the Polyfill, the rest are using native. I included a line in the polyfill that returns immediately if `'Promise' in self` is true. But in IE11 it seems to polyfill the entire spec (that we need) very well. No squeaks in console.log, and IE11 was able to chug faster through an intensive title search as well than with Q, so there don't appear to be any issues with setting up tens of thousands of nested Promises with this polyfill. Polyfill is also a lot smaller than Q.


www/js/app.js
```diff
@@ -963,26 +963,24 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
...
+                        reject("HTTP status " + request.status + " when reading " + url);
+                    }
                 }
-            }
-        };
-        request.onabort = request.onerror = deferred.reject;
-        request.send();
-        return deferred.promise;
+            };
+            request.onabort = request.onerror = deferred.reject;
```
Jaifroid, line 981: This should read `= reject;` at end of line (delete `deferred.`).

Jaifroid:
It turned out that WebP-Hero loaded its own Promise polyfill, and also an Array.from polyfill. This was causing some conflicts with browsers that load the WebP polyfill (e.g. Windows Mobile, IE11). So I have removed the WebP-Hero polyfill, as it is not guaranteed to load in ZIMs that are not using WebP, for example, and instead load the polyfills.io `Promise` and `Array.from polyfills` explicitly. Each of them contain a test that returns immediately if the polyfill is not needed, so they are not unnecessarily loaded.

Jaifroid:
To be clear, RequireJS does "load" the polyfills, but they contain a test at the top which exits if they are not needed, before the prototypes are polyfilled. It would be better if there were a way of conditionally loading JS with RequireJS, so that the polyfills are not even requested by the browser. I could do it by loading them dynamically using standard DOM methods, but that then bypasses RequireJS.

Well, we don't have to solve that now, it's only relevant if we decide to work on this experimental PR. I am seeing what "feels" like significant speedup in modern browsers, and also very definitely in Edge Classic, but I haven't done any objective benchmarking yet.

Jaifroid:
Because things run faster, some race conditions are emerging. One concerns the decompressors not being instantiated in time for the first decompression, since the code is arriving at the point of first decompression needed much sooner than it used to. Since we have a the `readSliceSingleThread` functions that check the decompressor is free, I added a simple check that the decompressor object is ready (latest commit). Looks like this (one for zstddec, one for xzdec):
```
        // Tests whether the decompressor is ready (initiated) and not busy
        if (zd && !busy) {
            return this.readSlice(offset, length);
        } else {
            // The decompressor is already in progress.
```
The addition is `zd &&`.

Jaifroid:
I've been struggling with crashes loading full English Wikipedia on UWP (i.e. Edge Legacy engine) -- this affects the Windows Mobile and Windows Desktop build. I eventually traced the crash to the `zstddec.js` decompressor. It seemed to be related to an out-of-memory error in the virtual machine's allocated memory. So I increased INITIAL_MEMORY in `zstddec.js` to 150,994,944 (9 x 16MiB), and this memory error goes away.

I find it bizarre that using native Promises should cause this issue. However, I notice that zstddec.js includes a Promise polyfill, so this could be another issue caused by conflicting polyfills (which using Q obviously sidesteps).

kelson42:
@Jaifroid With which ZIM file do you run your tests? Are you aware about https://github.com/openzim/libzim/pull/571 ?

Jaifroid:
@kelson42 I am testing with `wikipedia_en_all_maxi_2021-02.zim` (which I know isn't the very latest, but it's only one month's difference from latest). I presume lowering the ZSTD compression level is not a breaking change, i.e. a decompressor that can handle higher compression levels should be able to handle lower levels? If not, we would need to recompile.

kelson42:
@Jaifroid This file uses the highest compression level. We have discovered in June that this highest compression under certain circumstances, in particular the 32 bits, then the memory consumption will be really high. This is why we have decide to change that and lower a bit the compression to lower a lot the memory oncsumption. For this reason I would recommend to try with a file made in July as well.

Jaifroid:
I can now reproduce this memory error in Chromium (Edge) - see screenshot below! So I will commit a version with higher allocated memory.

@kelson42 July ZIMs work fine in Kiwix JS (I just tested with yesterday's `wikipedia_en_astronomy_maxi_2021-07.zim`).

However, I feel it's important that Kiwix JS should be able to cope with the most extreme case like this Full English ZIM. The now ancient Q library seems to have better memory optimization than native Promises, bizarrely, or at least in this edge case which is to do with how Emscripten uses Promises. We seemingly have a trade-off between speed of execution and memory consumption.

image

mossroy:
I've tested on my Firefox OS device : a few changes were needed on FirefoxOS-specific code (the switch from Q to Promise was not done there), I'll commit them on this branch.
It seems to work well on my device. I did not notice any difference regarding performance.
I did not check, but it should use its native implementation of Promises (Firefox OS 2.5 is based on Gecko 44, and Gecko supports Promises since version 29, based on https://caniuse.com/promises)

Jaifroid:
@mossroy Great! Thanks for doing those -- I hadn't noticed them due to my FFOS simulator having been wiped.
I'll try to add some benchmarking today so we can see if the need for slightly increased Emscripten memory stack is counterbalanced by performance gains I think I'm seeing in desktop browsers, and in UWP (but it's easy to be fooled).

Jaifroid:
I've replaced the `Array.from()` polyfill, which didn't work with Internet Explorer (!) in the tests suite. It was using an unsupported method in IE11. I found https://stackoverflow.com/a/62682524/9727685 which is a much simpler and easier to understand polyfill, and on quick tests with IE11 (probably the only browser that uses this polyfill), it works fine. It's also a lot less code. It only polyfills `Array.from` if it doesn't already exist.

There is one remaining issue in IE11: local tests in that browser fail at "All Unicode letters should be uppercase" (see screenshot). I don't see what that test failure has to do with this PR. I wonder if Q had some Unicode polyfilling logic in it...

image

Jaifroid:
I had the idea of using the tests suite, running locally, for some quick benchmarking.

Methodology: I exited all browsers, switched to the Use Promise polyfill branch/PR, and opened an In-Private / Incognito window in the browser I wished to test, navigating to localhost/Repos/kiwix-js/tests/ . I then closed the browser, switched branch to master, and re-openen and In-Private window to re-run the tests in master.

Caveat 1: Microsoft has removed Edge Legacy from Windows 10 completely. I had to run the Edge Legacy test in a Hyper-V VM that MS provides for testing Edge Legacy. I also had to install a mini web server (moongose.exe), node, etc.,in order to run those tests. As it's a VM, performance is appreciably slower.

Caveat 2: I've reported the first test results I achieved in each case after opening the In-Private window. However, on some browsers I re-ran the tests a few times, and got different timings even for the same branch, some wildly different. A true test would need to automate the tests and run them several times, logging the results, to get an average.

```
Kiwix-js Unit test

Edge Chromium Version 91.0.864.70 (Official build) (64-bit)

'Use Promise polyfill': 22 tests completed in 372 milliseconds, with 0 failed, 0 skipped, and 0 todo.
master branch:          22 tests completed in 481 milliseconds, with 0 failed, 0 skipped, and 0 todo.

Firefox 90.0 (64-bit)

'Use Promise polyfill': 22 tests completed in 279 milliseconds, with 0 failed, 0 skipped, and 0 todo.
master branch:          22 tests completed in 294 milliseconds, with 0 failed, 0 skipped, and 0 todo.

Internet Explorer 11

'Use Promise polyfill': 16 / 22 tests completed in 389 milliseconds, with 1 failed, 0 skipped, and 0 todo.
master branch:          16 / 22 tests completed in 430 milliseconds, with 1 failed, 0 skipped, and 0 todo.

Microsoft Edge 44.17763.1.0 Legacy **in Hyper-V VM** (not reliable!)
'Use Promise polyfill': 22 tests completed in 539 milliseconds, with 0 failed, 0 skipped, and 0 todo.
master branch:          22 tests completed in 801 milliseconds, with 0 failed, 0 skipped, and 0 todo. 
```

mossroy:
So it seems to show that the use of native Promises improves a bit the performance on all browsers that support them (which looks logical)
What is a good surprise is that the Promise Polyfill also looks a bit faster than Q on IE11.

Regarding the unit test failing on IE11, it seems to fail the same on master.
We had disabled unit tests on IE11 because of its lack of Promise support (see line 62 of tests/karma.conf.saucelabs.js), so we never checked that on the CI tests.
It might "simply" be because IE11 does not handle Unicode exactly as the other browser : you probably can log what `util.allCaseFirstLetters(testString6, "full"` outputs on IE11.

In any case, the Promise polyfill might allow us to re-enable unit tests on IE11 in our CI, which would be great.

Jaifroid:
Yes, I hadn't run the test in IE11 on master when I wrote that above. I'm sure I can fix that test: it's probably an obscure Unicode character I added to the test string that didn't exist at the time that IE11 was last updated. I agree it would be good to see if we can get automated tests running on IE11 again. With appropriate polyfills we may be able to continue to support IE11 for some time. The main thing that can't be polyfilled on IE11 are `async` functions and (I think) `await`. Mostly these are sugar for Promises, but increasingly a lot libraries are using them as standard, and a lot of sample code on MDN is using them (though it can almost always be worked around with a Promise syntax).

Jaifroid:
I'm going to rebase this PR on master because any further performance tests we run on this PR would be meaningless in comparison with master, given that master now has optimized and faster search. Anyone tracking this PR will need (after rebase) to reset local copy with `git reset --hard origin/master` or similar.

Jaifroid:
Now fully up-to-date with master.

Jaifroid:
I fixed the local (browser-based) tests for IE11 mentioned in https://github.com/kiwix/kiwix-js/pull/736#issuecomment-882026264. The problem was IE11 being unable to transform the Greek sigma variant that is used at the end of words to upper-case. It's an edge case, so I simply eliminated that letter. It does mean that upper-case transformations in title search do not produce the correct result in many Greek-language searches on, say, a Greek Wikipedia, but the great majority of Wikipedia searches do not require searching for all-caps. The only alternative would be to add a workaround specifically for the sigma variant especially for IE11, and I don't think that's something we should do (patch IE11's Unicode support). All other browsers transform the letter correctly. The tests will now pass on IE11 because I have eliminated that particular letter from the test case.

Jaifroid:
I opened #741 for re-enabling online Unit Tests for IE11.

@mossroy would you be in favour of merging this PR? It appears to demonstrate a performance improvement on modern browsers, where native Promises are more optimized than Q, and it improves performance on IE11 probably due to the smaller size of the Promise polyfill compared to the full-feature Q. Please note that I rebased this PR and force-pushed the branch 3 days ago (in case you want to do futher tests).

If we merge this, I will re-base #740 so that we can be sure that the new asm/wasm compiles are using the correct size of Emscripten memory stack.

www/js/lib/arrayFromPolyfill.js
```diff
@@ -0,0 +1,35 @@
+﻿/**
+ * Simple Array.from polyfi8ll (with Set support) from https://stackoverflow.com/a/62682524/9727685
```
Jaifroid, line 2: typo: poyfi8ll

mossroy (approved):
I tested again on my Firefox OS device and did not see any issue. I checked the webp polyfill with https://github.com/openzim/mwoffliner/files/5451044/zimfile_with_webp.zip. I suppose you will test it on IE11?

I'm in favor of merging this (with the small remarks I added). It's a welcome cleanup, and allows to use "modern" APIs

www/js/lib/promisePolyfill.js
```diff
@@ -0,0 +1,880 @@
+﻿/*
+	Yaku v0.19.3
+	(c) 2015 Yad Smood. http://ysmood.org
+	License MIT
+*/
+/*
```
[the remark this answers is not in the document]
Jaifroid: I just took this from https://polyfill.io/v3/ . The proper way to include this dependency would be with npm, I guess, but we're still having to add dependencies manually (with RequireJS). The two headers are in the original file. Maybe polyfills.io did some amalgam? In any case, I don't think I have the right to change the copyright headers and licence info...

Jaifroid:
I've adjusted the code to load the Polyfills in a cleaner way. I now load them conditionally in init.js if they are needed, so in most cases they will never even be fetched by the browser. It's obvious in retrospect, but I only saw how to do it via the conditional loading of asm or wasm in the other draft PR. It's also cleaner because if any upgrade of the dependencies is released (unlikely, I know), it's just a clean drop-in replacement. But we do really need to modernize dependency loading with #554 or similar.

Issue #739, opened 2021-07-19
Add instructions for building Emscripten binaries with docker


Jaifroid:
While looking to do some testing of #652, I stumbled upon a new official docker image for the Emscripten stack:

https://hub.docker.com/r/emscripten/emsdk

This ought to be very useful for #508, #511, #652, etc. With docker desktop (for example), you just type

`docker pull emscripten/emsdk`

This image is very new, I gather.

mossroy:
I had seen. These images exist since one year, and it's probably the way to go for all future emscripten compilation
I had started using one in https://github.com/mossroy/libzim_wasm/tree/update_to_emsdk_2.0.15 but did not have the time to finalize it (I had C compile issues, probably due to our code, not to the docker image)

Jaifroid:
Ah, OK, glad to know you've tried out this method!

I had a lot of difficulty getting emsdk to run stably on Windows (natively), as it's obviously not the dev's main target and things were always a bit broken with each update. So I switched to using WSL, which required me to install an older node verstion to run (at that point in time the emsdk was broken with latest node), but now that WSL instance is badly out of date and node complains it needs a newer version to do anything. I probably misconfigured something...

All in all, this docker container should simplify things a lot by providing a stable and predictable environment supported by the dev. It will be important, if we solve #508, that we can easily recompile libzim on a regular basis in an at least semi-automated way.

Issue #652, opened 2020-09-27
Compare performance of asm and wasm versions of zstandard decompressor


Jaifroid:
A wasm compile of zstandard is available in https://github.com/kiwix/kiwix-js/tree/Test-wasm-version-of-zstandard.
It should be tested on different platforms supporting wasm, and if performance warrants it, we should find a way for the browser to load the wasm version if it is supported, or the asm version if not.

Jaifroid:
I started to do some work on this. I worked out how to load the asm or the wasm version conditionally (same conditional loading as used at the top of `uiUtil.js`, applied in `zstdec_wrapper.js`) , and that works fine. However the issue with the increased memory requirements mentioned in #736 affects the wasm image as well (I was testing on a version that uses the Polyfill rather than Q), and that can't be hacked by changing values in the wasm binary (it really is a binary, not a script representation). So I need to rebuild the asm and the wasm with the latest Emscripten, for which  #739 will be useful.

Issue #388, opened 2018-06-02
Enable installing Kiwix JS as a Progressive Web App


Jaifroid:
A PWA is usually an offline version of a Web page/site. The Service Worker will cache content and only refresh it when it has internet access.

While our model is somewhat different (never need access to a server), making Kiwix JS work as a PWA could be an easier way of distributing the app, instead of using proprietary browser extensions. In our case, we would serve Kiwix JS from, say https://m.kiwix.org or https://kiwix.org/kiwix-js . But once downloaded, with or without a packaged ZIM, the app would work offline except for checking for code updates when the user asks it to (or on a schedule).

Google has gone all-in on PWAs. See https://developers.google.com/web/progressive-web-apps/ on the advantages, in particular:

> Progressive Web Apps are installable and live on the user's home screen, without the need for an app store. 

Microsoft is also heavily supporting the model. See https://www.pwabuilder.com/ (a Microsoft service). MS is also automatically placing major PWAs in the Microsoft Store, though developers can own the code if they prefer. Nevertheless, PWAs don't need a store model, and are intended to do away with that model.

I think, for Kiwix JS, such a system would require an easy way to download ZIM contents, as suggested in #114.

It's unclear to me at the moment whether accessing Kiwix JS as a PWA would overcome the need to pick the ZIM file (if packaged). I believe there is a maximum size for a PWA of approx. 20GB, but it can be smaller depending on free disk space of the user. If sufficient space is available, the ZIM could be "cached" (stored in indexedDB), enabling packaged apps.

EDIT: I think the ZIM would still need to be in the file system, which implies having to download it as an extra step, because we can't have binary access to files in indexedDB without loading the whole file into memory...

kelson42:
@Jaifroid Sounds to be a good idea to make the overall UX even smoother.

mossroy:
Well, it might be another way to distribute the app, but I don't see as an enhancement in terms of UX.

We already have an online version of kiwix-js (used for testing) : https://kiwix.github.io/kiwix-js. It's hosted on github, I update it after each release.
It's true that we could probably make this web page a PWA, so that it becomes available when offline. It could be another way to distribute the app (without the need to use browser extensions, and without their CSPs).

But, in terms of UX, I see it more as a regression compared to our already-existing modes.
[...]

PR #740, opened 2021-07-20
Recompile Emscripten binaries, implement WASM with fallback, and add instructions and scripts

Jaifroid:
This PR is for implementing #739.

I've done the Windows implementation and documentation, with a short PowerShell script that runs everything. It works like a dream -- so much easier than what I had to do before!

@mossroy Would you be able to do the equivalent linux script (no rush)? It's probably just a one-liner and very easy for you to test on your linux machines. Take a look at the script `Compile-Zstddec.ps1` in this PR.

I'm not sure whether we should add equivalent xzdec compilation to this PR to make it generic. For the moment it's just zstddec.


scripts/Compile-Zstddec.ps1
```diff
@@ -0,0 +1,14 @@
+# This PowerShell script (principally for Windows 10/11) compiles zstddec.js using docker according to the values set in 
+# /emscripten/ztandard/compile.sh . Please be sure to review the commandline in compile.sh before running this script, to
+# be sure it's doing what you want.
+#
+# Prerequisites:
+#
+#     * Ensure WSL 2 is enabled in Windows (WSL = Windows Subsystem for Linux) 
+#     * Install docker desktop on Windows: choose the WSL 2 docker backend (not the standard/legacy one)
+#     * Open docker and ensure the docker daemon is running correctly (icon should be green)
+#     * Ensure docker desktop is signed in to your dockerhub account
```
mossroy, line 10: It's probably not necessary, except that downloading from the docker hub will be faster if signed in

scripts/Compile-Zstddec.ps1
```diff
@@ -0,0 +1,14 @@
...
+#     * Run this script by opening a PowerShell terminal, cd to the kiwix-js repo and run './scripts/Compile-Zstddec.ps1'
```
mossroy, line 11: If I understand correctly your line below, you properly get the directory of the repo, so I suppose your script could be ran from any current directory?
That's what I'm going to do in bash

emscripten/zstandard/compile.sh
```diff
@@ -1 +1,5 @@
-emcc --memory-init-file 0 -O3 --closure 1 -s ENVIRONMENT="web" -s WASM=0 -s MALLOC="emmalloc" -s TOTAL_MEMORY=140247040 -s FILESYSTEM=0 -s DOUBLE_MODE=0 -s DYNAMIC_EXECUTION=0 -s MIN_IE_VERSION=11 -s EXPORT_NAME="ZD" -s MODULARIZE=1 -s EXPORTED_FUNCTIONS="['_malloc', '_free', '_ZSTD_createDStream', '_ZSTD_initDStream', '_ZSTD_decompressStream', '_ZSTD_isError', '_ZSTD_getErrorName', '_ZSTD_freeDStream', '_ZSTD_DStreamInSize', '_ZSTD_DStreamOutSize']" -s EXPORTED_RUNTIME_METHODS="['cwrap']" *.c -o zstddec.js
+echo "Compiling ASM version zstddec-asm.js"
+emcc --memory-init-file 0 -O3 --closure 1 -s ENVIRONMENT="web" -s WASM=0 -s MALLOC="emmalloc" -s TOTAL_MEMORY=150994944 -s FILESYSTEM=0 -s DOUBLE_MODE=0 -s DYNAMIC_EXECUTION=0 -s MIN_IE_VERSION=11 -s EXPORT_NAME="ZD" -s MODULARIZE=1 -s EXPORTED_FUNCTIONS="['_malloc', '_free', '_ZSTD_createDStream', '_ZSTD_initDStream', '_ZSTD_decompressStream', '_ZSTD_isError', '_ZSTD_getErrorName', '_ZSTD_freeDStream', '_ZSTD_DStreamInSize', '_ZSTD_DStreamOutSize']" -s EXPORTED_RUNTIME_METHODS="['cwrap']" *.c -o zstddec-asm.js
```
mossroy, line 2: According to codefactor (https://www.codefactor.io/repository/github/kiwix/kiwix-js/issues?page=4), it would be better to use `./*.c` instead of `*.c` : I'll change that

mossroy (commented):
I'm on the bash script

scripts/Compile-Zstddec.ps1
Jaifroid: I tried it without, but it fails saying it needs username and password. Seems like that one doesn't allow anonymous download. Rather than providing name/pwd in the script (as they would be personal) I thought it best just to ask the user on Windows to ensure docker desktop is logged in. May be different on linux.

scripts/Compile-Zstddec.ps1
Jaifroid: Yes, `$PSScriptRoot` always resolves to the full path of the directory the running script is in.

emscripten/zstandard/compile.sh
Jaifroid: OK, I'll test it in Windows once you've made your commit. I ignored some files in Codefactor, but can unignore them if they're useful. Actually, I think those ones are only the binaries / asm. And I should probably delete them from this PR?

scripts/Compile-Zstddec.ps1
```diff
@@ -0,0 +1,14 @@
...
+
+$repo_dir = ($PSScriptRoot -replace '[\\/]scripts[\\/]*$', '')
+docker container run -v $repo_dir\:/project -w /project emscripten/emsdk /bin/sh -c 'cd emscripten/zstandard/; ./compile.sh'
```
mossroy, line 14: I would advise to pass a specific docker image version, so that the build is reproducible. I'll do that in my bash script

mossroy:
I've commited my script.
It's reproducible : if I run the build several times, and from different computers, it gives the same result

emscripten/zstandard/compile.sh
mossroy: I think we should keep the generated files in this PR, and we should even replace the ones in the www directory with our newly-generated before merging.
And the same kind of scripts could be made available for xz

Jaifroid:
Great! I'll try to reproduce mine in Windows sandbox, as the only other machine I have is very ancient and would get clogged up if I try to install docker desktop on it.

Jaifroid:
I'll do some testing of these versions, and in particular try to get some results for #652 (comparing the wasm and the asm versions). I needed to recompile due to the memory issue with the Promise polyfill branch (as I wanted to test with that version which is faster). On a quick test with the old wasm from last year (in one browser), just judging by how it "feels" as a user browsing a ZIM, the wasm version is still not any faster than asm, but this might have changed with the newer Emscripten.

However, note that if we don't merge the Polyfill PR, then we probably don't need to compile these with the higher memory value I've included in this PR, as we only need more Emscripten memory for stability with that PR.

So long as the wasm version isn't slower, I could push my code that selects wasm if the browser supports it, or asm if it doesn't. It's not complicated, a couple of lines. Would it be best to push that to this PR (and change title) or to separate one? I think that one would depend on this one (and they both to some extent depend on the Promise Polyfill PR).

Final step would be to clone these scripts to recompile xzdec and maybe include a wasm version of that too.

Jaifroid:
Newly compiled `zstddec-asm.js` is failing on IE11 due to a (probably) trivial issue (see screenshot). May need to open an issue at Emscripten, but should check first that the flags haven't changed. I definitely included `-s MIN_IE_VERSION=11` in the ASM version (I removed it for the wasm version). 

image

Jaifroid:
I opened https://github.com/emscripten-core/emscripten/issues/14700. However, that GitHub is swamped with issues and PRs, so it wouldn't surprise me if this is overlooked. In the meantime, we can easily polyfill the `startsWith` method from https://developer.mozilla.org/en-US/docs/web/javascript/reference/global_objects/string/startswith#polyfill.

Jaifroid:
I confirm that the only polyfill needed to allow this asm to run on IE11 is:
```
if (!String.prototype.startsWith) {
    Object.defineProperty(String.prototype, 'startsWith', {
        value: function(search, rawPos) {
            var pos = rawPos > 0 ? rawPos|0 : 0;
            return this.substring(pos, pos + search.length) === search;
        }
    });
}
```
I'll add it to this PR in `zstddec_wrapper.js`, pending any comments from Kripken.

Jaifroid:
Kripken did reply on that issue, suggesting I make a PR for the polyfill... I don't know whether I can do that (I've actually suggested a simpler alternative), but in any case it looks like I should add the very short polyfill -- maybe to zimfile.js -- with a note to remove it if it's not needed in a future Emscripten release.

scripts/Compile-Zstddec.ps1
Jaifroid: Done.

Jaifroid:
I've added in the newly compiled wasm and asm modules (zstandard only, not xzdec yet), and I've made them conditionally loadable, so that IE11 still works and uses ASM, and browsers capable of running WebAssembly use the WASM. The WASM version is not any slower than ASM (in any one browser) on my quick subjective tests, and *may* be a little faster. What's sure is that it is a much smaller module to load and should run on a dedicated low-level system (though ASM apparently runs on its own precompiled thread too).

Just bear in mind that this branch doesn't have the Promise polyfill yet (that's a separate PR), so it's still using Q, and doesn't benefit from the speedup we see in that branch.

Now, should we try to do the same for xzdec? The compile string will need to be changed, because we last compiled on an older version of EMSDK, and some of the options have changed names.

Jaifroid:
@mossroy Just to let you know, we need to modularize xzdec, which also means refactoring xzdec_wrapper, to get it to work in the same way as zstddec. I need to push a change to the EMCC compile command to do that.

mossroy:
@Jaifroid : OK sorry I leave the keyboard for now, then.
I've commited the corresponding bash script for xzdec, that seems to work well
I had also commited the output asm and wasm files, feel free to regenerate after your changes

Jaifroid:
No problem, your changes won't conflict with what I'm doing, but we will need to recompile xzdec again when I've pushed the relevant changes (but that's easy now).

mossroy:
Congrats for finding and reporting a bug in emscripten! If you don't manage to do the PR they suggest, I might help

Jaifroid:
Feel free! I won't be working on that (handling what needs to be done with our code is plenty enough for me!!). On that issue, there is an Emscripten dev who seems to have taken an interest, but it's not clear if they are proposing to do the work or not...

Jaifroid:
Hmm. Latest commits are failing on one browser (Chrome 58, Window 7) in compiling the WASM -- see below. This is probably one of the first browsers to provide WASM support and it may well have been an incomplete in that version. How should we handle this? Some kind of fallback to ASM if WASM fails to compile? Might be hard to code for that...

```
Chrome 58.0.3029.81 (Windows 7) WARN: 'failed to asynchronously prepare wasm: CompileError: Compiling WASM function #23:<?> failed:Result = inconsistent arity in br_table target 9 (previous was 0, this one 1) @+2624
'
......................
Chrome 58.0.3029.81 (Windows 7) WARN: CompileError: Compiling WASM function #23:<?> failed:Result = inconsistent arity in br_table target 9 (previous was 0, this one 1) @+2624

CompileError: Compiling WASM function #23:<?> failed:Result = inconsistent arity in br_table target 9 (previous was 0, this one 1) @+2624

    at http://localhost:9876/base/www/js/lib/xzdec-wasm.js:14:530
    at <anonymous>
```

Jaifroid:
I thought I'd be clever and get round the Chrome 58 WASM compile error by detecting it and reloading the app in ASM mode. It worked, but Sauce Labs doesn't like it and complains (see screenshot)...

So now I'm stumped.  I downloaded a version of Chrome 58 and it indeed cannot compile the WASM. The problem is with that early version of Chrome, not with our modern WASM. My solution works for the user: they just get a momentary reload during the load sequence, which is hardly noticeable. But it doesn't work for testing. Any ideas as to how I can abort and change the RequireJS definition "on the fly", in a generic way, without doing `window.reload()`?

Last resort might be to change the Chrome version we're testing against? I don't suppose we should introduce some specific test for a specific build of this browser in the code (if that's possible), just to get around testing? 

image

Jaifroid:
OK, I ditched the reload idea, and fixed it instead with a fallback `require()` statement. I still set a value in `localStorage` to make sure ASM is used next time with that module. Code may need some refinement, but the principle works.
Not sure if I should do the same for `zstddec` (that one seems to load without problem in that browser, though we don't actually test decoding with it).

mossroy:
Your latest approach (use requireJS to load ASM if loading WASM fails) looks the best one to me.
But I don't think it's a good idea to store a "boot-with-asm" flag. Think of the scenario of a Chrome 58 on which you open kiwix-js, and then Chrome is upgraded to latest : it will keep using the ASM version instead of the WASM.
I understand you try to avoid useless CPU cycles, but we're talking about a very rare case (only one specific version of Chrome) : I think it's better to let it try to tun the WASM version every time

www/js/lib/xzdec_wrapper.js
```diff
@@ -20,11 +20,55 @@
...
+     var xzdec;
+     XZ().then(function (instance) {
+         // Instantiate the xzdec object
+         xzdec = instance;
+         if (~rqDefXZ.indexOf('xzdec-wasm')) {
+             localStorage.deleteItem(params.keyPrefix + 'boot-with-asm');
+         }
+     }).catch(function (err) {
+         if (/CompileError.+?WASM/i.test(err.message)) {
+             console.log("WASM failed to load, falling back to ASM...");
```
mossroy, line 61: Logging the error message here (as a second parameter) could be useful

www/js/lib/xzdec_wrapper.js
```diff
@@ -20,11 +20,55 @@
...
+     */
+     var xzdec;
+     XZ().then(function (instance) {
+         // Instantiate the xzdec object
+         xzdec = instance;
+         if (~rqDefXZ.indexOf('xzdec-wasm')) {
+             localStorage.deleteItem(params.keyPrefix + 'boot-with-asm');
+         }
+     }).catch(function (err) {
+         if (/CompileError.+?WASM/i.test(err.message)) {
```
mossroy, line 60: If the error does not match this pattern, it is currently silently ignored : it should be at least logged

mossroy (commented):
I would be in favor of adding the same fallback to ASM if loading the WASM file fails

Jaifroid:
> But I don't think it's a good idea to store a "boot-with-asm" flag. Think of the scenario of a Chrome 58 on which you open kiwix-js, and then Chrome is upgraded to latest : it will keep using the ASM version instead of the WASM.

I agree. It has drawbacks, so better to test each time.

mossroy:
I did not have the time to work on https://github.com/emscripten-core/emscripten/issues/14700
The PR code is probably not complicated to do.
What would take more time is to test it does what we need. It involves setting up an emscripten development environment, and being able to run our emscripten compilation with a newly-compiled emscripten binary.
As they say they don't have any CI on IE11, I suppose it's not possible to add an automatic test

Jaifroid:
We did our duty by reporting it! We can also test any fix they come up with here.

mossroy:
I would love to do this PR, just to be able to say "I contributed to emscripten", even if it would be a very small contribution ;-)
Maybe next week if I'm lucky

Jaifroid:
I have rebased on master, resolving conflicts, and have force-pushed this branch. Please re-synchronize branch.

Jaifroid:
It may be worth noting here that there are some warnings in Console (in both Firefox and Chromium) about failure to "stream compile" the wasm due to incorrect mime type. The compiler falls back to using ArrayBuffer instantiation. Clearly this is for instances where a large wasm is streamed from the Web, and is probably not relevant for us, But where we have control over the content types (possibly in extensions) we should see if we can set the right mime type for the wasm. If we cache the wasm in the Service Worker cache (e.g. #388), we may be able to set the mime type of the cached Response locally as well.

image

www/js/lib/xzdec_wrapper.js
```diff
@@ -20,11 +20,56 @@
  * along with Kiwix (file LICENSE-GPLv3.txt).  If not, see <http://www.gnu.org/licenses/>
  */
 'use strict';
-define(['xzdec'], function() {
-    // DEV: xzdec.js emits a global Module variable, which cannot be set in requireJS function line above, though it can be loaded in definition
-    var xzdec = Module;
-    xzdec._init();
-    
+
+// DEV: Put your RequireJS definition in the rqDef array below, and any function exports in the function parenthesis of the define statement
```
Jaifroid, line 24: rqDefXZ array

Jaifroid:
At the moment we load both zstddec and xzdec (either wasm or asm) at startup of the app, regardless of whether the compression type is encountered. This probably makes sense when there are lots of xz-encoded ZIMs still around. But I wonder if we might think about loading one or the other dynamically when the compression-type is first encountered in a ZIM, which would mean requiring the decoder just-in-time during a cluster read. It may not be so easy, and maybe it's a separate issue?

mossroy:
I agree that there's no need to load both decoders in memory, as only one can be used at a time.
It would be better in a separate issue IMHO because we never tested to unload a decoder.
Of course, it would become obsolete if we manage to use libzim/kiwix-lib

Jaifroid:
OK, I opened #744 for this.

This PR needs to be tested very thoroughly, particularly with repeated article loads, repeated ZIM loads, and in both jQuery and SW modes, to be sure that the new compiles are working as intended. However, I'll take it out of draft now.

mossroy:
About the streaming feature of webassembly : as you said, it's not relevant for us as the wasm file is local.
It would be easier (and maybe faster) to just disable this streaming feature.
Unfortunately I did not find such option in emscripten.

It might be technically possible by using `WebAssembly.instantiate` instead of `WebAssembly.instatiateStreaming` (see https://developer.mozilla.org/fr/docs/Web/JavaScript/Reference/Global_Objects/WebAssembly), in our *dev-wasm.js
I did not test, though.
If it works, we should report that to emscripten, and consider adding a step in our compilation script to do that change, until it's implemented in emscripten.

www/js/lib/xzdec_wrapper.js
```diff
@@ -20,11 +20,56 @@
...
+     var xzdec;
+
+     var instantiateDecoder = function (instance) {
+         xzdec = instance;
+     };
+
+     XZ().then(instantiateDecoder)
+     .catch(function (err) {
+         console.debug(err);
+         if (/CompileError.+?WASM/i.test(err.message)) {
```
mossroy, line 61: I'm wondering is we should not always try to load the ASM version if the WASM fails, regardless of the error message (that could be different, depending on the browser).
If not, the log level should be higher IMHO when the error message does not match

www/js/lib/zimfile.js
```diff
@@ -21,6 +21,20 @@
  */
 'use strict';
 
+/**
+ * Add Polyfill currently required by IE11 to run zstddec-asm and xzdec-asm
+ * See https://github.com/emscripten-core/emscripten/issues/14700
+ * If this is resolved upstream, remove this polyfill
```
mossroy, line 27: It would be good to say where this code has been copied from (as it's not ours). I know it's only a few lines, but it's always better

mossroy (commented):
I could not test, just added a few comments in the code

www/js/lib/xzdec_wrapper.js
Jaifroid: OK, yes you're right, we don't need to know exactly why instantiation failed, just that it failed.

www/js/lib/zimfile.js
Jaifroid: Source (MDN) added.

Jaifroid:
I've had a look in `settings.js` and found 

* [MINIMAL_RUNTIME_STREAMING_WASM_COMPILATION = 0;](https://github.com/emscripten-core/emscripten/blob/main/src/settings.js#L1715) and 
* [MINIMAL_RUNTIME_STREAMING_WASM_INSTANTIATION = 0;](https://github.com/emscripten-core/emscripten/blob/main/src/settings.js#L1734).

However, oddly, it says that streamin instantiation is disabled by default. The only thing I can think is that we have not set the option for the MINIMAL_RUNTIME (above those two options).

It could be worth experimenting with? I don't know if it's necessary just to eliminate a fallback warning in browser consoles, and of course our code might not work with the minimal runtime as I think we possibly need XHR to load the WASM binary.

mossroy:
This branch works fine on my Firefox OS device. It chooses the ASM versions (as WASM is unsupported) and does not complain. I did not see any obvious performance difference.

It might be worth testing the "minimal runtime" option you suggest.
Anyway, I don't consider the console warning to be a blocker, if we don't manage to get rid of it.
If you prefer, this PR could be merged, and we might investigate further with another issue/PR.

Another improvement could be to display in the Settings section : 
- if WASM is supported by the browser
- which version (ASM or WASM) is used for each decoder
But it could be a different issue too, and is not a blocker at all

Jaifroid:
Yes, populating the API panel could provide good diagnostic info to savvy users.

Jaifroid:
I've added code to populate the API Status Panel. It was trickier than I thought to get right given the async nature of the errors happening. I've included in zstddec-wrapper and xzdec-wrapper some `throw` statements that can be uncommented to simulate what would happen if there is an error loading one or other of the decoders. To test error reporting, uncomment both statements in one or both of the files, save and reload fully (usually ctrl-shift-r in the dev panel, though sometimes even that doesn't defeat caching...). 

Some images below of what would be seen in a) an error situation, and b) normal operation. Please note that the actual decoder being used ('ZSTD' or 'XZ') in square brackets after the machine type ('WASM' or 'ASM') will only be displayed once a ZIM file has been loaded. It will always report the last decompression type used. Though extremely unlikely in practice, a ZIM could hypothetically use ZSTD for some clusters and XZ for some others. The only way to know the decompression type is at the moment a Decompressor is requested. Code to deal with any particular decompression type is set up in its wrapper, so the code shoudl be extensible.

image

image

Jaifroid:
I've run out of time to deal with the console warning, i.e. to experiment with different values when compiling xzdec and zstddec. We've made it fairly easy to do that (it's the testing that would be time-consuming). So I think it's best to set that as a new issue.

@mossroy Please let me know if the implementation of the Status Panel code is satisfactory for merging (in due course). I have tested it on IE11, Firefox latest, Edge Chromium. I have yet to test on Edge Legacy or UWP.

Jaifroid:
Just noticed a silly minor error... Will push a very small change.

www/js/lib/xzdec_wrapper.js
```diff
@@ -1,7 +1,7 @@
-/**
- * xzdec_wrapper.js: Javascript wrapper around compiled xz decompressor.
+﻿/**
+ * zstddec_wrapper.js: Javascript wrapper around compiled zstd decompressor.
```
Jaifroid, line 2: Revert incorrect header for this file (xzdec, not zstddec).

www/js/lib/uiUtil.js
```diff
@@ -429,6 +429,18 @@ define(rqDef, function() {
         });
     }
 
+    // Reports an error in loading one of the ASM or WASM machines to the UI API Status Panel
+    // This can't be done in aoo.js because the error occurs after the API panel is first displayed
```
Jaifroid, line 433: Funny typo - app.js, not aoo.js

www/js/lib/uiUtil.js
```diff
@@ -429,6 +429,18 @@ define(rqDef, function() {
...
+    function reportAssemblerErrorToAPIStatusPanel(decoderType, error) {
+        // Report error to API panel because error is produced asynchronously after panel is first displayed
```
Jaifroid, line 435: Redundant comment already explained in function header. Delete.

www/js/lib/zimfile.js
```diff
@@ -21,6 +21,32 @@
...
+            var pos = rawPos > 0 ? rawPos|0 : 0;
+            return this.substring(pos, pos + search.length) === search;
+        }
+    });
+}
+
+/**
+ * A global variable to track the assembler machine type and the last used decompressor (for reporting to the API panel)
+ * This is populated in the Emscripten wrappers
+ * @type {Object}
```
Jaifroid, line 42: Add object properties to type definition?

Jaifroid:
```
     * @property {String} assemblerMachineType The assembler machine type supported and/or loaded by this app: 'ASM' or 'WASM'
     * @property {String} decompressorLastUsed The decompressor that was last used to decode a compressed cluster (currently 'XZ' or 'ZSTD')
     * @property {String} errorStatus A description of any detected error

```

Jaifroid:
@mossroy Apart from the minor changes I've indicated in comments above, are you happy for me to squash-merge?
No rush -- it's August!

mossroy:
It seems to work well on my FirefoxOS device :-)
I did not look at the code yet

www/js/lib/zimfile.js
mossroy: Good idea

mossroy (approved):
I did not have a very deep look, but it looks good to me

www/js/lib/xzdec_wrapper.js
```diff
@@ -28,10 +28,12 @@ var rqDefXZ = [];
...
 // Select asm or wasm conditionally
 if ('WebAssembly' in self) {
-    console.debug('Using WASM xz decoder');
+    console.debug('Instantiating WASM xz decoder');
+    params.decompressorAPI.assemblerMachineType = 'WASM';
     rqDefXZ.push('xzdec-wasm');
 } else {
-    console.debug('Using ASM xz decoder');
+    console.debug('Instantiating ASM xz decoder');
+    params.decompressorAPI.assemblerMachineType = 'ASM';
```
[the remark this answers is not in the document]
Jaifroid: AFAIK, the full names are Asm.js and WebAssembly, but everywhere you see ASM and WASM for short, as in the binaryen encoder `asm2wasm`. It makes for a more concise panel if we can use ASM vs WASM as codes for quick recognition for anyone looking at the API panel. However, I can expand the console comments to give the full name.

www/js/app.js
```diff
@@ -464,6 +464,16 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
         settingsStoreStatusDiv.classList.remove('apiAvailable', 'apiUnavailable');
         settingsStoreStatusDiv.classList.add(params.storeType === 'none' ? 'apiUnavailable' : 'apiAvailable');
         apiPanelClass = params.storeType === 'none' ? 'card-warning' : apiPanelClass;
+        // Update Decompressor API section of panel
+        var decompAPIStatusDiv = document.getElementById('decompressorAPIStatus');
+        apiName = params.decompressorAPI.assemblerMachineType;
+        if (apiName && params.decompressorAPI.decompressorLastUsed) {
+            apiName += ' [&nbsp;' + params.decompressorAPI.decompressorLastUsed + '&nbsp;]';
+        }
+        apiPanelClass = params.decompressorAPI.errorStatus ? 'card-danger' : apiName ? apiPanelClass : 'card-warning';
```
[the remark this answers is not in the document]
Jaifroid: I'm wary of cluttering the API panel, as I think each API's info should fit on a single line at least in an average desktop browser. Therefore I'd be more in favour of using the orange colour for ASM. The panel is only really for people who understand something about HTML/JS, and those people will know what these messages mean (or can find out easily if they are interested). The average user will never look at this panel.

www/js/lib/xzdec_wrapper.js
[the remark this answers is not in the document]
mossroy: All right, no problem. Don't bother to change the console comments, they are already meaningful

www/js/app.js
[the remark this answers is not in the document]
mossroy: OK I agree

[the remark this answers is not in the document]
Jaifroid: I've implemented turning the colour orange (both the text and the panel) if the API is 'ASM', for which I've added a new class name `apiSuboptimal` in `app.css`. I've done a quick test with IE11, artificially setting the Service Worker availability to true in order to test that this class setting works, and it seems to. Ideally we'd need to test on a broswer that has Service Worker, but does not have WASM... Not sure if there is one.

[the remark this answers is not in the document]
mossroy: I tested on my Firefox OS device : it displays as expected "Decompressor API: ASM [ ... ]" in orange, as it does not support WASM.
BUT it's also in orange on Chromium and Firefox, with "Decompressor API: WASM [ ... ]"
It's probably because you test that `apiName === 'WASM'`, where there can be a suffix [ ... ] : I suppose it should be a "starts with" test instead of equality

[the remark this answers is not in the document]
Jaifroid: Apologies, I did this in a rush this afternoon and didn't actually test loading a ZIM... Will correct early tomorrow, so we can merge before release.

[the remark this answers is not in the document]
mossroy: Perfect, no problem!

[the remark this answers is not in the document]
Jaifroid: Hopefully now fixed. I simply moved the place where the decompressor type is added to the API name, so `apiName = params.decompressorAPI.assemblerMachineType` at the point of testing.

mossroy:
It's working fine on Firefox OS (orange) and latest Firefox/Chromium (green).
Ready to be merged for me

Jaifroid:
OK, will do.

PR #742, opened 2021-07-23
Update Karma to 6.3.4

Jaifroid:
This is to test whether we can upate Karma to 6.3.4 to eliminate vulnerabilities.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -12,7 +12,7 @@
         "karma-qunit": "^4.1.1",
         "karma-requirejs": "^1.1.0",
         "karma-sauce-launcher": "^4.3.4",
-        "nightwatch": "^1.5.1",
+        "nightwatch": "1.5.1",
         "qunit": "^2.13.0",
         "requirejs": "2.3.3"
       },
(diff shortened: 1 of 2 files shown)


Jaifroid:
Connection to the testing instance is refused because the server opens port 8080, but the tests attempt to connect on port 80. However, a search of all of our code fails to find a reference to port 80. All the server access in the tests is already 8080. I wonder where port 80 comes from.

The log shows the http-server starting on 8080:
```
>Run ./node_modules/.bin/http-server . &
Starting up http-server, serving .
Available on:
  http://127.0.0.1:8080
  http://10.1.1.52:8080
Hit CTRL-C to stop the server
```
And here are the refused attempts to connect on port 80:

image

@Krinkle Is this something you could kindly look at?

Krinkle:
@Jaifroid This patch is meant to update Karma. Karma is used for the QUnit tests, which are passing even in this pull request.

Screenshot

The failure is coming from the end-to-end test run via Nightwatch:

Screenshot

I suspect that the problem is due to the unrelated upgrades happening in the same patch. For example, nightwatch is updated from 1.5 to 1.7 by your patch currently. This is because it is specified as `^1.5` currently in package.json, which means "1.5 or later, within 1.x", and npm updates these automatically.

I would recommend removing the `^` or `~` symbol at the start of dependency versions in package.json (as I did previously for you with `requirejs`), and then run `rm -rf node_modules/ && npm install`. That should get you a similar patch, but only ugrading karma (and any indirect dependencies), and thus confirm that it passes with newer Karma.

Jaifroid:
@Krinkle Thank you for the advice, I'll try!

Jaifroid:
OK, I see what the issue is: to fix vulnerabilities, I would need to run `npm audit fix --force`, but this would then force-upgrade nightwatch to 1.7.7 because 1.5.1 depends upon packages that have 5 "high severity" vulnerabilities.... Doing that will cause the tests to fail. Looks a bit like Catch 22. 😯

image

Jaifroid:
Well at least Karma is now upgraded without issue, thanks to @Krinkle's advice. Maybe we need a separate issue to work on Nightwatch.

Jaifroid:
Thank you @Krinkle!

PR #749, opened 2021-08-22
Create codeql-analysis.yml

Jaifroid:
@mossroy This is a new security feature of GitHub. It works with JavaScript. A bot does some automatic code analysis after each commit to master. It could be useful. I thought we could try it out. We can always turn it off if it's not useful.

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -32,7 +32,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        language: [ 'cpp', 'javascript' ]
+        language: [ 'javascript' ]
         # CodeQL supports [ 'cpp', 'csharp', 'go', 'java', 'javascript', 'python' ]
         # Learn more:
         # https://docs.github.com/en/free-pro-team@latest/github/finding-security-vulnerabilities-and-errors-in-your-code/configuring-code-scanning#changing-the-languages-that-are-analyzed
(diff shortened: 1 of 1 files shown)


Jaifroid:
(I will remove the cpp analysis, which obviously fails, leaving only JS analysis.)

.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,71 @@
...
+    branches: [ master ]
+  pull_request:
+    # The branches below must be a subset of the branches above
+    branches: [ master ]
+  schedule:
+    - cron: '44 12 * * 3'
+
+jobs:
+  analyze:
+    name: Analyse
```
mossroy, line 25: Why did you change this to "Analyse" instead of the original "Analyze"?

mossroy (approved):
This can be interesting

.github/workflows/codeql-analysis.yml
Jaifroid: Because I'm a linguist, and etymologically "analyze" is completely incorrect spelling! The root is Greek -lysis, as in analysis, dialysis, and is not the same root as other -ise/ize verbs in English, French, and Romance languages. OED is very specific that "analyze" is wrong even in US English (though you see it everywhere)!

This is just me bing a pedant. 😯

mossroy: I had no idea and trust you 100% on this

Issue #752, opened 2021-08-22
Allow wasm-eval instead of unsafe-eval in the browser extension manifest


mossroy:
This is a follow-up of #751 : `unsafe-eval` was needed to allow WASM in chromium extensions.
The goal of this issue to allow only what is necessary.

Problem is that the CSP that forced us to do that has been introduced at the beginning of 2017 (https://codereview.chromium.org/2646713002/), and the `wasm-eval` has been implemented at the end of 2017 (https://bugs.chromium.org/p/v8/issues/detail?id=7041).
So switching to `wasm-eval` would break compatibility with the chromium-based browsers of 2017.

So I suggest to also try to trap this specific exception, and automatically fallback to ASM in this case. It would only occur for these browsers of 2017 (rare case), and would allow them to work anyway. It's not a big issue if they use ASM instead of WASM.

mossroy:
Maybe another way would be to sandbox the page that loads the wasm : https://developer.chrome.com/docs/extensions/mv2/sandboxingEval/
But this page is currently www/index.html so it's probably complicated to sandbox it

Jaifroid:
There are a couple of Chromium-only wasm-related CSP directives. What is needed under MV3 (Manifest V3) is incorporated into #984. Note that there is no eval allowed at all under that manifest version. Therefore, realistically, to run Zimit-based archives, we needed to introduce the PWA into the Chromium extension, as an option, alongside "local" use of the Service Worker. This code has now been merged into main.

PR #750, opened 2021-08-22
Update CHANGELOG for v3.2.0

Jaifroid:
I think it's best to add the TODO for the next version as part of the step for preparing source code for v3.3.0, and not show it in the tagged version of v3.2.0. However, if you don't agree, I'll happily replace it now.

PR #754, opened 2021-08-22
Add a CSP to allow WASM in Chromium extensions

### LINKED ISSUE #751: WASM compilation is disabled by default in Chromium extensions
The decompressor API is red and its status is "Error loading ZSTD decompressor!".
In the browser console :

> Could not instantiate any ZSTD decoder! RuntimeError: abort(CompileError: WebAssembly.instantiate(): Wasm code generation disallowed by embedder). Build with -s ASSERTIONS=1 for more info.
> at A (zstddec-wasm.js:15)
> at zstddec-wasm.js:18

And nothing works afterwards.

It comes from a CSP applied on Chromium extensions (not on Firefox extensions), for security reasons.
It can be overridden by adding a line in manifest.json, see https://stackoverflow.com/questions/48523118/wasm-module-compile-error-in-chrome-extension . It seems to work with both `unsafe-eval` or `wasm-eval`, and does not seem to annoy Firefox.

Is it risky to do that?
Should we use `unsafe-eval` (that probably works with all versions of Chromium, but opens more security risks) or `wasm-eval` (that opens less security risks, but might not work with all versions of Chromium)?

mossroy:
Fixes #751

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -37,7 +37,7 @@
         "scripts": ["webextension/backgroundscript.js"]
     },
 
-    "content_security_policy": "script-src 'self' 'wasm-eval'; object-src 'self'",
+    "content_security_policy": "script-src 'self' 'unsafe-eval'; object-src 'self'",
     
     "author": "mossroy",
     "homepage_url": "https://www.kiwix.org",
--- a/manifest.json
+++ b/manifest.json
@@ -1,7 +1,7 @@
 {
     "manifest_version": 2,
     "name": "Kiwix",
-    "version": "3.2",
+    "version": "3.2-WIP",
 
     "description": "Kiwix : offline Wikipedia reader",
     
--- a/manifest.json
+++ b/manifest.json
@@ -37,7 +37,7 @@
         "scripts": ["webextension/backgroundscript.js"]
     },
 
-    "content_security_policy": "script-src 'self' 'wasm-eval';",
+    "content_security_policy": "script-src 'self' 'wasm-eval'; ; object-src 'self'",
     
     "author": "mossroy",
     "homepage_url": "https://www.kiwix.org",
--- a/manifest.json
+++ b/manifest.json
@@ -36,6 +36,8 @@
     "background": {
         "scripts": ["webextension/backgroundscript.js"]
     },
+
+    "content_security_policy": "script-src 'self' 'wasm-eval';",
     
     "author": "mossroy",
     "homepage_url": "https://www.kiwix.org",
--- a/manifest.json
+++ b/manifest.json
@@ -1,7 +1,7 @@
 {
     "manifest_version": 2,
     "name": "Kiwix",
-    "version": "3.2-WIP",
+    "version": "3.2",
 
     "description": "Kiwix : offline Wikipedia reader",
     
@@ -37,7 +37,7 @@
         "scripts": ["webextension/backgroundscript.js"]
     },
 
-    "content_security_policy": "script-src 'self' 'wasm-eval'; ; object-src 'self'",
+    "content_security_policy": "script-src 'self' 'wasm-eval'; object-src 'self'",
     
     "author": "mossroy",
     "homepage_url": "https://www.kiwix.org",


mossroy:
The UI test now fail on edge (instead of edge44 this morning for #750, where edge was passing).
It's always the "Error ECONNRESET: socket hang up" symptom

Jaifroid:
Now it's firefox45 failing with ECONNRESET... (I just re-ran tests). Sign.

Jaifroid:
Tests passing now. Is it not necessary to trap any potential error with unsafe-wasm (as opposed to unsafe-eval), or do older browsers just ignore it? Also, I'm wondering why our fallback doesn't kick in if the Promise fails (without unsafe-wasm)... However, I haven't tested. I'll try to test on Edge Chromium shortly to be sure this won't block release.

Jaifroid:
I've now tested on Edge 93.0, and the unpacked extension based on this PR runs the WASM version without complaining. So as far as I am concerned, this can be merged. Do we need to do anything about pre-2017 Chromium?

Jaifroid (approved):

mossroy:
I realize I did not commit the right value : shame on me.
This PR was meant to use `unsafe-eval` (supported by all browsers) instead of `wasm-eval` (supported by chromium-based browsers since 2018). Switching to `wasm-eval` is for #752
Will fix

Jaifroid:
No problem, I'll test this version. The only thing is we need to be sure the Stores will allow extensions with unsafe-eval, though I guess we could always try and revert to usafe-wasm if the Chrome and/or Edge stores refuse.

Jaifroid:
Sorry, I meant *wasm-eval

Jaifroid:
Working fine in dev mode with `unsafe-eval`.

image

mossroy:
I had a look at why our javascript code does not gracefully fallback to ASM in this case.
In fact, it does trap the error with WASM, and tries to fallback to ASM.
But if you look inside zstddec-asm.js, you see that it tries to load a file zstddec-asm.wasm (that does not exist) if WebAssembly is supported by the browser. So I suppose it somehow calls the eval function that is forbidden by CSP

Jaifroid:
Oh, OK. Strange. I wonder if we need another compile flag to prevent that happening, something that says "use ASM.JS only". I'll see if there's an obvious flag in the settings.

Jaifroid:
Possibly we could try `DYNAMIC_EXECUTION=0`. See https://github.com/emscripten-core/emscripten/blob/main/src/settings.js#L1184. Description below seems to fit:

```
// When set to 0, we do not emit eval() and new Function(), which disables some functionality
// (causing runtime errors if attempted to be used), but allows the emitted code to be
// acceptable in places that disallow dynamic code execution (chrome packaged app,
// privileged firefox app, etc.). Pass this flag when developing an Emscripten application
// that is targeting a privileged or a certified execution environment, see
// Firefox Content Security Policy (CSP) webpage for details:
// https://developer.mozilla.org/en-US/Apps/Build/Building_apps_for_Firefox_OS/CSP
// When this flag is set, the following features (linker flags) are unavailable:
//  --closure 1: When using closure compiler, eval() would be needed to locate the Module object.
//  -s RELOCATABLE=1: the function Runtime.loadDynamicLibrary would need to eval().
//  --bind: Embind would need to eval().
// Additionally, the following Emscripten runtime functions are unavailable when
// DYNAMIC_EXECUTION=0 is set, and an attempt to call them will throw an exception:
// - emscripten_run_script(),
// - emscripten_run_script_int(),
// - emscripten_run_script_string(),
// - dlopen(),
// - the functions ccall() and cwrap() are still available, but they are restricted to only
//   being able to call functions that have been exported in the Module object in advance.
// When set to -s DYNAMIC_EXECUTION=2 flag is set, attempts to call to eval() are demoted
// to warnings instead of throwing an exception.
// [link]
var DYNAMIC_EXECUTION = 1;
```

mossroy:
Yes, it's certainly worth testing!

Jaifroid:
Drat, we already compile with `DYNAMIC_EXECUTION=0`!!!!! So I wonder why we're getting this error.

Jaifroid:
I wonder if the setting `--closure 1` overrides or conflicts with disabling dynamic execution.

Jaifroid:
I recompiled the ZSTD wasm version with `--closure 0`, but unfortunately it still blocks WASM loading when I remove the `unsafe-eval` in the CSP. I could keep hunting for settings that might help, or we could go ahead wiht `unsafe-eval` and see if it gets into the Stores? We could probably remove eval code ourselves from the WASM js, but then we'd have a non-reproducible compile (at least not without adding extra steps to the helper scripts). We could also disable the testing/fallback in the WASM javascript loader, which would have the same disadvantage, and would mean that Chromium extensions would always use ASM.js instead of WASM (which would be a shame, though would have no direct impact on the end user).

mossroy:
OK. Maybe we should merge this PR, release v3.2.0 and investigate improvements afterward, in #752
The security risk is there, but appears to be very low in our case.

Jaifroid:
Agreed.

PR #758, opened 2021-08-29
Update source code to v3.3-WIP

Jaifroid:
@mossroy I think this is all the places where we update the version number. This is for merge into master.

As a separate point, I propose that the implementation on GitHub pages should be set to the release tag, rather than to master with the source code. updated to -WIP. If you're happy with that, I can do that.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ Please note that this application has changed its name over time.
 It was first called Evopedia (and was using the file format of Evopedia).
 Then it was renamed Kiwix-html5 (and uses ZIM files), and then again was renamed to Kiwix-JS.
 
+## Kiwix-JS v3.3.0
+
+Released on *TODO*
+
 ## Kiwix-JS v3.2.0
 
 Released on *2021-08-22*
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,10 @@ Please note that this application has changed its name over time.
 It was first called Evopedia (and was using the file format of Evopedia).
 Then it was renamed Kiwix-html5 (and uses ZIM files), and then again was renamed to Kiwix-JS.
 
+## Kiwix-JS v3.3.0
+
+Released on *TODO*
+
```
mossroy, line 10: Maybe you could prepare the line for the detailed changelog too, like in https://github.com/kiwix/kiwix-js/commit/99ec91fb495dfbb1d4f34c1ed8774f938e1d6800#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4edR11

mossroy (approved):
Looks good.

You're right about what we publish on gh-pages : it would be much better to have the release version

Jaifroid:
JFYI I hard-reset the gh-pages branch to tag 3.2.0. However, it still has the -WIP on the end of the header title, because that was in the tag. Maybe for the next branch we can create a tagged release without -WIP, and only insert it when we update the source code to the next version. It's very minor.

Jaifroid:
Hmm. Tests keep failing, despite restarting three times. Clearly nothing to do with this PR. I'll wait for Sauce Labs to stabilize.

PR #759, opened 2021-09-03
Document workaround for image download

### LINKED ISSUE #694: Can't save an image in the Kiwix extension
You can't save an image from a Wikipedia article in the Kiwix web browser **extension**.
After clicking on the **Save Image line**, absolutely nothing happens:
Kiwix Save Image

mossroy:
Fixes #694

### DIFF
--- a/www/index.html
+++ b/www/index.html
@@ -104,6 +104,7 @@
                         <li><a href="#aboutApp">About this App</a></li>
                         <li><a href="#usage">Usage</a></li>
                         <li><a href="#searchSyntax">Title Search Usage</a></li>
+                        <li><a href="#imageDownload">Image Download</a></li>
                         <li><a href="#privacy">Privacy Policy</a></li>
                         <li><a href="#expert">Expert Settings and Technical Information</a>
                             <ul>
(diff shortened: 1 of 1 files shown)


www/index.html
```diff
@@ -209,6 +209,17 @@ <h3 id="searchSyntax">Title search usage</h3>
                     </p>
                     <p style="text-align: right"><a href="#contents">↑ Back to Contents</a></p>
                     
+                    <h3 id="imageDownload">Image download</h3>
+                    <p>
+                        It's unfortunately not possible to download an image (from a page of a ZIM file) with the standard feature of browsers (usually right-click -> "Save
```
Jaifroid, line 214: Very slight modifications for style:

> Unfortunately it is not possible to download an image (from an article in a ZIM archive) using the standard method in most browsers (usually right-click -> "Save image as...").

www/index.html
```diff
@@ -209,6 +209,17 @@ <h3 id="searchSyntax">Title search usage</h3>
...
+                        image as..."). See <a href="https://github.com/kiwix/kiwix-js/issues/694">issue 694</a>.
+                    </p>
+                    <p>
+                        As a workaround, it's possible to choose "Copy Image" in the context menu (after the right-click), then paste it in any suitable software (Paint, Gimp
+                        etc, or even a Word processor), then save it.
```
Jaifroid, line 219: Slight revision to indicate that the workaround is for browser extensions (e.g. right-click doesn't work in UWP apps):

> One workaround (at least in browser extensions) is to choose "Copy Image" in the context menu (after right-click or long-press)...

Jaifroid:
@mossroy I've made some suggestions for slight revision, but otherwise this looks great. I wondered if you thought this section merits an entry in the Contents at the top of the page?

Jaifroid (approved):

mossroy:
Thanks for your suggestions, I also added it in Contents.

PR #765, opened 2021-09-22
Tweak regular expression that detects ZIM URIs

Jaifroid:
This is a hypothetical but realistic bugfix. I realized that the ZIM URI detection in jQuery mode should theoretically exclude some other potential URI types that could be found in some ZIMs. We currently exclude only URIs beginning 'http' (to include 'https') from processing as ZIM URIs -- i.e. URIs that point to external resources. However, this could lead to errors or unnecessary processing with URIs with the protocol 'data:'. These are perfectly conceivable for either CSS or images, so should be excluded explicitly from processing.

I know that Firefox OS doesn't allow data URIs for CSS (I think), but does allow them for images (I think) -- or is it the other way round? But that is rather specific to that OS.

I did wonder also whether we should add 'javascript:' as another protocol that can (and is) found in ZIMs. But maybe it's better to disable JavaScript links in jQuery mode, as they could reference scripts that are not loaded in that mode, and I think they are not allowed to run by some CSPs in any case (they may still be allowed in the iframe).

To be clear, I don't classify this as an 'enhancement' to jQuery mode. It's a small tweak to prevent potential bugs especially with newer ZIM types. Unfortunately I haven't identified an actual case. There are some data URIs in CSS, but we currently don't process CSS at all.

### DIFF
--- a/www/js/app.js
+++ b/www/js/app.js
@@ -1284,10 +1284,11 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
     // Regex below finds images, scripts, stylesheets and tracks with ZIM-type metadata and image namespaces [kiwix-js #378].
     // It first searches for <img, <script, <link, etc., then scans forward to find, on a word boundary, either src=["'] or href=["']
     // (ignoring any extra whitespace), and it then tests the path of the URL with a non-capturing negative lookahead that excludes
-    // URLs that begin 'http' (i.e. non-relative URLs). It then captures the whole of the URL up until either the opening delimiter
-    // (" or ', which is capture group \3) or a querystring or hash character (? or #). When the regex is used below, it will be further
-    // processed to calculate the ZIM URL from the relative path. This regex can cope with legitimate single quote marks (') in the URL.
-    var regexpTagsWithZimUrl = /(<(?:img|script|link|track)\b[^>]*?\s)(?:src|href)(\s*=\s*(["']))(?!http)(.+?)(?=\3|\?|#)/ig;
+    // URLs that begin 'http' (i.e. non-relative URLs) or 'data:' images/css. It then captures the whole of the URL up until either
+    // the opening delimiter (" or ', which is capture group \3) or a querystring or hash character (? or #). When the regex is used
+    // below, it will be further processed to calculate the ZIM URL from the relative path. This regex can cope with legitimate single
+    // quote marks (') in the URL.
+    var regexpTagsWithZimUrl = /(<(?:img|script|link|track)\b[^>]*?\s)(?:src|href)(\s*=\s*(["']))(?!http|data:)(.+?)(?=\3|\?|#)/ig;
     // Regex below tests the html of an article for active content [kiwix-js #466]
     // It inspects every <script> block in the html and matches in the following cases: 1) the script loads a UI application called app.js;
     // 2) the script block has inline content that does not contain "importScript()", "toggleOpenSection" or an "articleId" assignment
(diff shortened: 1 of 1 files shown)


Jaifroid:
While we're at it, should we add other protocols: ftp: and gopher: ;-) ?

Or we could make it generic, and detect just a URI with a colon by using this negative lookahead: `(?![^:]+:)` (that means, "look ahead for anything that's **not** a colon at least once followed by a colon, and assert that the regex should **not** match this URI if that's the case". It would come here (NB untested):
```
var regexpTagsWithZimUrl = /(<(?:img|script|link|track)\b[^>]*?\s)(?:src|href)(\s*=\s*(["']))(?![^:]+:)(.+?)(?=\3|\?|#)/ig;
                                                                                             ^^^^^^^^^^
```

mossroy:
I agree with your classification of this issue/PR, no problem.

I would be in favor of a more generic regexp as you suggest.
According to https://en.wikipedia.org/wiki/Uniform_Resource_Identifier#Syntax , the scheme can only be alphanumeric or signs +, . or - : you should match this more precisely (instead of `[^:]+`), because colons are allowed in the path afterwards. Try with something like `[a-zA-Z][a-zA-Z0-9+.-]+` (untested)

Jaifroid:
OK, thanks. I'll test `(?![a-zA-Z][a-zA-Z0-9+.-]+:)`. The schema seems to suggest that `a:` would be an invalid protocol, whereas `aa:` would be valid, hence I'll use a `+` instead of a `*` as the repeat modifier before the colon.

Jaifroid:
I've tested in RegexBuddy with some test data. Because the regex is case-insensitive, we don't need to specify the upper-case ranges, the regex does this for us. So it's sufficient to have `[a-z][a-z0-9+.-]+:`.

Running it on the test data (from a ZIM article), it immediately identified a protocol we didn't think of: `href="geo:44.35,-68.2167"`. So it's a good idea to go generic here.

Jaifroid:
The reference to geo: links reminded me that in v2.2.0 we have this for the CHANGELOG:

`* NEW: Make geo: and tel: links work.`

I remember you working on this, but I can't find a reference to geo: or tel: links in the code (not explicitly). I think this means they are handled in the same way as external http(s): links, and opened in a new tab (in jQuery mode). The closed issue for it is #313, and it is referenced as fixed in #319. 

If we are changing the position of these links in jQuery mode (not putting them into a `data-kiwixurl`) attribute, then we need to be careful not to cause a regression with later processing of links in jQuery mode.

Jaifroid:
So, this code doesn't handle ZIM anchors: that is now done separately. Since we won't get image, script, link or track tags with geo: or tel:, but we might get them with data: or javascript:, I think the proposed code change will not have unintended consequences.

mossroy:
There are `tel:` links in wikivoyage ZIM files, at least wikivoyage_fr_all_maxi_2020-10.zim, at the end of most articles
They work well on my Firefox OS device, both in this branch and in master : it opens the Dialer of the phone.
These links are indeed opened in a new "tab" (a new window for FirefoxOS), which is not necessary but not a big issue.

mossroy (approved):
Looks good!
Ran a few basic tests on the Firefox OS device and did not see any regression

Jaifroid:
OK, thanks. I've tested in Edge UWP, Edge Chromium and IE11.

Jaifroid:
@mossroy Before merging, we should briefly consider an edge case. This is the case of a data URI that contains a WebP-encoded image. It would look like: ``.

The point of the current PR is to allow any `data:image` URIs (and CSS) to be rendered by the browser immediately without further processing, but in the case of old browsers that don't support WebP natively, the polyfill will not find WebP data URIs (in any case, it only supports actual files or `Uint8Arrays`, and doesn't work with data URIs even though it creates data URIs).

This is very much an edge case for old browsers which (with one exception) only work in jQuery mode. I think it would be an "enhancement" for rare cases rather than a bug fix, and not completely straightforward (it would require converting the data URI into a `Uint8Array`). So according to our policy we should NOT patch this. Do you agree?

mossroy:
I agree with you that we should not bother with this edge case, as it's very rare, and probably not for the main images

Issue #513, opened 2019-06-01
ZIM files larger than 2GB can not be handled by the wasm version of libzim/kiwix-lib


mossroy:
(description updated on 2021-10-30)

It's a serious problem as there are a lot of ZIM files larger than 2GB. For example : https://download.kiwix.org/zim/wikipedia/?C=S;O=D

A first upstream bug was filed : https://github.com/emscripten-core/emscripten/issues/5250 and is marked as fixed.
But it did not seem to be enough for our need.
I opened another one : https://github.com/emscripten-core/emscripten/issues/15394

Jaifroid:
Ah good, I read that thread with interest, as I was worried that maybe the whole file was being read into memory. Good to hear it may be an arbitrary limitation, and that WORKERFS doesn't work by reading the whole file into memory.

mossroy:
We should test again with a recent version of emscripten (>=1.38.35) because https://github.com/emscripten-core/emscripten/pull/8759 might have solved that

mossroy:
I'll need to switch to another emscripten docker image : https://hub.docker.com/r/trzeci/emscripten
But the recent emscripten versions are not available, they're working on it : https://github.com/trzecieu/emscripten-docker/issues/46

mossroy:
I created a new branch `emscripten-upgrade-for-large-file-support` (https://github.com/mossroy/libzim_wasm/tree/emscripten-upgrade-for-large-file-support) using this other emscripten docker image (version 1.38.43)
Our docker image can be built and used, but emscripten compilation of our code fails
with :
```
warning: unexpected argument type i32 at index 1 in call to '__fseeko',
 should be i64
warning: unexpected return type i32 in call to '__ftello', should be i6
4
warning: unexpected argument type i32 at index 1 in call to '__fseeko',
 should be i64
warning: unexpected argument type i32 at index 1 in call to '__fseeko',
 should be i64
warning: unexpected argument type i32 at index 3 in call to 'pread', sh
ould be i64
warning: invalid item (maybe a typo?) in EXPORTED_RUNTIME_METHODS: ALLOC_STATIC
error: asm2wasm seeing an invalid argument type at index 1 (this will not valida
te) in call from $__ZNSt3__213basic_filebufIcNS_11char_traitsIcEEE7seekoffExNS_8
ios_base7seekdirEj to $___fseeko (this is likely due to undefined behavior in C,
 like defining a function one way and calling it in another, which is important
to fix)
[...]

PR #766, opened 2021-10-28
Libzim experiments

mossroy:
It uses the libzim only for getting content (articles and assets), but without setting the MIME type for now (because the C code does send this information for now)
For each content requested, it first resolves redirections without libzim (because it's not implemented in our C code either)
All the search features are without libzim for now
Error handling is bad when using libzim

### DIFF
--- a/www/js/app.js
+++ b/www/js/app.js
@@ -2102,39 +2102,31 @@ function handleMessageChannelMessage (event) {
             console.warn('Title ' + title.replace(/^(.{1,160}).*/, '$1...') + ' not found in archive.');
             // DEV: We send null for the content, so that the ServiceWorker knows that the article was not found (as opposed to being merely empty)
             messagePort.postMessage({ action: 'giveContent', title: title, content: null, zimType: selectedArchive.zimType });
-        } else if (dirEntry.isRedirect()) {
-            selectedArchive.resolveRedirect(dirEntry, function (resolvedDirEntry) {
-                var redirectURL = resolvedDirEntry.namespace + '/' + resolvedDirEntry.url;
-                // Ask the ServiceWorker to send an HTTP redirect to the browser.
-                // We could send the final content directly, but it is necessary to let the browser know in which directory it ends up.
-                // Else, if the redirect URL is in a different directory than the original URL,
-                // the relative links in the HTML content would fail. See #312
-                messagePort.postMessage({ action: 'sendRedirect', title: title, redirectUrl: redirectURL });
-            });
+        } else if (dirEntry.isRedirect) {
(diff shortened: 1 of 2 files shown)


Jaifroid:
I have played a bit with this proof-of-concept. First of all, congratulations -- it's a great achievement to have been able to compile libzim with Emscripten and to have overcome the obstacles with stability and the amount of data returned.

While it does "seem" slower than our current implementation, it is not unacceptably slow (at least on a modern PC), and more work can be done to find out where the bottlenecks are. It looks like there could be some positive news coming in the not-too-distant future on #513, and that could well provide faster file access.

This is looking very positive!

mossroy:
I've rebased this branch, and updated the wasm of libzim, which now allows to use any size of ZIM file, and is based on libzim 7.2.2

mossroy:
Rebased on master

Jaifroid:
Out of curiosity as to whether the libzim port might improve #581, I tried the prototype on an Android phone (with latest Chromium browser). At the moment it does not seem able to do an article count, a search or display an article. I tried with a small Ray Charles and also with two large ZIMs (>4GB). It doesn't seem to be the 4GB issue, as the Ray Charles didn't work either. I didn't connect a debugger to the phone, so I'm not sure what the issue is, but thought I'd document this for further investigation.

Jaifroid:
Rebased this branch/PR on master.

Jaifroid:
Interestingly, I'm getting the same scroll-lock issue when using libzim to read a ZIM with a graphical front page as reported in https://github.com/kiwix/kiwix-js-windows/issues/475. It affects this branch in SW mode, but also jQuery mode (which does not use the libzim WASM).

Jaifroid:
@Rishabhg71 I've force-pushed this PR in order to get rid of the distracting spacing and layout differences, and also to revert the libzim versions to the same that are in main, so you'll need to reset your local copy to origin. You'll see that, because we already have code to load the libzim WASM for full-text search, actually this PR is very simple, only a few lines intervening in the message channel. For now, I think all you need to do is to make the logic conditional, based on the dropdown you're adding, so that users and developers can easily switch between current implementation and trying out the different libzim versions.

This will require some logic in zimArchive.js, where libzim is currently loaded. You can see the implementation in the PWA, which lets users choose between DEV / ASM / WASM versions of libzim (for the purposes of search, but the loading logic is the same even though your purpose will be to allow reading the contents of a ZIM with the chosen libzim version (as well as search).

Jaifroid:
Force-pushed again after rebase on main (with javascript-libzim v0.5).

Jaifroid:
Branch is now superseded by #1160.

Issue #764, opened 2021-09-17
Provide a workaround to enable ServiceWorker mode in Firefox Extension


Jaifroid:
This is an explanation form #727 of what this would look like:

@mossroy I think I have a workaround for Firefox Extension users that allows them to access the Service Worker mode. I've done a bit of experimenting today which confirms that:

* Service Worker mode works fine in a Firefox Extension if the extension opens a tab that points to a remote PWA (served over https);
* If the PWA is coded correctly (to cache its own code), the extension works entirely offline (with Service Worker of course). 

This means that I believe it would be feasible to adopt the same solution I've adopted with the UWP app, viz.:

* In jQuery mode the extension/app runs from local code (in this case from a moz-extension: URL);
* If the user turns on Service Worker, a banner can be shown which warns the user that one-time access to the secure server will be needed to enable SW mode, so that the app can cache its own code;
* If the user agrees, the app will access the remote server;
* On startup, if SW mode is still on, the app will bootstrap itself from local code to its cached (remote) code, effectively with a redirect or simply by opening the remote tab if it detects SW mode;
* If the user wishes to switch back to jQuery mode, the user will be alerted that the app will switch back to local code, and will be asked to give permission to proceed.

Below are some screenshots of how this works in the UWP app. I think it is a feasible model for Firefox Extensions. It would require #388 to be completed first. It is a workaround rathet than a solution, but one that is perfectly viable.

SWITCHING FROM JQUERY TO SW MODE:

![image](https://user-images.githubusercontent.com/4304337/120119534-18c5ce00-c190-11eb-975a-667f87c7c43d.png)

SWITCHING BACK FROM SW MODE TO JQUERY MODE:

![image](https://user-images.githubusercontent.com/4304337/120119556-41e65e80-c190-11eb-811f-3309f84d26c4.png)

To be clear, apart from some controlling code (in `backgroundscript.js`) that detects the mode of operation and opens the appropriate tab, what this workaround would do is to act as a glorified bookmark to the PWA version. The clever work is done by the Service Worker and the Cache API, not by the extension itself. But for the user, this makes little difference other than that they will notice an https:// address in their browser instead of a moz-extension: address. Above all, this workaround is confirmed to work fully offline once the app has cached its code.

_Originally posted by @Jaifroid in https://github.com/kiwix/kiwix-js/issues/727#issuecomment-851060197_

Jaifroid:
[...]

PR #771, opened 2021-10-30
Enable Service Worker mode in Firefox extension via workaround

Jaifroid:
This is the PR for #764. It enables accessing an implementation of Kiwix JS as an offline-first Progressive Web App.

It works like this:

* User must be in a Firefox extension
* User attempts to switch to Service Worker mode
* App warns user that the app will reload as a PWA and asks for confirmation
* App tests access to the remote server (defined in `init.js`)
* If server is available (or code is already pre-cached), app navigates to server address (and pre-caches code if necessary) - it also stores the address of the extension which launched it
* If user switches back to JQuery mode, a warning is shown and confirmation requested
* App navigates back to the stored extension address

For testing purposes, the address of the PWA is currently `http:/localhost:8080/`. **To test, change this (in `init.js`) to the address of your development server, if necessary.**


service-worker.js
```diff
@@ -59,54 +77,151 @@ var regexpExcludedURLSchema = /^(?:chrome-extension|example-extension):/i;
...
-    // "Claiming" the ServiceWorker is necessary to make it work right away,
-    // without the need to reload the page.
-    // See https://developer.mozilla.org/en-US/docs/Web/API/Clients/claim
-    event.waitUntil(self.clients.claim());
+    console.log('[SW] Claiming clients for current page');
+    // Check all the cache keys, and delete any old caches
+    event.waitUntil(
+        caches.keys().then(function (keyList) {
+            return Promise.all(keyList.map(function (key) {
+                console.log('[SW] Current cache key is ' + key);
```
Jaifroid, line 171: Change to console.debug

service-worker.js
```diff
@@ -59,54 +77,151 @@ var regexpExcludedURLSchema = /^(?:chrome-extension|example-extension):/i;
...
+    var cache = /\.zim\w{0,2}\//i.test(rqUrl) ? ASSETS_CACHE : APP_CACHE;
+    event.respondWith(
+        // First see if the content is in the cache
+        fromCache(cache, rqUrl).then(function (response) {
+            // The response was found in the cache so we respond with it 
+            return response;
+        }, function () {
+            // The response was not found in the cache so we look for it in the ZIM
+            // and add it to the cache if it is an asset type (css or js)
+            if (/\.zim\w{0,2}\//i.test(rqUrl) && regexpZIMUrlWithNamespace.test(rqUrl)) {
```
Jaifroid, line 199: If we change to prepending the ZIM name as defined in the ZIM's metadata, we should change this regex (as it won't need to account for split files).

service-worker.js
```diff
@@ -59,54 +77,151 @@ var regexpExcludedURLSchema = /^(?:chrome-extension|example-extension):/i;
...
-    // If event.respondWith() isn't called because this wasn't a request that we want to handle,
-    // then the default request/response behavior will automatically be used.
+                }).catch(function (msgPortData, title) {
+                    console.error('Invalid message received from app.js for ' + title, msgPortData);
+                    return msgPortData;
+                });
+            } else {
+                // It's not a ZIM URL
+                return fetch(event.request).then(function (response) {
+                  // If request was success, add or update it in the cache, but be careful not to cache the ZIM archive itself!
```
Jaifroid, line 214: "was successful" (not "was success")

service-worker.js
```diff
@@ -208,49 +326,51 @@ function removeUrlParameters(url) {
...
-    return caches.open(CACHE_NAME).then(function (cache) {
-        return cache.match(request).then(function (matching) {
+    if (!useCache && cache === ASSETS_CACHE) return Promise.reject('disabled');
+    return caches.open(cache).then(function (cacheObj) {
+        return cacheObj.match(requestUrl).then(function (matching) {
             if (!matching || matching.status === 404) {
                 return Promise.reject('no-match');
             }
-            console.log('[SW] Supplying ' + request.url + ' from ' + CACHE_NAME + '...');
+            console.log('[SW] Supplying ' + requestUrl + ' from ' + cache + '...');
```
Jaifroid, line 342: Maybe change to console.debug

service-worker.js
```diff
@@ -208,49 +326,51 @@ function removeUrlParameters(url) {
...
-function updateCache(request, response) {
+function updateCache(cache, request, response) {
     // Prevents use of Cache API if user has disabled it
-    if (!useCache) return Promise.resolve();
-    return caches.open(CACHE_NAME).then(function (cache) {
-        console.log('[SW] Adding ' + request.url + ' to ' + CACHE_NAME + '...');
-        return cache.put(request, response);
+    if (!useCache && cache === ASSETS_CACHE) return Promise.resolve();
+    return caches.open(cache).then(function (cacheObj) {
+        console.log('[SW] Adding ' + request.url + ' to ' + cache + '...');
```
Jaifroid, line 359: console.debug

www/article.html
```diff
@@ -2,6 +2,7 @@
 <html>
     <head>
         <meta charset="utf-8">
+        <meta http-equiv="Content-Security-Policy" content="default-src http: https: file: data: blob: chrome-extension: ms-appx-web: 'unsafe-inline' 'unsafe-eval';">
```
Jaifroid, line 5: Redundant from a previous test, remove.

www/index.html
```diff
@@ -6,6 +6,7 @@
         <title>Kiwix</title>
         <meta name="description" content="Offline Wikipedia reader">
         <meta name="viewport" content="width=device-width, initial-scale=1.0">
+        <meta http-equiv="Content-Security-Policy" content="default-src http: https: file: data: blob: chrome-extension: ms-appx-web: 'unsafe-inline' 'unsafe-eval';">
```
Jaifroid, line 9: Probably redundant?

www/index.html
```diff
@@ -55,7 +56,7 @@
         <section id="search-article" role="region">
             <header id="top">
                 <nav class="navbar navbar-expand-md bg-light" role="navigation">
-                    <a class="navbar-brand">Kiwix 3.3-WIP</a>
+                    <a id="appVersion" class="navbar-brand">Kiwix</a>
```
Jaifroid, line 59: Now setting this dynamically from the appVersion defined in init.js. We will need to change script that updates this programmatically when preparing a new release.

www/js/app.js
```diff
@@ -552,7 +577,15 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
...
             tmpMessageChannel.port1.onmessage = handleMessageChannelMessage;
             // Send the init message to the ServiceWorker, with this MessageChannel as a parameter
-            navigator.serviceWorker.controller.postMessage({'action': 'init'}, [tmpMessageChannel.port2]);
+            if (navigator.serviceWorker.controller) {
+                navigator.serviceWorker.controller.postMessage({
+                    'action': 'init'
+                }, [tmpMessageChannel.port2]);
+            } else {
+                console.error('The Service Worker is active but is not controlling the current page! We have to reload.');
+                window.location.reload();
```
Jaifroid, line 586: This is occasionally necessary when the Service Worker stalls (not properly restarted).

www/js/app.js
```diff
@@ -569,19 +602,38 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
...
-                // is indeed unregistered but still active...
-                // So we have to disable it manually (even if it's still registered and active)
-                navigator.serviceWorker.controller.postMessage({'action': 'disable'});
-                messageChannel = null;
+            if (params.extensionURL) {
+                // We are in an extension, and the user may wish to revert to local code
+                var message = 'This will switch to using locally packaged code only. Configuration settings may be lost.\n\n' +
+                'WARNING: App will re-load in JQuery mode!';
+                var launchLocal = function () {
+                    settingsStore.setItem('allowInternetAccess', false, Infinity);
```
Jaifroid, line 612: Storing allowInternetAccess is being used as a passthrough. It is more important in the local extension, where it will allow the app to go straight through to the Service Worker mode without checking first with the user. It is set here just to keep the logic synchronized.

www/js/app.js
```diff
@@ -650,23 +710,20 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
         }
         $('input:radio[name=contentInjectionMode]').prop('checked', false);
         $('input:radio[name=contentInjectionMode]').filter('[value="' + value + '"]').prop('checked', true);
-        contentInjectionMode = value;
         // Save the value in the Settings Store, so that to be able to keep it after a reload/restart
-        settingsStore.setItem('lastContentInjectionMode', value, Infinity);
+        settingsStore.setItem('contentInjectionMode', value, Infinity);
```
Jaifroid, line 714: Changing lastContentInjectionMode to contentInjectionMode simplifies passing this value between local and SW code. I think it is always synchronized with the current contentInjectionMode, so the description is not wrong.

mossroy (commented):
I could not test yet, and did not have a deep look at the code.
But here is a first batch of comments on the code

mossroy:
Regarding the automatic publishing of the PWA code, here are some info we might need to finish it (I put them here so that we don't forget) : 

- Documentation of the image made by Renaud, that is used by Github actions : https://github.com/openzim/docker-publish-action
- This builds our nginx image with the content in it, and publishes it on https://hub.docker.com/r/kiwix/kiwix-moz-extension
- There is a hook on docker hub that triggers when a new image is published. It updates the container https://moz-extension.kiwix.org/current (deployed with Sloppy), with the image that has the tag `latest` on docker hub

For testing this PR, we probably need to change the .github/workflows/publish-docker.yaml to make it run on all pushes (by adding "push, " before "published" in the release:types:

When we'll be ready to merge this PR, we should configure it to run only on tags, and configure the tag-pattern accordingly.

If we later decide that we should "freeze" the code of the PWA, we might do it by changing the Dockerfile, to include the different versions in it (besides the "current" directory) : an idea might be to uncompress the content of each file from https://download.kiwix.org/release/browsers/firefox/ (an XPI is a ZIP file)

service-worker.js
```diff
@@ -59,54 +77,151 @@ var regexpExcludedURLSchema = /^(?:chrome-extension|example-extension):/i;
...
+  "www/js/lib/zimArchive.js",
+  "www/js/lib/zimArchiveLoader.js",
+  "www/js/lib/zimDirEntry.js",
+  "www/js/lib/zimfile.js",
+  "www/js/lib/fontawesome/fontawesome.js",
+  "www/js/lib/fontawesome/solid.js"
+];
+
+// Conditionally load the correct assembler binaries
+if ('WebAssembly' in self) {
```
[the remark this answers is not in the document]
Jaifroid: Done (pending push).

www/js/app.js
```diff
@@ -84,14 +84,37 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
...
         'prefix': '', // A field to hold the original search string
         'status': '',  // The status of the search: ''|'init'|'interim'|'cancelled'|'complete'
         'type': ''    // The type of the search: 'basic'|'full' (set automatically in search algorithm)
     };
-    
+    appstate['updateNeeded'] = false; // This will be set to true if the Service Worker has an update waiting
+    // Set display of app version
+    document.getElementById('appVersion').innerHTML = 'Kiwix ' + params.appVersion;
+
+    // Apply any override parameters in querystring (done as a self-calling function to avoid creating global variables)
```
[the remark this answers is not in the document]
Jaifroid: We could pass all the user's settings to the PWA (and vice versa) this way for consistency of experience. This is why I've implemented it generically, though I haven't yet coded passing more settings than two that are vital. In practice, there are some core ones we might pass, like dark theme, and so on.

Currently being sent (and needed) are: the requested `contentInjectionMode` and the passthrough flag `allowInternetAccess`.

www/js/app.js
```diff
@@ -593,52 +645,60 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
...
-                    // We need to wait for the ServiceWorker to be activated
-                    // before sending the first init message
-                    var serviceWorker = reg.installing || reg.waiting || reg.active;
-                    serviceWorker.addEventListener('statechange', function(statechangeevent) {
-                        if (statechangeevent.target.state === 'activated') {
-                            // Remove any jQuery hooks from a previous jQuery session
-                            $('#articleContent').contents().remove();
-                            // Create the MessageChannel
+                } else {
+                    if (protocol === 'moz-extension:') {
```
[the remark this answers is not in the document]
Jaifroid: Actually I originally ran `launchMozillaExtensionServiceWorker()` from the `catch` function of the Promise (around line 690 below, probably shifted once I push review changes). However, this was causing me problems, including timing problems, so I shifted it here. I could certainly try again putting it in the `catch`, because some of the timing issues I was having were later resolved by different code, so maybe it would work cleanly now.

www/js/app.js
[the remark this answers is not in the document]
Jaifroid: I've added a comment (pending push).

www/js/app.js
```diff
@@ -569,19 +602,38 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
      * @param {String} value The chosen content injection mode : 'jquery' or 'serviceworker'
      */
     function setContentInjectionMode(value) {
+        contentInjectionMode = value;
         if (value === 'jquery') {
-            if (isServiceWorkerReady()) {
```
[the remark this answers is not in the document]
Jaifroid: Agreed. I'll check what the effects are on Chromium of not disabling it first.

Jaifroid:
I forgot to demonstrate how a user is notified that an update is waiting. See screenshot below. This will not show immediately on loading the app, but (if the client is online) will show after the user has navigated within the app (e.g. going to About tab and back to Configuration, or after loading a ZIM, browsing, and going back to Configuration). The user doesn't have to do anything to update. It will simply update itself next time the user starts the app.

This can be tested from localhost without committing, by changing the `appVersion` in `app.js` AND in `service-worker.js`. Or to test on the nginx server, do the same, commit, and dispatch a new Docker workflow. I have bumped to 3.2.2 to test this on the server.

www/js/app.js
```diff
@@ -64,34 +64,90 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
...
+    params['appVersion'] = '3.2.2'; // **IMPORTANT** Ensure this is the same as the version number in service-worker.js
+    // The PWA server (currently only for use with the Mozilla extension)
+    params['PWAServer'] = 'https://moz-extension.kiwix.org/current/'; // Include final slash!
+    // params['PWAServer'] = 'http://localhost:8080/'; // DEV: Uncomment this line (and adjust) for local testing
+    // A parameter to determine the Settings Store API in use
+    params['storeType'] = settingsStore.getBestAvailableStorageAPI();
     params['hideActiveContentWarning'] = settingsStore.getItem('hideActiveContentWarning') === 'true';
     params['showUIAnimations'] = settingsStore.getItem('showUIAnimations') ? settingsStore.getItem('showUIAnimations') === 'true' : true;
-    document.getElementById('hideActiveContentWarningCheck').checked = params.hideActiveContentWarning;
-    document.getElementById('showUIAnimationsCheck').checked = params.showUIAnimations;
```
Jaifroid, line 73: These UI settings have been moved to dedicated section below, as they need to be set after any overrides.
It is also neater to have all params in an uninterrupted list.

www/js/app.js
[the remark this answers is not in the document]
Jaifroid: OK, I've now checked, and we can't completely disable the Service Worker in case a user is accessing the app from an online implementation. In that case, it will now function as a PWA, and will need access to the Service Worker. This will be the case if a user accesses the implementation from https://kiwix.github.io/kiwix/kiwix-js, or (less likely) https://moz-extension.kiwix.org. However, I've added logic below to ensure that the ASSETS_CACHE is completely turned off and deleted. Note that a Chrome extension won't cache its own code in SW mode because it cannot use the Cache API, so we don't need to blacklist it explicitly.

www/js/app.js
```diff
@@ -650,25 +751,11 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
...
-        setContentInjectionMode(lastContentInjectionMode);
-    }
-    else {
-        setContentInjectionMode('jquery');
-    }
-    
-    var serviceWorkerRegistration = null;
-    
-    // We need to establish the caching capabilities before first page launch
-    refreshCacheStatus();
```
Jaifroid, line 671: We now instantiate the Service Worker towards the top of app.js, in the block "Set the State and UI Settings associated with parameters".

www/js/app.js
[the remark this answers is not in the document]
Jaifroid: Now done.

Jaifroid:
Summary (before I forget) of changes/improvements since initial version at Hackathon:

* Added a failsafe, so that the Extension will ask for confirmation if it detects that the last attempt to switch to Service Worker mode failed. This works by sending back a specific querystring to the Extension (from the PWA) signalling success. A minimal amount of code is loaded into a hidden iframe to process this and write it to the Extension's localStorage.
* Significant (or visually striking) app parameters are now passed between the Extension and the PWA (and vice versa).
* It proved impossible to pass the ASSETS_CACHE string early enough to the Service Worker, so it is now hard coded in both app.js and service-worker.js. I've instead added a check that will provide console.error notification if the values diverge.
* Added a redirect from the root of the remote server to the current implementation (though it is out of scope of the Service Worker, so the redirect itself will not work offline). This is neater than seeing the default welcome page.
* Added a reference to the workaround in the explanation about the difference between JQuery and SW modes (though I haven't documented thoroughly, I'm not sure it's necessary).
* Provided a functioning PowerShell script, `./scripts/Build-DockerContainer.ps1` which programmatically rebuilds the container using the Rest API. It would be simple to make a similar (or less elaborate) Linux script also using curl.

Jaifroid:
TODO (evolving list):

- [ ] Fix tests

Jaifroid:
@mossroy Would you say #772 should be part of this PR, or a separate PR?

service-worker.js
Jaifroid: I have changed the corresponding function in app.js so that it uses the zimFile property `name` that we define when the archive is first loaded. Therefore the filename will never be `.zimaa`, etc., any more, so we do not need to test for two letters after the .zim extension. There actually is no ZIM metadata (in the M namespace) that corresponds to the exact archive version. It would be possible to construct one from `name` and `date`, which might be preferable to relying on the abstracted filename? See https://wiki.openzim.org/wiki/Metadata

mossroy:
To, me #772 is a separate PR

service-worker.js
mossroy: I think that the standardized ZIM file name is enough

Jaifroid: OK

Jaifroid:
I've added a script to CI that tests the integrity of the duplicate values in app.js and service-worker.js. Below are examples of the script passing correctly, and the script failing when submitting deliberately erroneous data.

image

image

Jaifroid:
I've now completed my TODO list for this PR. I've also created a separate PR #774 to test the offline file list in CI, but that PR can only be reviewed and merged after this one. As far as I'm concerned, the implementation is complete. I would recommend that, once approved, this PR (together with #774) should form the basis of a new release, but without making Service Worker the default yet. It would be good to consider this to be experimental for a while, as less obvious gotchas may take some actual real-world use to flush out.

www/index.html
Jaifroid: Done.

mossroy:
I don't forget this PR. I'll try to have a look at it as soon as I can

www/js/app.js
```diff
@@ -569,19 +651,45 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
...
             }
             refreshAPIStatus();
-            // User has switched to jQuery mode, so no longer needs CACHE_NAME
-            // We should empty it to prevent unnecessary space usage
-            if ('caches' in window) caches.delete(CACHE_NAME);
+            // User has switched to jQuery mode, so no longer needs ASSETS_CACHE
+            // We should empty it and turn it off to prevent unnecessary space usage
+            if ('caches' in window && isMessageChannelAvailable()) {
+                var channel = new MessageChannel();
+                if (isServiceWorkerAvailable && navigator.serviceWorker.controller) {
```
mossroy, line 686: This line generates a blocking error on browsers that do not support ServiceWorkers (like my Firefox OS device).
It can be easily fixed this way : 
```suggestion
                if (isServiceWorkerAvailable && navigator.serviceWorker && navigator.serviceWorker.controller) {
```

mossroy: Apart from that, I did not see any regression on my Firefox OS device.
Of course, as it does not support Serviceworkers, I could not test this feature on it at all

www/js/lib/uiUtil.js
```diff
@@ -402,7 +473,7 @@ define(rqDef, function() {
         }
         // If we are in Config and a real document has been loaded already, expose return link so user can see the result of the change
         // DEV: The Placeholder string below matches the dummy article.html that is loaded before any articles are loaded
-        if (document.getElementById('liConfigureNav').classList.contains('active') &&
+        if (document.getElementById('liConfigureNav').classList.contains('active') && doc &&
```
mossroy, line 476: Is this change done on purpose?

www/js/lib/uiUtil.js
```diff
@@ -225,6 +225,77 @@ define(rqDef, function() {
...
+    function spinnerDisplay(show, message) {
+        var searchingArticles = document.getElementById('searchingArticles');
+        var spinnerMessage = document.getElementById('cachingAssets');
+        if (show) searchingArticles.style.display = 'block';
+        else searchingArticles.style.display = 'none';
+        if (message) {
+            spinnerMessage.innerHTML = message;
+            spinnerMessage.style.display = 'block';
+        } else {
+            spinnerMessage.innerHTML = 'Caching assets...';
```
mossroy, line 294: It's a bit strange to have this string here

scripts/create_all_packages.sh
```diff
@@ -58,7 +58,6 @@ else
    sed -i -e "s/$VERSION_TO_REPLACE/$MAJOR_NUMERIC_VERSION/" tmp/manifest.json
 fi
 sed -i -e "s/$VERSION_TO_REPLACE/$VERSION/" tmp/manifest.webapp
-sed -i -e "s/$VERSION_TO_REPLACE/$VERSION/" tmp/www/index.html
```
mossroy, line 61: Shouldn't we replace the version in app.js and service-worker.js?

www/js/app.js
```diff
@@ -64,34 +65,100 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
...
-    // Set parameters and associated UI elements from the Settings Store
-    // DEV: The params global object is declared in init.js so that it is available to modules
-    params['storeType'] = settingsStore.getBestAvailableStorageAPI(); // A parameter to determine the Settings Store API in use
+    /**
+     * Set parameters from the Settings Store, together with any defaults
+     * Note that the params global object is declared in init.js so that it is available to modules
+     * WARNING: Only change these paramaeters if you know what you are doing
+     */
+    // The current version number of this app
+    params['appVersion'] = '3.2.3'; // **IMPORTANT** Ensure this is the same as the version number in service-worker.js
```
mossroy, line 74: Shouldn't we use the same "3.3-WIP" string as elsewhere?
I suppose you used these version numbers for testing purpose?

www/js/app.js
```diff
@@ -64,34 +65,100 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
...
     // A parameter to set the app theme and, if necessary, the CSS theme for article content (defaults to 'light')
     params['appTheme'] = settingsStore.getItem('appTheme') || 'light'; // Currently implemented: light|dark|dark_invert|dark_mwInvert
-    document.getElementById('appThemeSelect').value = params.appTheme;
-    uiUtil.applyAppTheme(params.appTheme);
     // A global parameter to turn on/off the use of Keyboard HOME Key to focus search bar
     params['useHomeKeyToFocusSearchBar'] = settingsStore.getItem('useHomeKeyToFocusSearchBar') === 'true';
-    document.getElementById('useHomeKeyToFocusSearchBarCheck').checked = params.useHomeKeyToFocusSearchBar;
-    switchHomeKeyToFocusSearchBar();
+    // A parameter to access the URL of any extension that this app was launched from
+    params['extensionURL'] = settingsStore.getItem('extensionURL');
```
mossroy, line 91: It might be more meaningful to have a name that explains it's the URL of the extension that called this app (in case of a PWA)

www/js/app.js
```diff
@@ -64,34 +65,100 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
...
+
     // An object to hold the current search and its state (allows cancellation of search across modules)
     appstate['search'] = {
         'prefix': '', // A field to hold the original search string
         'status': '',  // The status of the search: ''|'init'|'interim'|'cancelled'|'complete'
         'type': ''    // The type of the search: 'basic'|'full' (set automatically in search algorithm)
     };
+
+    // A Boolean to store the update status of the PWA version (currently only used with Firefox Extension)
+    appstate['updateNeeded'] = false; // This will be set to true if the Service Worker has an update waiting
```
mossroy, line 103: Same thing : here. It might be renamed to `pwaUpdateNeeded` for example

www/js/app.js
```diff
@@ -64,34 +65,100 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
...
     
+    /**
+     * Apply any override parameters that might be in the querystring.
+     * This is used for communication between the PWA and any local code (e.g. Firefox Extension), both ways.
+     * It is also possible for DEV (or user) to launch the app with certain settings, or to unset potentially
+     * problematic settings, by crafting the querystring appropriately.
+     */
+    (function overrideParams() {
+        var rgx = /[?&]([^=]+)=([^&]+)/g;
```
mossroy, line 112: Use a more meaningful variable name, that gives the purpose of the regexp

www/js/app.js
```diff
@@ -64,34 +65,100 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
...
+            frame.id = 'kiwixComm';
+            frame.style.display = 'none';
+            document.body.appendChild(frame);
+            frame.src = params.extensionURL + '/www/index.html'+ message;
+            // Now remove redundant frame. We cannot use onload, because it doesn't give time for the script to run.
+            setTimeout(function () {
+                var kiwixComm = document.getElementById('kiwixComm');
+                // The only browser which does not support .remove() is IE11, but it will never run this code
+                if (kiwixComm) kiwixComm.remove();
+            }, 3000);
```
mossroy, line 144: Instead of creating a dummy iframe, waiting an arbitrary time etc, maybe you could simply use an XHR request?

www/js/app.js
```diff
@@ -64,34 +65,100 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
...
+        }
+    })();
+
+    /**
+     * Set the State and UI settings associated with parameters defined above
+     */
+    document.getElementById('hideActiveContentWarningCheck').checked = params.hideActiveContentWarning;
+    document.getElementById('showUIAnimationsCheck').checked = params.showUIAnimations;
+    document.getElementById('titleSearchRange').value = params.maxSearchResultsSize;
+    document.getElementById('titleSearchRangeVal').innerHTML = params.maxSearchResultsSize;
```
mossroy, line 154: To avoid the code scanning warning, you could escape the content before putting it in `.innerHTML` (even if, in our case, we know it will always be a number)

www/js/app.js
```diff
@@ -485,9 +553,9 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
      * If Service Worker is not available, the attributes of the memory cache are returned instead
      * @returns {Promise<Object>} A Promise for an object with cache attributes 'type', 'description', and 'count'
      */
-    function getCacheAttributes() {
+    function getAssetsCacheAttributes() {
         return new Promise(function (resolve, reject) {
-            if (contentInjectionMode === 'serviceworker') {
+            if (params.contentInjectionMode === 'serviceworker' && navigator.serviceWorker.controller) {
```
mossroy, line 558: It might be not strictly necessary here, but you might test `navigator.serviceWorker` too, before testing `navigator.serviceWorker.controller`

www/js/app.js
```diff
@@ -497,17 +565,17 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
                     else resolve(cache);
                 };
                 // Ask Service Worker for its cache status and asset count
-                navigator.serviceWorker.controller.postMessage({
+                if (navigator.serviceWorker.controller) navigator.serviceWorker.controller.postMessage({
```
mossroy, line 568: Is this test necessary, as it already has been tested a few lines above?

mossroy (commented):
Today I found some time to look at the code.
I'll try to test later.

This is a really huge and complicated PR.
This complexity is not by accident : you have to handle a lot of different cases to reach our goal.

But I'm frightened by all the code it needs, and the difficulty we might have to maintain it.
I don't have an obvious solution for that, I'll have to think about it.

It also adds some weight to our already-very-heavy app.js file.
One day we'll have to find a way to split it in different pieces, with cleaner responsibilities

www/js/app.js
Jaifroid: OK, that's helpful, thank you.

Jaifroid: It seems anomalous that `isServiceWorkerAvailable` returns true on this device, but `navigator.serviceWorker` returns false!

.github/workflows/CI.yml
```diff
@@ -24,6 +24,12 @@ jobs:
       # Clone the repo and checkout the commit for which the workflow was triggered
       - uses: actions/checkout@v2
 
+      - name: Test integrity of duplicate values
+        shell: bash
+        run: |
+          chmod +x ./scripts/test_duplicate_values.sh
+          ./scripts/test_duplicate_values.sh
+
```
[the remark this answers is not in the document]
Jaifroid: #774 checks that the list of files in the Service Worker is consistent with the files in the filesystem. This test is different: it checks that certain duplicated values between `app.js` and `service-worker.js` are consistent with each other. It's a way to ensure a dev doesn't change a value in one place without changing a required corresponding value. It could be a separate PR, but conceptually, for me, it belongs with the changes being made here.

.github/workflows/publish-docker.yaml
```diff
@@ -0,0 +1,39 @@
+# Workflow to update docker image for moz-extension.kiwix.org (since docker autobuild has been disabled)
+
+name: Docker
```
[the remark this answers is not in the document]
Jaifroid: Ah, OK, I think this was copied from the provided workflow, but we could change to "Publish docker container for Mozilla extension".

index.nginx.html
```diff
@@ -0,0 +1,19 @@
+<!doctype html>
```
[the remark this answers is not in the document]
Jaifroid: I can try.

scripts/Build-DockerContainer.ps1
```diff
@@ -0,0 +1,92 @@
+param (
```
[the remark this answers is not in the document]
Jaifroid: It uses the REST API to dispatch the Docker build and update the container in the cloud, with appropriate values extracted from the app's code. It would be equivalent to a set of grep commands to read values from the file system followed by a curl command that dispatches the build.

This script isn't strictly necessary. It's just a utility for a developer (me!) who needs to update and test from a specific branch. It's possible to dispatch the docker build from the GitHub workflows menu (on GitHub) instead. But this does a few checks for consistency, so for me it's reassuring to have.

I will add a description.

scripts/test_duplicate_values.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/bash
```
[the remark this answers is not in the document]
Jaifroid: See comment above. For me it feels like this belongs to this PR.

service-worker.js
```diff
@@ -59,54 +77,144 @@ var regexpExcludedURLSchema = /^(?:chrome-extension|example-extension):/i;
...
+];
+
+// Process install event
+self.addEventListener("install", function (event) {
+    console.debug("[SW] Install Event processing");
+    // DEV: We can't skip waiting because too many params are loaded at an early stage from the old file before the new one can activate...
+    // self.skipWaiting();
+    // We try to circumvent the browser's cache by adding a header to the Request
+    var requests = precacheFiles.map(function (url) {
+        // Ensuring all files are explicitly versioned via the querystring helps to prevent browser caching too
```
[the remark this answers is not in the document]
Jaifroid: Yes, I guess it says the same thing but adds some detail on the mechanism. I can unify the comments.

service-worker.js
```diff
@@ -59,54 +77,144 @@ var regexpExcludedURLSchema = /^(?:chrome-extension|example-extension):/i;
...
+    "www/js/lib/zstddec-wasm.wasm"
+];
+
+// Process install event
+self.addEventListener("install", function (event) {
+    console.debug("[SW] Install Event processing");
+    // DEV: We can't skip waiting because too many params are loaded at an early stage from the old file before the new one can activate...
+    // self.skipWaiting();
+    // We try to circumvent the browser's cache by adding a header to the Request
+    var requests = precacheFiles.map(function (url) {
```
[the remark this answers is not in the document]
Jaifroid: OK

service-worker.js
```diff
@@ -50,7 +68,7 @@ var regexpCachedContentTypes = /text\/css|text\/javascript|application\/javascri
  * 'example-extension' is included to show how to add another schema if necessary
  * @type {RegExp}
  */
-var regexpExcludedURLSchema = /^(?:chrome-extension|example-extension):/i;
+var regexpExcludedURLSchema = /^(?:file|chrome-extension|example-extension):/i;
```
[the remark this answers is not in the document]
Jaifroid: This disables use of Cache API when a particular schema is used. Currently listed are those that cause exceptions if an attempt is made to access Cache. Would we want a moz-extension schema that was fully supported by Firefox not to be able to use Cache API? In some ways it's immaterial, because the code will never currently run, and it's hare for us to test what would happen if Service Workers were supported in Firefox extensions!

service-worker.js
```diff
@@ -59,54 +77,144 @@ var regexpExcludedURLSchema = /^(?:chrome-extension|example-extension):/i;
...
+    if (cache === ASSETS_CACHE && !fetchCaptureEnabled) return;
+    event.respondWith(
+        // First see if the content is in the cache
+        fromCache(cache, rqUrl).then(function (response) {
+            // The response was found in the cache so we respond with it 
+            return response;
+        }, function () {
+            // The response was not found in the cache so we look for it in the ZIM
+            // and add it to the cache if it is an asset type (css or js)
+            if (/\.zim\//i.test(rqUrl) && regexpZIMUrlWithNamespace.test(rqUrl)) {
```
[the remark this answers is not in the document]
Jaifroid: OK, will test.

www/js/lib/uiUtil.js
Jaifroid: I'm not sure these changes belong in this PR, I can't remember why they've changed. I'll check. Maybe some incidental code cleanup.

scripts/create_all_packages.sh
Jaifroid: In index.html we set these dynamically, hence I removed this line. I think you're right that this script could be used to set the final release values in app.js and service-worker.js.

www/js/app.js
Jaifroid: Yes, it was for testing. Will revert.

www/js/app.js
Jaifroid: OK.

www/js/app.js
Jaifroid: I can try!

www/js/app.js
Jaifroid: I'll check, it could be left over from a previous iteration of the code.

Jaifroid:
Thanks you very much, @mossroy, for taking the time to review this PR and for all the helpful suggestions. On the code length and complexity, I think this is because what is effectively done by this PR is to enable this app to run generically as a PWA. One way to think of it is that it future-proofs the app should we ever decide to expose and advertise the ability to run (and, with a little bit more code, install) the app as a PWA. This technology represents a viable alternative to the browser extension format, and gives us options in the future. Obviously the immediate benefit is allowing Firefox Extension users to run Service Worker mode, which unblocks a lot of progress we want to make with this app, in particular making SW mode the default and deprecating jQuery mode in a graceful way.

It would be a very small step from this PR to making the app generically available as an installable, offline-first PWA. So I don't think we need to consider this PR only as a workaround for an edge case. What we are doing is using that edge case as a way of introducing and (cautiously) testing a modern technology in the app that could increase access and discoverability in the future.

www/js/app.js
Jaifroid: OK, I seem to remember why I didn't choose XHR. It's because we are making a cross-origin request, and I believed that only iframes are allowed to load cross-origin sites (and we don't need to access the contents of the iframe, merely send a one-way message).  It could be that the message would be received with XHR even if the response is blocked? I could try XHR to find out, if it is conceptually neater than using an iframe?

www/js/app.js
mossroy: That's right!
In fact, it's because the existing code was not calling the function `isServiceWorkerAvailable`, but only testing it existed...

So I suggest this fix instead (tested on my Firefox OS device)

```suggestion
                if (isServiceWorkerAvailable() && navigator.serviceWorker.controller) {
```

.github/workflows/CI.yml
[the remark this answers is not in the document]
mossroy: OK, sorry for the confusion.
You're right : it should belong to this PR

www/js/app.js
mossroy: It's worth testing with XHR. It would be much cleaner in my opinion, at least because we can handle errors on this HTTP call.
I also dislike the `setTimeout` here because we have no guarantee on how much time the HTTP call can take : in worst case we might close the iframe before the HTTP call could be sent

index.nginx.html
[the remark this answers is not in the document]
Jaifroid: I've moved this and have tested that it builds correctly by dispatching the workflow.

Jaifroid:
@mossroy There is a practical issue I think we need to decide on regarding how we test new releases once we have a published release. Currently, as we have no published release, there is no impediment to testing out our code by pushing it to the /current/ directory on the container. However, this won't be possible after it is published, since we mustn't push test code to a live implementation.

Therefore, I've been thinking that we need to be able to push code from master to the "current" directory, and code from any branch to another directory, possibly "dev". However, we need to be able to push both at the same time, as we create the container from scratch each time we run the `docker-publish-action`.

How would you recommend we deal with this? I noticed the openzim/docker-publish-action we are using supports passing `build-args` (see documentation at https://github.com/openzim/docker-publish-action/blob/main/README.md), so we can avoid hard-coding the `current` directory in the dockerfile (`dockerfile-moz-extension.pwa`) and use a variable instead (see https://stackoverflow.com/a/37639938). But there may be an obvious and simpler solution I'm overlooking...

This could be a separate issue.

www/js/app.js
Jaifroid: Oops, silly me! Thanks for spotting that.

Jaifroid:
@mossroy I've thought of a solution regarding my comment above (about maintaining a test server) that would keep things fairly simple. If we used the GitHub Pages implementation of this app as our test server, then it would mean we could test new releases by pointing the extension to that server, and then switch to the production server (the docker-served version) on release. It would mean that the GitHub Pages version would not always remain synchronized with new releases, and would sometimes have test code on it, but we could make that clear in the README. Would that be a good way forward?

Jaifroid:
I've added some logic to the `publish-docker.yaml` to enable overriding the `appVersion` in the app if the tag or the override value matches `^[0-9.]+`, i.e. if it looks like a deliberate release. This rewrites the `appVersion` declared in app.js and in `service-worker.js` (only in the container version, not in the Repo). It also can be dispatched from the GitHub workflow (select this branch as it's not in master yet):

image

service-worker.js
```diff
@@ -59,54 +77,144 @@ var regexpExcludedURLSchema = /^(?:chrome-extension|example-extension):/i;
...
-        );
-    }
-    // If event.respondWith() isn't called because this wasn't a request that we want to handle,
-    // then the default request/response behavior will automatically be used.
+                }).catch(function (msgPortData, title) {
+                    console.error('Invalid message received from app.js for ' + title, msgPortData);
+                    return msgPortData;
+                });
+            } else {
+                // It's not a ZIM URL
```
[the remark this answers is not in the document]
Jaifroid: This is OK, we still want to fetch it, and then we decide if we want to cache it based on a more refined test below. I've updated comment (pending push) to be clearer.

www/js/app.js
Jaifroid: Done

www/js/lib/uiUtil.js
Jaifroid: Yes, this was done on purpose because I found an exception was being thrown on doc.title in certain circumstances before a ZIM was loaded, and it blocked loading of configuration values, I believe when communicating across implementations.

www/js/lib/uiUtil.js
Jaifroid: These functions do belong here. I have to show the spinner when the user requests to use Service Worker in the extension, while the app is checking that it has access to the server. This can take a few seconds and can be disconcerting otherwise. So I tidied up the spinnerDisplay routine in the app and made a function for it, which allows simplification of code at that point (and elsewhere in the app).

www/js/app.js
Jaifroid: Removed.

www/js/app.js
Jaifroid: @mossroy I've tried to use XHR with the following code:
```
            var commReq = new XMLHttpRequest();
            var commUrl = params.referrerExtensionURL + '/www/index.html'+ message;
            commReq.addEventListener('load', function (e) {
                console.debug('XHR should have loaded', e);
            });
            commReq.addEventListener('error', function (err) {
                console.error('There was an error communicating with the extension!', err);
            });
            commReq.open('GET', commUrl);
            commReq.send();
```
Unfortunately, it doesn't work. No error is thrown, and the URL seems to load successfully (I get the console message), but it is merely loading index.html and not running attached JavaScript, so init.js in the extension never loads in order to process the querystring and receive the success message.

This is in fact why I chose iframe in the first place, because an iframe can load, sandboxed, a page and run JS on the page. According to https://stackoverflow.com/questions/3728798/running-javascript-downloaded-with-xmlhttprequest, XMLHttpRequest can't do this: we would have to eval any scripts on the page, but they would run in the wrong context, so it would be useless (as well as dangerous). 

An answer [here](https://stackoverflow.com/questions/41110413/javascript-not-executing-when-loaded-via-xmlhttprequest-but-does-when-loaded-wit) suggests that jQuery can do a hack to run javascript from the remote page, but if you look at the source of the jquery function, you see that it loads scripts into the current document, which is precisely what we don't want to happen (wrong context again).

So I think we're stuck with the iframe method. I took care only to load index.html and init.js (not the whole app), and also it's communicating with a local extension, not over the Web, so 3 full seconds in the timeout should be enough time even on the slowest systems before removing the iframe. However, I can increase it to 5 seconds, or even 10 seconds if you prefer. The user won't notice, it'll be eliminated in the background.

Jaifroid:
@mossroy I've completed the current round of code review as far as is possible (see [comment on XHR issue](https://github.com/kiwix/kiwix-js/pull/771#discussion_r771967688) above as the only issue I couldn't resolve, but for which I think there are good reasons).

I'd like to push the updated `publish-docker.yaml` to master now, if that's OK, as it is needed to remove the code conflict between this branch and master. The one we have on master is now outdated.

mossroy:
> @mossroy I've thought of a solution regarding my comment above (about maintaining a test server) that would keep things fairly simple. If we used the GitHub Pages implementation of this app as our test server, then it would mean we could test new releases by pointing the extension to that server, and then switch to the production server (the docker-served version) on release. It would mean that the GitHub Pages version would not always remain synchronized with new releases, and would sometimes have test code on it, but we could make that clear in the README. Would that be a good way forward?

You're right about the issue.
Your proposal looks like a good idea.
We can easily change the github pages version by moving the tag gh-pages.
I don't think we have ever publicly mentioned this GitHub Pages version as an "official" version, so we don't have a strict commitment of stability on it

mossroy:
> I'd like to push the updated `publish-docker.yaml` to master now, if that's OK, as it is needed to remove the code conflict between this branch and master. The one we have on master is now outdated.

You can

mossroy:
I've tested this PR on Firefox, inside a temporary extension.
On first launch, it displays a popup saying "The last attempt to launch the PWA appears to have failed. Do you wish to try again?". I suppose it's a small bug? Or maybe because of my previous tests?

After that, it works nicely.
However, I've encountered a (probably minor) issue : I've used the ServiceWorker/PWA mode, then went offline. Then I used the Firefox feature to remove the local data of https://moz-extension.kiwix.org. After that, if I try to run the extension (while being offline), it uses the https://moz-extension.kiwix.org URL, which fails, and displays an error page (impossible to connect to this address). It's only since the second attempt to launch the extension that it properly detects that the PWA could not be launched previously, and suggests to try again (and allows to use jQuery mode). I understand it might be not possible to fix this...

I also tested it on Chromium, inside a temporary extension. There is a minor side effect : when I first launch it, and switch to ServiceWorker mode, the page is reloaded, I don't know why. It was not necessary before.

.github/workflows/CI.yml
```diff
@@ -24,6 +24,12 @@ jobs:
       # Clone the repo and checkout the commit for which the workflow was triggered
       - uses: actions/checkout@v2
 
+      - name: Test integrity of duplicate values
```
mossroy, line 27: It might be a good idea to be more specific on which duplicate values we are checking here

scripts/Build-DockerContainer.ps1
```diff
@@ -0,0 +1,120 @@
+# This is a utility script which helps developers choose sensible values for updating the online Docker-based implementation
+# of this app while testing and developing code in a specific branch. It checks app.js and service-worker.js for consistency,
+# and checks that that the underlying branch of a PR has been checked out (rather than the PR itself). It then calls the
```
mossroy, line 3: Typo : duplicated "that"

mossroy (approved):
I think we can merge this PR (see my last comments anyway)
We don't take much risk of regression as it's a new mode.
As you suggested, we could release a new version, without changing the default mode, to allow us for more testing before switching.

Jaifroid:
Happy New Year! Thanks for these tests. I reply to each point below. 

> I've tested this PR on Firefox, inside a temporary extension. On first launch, it displays a popup saying "The last attempt to launch the PWA appears to have failed. Do you wish to try again?". I suppose it's a small bug? Or maybe because of my previous tests?

This may be to do with previous tests. However, I do need to do some rigorous testing of the logic under different conditions, including upgrades, to be sure to minimize this kind of thing. It's essential to have this check, so that users don't end up with an unusable app that keeps redirecting to the Service Worker version when they are fully offline and they may have cleared their cache or something else that has deleted the `appCache`.

> After that, it works nicely. However, I've encountered a (probably minor) issue : I've used the ServiceWorker/PWA mode, then went offline. Then I used the Firefox feature to remove the local data of https://moz-extension.kiwix.org. After that, if I try to run the extension (while being offline), it uses the https://moz-extension.kiwix.org URL, which fails, and displays an error page (impossible to connect to this address). It's only since the second attempt to launch the extension that it properly detects that the PWA could not be launched previously, and suggests to try again (and allows to use jQuery mode). I understand it might be not possible to fix this...

This is related to above. It's the code that tests access to the Service Worker / appCache. In order to have a smooth startup experience in SW mode, we assume that we have access to this (offline) if we launched successfully last time. So the app tries to go straight through to https://moz-extension.kiwix.org/... It sets a "fail" flag which is only removed if the Service Worker version successfully loaded. We have to hope the user will try restarting the app if they run into a problem (like they deleted their cache) in which case the app will know it failed last time, and will ask the user what to do. What you describe should be quite rare. I'll see if there is any other unobtrusive and quick way to check that we have access to the SW before launching it. We could at least check that `appCache` exists and is accessible, but it means adding more code in `init.js`, as we need to reload the page very early on in the app boot process.

> I also tested it on Chromium, inside a temporary extension. There is a minor side effect : when I first launch it, and switch to ServiceWorker mode, the page is reloaded, I don't know why. It was not necessary before.

Yes, this is actually correcting an issue whereby the Service Worker is not properly registered, and which causes an exception in our code (that is swallowed up inside a Promise, if I remembered rightly). I think it's a timing issue. I've put a [comment at the point where this sometimes happens](https://github.com/kiwix/kiwix-js/pull/771/files#diff-484c9e8aa09e09b8eb1fafaf55e8eadcd0eaab230e38d02bb9440841bfa090a4R632). I may be able to avoid this reload by reinstating `self.clients.claim()` in the Service Worker (there was a reason I had to eliminate it, but I can't remember right now).

Jaifroid:
@mossroy As discussed, I am providing code for publishing a branch to gh-pages for testing purposes (in the workflow that publishes the extension). I need to test this code (and fine-tune), but doing this will set our gh-pages implementation to this branch (hopefully -- it is the intended behaviour). Is it OK for me to do this? Details below.

I have provided a "target" for the workflow that publishes the extension, to support publishing the extension either to the gh-pages branch or to the docker container. The option is provided in the manual dispatch (see screenshot below). I need to test the manual dispatch. The code is incomplete: I need to provide for updating **both** versions when we publish a tag, but I want to test manually first.

Once it all works, I'll give this workflow a new name: "Publish Mozilla extension", as it will no longer be just for publishing the Docker container.

image

.github/workflows/CI.yml
Jaifroid: I've added another test (for correctness of PWAServer in app.js) and have added comments to this script to make clearer what we are doing. Script still needs to be tested.

Jaifroid:
> We could at least check that `appCache` exists and is accessible, but it means adding more code in `init.js`, as we need to reload the page very early on in the app boot process.

@mossroy Although we can't access `appCache` directly from the (real) Extension, we can attempt to load an image that should have been cached by the PWAServer before we jump to it. If the appCache has the image, the Service Worker should run and serve it back to us. We already had the code for this, so in fact it didn't involve any extra coding. I've added these checks in https://github.com/kiwix/kiwix-js/pull/771/commits/60cff9d899ae367ccb594125ea27f23f3e061c62. It seems to add only a very slight delay to the launch of the PWA from an extension, so it seems acceptable. It should help with the issue you describe, so long as we don't have a situation where the Service Worker gets put into some kind of deep sleep by the system. I have to test thoroughly, to be sure that this won't start adding annoying prompts for the user.

Jaifroid:
Quick image to show outcome of new integrity checks (in the CI workflow):

image

Jaifroid:
> I also tested it on Chromium, inside a temporary extension. There is a minor side effect : when I first launch it, and switch to ServiceWorker mode, the page is reloaded, I don't know why. It was not necessary before.

I believe this is fixed with https://github.com/kiwix/kiwix-js/pull/771/commits/e78ac15b87b53c50a672f6b33f2c7d68c26836c0 according to my tests. I'm not seeing the reload happening now in the Chrome Extension. Effectively, this was a regression due to my having removed `self.clients.claim()` along with `skipWaiting()` (which did need to be removed in order to have clean upgrades after a use has closed the app).

Jaifroid:
Slight bug in Chromium with this PR. Caching API information shows "undefined" (see screenshot), whereas it should say the cache is "Custom" and have a - in the assets count box.

image

Jaifroid:
Tests (in progress):

- [ ] IE11: works for jQuery mode
- [ ]

Jaifroid:
@mossroy Quick question: given that we will use GitHub Pages as a testbed, do we want to update the implementation there automatically on each push to master? It **will** of course be updated on release, and can also be updated by (manual) workflow dispatch for testing specific versions in branches or PRs.

NB this would **not** apply to the docker build, as it must remain synchronized with releases (though we can override manually if we have to correct errors).

mossroy:
> @mossroy Quick question: given that we will use GitHub Pages as a testbed, do we want to update the implementation there automatically on each push to master? It **will** of course be updated on release, and can also be updated by (manual) workflow dispatch for testing specific versions in branches or PRs.
> 
> NB this would **not** apply to the docker build, as it must remain synchronized with releases (though we can override manually if we have to correct errors).

If you feel it would be helpful, no problem for me

Jaifroid:
@mossroy I'm ready to merge this now. I have completed extensive testing as listed in https://github.com/kiwix/kiwix-js/pull/771#issuecomment-1004228400. As a result of both your last review and these tests, I've made several minor improvements and adjustments as listed in the PR's commit log from 2nd January. These are all small things and some were experiments.

The only area I cannot properly test is the long-term durability of the PWA. It does appear to be stable through browser restarts, but only time will tell, once we have a durable extension, how well this holds up. My experience elsewhere is that the PWA can persist offline for a very long time if the user isn't in the habit of clearing out browser data regularly (at least in Chromium). If they are, well they probably know what they're doing.

Let me know when you're happy for me to squash and merge.

mossroy (approved):
You can squash/rebase/merge this.

PR #773, opened 2021-10-30
Add docker file to support Firefox extension

Jaifroid:
This should set up the docker.


Jaifroid:
This is now included in #771.

Issue #772, opened 2021-10-30
Add a step in our CI to check the list of pre-cached files for the PWA-ServiceWorker


mossroy:
It should be possible to compare the list of files that exist in www directory, and the list of pre-cached files in service-worker.js file.

If one file is missing, it should break the CI

Jaifroid:
I'm wondering how this would work. Wouldn't we need to provide a list of files needed by the app for the testing engine. In which case, how is that different from providing the list of files in `service-worker.js`? We would then have to maintain two lists of files instead of one?

I did think maybe we could use Node APIs to read the contents of the `www` directory, because we probably (though not necessarily) need all the files that are in that directory and subdirectories. However, we also need some files that are not in that directory, but not all of the files. The needed files are listed in our dockerfile, but I'm not sure how easily it translates into a format Node APIs can use.

mossroy:
My idea was to read the list of files in the www directory, apply some additions/exclusions if necessary, then compare this list with the hard-coded list of cached files.
In the CI, we are free to use the technology we want to do that (not necessarily javascript). It might be a bash script for example

Jaifroid:
#774 is implemented and working. However, it can't be merged until #771 is merged (and then it may need to be rebased on master).

PR #774, opened 2021-11-23
Check offline files list through CI

Jaifroid:
Testing CI solution for #772 .

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -110,3 +110,12 @@ jobs:
           ./node_modules/.bin/nightwatch -c nightwatch.js --env edge40,edge44
           ./node_modules/.bin/nightwatch -c nightwatch.js --env firefox45,chrome58,ie11
           pkill node || echo "Node process not running (anymore)"
+
+  # Tests the list of offline files in the Service Worker and compares it to the contents of the www directory
+  test-offline-files:
+    runs-on: windows-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Test the list of offline files in Service Worker
+        shell: pwsh
+        run: ./scripts/Check-OfflineFilesList.ps1
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -16,8 +16,9 @@ jobs:
   # You can run these same tests locally in your own developer
   # environment via `npm ci && npm test`.
   tests-basic:
-    # Packages 'firefox' and 'google-chrome' are pre-installed.
-    #
+    # We want to run on external PRs, but not on our own internal PRs as they'll be run by the push to the branch
+    if: github.event_name == 'push' || github.event.pull_request.head.repo.full_name != github.repository
+    # Packages 'firefox' and 'google-chrome' are pre-installed:
     # https://github.com/actions/virtual-environments/blob/ubuntu20/20201210.0/images/linux/Ubuntu2004-README.md
     runs-on: ubuntu-20.04
     steps:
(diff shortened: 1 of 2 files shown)


Jaifroid:
Script confirmed to be working via pushing incorrect list. Screenshots show: 1. script failing due to missing file; 2. script passing.

image

image

Jaifroid:
Please note this PR is based on #771 and that PR should be merged first.

scripts/Check-OfflineFilesList.ps1
```diff
@@ -0,0 +1,41 @@
...
+  if (-Not ($ListOfSWFiles -ccontains $_ -or $ListOfExemptions -contains $_)) {
+    "ERROR! The file '$_' is not in the list of offline files in service-worker.js`n"
+    $MissingFiles = $true
+  }
+}
+if ($MissingFiles) {
+  Write-Host "** Please add the missing file(s) listed above to service-worker.js **`n" -ForegroundColor red
+  exit 1
+} else {
+  Write-Host "All non-exempt files in /www are listed in service-worker.js`n" -ForegroundColor green
```
Jaifroid, line 39: I disagree with the CodeFactor warning below. First, we know the context this script will mostly run in (GitHub shell), where it works. Second, `Write-Host` is the only command that allows setting colours in the terminal (the recommended as substitutes do not accept the `-ForegroundColor` switch). Finally, the output is captured perfectly well by the GitHub Web UI. The script can also be run locally by the user (in Terminal), and is confirmed to work on W10 and W11.

Jaifroid:
Rebased on master.

scripts/Check-OfflineFilesList.ps1
```diff
@@ -0,0 +1,49 @@
...
+  This script is part of Kiwix. It checks that the list of offline files precached by the Service Worker is complete
+  It is intended to be used primarily in CI testing, but can also be run from the commandline in open-source PS Core
+  https://github.com/PowerShell/PowerShell/releases/
+** #>
+
+# Provide the path and name of the Service Worker file relative to the Repository root
+$SWFile = 'service-worker.js'
+
+# List below any files, relative to Repository root, that are not required by the offline precache
+$ListOfExemptions = (
```
mossroy, line 11: Could you remind me why these files are exempted?

.github/workflows/CI.yml
```diff
@@ -110,3 +111,14 @@ jobs:
           ./node_modules/.bin/nightwatch -c nightwatch.js --env edge40,edge44
           ./node_modules/.bin/nightwatch -c nightwatch.js --env firefox45,chrome58,ie11
           pkill node || echo "Node process not running (anymore)"
+
+  # Tests the list of offline files in the Service Worker and compares it to the contents of the www directory
+  test-offline-files:
+    # We want to run on external PRs, but not on our own internal PRs as they'll be run by the push to the branch
+    if: github.event_name == 'push' || github.event.pull_request.head.repo.full_name != github.repository
+    runs-on: windows-latest
```
mossroy, line 119: In the short term, it's not a problem to run this script on Windows.
In the mid or long term, it might be : if one day we switch to another CI that would only support linux, and/or if we want to avoid having to pay a Windows license.

I understand you're more comfortable with PowerShell than with Bash, so don't rewrite this script. It should be easy to rewrite later if necessary.

NB : I know PowerShell can run on Linux, but it's not commonly installed on linux distros

mossroy (approved):
I could no test, and do not know PowerShell enough to really review.
But the logic looks fine and I know you tested it.

scripts/Check-OfflineFilesList.ps1
Jaifroid: Because these are files that are never used by our code. Maybe they are used as part of ancillary packaging for apps, for example, but will never actually be needed to run the app, so it would be a waste of valuable cache space (and download time) to load them, especially over slow bandwidth. If we don't exempt them, then the script will thrown an error, because these files are not loaded by the Service Worker.

.github/workflows/CI.yml
Jaifroid: I understand. I didn't feel I knew enough bash to be comfortable getting things right (or it would have taken a very long time and a lot of testing, trial-and-error). I'm sure you could rewrite it based on the same logic and regexes in about 10 minutes, but as you say, that can wait till we need to do it (hopefully not!).

PR #783, opened 2021-12-14
Disable dragging of some elements

### LINKED ISSUE #782: Dragging of elements may cause unexpected behavior
Few elements like the navigation bar on the bottom and Kiwix logo are draggable which can cause unexpected behavior and can also interfere with the drag and drop feature of ZIM files.

https://user-images.githubusercontent.com/77742477/146038674-c36be025-992f-4531-8186-acaa385548d9.mp4

try-catch-stack:
Fixes #782
Added the attribute `draggable=false` to the bottom navigation bar to prevent it from causing unexpected behavior.

### DIFF
--- a/www/index.html
+++ b/www/index.html
@@ -601,10 +601,10 @@
                     </div>
                 </div>
                 <div id="navigationButtons" class="btn-group btn-block">
-                    <a href="#" class="btn btn-lg" draggable="false" id="btnHomeBottom" title="Home"><i class="fas fa-home"></i></a>
-                    <a href="#" class="btn btn-lg" draggable="false" id="btnBack" title="Back"><i class="fas fa-arrow-left"></i></a>
-                    <a href="#" class="btn btn-lg" draggable="false" id="btnForward" title="Forward"><i class="fas fa-arrow-right"></i></a>
-                    <a href="#top" class="btn btn-lg" draggable="false" id="btnTop" title="Top"><i class="fas fa-arrow-up"></i></a>
+                    <a href="#" class="btn btn-lg" draggable="false" id="btnHomeBottom" title="Home"><i class="fas fa-home" draggable="false"></i></a>
+                    <a href="#" class="btn btn-lg" draggable="false" id="btnBack" title="Back"><i class="fas fa-arrow-left" draggable="false"></i></a>
+                    <a href="#" class="btn btn-lg" draggable="false" id="btnForward" title="Forward"><i class="fas fa-arrow-right" draggable="false"></i></a>
+                    <a href="#top" class="btn btn-lg" draggable="false" id="btnTop" title="Top"><i class="fas fa-arrow-up" draggable="false"></i></a>
                 </div>
             </footer>
         </section>
--- a/www/index.html
+++ b/www/index.html
@@ -601,10 +601,10 @@
                     </div>
                 </div>
                 <div id="navigationButtons" ondragstart="return false;" class="btn-group btn-block">
-                    <a href="#" class="btn btn-lg" draggable="false" id="btnHomeBottom" title="Home"><i class="fas fa-home" draggable="false"></i></a>
-                    <a href="#" class="btn btn-lg" draggable="false" id="btnBack" title="Back"><i class="fas fa-arrow-left" draggable="false"></i></a>
-                    <a href="#" class="btn btn-lg" draggable="false" id="btnForward" title="Forward"><i class="fas fa-arrow-right" draggable="false"></i></a>
-                    <a href="#top" class="btn btn-lg" draggable="false" id="btnTop" title="Top"><i class="fas fa-arrow-up" draggable="false"></i></a>
+                    <a href="#" class="btn btn-lg" id="btnHomeBottom" title="Home"><i class="fas fa-home"></i></a>
+                    <a href="#" class="btn btn-lg" id="btnBack" title="Back"><i class="fas fa-arrow-left"></i></a>
+                    <a href="#" class="btn btn-lg" id="btnForward" title="Forward"><i class="fas fa-arrow-right"></i></a>
+                    <a href="#top" class="btn btn-lg" id="btnTop" title="Top"><i class="fas fa-arrow-up"></i></a>
                 </div>
             </footer>
         </section>
--- a/www/index.html
+++ b/www/index.html
@@ -96,7 +96,7 @@
             <article class="view-content">
                 <div id="about" style="display: none;" class="container">
                     <h2 id="contents">
-                        <img src="img/icons/kiwix-60.png" alt="Kiwix icon"/>
+                        <img src="img/icons/kiwix-60.png" draggable="false" alt="Kiwix icon"/>
                         Kiwix : offline Wikipedia viewer
                     </h2>
                     <p>Official site : <a href="https://www.kiwix.org/" target="_blank">https://www.kiwix.org&nbsp;<img src="img/Icon_External_Link.png" /></a></p>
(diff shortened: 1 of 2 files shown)


mossroy (approved):
Thanks for this PR.
It indeed fixes this issue, that I had never noticed

kelson42:
@try-catch-stack This PR would benefit of a rebase on latest git head master before merging.

try-catch-stack:
> @try-catch-stack This PR would benefit of a rebase on latest git head master before merging. I

I fetched and merged the changes from the master branch , should I rebase or this would be fine?

Jaifroid:
@try-catch-stack I think there's no need to rebase because there are no conflicts now and we will squash on merging. However I'd like to test on different browsers first, and will get back to you.

Jaifroid (changes requested):
I've now tested in IE11 (actually in IE mode in Edge Chromium), and unfortunately it doesn't work, even though the draggable attribute appears to be supported by IE11: see https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/draggable. This could be an artefact of the IE mode, or it could be that you need to add the draggable attribute also to the `<i></i>` elements that render the icons.

@try-catch-stack are you able to test in real IE11, or IE mode?

Jaifroid:
I can now confirm that the issue is that the draggable attribute needs to be added to the innermost draggable element, which in the case of the icons along the bottom is the `<i></i>` tags. @try-catch-stack Please check whether it is enough to add the draggable attribute only to those tags and not to the outer tags they are currently in (for the browsers you can test on). It would be ideal if we don't have to repeat the attribute. When you've tested, kindly confirm which browsers you've tested on.

try-catch-stack:
I have tested on Chrome,Edge, and IE11 (both IE and IE mode on edge) . I can confirm that adding `draggable=false` attribute only on the outer tag doesn't seem to work on both , IE mode and real IE. Neither does adding it only on the inner `<i></i>` tags work as the whole outer div acts as the navigation link which can be dragged. The only workaround  that seems to work is adding the attribute on both inner and outer tags.

Jaifroid:
Thank you for the change. I confirm it is now working on IE11. However, unfortunately, it's not working on Firefox. I just tested on Firefox 95.0.1 on Windows, but the elements on the bottom bar are still draggable. I think this might be because bootstrap replaces the `<i></i>` tags with an SVGs, and possibly in Firefox specifically, this adds a further inner element `<path ... />` which doesn't have the draggable attribute. See screenshot, where you can also see that the two `draggable="false"` attributes have been correctly applied to both outer elements, but not to this innermost element.

image

Jaifroid:
Since you can't add the draggable attribute in the source (the code is generated dynamically), you only have two options: 1) add the draggable element in JavaScript after the page has loaded and Bootstrap has rendered the elements; or 2) use CSS to prevent dragging of this element. The latter is a cleaner solution. You could try adding this to `app.css` for the `path` element. NB I don't know if this will work, you'd have to experiment. But the relevant CSS would be `user-drag: none;`. See also https://stackoverflow.com/a/55761346/9727685, but some of those directives would be overkill.

Jaifroid:
Also, note that one user added a comment to that stackoverflow answer saying that neither the draggable attribute nor the CSS seem to work in Firefox, but that could be a specific case. It's worth testing. I would be very surprised if Firefox has no clean way of preventing drag.

Jaifroid:
@try-catch-stack  see https://stackoverflow.com/questions/26356877/html5-draggable-false-not-working-in-firefox-browser for discussion of this issue in Firefox. I think the solutions I suggested above won't work. You might try one of the others suggested here.

try-catch-stack:
@Jaifroid Thanks for pointing out the issue. Seems , it was some issue specific to how  firefox deals with draggable attribute and not the `<path...>` tag as this tag is present on all the browsers.I did some research and found that there is this attribute for onDrag event `ondragstart="return false;"` which is triggered when a user tries to drag an element. This attribute solves the issue when added to the parent `div`.
Check https://www.w3schools.com/jsref/event_ondragstart.asp for reference .

Jaifroid:
@try-catch-stack That's good news! Just a quick note however that we might have an issue with including inline JS inside the HTML due to the CSP of the Firefox OS which this app is still building for. @mossroy could you kindly confirm whether instead of  `ondragstart="return false;"` on the div, it would be better to add an event listener in app.js for this div (which has a unique ID)?

Also @try-catch-stack could you please check whether using this event makes the `draggable="false"` attributes redundant, or are they still needed for other browsers?

try-catch-stack:
Yes @Jaifroid the `draggable=false` attribute can now be removed from all the tags. I have tested on chrome,edge,IE and firefox and it works fine on all of them. Just need a confirmation on that inlineJS thing or should I add an event listener.

Jaifroid:
I've tested also and it's working really well on all browsers.
My memory is that using inline events in the HTML could be a blocker for Firefox OS, but I may be misremembering.
Let's request a new review from @mossroy .

Jaifroid (approved):
It's working for me, but it may be better to use event listener in app.js.

mossroy (changes requested):
Thanks for this PR.

I did not test, but it's probable that this inline javascript could be blocked by some CSP (not only the one of Firefox OS : browser extensions also have some).
This event handling should be moved to app.js.

You'll need to add an id for the kiwix icon (there's already one for the navigation buttons), and add the event handler to these 2 DOM elements (with a meaningful comment) in app.js.

try-catch-stack:
Fine @mossroy . I will add the event listener in app.js . Is there any specific place in app.js or a function that you want me to add it in?

www/js/app.js
```diff
@@ -393,6 +393,7 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
             document.getElementById('modes').scrollIntoView();
         }, 600);
     });
+    document.getElementById('navigationButtons').ondragstart=function () {return false;}
```
mossroy, line 396: Please add a comment above to explain the purpose

mossroy (commented):
Cool! It seems to be working on my Firefox desktop.
I tested on FirefoxOS, but the images seem to be never draggable on this platform. In any case, it causes no regression on it.

Is it possible to disable dragging of the Kiwix icon with the same javascript, instead of the draggable attribute?
It would make the code more consistent (you'll have to set an id to this HTML tag in index.html, as it currently does not have any)

www/js/app.js
try-catch-stack: Done

www/index.html
```diff
@@ -96,7 +96,7 @@
             <article class="view-content">
                 <div id="about" style="display: none;" class="container">
                     <h2 id="contents">
-                        <img src="img/icons/kiwix-60.png" alt="Kiwix icon"/>
+                        <img src="img/icons/kiwix-60.png" id='kiwixLogo' alt="Kiwix icon"/>
```
mossroy, line 99: Please use `"` instead of `'` to be consistent
```suggestion
                        <img src="img/icons/kiwix-60.png" id="kiwixLogo" alt="Kiwix icon"/>
```

www/js/app.js
```diff
@@ -394,6 +394,11 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
         }, 600);
     });
 
+    //Adds an event listener to kiwix logo and bottom navigation bar which gets triggered when these elements are dragged.
+    //Returning false prevents their dragging(which can cause some unexpected behavior)
```
mossroy, line 398:
```suggestion
    //Returning false prevents their dragging (which can cause some unexpected behavior)
    //Doing that in javascript is the only way to make it cross-browser compatible
```

mossroy (approved):
Looks good!
I suggested a few minor fixes

mossroy (approved):
Looks good to me.
If you confirm it works on all browsers, including IE, it can be merged for me

Jaifroid:
Just running the automated tests, and I'll do some manual too.

try-catch-stack:
Yes I have tested on the browsers including IE and it works fine for me.

Jaifroid (approved):
I corroborate. I have tested in Firefox in JQuery and SW modes; in IE11, in Edge IE Mode, and in Edge (Chromium). All working correctly.

Jaifroid:
@try-catch-stack would you like me to squash and merge now?

try-catch-stack:
Yaa sure @Jaifroid  you can squash and merge now

Jaifroid:
Done! Many thanks for this PR.

Issue #729, opened 2021-06-06
Support 'auto' dark theme that reflects OS settings


Jaifroid:
There are now CSS media queries and corresponding DOM functions in Firefox and Chromium (and Safari) that can detect the system-wide / OS 'dark mode' settings in both Windows and Mac Mojave (I haven't checked for Linux yet). See: https://caniuse.com/?search=prefers-color-scheme.

I've tested this in Kiwix JS Windows, and simple code like this can register to be notified of a change in the system preferences, and adapt the auto theme accordingly:
```
var uiSettings = window.matchMedia('(prefers-color-scheme:dark)');
uiSettings.onchange = function () {
    if (params.cssUITheme == 'auto') cssUIThemeGetOrSet('auto');
}; 
```
The bit inside the `onchange` function would need to be adapted to Kiwix JS code, but we have the methods for switching themes. Some OS's allow auto switching themes according to time of day, for example, so Kiwix would auto-adapt when the system changes the theme, or simply when the user switches the OS theme manually. I've tested this on Windows 10, with the global Light/Dark theme in Settings.

Theme should of course default to Light if the user is using an older browser where the OS theme cannot be detected. The user can override by choosing Light or Dark explicitly, instead of, say, Auto, in Kiwix JS Configuration.

ShubhashishMaiti:
Hello Sir, I am interested in working on this issue. Could you please assign this issue to me?

Jaifroid:
@ShubhashishMaiti That would be helpful. Please note that the sample code I gave above was for a related app, Kiwix JS Windows, where this is implemented. It would need to be adapted to the parameters used in Kiwix JS, but the principle is the same.

Maneet-k:
Sir, I wanna work on this issue. Can I help you in this ?

Maneet-k:
@mossroy are nightwatch.js and nightwatch_runner.js  ,the files I should look into?

mossroy:
No, these files are used for automated testing, and have nothing to do with a night theme

Maneet-k:
@mossroy ohh my bad....... would you please mind me telling the files i should look into. That'll be really helpful🤧

Jaifroid:
[...]

PR #786, opened 2021-12-30
update app.js

Maneet-k:
[#729 ](https://github.com/kiwix/kiwix-js/issues/729)
at 101 line

### DIFF
--- a/www/js/app.js
+++ b/www/js/app.js
@@ -98,6 +98,15 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
     
     // Unique identifier of the article expected to be displayed
     var expectedArticleURLToBeDisplayed = "";
+ 
+    var darkPreference = window.matchMedia("(prefers-color-scheme : dark");
+    darkPreference.addEventListener("change", e=> {
+            if (params.appTheme == "auto") {
+            let newColorScheme = e.matches ? "dark_invert" : "light";
+            uiUtil.applyAppTheme(newColorScheme);
+            }  
+            else  uiUtil.applyAppTheme(params.appTheme);     
+        });
     
     /**
      * Resize the IFrame height, so that it fills the whole available height in the window


www/js/app.js
```diff
@@ -98,6 +98,15 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
     
     // Unique identifier of the article expected to be displayed
     var expectedArticleURLToBeDisplayed = "";
+ 
+    var darkPreference = window.matchMedia("(prefers-color-scheme : dark");
+    darkPreference.addEventListener("change", e=> {
```
Jaifroid, line 103: Don't use arrow functions, as they are not compatible with all the JavaScript versions we support.

www/js/app.js
```diff
@@ -98,6 +98,15 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
...
+            if (params.appTheme == "auto") {
+            let newColorScheme = e.matches ? "dark_invert" : "light";
```
Jaifroid, line 105: Use `var` instead of `let`, again for compatibility reasons.

www/js/app.js
```diff
@@ -98,6 +98,15 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
     
     // Unique identifier of the article expected to be displayed
     var expectedArticleURLToBeDisplayed = "";
+ 
+    var darkPreference = window.matchMedia("(prefers-color-scheme : dark");
+    darkPreference.addEventListener("change", e=> {
+            if (params.appTheme == "auto") {
```
Jaifroid, line 104: You probably want to set appTheme to `auto_invert` (in `index.html`) rather than `auto`. This will allow it to be extended to support `auto_mwInvert`.

www/js/app.js
Jaifroid: Since you will need to handle this variable in `uiUtil`, you will need to use `params.newColourScheme` (this is because the params object is global and is available outside `app.js`, whereas `newColorScheme` would be available only in `app.js`).

Jaifroid (changes requested):
@Maneet-k Please see comments in the code. As a general comment, the code you've added to `app.js` looks like it should work (with my suggested changes), but you will need to add code to `uiUtil.applyAppTheme` to handle the values you have set, and in particular to handle` params.newColourScheme`.  And of course you'll need to provide the auto option in `index.html`.

www/js/app.js
Maneet-k: Ohk Sir.

PR #793, opened 2022-01-10
Auto Dark mode Theme added

### LINKED ISSUE #729: Support 'auto' dark theme that reflects OS settings

Maneet-k:
Fixes #729


Jaifroid:
@Maneet-k Where are you with this PR? You haven't requested a review, as far as I'm aware.

Maneet-k:
> @Maneet-k Where are you with this PR? You haven't requested a review, as far as I'm aware.

Mr. @Jaifroid I wasn't aware that I needed to ask for review. :/

Maneet-k:
> > @Maneet-k Where are you with this PR? You haven't requested a review, as far as I'm aware.
> 
> Mr. @Jaifroid I wasn't aware that I needed to ask for review:/ Could you review this code.
I guess I need to make 
a new PR( if the code is correct). Can you review the changes then I can raise a PR that'll be up to date with the master repo

Jaifroid:
@Maneet-k You don't need to make a new PR so long as this one can be merged without conflicts. Currently it clearly says it has no conflicts with the base branch. I'll review asap.

Jaifroid (changes requested):
@Maneet-k When it works, this is good. However, it doesn't work consistently. If I start the app in Light mode (the default), and switch to Auto, it works first time. but then if I switch my OS's theme, the app doesn't switch to follow my OS light/dark theme.

If I leave it on Auto and **restart** the app, then the auto switching works fine. But if I switch the app back to light, restart the app, then the event listener never gets registered.

You need to change your overall design so that the event listener is registered whenever the user switches to Auto. If you don't register the event listener then, then the app will never know that the OS has changed theme.

www/js/lib/uiUtil.js
```diff
@@ -418,6 +418,8 @@ define(rqDef, function() {
      * @param {String} theme The theme to apply (light|dark[_invert|_mwInvert])
      */
     function applyAppTheme(theme) {
+        if (theme === "auto_invert") theme = window.matchMedia('(prefers-color-scheme:dark)').matches ? "dark_invert" : "light" 
+        if (theme === "auto_mwInvert") theme = window.matchMedia('(prefers-color-scheme:dark)').matches ? "dark_mwInvert" : "light"
```
Jaifroid, line 422: Ditto about repeated code here. You don't need two tests. Just test the `auto` part as indicated above.

Maneet-k: I agree Mr. @Jaifroid , it would work in the above case, But in this case I am setting `theme` variable's value.

Maneet-k:
> @Maneet-k When it works, this is good. However, it doesn't work consistently. If I start the app in Light mode (the default), and switch to Auto, it works first time. but then if I switch my OS's theme, the app doesn't switch to follow my OS light/dark theme.
> 
> If I leave it on Auto and **restart** the app, then the auto switching works fine. But if I switch the app back to light, restart the app, then the event listener never gets registered.
> 
> You need to change your overall design so that the event listener is registered whenever the user switches to Auto. If you don't register the event listener then, then the app will never know that the OS has changed theme.

@Jaifroid could you please review these commits whenever you have free time. I guess it should be working this time.

www/js/lib/uiUtil.js
```diff
@@ -418,6 +418,8 @@ define(rqDef, function() {
      * @param {String} theme The theme to apply (light|dark[_invert|_mwInvert])
```
Jaifroid, line 418: Please update this description above to include the new values.

www/js/app.js
```diff
@@ -166,7 +165,18 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
     
     // Unique identifier of the article expected to be displayed
     var expectedArticleURLToBeDisplayed = "";
-    
+ 
+    // define variable for dark preference for matchMedia
+    var darkPreference = window.matchMedia('(prefers-color-scheme:dark)');
+    if (!window.matchMedia) document.getElementById('appThemeSelect').options[0].style.display = "none";
+    if (!window.matchMedia) document.getElementById('appThemeSelect').options[1].style.display = "none";
```
Jaifroid, line 172: Again here, just use a single `if` condition, and inside it set the two options to hidden. In general, if you find you are doing the same test twice, you are probably writing inefficient code!

www/js/lib/uiUtil.js
Jaifroid: OK, but you can still make the code more concise by defining `var darkPreference = window.matchMedia.....`, then do your tests on `darkPreference`

www/js/app.js
```diff
@@ -166,7 +165,18 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
     
     // Unique identifier of the article expected to be displayed
     var expectedArticleURLToBeDisplayed = "";
-    
+ 
+    // define variable for dark preference for matchMedia
+    var darkPreference = window.matchMedia('(prefers-color-scheme:dark)');
+    if (!window.matchMedia) document.getElementById('appThemeSelect').options[0].style.display = "none";
```
Jaifroid, line 171: Don't test on `window.matchMedia` because every browser supports it, but not every browser supports `prefers-color-scheme`. You need to test on the latter. See https://caniuse.com/prefers-color-scheme vs https://caniuse.com/matchmedia.

Maneet-k: > Don't test on `window.matchMedia` because every browser supports it, but not every browser supports `prefers-color-scheme`. You need to test on the latter. See https://caniuse.com/prefers-color-scheme vs https://caniuse.com/matchmedia.

Ohk got it!

www/js/app.js
Maneet-k: @Jaifroid, `document.getElementById('appThemeSelect').options[0].style.display = "none";` 
Should I write these expressions as it is or use the `for` loop.

Jaifroid: I think `for` is overkill here. If there were several options, yes, but just two it's OK just to set them both in the `if` block.

Maneet-k: Ohkay sir

www/js/app.js
```diff
@@ -166,7 +165,22 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
...
     // Unique identifier of the article expected to be displayed
     var expectedArticleURLToBeDisplayed = "";
-    
+ 
+    // define and store dark preference for matchMedia
+    var darkPreference = window.matchMedia('(prefers-color-scheme:dark)');
+    // if 'prefers-color-scheme' is not supported in the browser
+    if (window.matchMedia('(prefers-color-scheme)').media === 'not all') {
+        document.getElementById('appThemeSelect').options[0].style.display = "none";
+        document.getElementById('appThemeSelect').options[1].style.display = "none";
```
Jaifroid, line 174: @Maneet-k Did you test this code in IE11 or IE mode? While it matches, setting the display of options 0 and 1 to "none" does nothing at all. They can still be selected in IE11, albeit with the arrow keys (the dropdown doesn't work). If the idea is to remove these options from the UI, then you need to remove them explicitly rather than just hide them. It's also not safe to use `options[0]` (i.e. by number), because in the future we could add more options, and Auto options might not always be the first two. It would be better to match them with `querySelectorAll` perhaps, and then remove them from the DOM, unless you can find a way to hide them that works.

Maneet-k:
Mr. @Jaifroid, I tested on opera, it's working now (Can't say about IE). Could you please review these commits and let me know if anything else I need to work on.

www/js/app.js
```diff
@@ -166,7 +165,24 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
...
     var expectedArticleURLToBeDisplayed = "";
-    
+ 
+    // define and store dark preference for matchMedia
+    var darkPreference = window.matchMedia('(prefers-color-scheme:dark)');
+    // if 'prefers-color-scheme' is not supported in the browser
+    if (window.matchMedia('(prefers-color-scheme)').media === 'not all') {
+        var optionsToBeRemoved = document.querySelectorAll('.autoNotAvailable');
+        for (var i = 0; i < optionsToBeRemoved.length; i++) {
+            optionsToBeRemoved[i].remove();
```
Jaifroid, line 175: Unfortunately `ChildNode.remove()` isn't available in IE11, so this code won't work. See https://caniuse.com/?search=remove . To test in IE11, download Microsoft Edge, and run it in "Internet Explorer mode". Opera isn't a good analogue for IE11 because it is based on Chromium.

www/index.html
```diff
@@ -466,6 +466,8 @@ <h3>Display settings</h3>
                                         <b>Select app theme</b> (content inversion is experimental):
                                     </label>
                                     <select class="form-control" id="appThemeSelect">
+                                        <option class="autoNotAvailable" value="auto_invert">Auto (generic)</option>
+                                        <option class="autoNotAvailable" value="auto_mwInvert">Auto (Wikimedia)*</option>
```
Jaifroid, line 470: This isn't a very intuitive class name. Why not just `class="auto"`, i.e. you just label the options as auto options?

www/js/app.js
```diff
@@ -166,7 +165,24 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
     
     // Unique identifier of the article expected to be displayed
     var expectedArticleURLToBeDisplayed = "";
-    
+ 
+    // define and store dark preference for matchMedia
+    var darkPreference = window.matchMedia('(prefers-color-scheme:dark)');
+    // if 'prefers-color-scheme' is not supported in the browser
+    if (window.matchMedia('(prefers-color-scheme)').media === 'not all') {
+        var optionsToBeRemoved = document.querySelectorAll('.autoNotAvailable');
```
Jaifroid, line 173: If you change classname above, change it here too.

Jaifroid (changes requested):
You're nearly there now @Maneet-k! Just a few minor changes, and unfortunately an incompatibility with IE11 to work around... I suggest testing IE11 support by using Microsoft Edge and running in Internet Explorer mode if you don't have access to Windows 10 (which has native IE11).

www/js/app.js
Maneet-k: Then I guess I should try `removeChild` method

Jaifroid (approved):
@Maneet-k The code looks good now. I haven't tested fully, but as far as the code is concerned, I'm happy to approve. Please request a review from mossroy now for sign-off.

www/js/lib/uiUtil.js
```diff
@@ -415,9 +415,12 @@ define(rqDef, function() {
      * For each contentTheme, a stylesheet must be provided in www/css that is named 'kiwixJS' + contentTheme
      * A rule may additionally be needed in app.css for full implementation of contentTheme
      * 
-     * @param {String} theme The theme to apply (light|dark[_invert|_mwInvert])
+     * @param {String} theme The theme to apply (auto[_invert|_mwInvert]|light|dark[_invert|_mwInvert])
      */
     function applyAppTheme(theme) {
+        var darkPreference = window.matchMedia('(prefers-color-scheme:dark)');
+        if (theme === "auto_invert") theme = darkPreference.matches ? "dark_invert" : "light"
```
mossroy, line 422: semicolons are missing at the end of these 2 lines

www/js/app.js
```diff
@@ -166,7 +165,24 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
...
+    // define and store dark preference for matchMedia
+    var darkPreference = window.matchMedia('(prefers-color-scheme:dark)');
+    // if 'prefers-color-scheme' is not supported in the browser
+    if (window.matchMedia('(prefers-color-scheme)').media === 'not all') {
+        var optionsToBeRemoved = document.querySelectorAll('.auto');
+        for (var i = 0; i < optionsToBeRemoved.length; i++) {
+            optionsToBeRemoved[i].parentNode.removeChild(optionsToBeRemoved[i]);
+        }
+    }
+    //Apply previously stored appTheme
```
mossroy, line 178: Minor: please add a space after `//` to match the style applied elsewhere

www/js/lib/uiUtil.js
```diff
@@ -415,9 +415,12 @@ define(rqDef, function() {
      * For each contentTheme, a stylesheet must be provided in www/css that is named 'kiwixJS' + contentTheme
      * A rule may additionally be needed in app.css for full implementation of contentTheme
      * 
-     * @param {String} theme The theme to apply (light|dark[_invert|_mwInvert])
+     * @param {String} theme The theme to apply (auto[_invert|_mwInvert]|light|dark[_invert|_mwInvert])
```
mossroy, line 418: When the user selects `Auto (wikimedia)*`, it should always display the warning `* Implements workarounds specific to Wikimedia ZIMs...` like it does when `Dark + invert (Wikimedia)*` is selected.
It does work when the OS theme is dark, but it's not displayed if the OS theme is not dark.

www/index.html
```diff
@@ -466,6 +466,8 @@ <h3>Display settings</h3>
                                         <b>Select app theme</b> (content inversion is experimental):
                                     </label>
                                     <select class="form-control" id="appThemeSelect">
+                                        <option class="auto" value="auto_invert">Auto (generic)</option>
```
mossroy, line 469: I find disturbing that the `auto*` options do not match the `dark*` options.

There is no `auto` option that allows to switch to `Dark (app only)` when the OS is in dark mode. I know we did not initially ask for it, but shouldn't it be added?

And the labels should match better. I suggest for example: `Auto (generic invert)` and `Dark (generic invert)`, `Auto (app only)` and `Dark (app only)`, `Auto (wikimedia invert)*` and `Dark (wikimedia invert)*`

www/index.html
```diff
@@ -466,6 +466,8 @@ <h3>Display settings</h3>
...
+                                        <option class="auto" value="auto_mwInvert">Auto (Wikimedia)*</option>
                                         <option value="light">Light</option>
```
mossroy, line 471: I know it's what @Jaifroid asked for, but I find misleading that the "Light" option (which is the default) is not the first in the list.

mossroy (changes requested):
Thanks for the PR.
It's close to the goal, but there are a few issues that need to be fixed (or discussed) in my opinion.
I only tested with Firefox on Ubuntu 20.04

www/index.html
Jaifroid: I just asked for Light to be the default, I think. I agree it can go first in the list (for that very reason). As long as it is the default value when the user starts the app for the first time.

www/index.html
Jaifroid: This ought to be pretty easy to implement within the existing schema.

www/js/lib/uiUtil.js
Jaifroid: I agree. I had thought this was working, but I realize now that it's because I had only checked for it when my OS was set to dark mode.

www/index.html
Maneet-k: I can tell @mossroy has better memory. Anyways,
@mossroy @Jaifroid then where you want to put the `auto` options : just after `light` option or in the end

Jaifroid: > I can tell @mossroy has better memory

Confirmed! 😉

Maneet-k: Hahaa XD

Maneet-k: >  @Jaifroid please let me know where you want to put the `auto` options : just after `light` option or in the end
What about this one

Maneet-k: > I think this is what @mossroy was implying:
> 
> 1. Light
> 2. Auto (app only)
> 3. Dark (app only)
> 2. Auto (generic invert)
> 3. Dark (generic invert)
> 4. Auto (wikimedia invert)*
> 5. Dark (wikimedia invert)*
> 
> However, @mossroy I do wonder if it's not a bit overwhelming and we could simplify, at least not have the `Auto (app only)` option?

I should have reloaded my tab before asking 👀
Yeah it seems to be bit scattered cuz the users will just want the auto light or dark to applied and those options we'll provide other options.So ,this is what i think should be implemented 
1.Light
2.Dark(app only) 
3. Dark(generic invert)
4. Dark (wikimedia invert)*
5. Auto (app only)
6. Auto (generic invert)
7. Auto (wikimedia invert)*

I wonder what's mossroy has on his mind

mossroy: What I had in mind was the last proposal above, from @Maneet-k 
However, I also find that it's a lot of options, it becomes confusing.

I was about to suggest to remove the "app only" option (dark or auto), but I'm not sure. I currently find it useless but what would happen if we have a ZIM file that handles dark mode itself (I don't know if we already have some): depending on how it is handled by the ZIM content, applying an invert stylesheet on a content that is already inverted could revert it to light mode...

We should find a way to have good default values for the user. Maybe it implies to have several form fields instead of only one.
For example, keep only Light/Dark/Auto options in this drop-down list, and have another drop-down list appearing when Dark or Auto is selected. This second drop-down list would have Generic invert/App only/Wikimedia invert options, with "Generic invert" selected by default (if we agree that it's the best default option)

Jaifroid: > @Jaifroid what do you say?

I like the two dropdown options, but I think it's more complex and should maybe be a separate PR after we've nailed this one in terms of the options we want.. Please follow mossroy's advice about the options using the current schema, and then later we can find a way to simplify if necessary. As long as we don't disrupt whatever settings users have currently chosen, I don't think it matters too much if we introduce more options now, and then simplify the UI for selecting them as a separate PR for the following release. Most users will probably just stick with light, or choose one of the two dark options. Hard-core fans of dark/light mode switching will understand the other options.

mossroy: I agree. It can be done in two steps

www/js/lib/uiUtil.js
```diff
@@ -415,9 +415,12 @@ define(rqDef, function() {
      * For each contentTheme, a stylesheet must be provided in www/css that is named 'kiwixJS' + contentTheme
      * A rule may additionally be needed in app.css for full implementation of contentTheme
      * 
-     * @param {String} theme The theme to apply (light|dark[_invert|_mwInvert])
+     * @param {String} theme The theme to apply (light|dark[_invert|_mwInvert]|auto[_invert|_mwInvert])
      */
     function applyAppTheme(theme) {
+        var darkPreference = window.matchMedia('(prefers-color-scheme:dark)');
+        var selectedThemeContent = theme.replace(/^[^_]*/, '');
+        if (/^auto/.test(theme)) theme = darkPreference.matches ? ("dark"+selectedThemeContent) : ("light"+selectedThemeContent);
```
Jaifroid, line 423: Minor, but please make use of quote marks and spacing match the style of this block in above line:

`('dark' + selectedThemeContent) : ('light' + selectedThemeContent)`

I'm slightly confused about the last option here `('light' + selectedThemeContent)`. This could produce a theme name like `light_mwInvert`, which is not a currently valid theme name. Wouldn't it be better for this last option just to be `'light'`?

Jaifroid (changes requested):
@Maneet-k Something appears to be broken. Did you test your changes thoroughly? Steps to reproduce:

* Make sure OS is set to dark
* Select Auto_mwInvert before loading any ZIM
* Load a Wikipedia ZIM (it should open with content dark)
* Switch OS to light theme

Result is a hybrid of light app with dark content, i.e. the content hasn't switched to light. See screenshot below.

image

Maneet-k:
> @Maneet-k Something appears to be broken. Did you test your changes thoroughly? Steps to reproduce:
> 
> * Make sure OS is set to dark
> * Select Auto_mwInvert before loading any ZIM
> * Load a Wikipedia ZIM (it should open with content dark)
> * Switch OS to light theme
> 
> Result is a hybrid of light app with dark content, i.e. the content hasn't switched to light. See screenshot below. I did this test in Service Worker mode, but it seems to affect JQuery too. Before asking for a review, please test thoroughly every option in the dropdown with actual content loaded, and for each one switch your OS to light and back to dark or vice versa to see that everything is operating as intended, and all UI effects are correct.
> 
> image

@Jaifroid Could you please tell if this bug was present [this commit](https://github.com/kiwix/kiwix-js/pull/793/commits/50e2841d80578d8ce54e17ffd1fd4d5f54d7d567), that'll be a help in debugging the code

Jaifroid:
> @Jaifroid Could you please tell me if this bug was present before last this commit, that'll be a help in debugging the code

Yes, I did the test earlier this morning, and pulled the latest commits to this PR when I did it. But please do test yourself and confirm.

www/js/lib/uiUtil.js
```diff
@@ -439,13 +449,6 @@ define(rqDef, function() {
         footer.classList.add(contentTheme || '_light');
         // Embed a reference to applied theme, so we can remove it generically in the future
         htmlEl.dataset.theme = theme;
-        // Hide any previously displayed help
-        var oldHelp = document.getElementById(oldTheme + '-help');
-        if (oldHelp) oldHelp.style.display = 'none';
-        // Show any specific help for selected contentTheme
-        var help = document.getElementById(theme + '-help');
-        if (help) help.style.display = 'block';
```
Jaifroid, line 447: @Maneet-k Why did you remove this code in the latest commit?

www/js/lib/uiUtil.js
Jaifroid: @Maneet-k You haven't addressed my specific question here. You are producing options that are invalid, it seems to me. This code can produce a theme `light_mwInvert` which makes no sense to me.

www/js/lib/uiUtil.js
Jaifroid: I see you've moved the code -- kindly explain why it was necessary.

www/js/lib/uiUtil.js
```diff
@@ -415,9 +415,12 @@ define(rqDef, function() {
      * For each contentTheme, a stylesheet must be provided in www/css that is named 'kiwixJS' + contentTheme
      * A rule may additionally be needed in app.css for full implementation of contentTheme
      * 
-     * @param {String} theme The theme to apply (light|dark[_invert|_mwInvert])
+     * @param {String} theme The theme to apply (light|dark[_invert|_mwInvert]|auto[_invert|_mwInvert])
      */
     function applyAppTheme(theme) {
+        var darkPreference = window.matchMedia('(prefers-color-scheme:dark)');
+        var selectedThemeContent = theme.replace(/^[^_]*/, '');
```
Jaifroid, line 422: `selectedThemeContent` is exactly the same code as `contentTheme` below. It looks like redundant code.

www/js/lib/uiUtil.js
```diff
@@ -428,6 +431,13 @@ define(rqDef, function() {
         var contentTheme = theme.replace(/^[^_]*/, '');
         var oldAppTheme = oldTheme.replace(/_.*$/, '');
         var oldContentTheme = oldTheme.replace(/^[^_]*/, '');
+        // Hide any previously displayed help
+        var oldHelp = document.getElementById(oldContentTheme + '-help');
+        if (oldHelp) oldHelp.style.display = 'none';
+        // Show any specific help for selected contentTheme
+        var help = document.getElementById(contentTheme + '-help');
+        if (help) help.style.display = 'block';
+        if(theme === 'light' + selectedThemeContent) contentTheme = null;
```
Jaifroid, line 440: This line looks like a fudge. First of all, above, you set `theme` to an invalid value (e.g. `light_mwInvert`), and here you remove the invalid part. It would be much better, I believe, if you just to set to a valid values above (e.g., just set to `light`). Also, note inconsistent spacing after `if` here.

Jaifroid (changes requested):
@Maneet-k It would be much easier for us to review if you explain changes you have made and list the tests you have done before you request further reviews. Please address all the comments we make on previous code reviews before requesting further review!

www/index.html
```diff
@@ -468,10 +468,13 @@ <h3>Display settings</h3>
                                     <select class="form-control" id="appThemeSelect">
                                         <option value="light">Light</option>
                                         <option value="dark">Dark (app only)</option>
-                                        <option value="dark_invert">Dark + invert (generic)</option>
-                                        <option value="dark_mwInvert">Dark + invert (Wikimedia)*</option>
+                                        <option value="dark_invert">Dark (generic invert)</option>
+                                        <option value="dark_mwInvert">Dark (wikimedia invert)*</option>
+                                        <option class="auto" value="auto">Auto (app only)</option>
+                                        <option class="auto" value="auto_invert">Auto (generic invert)</option>
+                                        <option class="auto" value="auto_mwInvert">Auto (wikimedia invert)*</option>
```
Maneet-k, line 475: `class=auto` in case, browser doesn't support `prefers-color-scheme` so is used to remove those options from the DOM

www/index.html
```diff
@@ -468,10 +468,13 @@ <h3>Display settings</h3>
...
                                     </select>
-                                    <p id="dark_mwInvert-help" class="appThemeInfo form-text text-muted">* Implements workarounds specific to Wikimedia ZIMs. Try generic option if there are display errors with recent ZIMs.</p>
+                                    <p id="_mwInvert-help" class="appThemeInfo form-text text-muted">* Implements workarounds specific to Wikimedia ZIMs. Try generic option if there are display errors with recent ZIMs.</p>
```
Maneet-k, line 477: changed from `dark_mwInvert` to` _mwInvert` , to get the wikimedia help for whenevr the wikimedia content theme is applied.(purpose : scope)

www/js/lib/uiUtil.js
Maneet-k: this variable is to store the content theme the user has chosen as we are gonna change the the theme values for `auto` themes (so these values are not gonna be same). if won't do this those values will be lost. And this variable is making the code more compact instead of using three `if` statements. Also I'm using this for wikimedia help to remove the invalid part from the `light` theme.

www/js/lib/uiUtil.js
Maneet-k: I moved the code cuz if I haven't moved it that invalid contentTheme would've applied and the jquery was also breaking because of that so I moved this before the content theme gets applied and I could change the invalid contenttheme value to null(auto's light case). this is all done to get the help displayed on the DOM

Maneet-k (commented):
@Jaifroid I've tested the code on edge, firefox and chrome and Getting required results.

www/js/lib/uiUtil.js
```diff
@@ -415,9 +415,12 @@ define(rqDef, function() {
      * For each contentTheme, a stylesheet must be provided in www/css that is named 'kiwixJS' + contentTheme
      * A rule may additionally be needed in app.css for full implementation of contentTheme
      * 
-     * @param {String} theme The theme to apply (light|dark[_invert|_mwInvert])
+     * @param {String} theme The theme to apply (light|dark[_invert|_mwInvert]|auto[_invert|_mwInvert])
      */
     function applyAppTheme(theme) {
+        var darkPreference = window.matchMedia('(prefers-color-scheme:dark)');
+        var selectedThemeContent = theme.replace(/^[^_]*/, '');
+        if (/^auto/.test(theme)) theme = darkPreference.matches ? 'dark' + selectedThemeContent : 'light' + selectedThemeContent;
```
Jaifroid, line 423: So, instead of setting an invalid value here, why not just set the last value in the ternary construct to `'light'` instead of `'light' + selectedThemeContent`?

www/js/lib/uiUtil.js
Jaifroid: So this is why it's best not to set invalid values in the first place! I think this needs to be tweaked so you set the correct values above, then you won't need to correct them, and you won't need to move this code.

Maneet-k: But then I'm unable to find any other way to get the help wikimedia to DOM when the auto(wikimedia) changes to light because it's easier to display that help on the screen but the removal is difficult if there's another way. So that's why I've done this. And how that value is invalid, if it is then the passed `auto` values are invalid too.

Jaifroid: OK, maybe I'm misunderstanding something, so I'll check properly in a local implementation and get back to you.

Maneet-k: yes @Jaifroid please. And I apologize for taking that much time.

Jaifroid: No need to apologize!

Jaifroid:
@Maneet-k I've had a bit of a play around with this, and came up with these suggested code changes that will remove the redundancy, and mean you don't have to set invalid values and then correct them. These changes aren't as extensive as they look, because some of them are just moving the "Help" part of the code back to where it was previously, which helps to keep the code structured. It keeps your innovations, but reduces the redundancy.

I'm not sure if I can commit this to your PR or not. I could try, or else make a patch file. Or you can look through it and do it manually if you agree with the code. Let me know what you prefer.

image

Jaifroid:
Here's a patch in case it's easier.

[0001-Update-uiUtil.js.zip](https://github.com/kiwix/kiwix-js/files/8008725/0001-Update-uiUtil.js.zip)

Jaifroid:
There's a line missing from the patch (to determine the actual old theme from "auto"). Hang on.

Jaifroid:
So, to above screenshot code, this needs adding:

image

And here's the patch:

[0002-Update-uiUtil.js.zip](https://github.com/kiwix/kiwix-js/files/8008779/0002-Update-uiUtil.js.zip)

NB I only tested quickly. There may be bugs.

Maneet-k:
Thanks @Jaifroid!
 
> NB I only tested quickly. There may be bugs.

Ohkay, I'll test this out and get back to you soon

Maneet-k:
@Jaifroid we still need to make `contentTheme = null` for auto themes and the system preferences theme is light

www/js/lib/uiUtil.js
Maneet-k: And I agree I didn't need to move the code because I could set the contentTheme to null after the help in this code too.

Jaifroid:
> @Jaifroid we still need to make `contentTheme = null` for auto themes and the system preferences theme is light after the wikimedia help is displayed on the DOM

I think I made it so that is done by the regexes at the beginning, so shouldn't be necessary.

Maneet-k:
> I think I made it so that is done by the regexes at the beginning, so shouldn't be necessary.

Jaifroid if we aren't changing the contentTheme to null it's giving hybrid results as light gets inverted because of that. 
Can I show you the code?

Maneet-k:
> > @Jaifroid we still need to make `contentTheme = null` for auto themes and the system preferences theme is light after the wikimedia help is displayed on the DOM
> 
> I think I made it so that is done by the regexes at the beginning, so shouldn't be necessary.

Also code is still the same only theme is broken down into appTheme + contentTheme

Jaifroid:
Ah, I see what you mean now! Apologies, I was rushing earlier. Please do commit your code!

Maneet-k:
> Ah, I see what you mean now! Apologies, I was rushing earlier. 

That's okay @Jaifroid , cuz it was all because of poor communication skills ;_; that I was enable to present my points (And I def. working on them). 

> Please do commit your code!

jaifroid I've committed the code, please tell if we need to make any other changes

Jaifroid (approved):
Thank you @Maneet-k , it's looking good now! Functionality is working well on quick tests. Three auto options are working as intended, and Auto (app only) is producing the correct hybrid if user puts OS into dark mode. I tested on Edge (Chromium) on Windows 10.

Jaifroid:
> cuz it was all because of my poor communication skills ;_; that I was enable to present my points (And I def. working on them).

Not at all! Thank you for working so patiently on this.

Maneet-k:
> Not at all! Thank you for working so patiently on this.

It was all because you never gave up and helped me in process of learning:) and this thing kept me going for this. THANKS @Jaifroid!
I guess all we need now is confirmation from mossroy

mossroy:
I'm not forgetting this PR.
I just need to find a bit more time to test it: maybe tomorrow

www/js/app.js
```diff
@@ -166,7 +165,24 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
     
     // Unique identifier of the article expected to be displayed
     var expectedArticleURLToBeDisplayed = "";
-    
+ 
+    // define and store dark preference for matchMedia
+    var darkPreference = window.matchMedia('(prefers-color-scheme:dark)');
+    // if 'prefers-color-scheme' is not supported in the browser
```
mossroy, line 171: The comment should explain that, in this case, the "auto*" options are not displayed to the user.
I did not understand at the beginning

www/js/app.js
```diff
@@ -166,7 +165,24 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
...
+    if (window.matchMedia('(prefers-color-scheme)').media === 'not all') {
+        var optionsToBeRemoved = document.querySelectorAll('.auto');
```
mossroy, line 173: Maybe there is a risk to match some other DOM elements that would use the same class name.
It would be safer to do that only on the drop-down list

mossroy (changes requested):
It seems to work well (only tested on latest Firefox on Ubuntu 20.04), thanks!
I also saw no regression on Firefox OS.

But I just added some remarks that should be fixed before merging IMHO. I suppose it should not be very complicated to fix.

www/index.html
```diff
@@ -468,10 +468,13 @@ <h3>Display settings</h3>
                                     <select class="form-control" id="appThemeSelect">
```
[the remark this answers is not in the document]
Maneet-k: In my opinion,Ig we don't need to tell them this, as there won't be any auto options for OS/devices that can't handle dark mode.
Also I didn't clearly understand like to whom this note be displayed(to the available one or to the not available one or to everyone)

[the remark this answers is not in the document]
Jaifroid: I think we can do this succinctly. I presume @mossroy means that some people might think "Auto" means that the app will decide when to display dark theme and when to display light theme. But in reality, "Auto" means "whatever is set in the Operating System".

So all we want is a little note in the HTML that says "Auto themes match the dark/light mode of your device." I guess this note can be hidden if auto themes are not available.

This can be done with the same `querySelectorAll()` statement that you use, e.g. `querySelectorAll('option.auto, p.kiwix-auto-description')` (if you give that paragraph the class name `class="kiwix-auto-description"`).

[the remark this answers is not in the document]
Maneet-k: okay @Jaifroid  , I got it👍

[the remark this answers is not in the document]
Maneet-k: jaifroid, I'm implementing a code that'll display this note for 4seconds when the user selects auto theme and after 4secs this note will be removed from the DOM. Would that be okay?

Maneet-k:
@Jaifroid this is what I meant before, if this isn't a good idea please tell me if this is looking overwhelming or something else .

Jaifroid:
> @Jaifroid this is what I meant before, if this isn't a good idea please tell me if this is looking overwhelming or something else .

OK, I'll test and let you know how it looks!

Maneet-k:
> OK, I'll test and let you know how it looks!

Ohkiee:)

www/js/lib/uiUtil.js
```diff
@@ -438,14 +441,21 @@ define(rqDef, function() {
...
         // Show any specific help for selected contentTheme
-        var help = document.getElementById(theme + '-help');
+        var help = document.getElementById(contentTheme.replace(/_/, '') + '-help');
         if (help) help.style.display = 'block';
-        
+        // Remove the contentTheme for auto themes whenever system is in light mode
+        if (/^auto/.test(theme) && appTheme === 'light') contentTheme = null;
+        // Hide any previously displayed discription for auto themes
+        var oldDescription = document.getElementById('kiwix-auto-description');
+        oldDescription.style.display = 'none';
```
mossroy, line 455: It might be safer to test if oldDescription is defined before using it here (just like elsewhere)

mossroy (approved):
This PR now looks good to me.
Tested again on my FirefoxOS device: auto options are not available (which is normal)
Tested on Firefox 97 on Ubuntu: follows the Ubuntu theme with no issue

Jaifroid:
I'll do some testing of the final version on Windows-based browsers. I may not be able to do this till weekend.

www/js/lib/uiUtil.js
Maneet-k: >I don't think there is a need to hide the note after a few seconds, but I leave this conversation open in case you want to discuss more about it

@mossroy I was just wanted to testing if this will look okay but later I decided on j=keeping it simple cuz this one was looking better.

>It might be safer to test if oldDescription is defined before using it here (just like elsewhere)

Ohkay, will move this 🚀

Jaifroid (approved):
I've now tested on Edge Chromium on Windows 11 and on Internet Explorer Mode. In the former, everything works as expected. In the latter, the Auto options don't show (this is by design, as the browser doesn't support the media query). In IE11 dark modes don't work properly, but this is a known issue with IE11 and pre-exists this PR. We originally decided not to fix it for IE11 because IE11 support is deprecated.

So, for me, this is ready to merge.

www/js/lib/uiUtil.js
Jaifroid: @Maneet-k Is this minor thing still pending?

Maneet-k: Oh I'm sorry @Jaifroid , it totally slipped my mind.

Maneet-k: @Jaifroid as suggested by you that we need to follow the style pattern. Don't you think as `oldHelp` was defined just before using it, So should `oldDescription` follow this pattern. It's just my reason for describing this here. If you want me to fix this, I'm on it.

mossroy: If there is no element in the DOM with name `kiwix-auto-description`, then this line will fail to set a style on it.

Maneet-k: Oh yess! I got it

mossroy: I don't see how your latest commit fixes the risk that oldDescription could be null. You just moved its instanciation, which does not fix anything.
`getElementById` can return null if the DOM element is not found. See https://developer.mozilla.org/en-US/docs/Web/API/Document/getElementById
If this happens, `oldDescription.style.display` will raise an exception.
You can avoid this by testing `if (oldDescription)` before this statement , like it's done for `description`, `help` etc.

www/js/lib/uiUtil.js
```diff
@@ -438,14 +441,21 @@ define(rqDef, function() {
...
         // Show any specific help for selected contentTheme
-        var help = document.getElementById(theme + '-help');
+        var help = document.getElementById(contentTheme.replace(/_/, '') + '-help');
         if (help) help.style.display = 'block';
-        
+        // Remove the contentTheme for auto themes whenever system is in light mode
+        if (/^auto/.test(theme) && appTheme === 'light') contentTheme = null;
+        // Hide any previously displayed description for auto themes
+        var oldDescription = document.getElementById('kiwix-auto-description');
+        if(oldDescription) oldDescription.style.display = 'none';
```
mossroy, line 455: Please add a space between `if` and `(oldDescription)` and we'll start the process of merging this PR.
There is a conflict with master that I will resolve

mossroy:
I've resolved the conflict (certainly introduced by what we have done on master in the meantime, unrelated to this PR)

Before merging, you would ideally rebase and squash this PR
@Maneet-k do you know how to do that?

If not, I'll squash and merge from the Github UI (which might not rebase, I don't remember)

Maneet-k:
> Before merging, you would ideally rebase and squash this PR @Maneet-k do you know how to do that?

ok @mossroy , I don't know how to do this But I'll try.
Could you please tell me if I need to squash the last commit that was done by you?

Maneet-k:
> > Before merging, you would ideally rebase and squash this PR @Maneet-k do you know how to do that?
> 
> ok @mossroy , I don't know how to do this But I'll try. Could you please tell me if I need to squash the last commit that was done by you?

Because I'm getting other commits in my git-rebase-todo file. So when I'm squashing commits(that are done by me)into the single one. The other commits are having conflicts with merging.

mossroy:
Well, it's probably not the easiest case for you to learn how to do that.
So I'll squash and merge with Github UI.

By the way, next time you create a PR on an open-source project, you should create a branch in your fork, instead of committing on your master.

Maneet-k:
> By the way, next time you create a PR on an open-source project, you should create a branch in your fork, instead of committing on your master branch.

Ohkay, Thanks!!

mossroy:
Thanks for your PR

Issue #794, opened 2022-01-10
Add checkbox or button to empty appCache in Cache API panel


Jaifroid:
This would mostly be a utility for development, but could also be useful for a user who needs to reset their app to the latest code for troubleshooting.

With the new code as a result of #771, in SW mode (and also in jQuery mode when being run from a secure server) the app will only update its own code if the appVersion number has changed. This complicates development because it is necessary to empty the appCache explicitly in DevTools before reloading the app's page, otherwise new code will not be displayed. A checkbox could be added in the Cache API panel under "Do not cache assets". This checkbox would read "Also empty app cache (refreshes app from server)". The checkbox would only appear if appCache is in use, and would be automatically unchecked if user turns "Cache assets" back on.

For an ordinary user, also, there may be times when the appCache has been filled incorrectly, or partial contents were cached. It would be useful to be able to "reset" the app from the UI.

PR #797, opened 2022-01-12
Provide methods for resetting the app and bypassing appCache

Jaifroid:
This is intended to address #794, because we have a number of new developers who will find the new functionality confusing.

This provides a UI for resetting the app (see screenshot) under Expert Settings.

The button "**Reset**" performs a full app reset consisting of:

* Warning the user and seeking confirmation
* Clearing any cookie entries
* Clearing any Local Storage entries
* Deleting the Cache API caches
* Deregistering any registered Service Workers
* Rebooting the app

Of course all settings are lost and reverted to defaults.

The **checkbox** is intended for Service Worker mode only (it will complain if set in JQuery mode). It bypasses the appCache by:

* Clearing the Cache API caches
* Preventing Service Worker from using the appCache, forcing it to fetch all app files

When re-enabling the appCache (by unticking the checkbox), it will:

* Deregister the current Service Worker
* Reboot the app in order to reinstall the Service Worker and repopulate the offline appCache.

image

### DIFF
--- a/service-worker.js
+++ b/service-worker.js
@@ -239,15 +239,18 @@ self.addEventListener('message', function (event) {
             outgoingMessagePort = null;
             fetchCaptureEnabled = false;
         }
+        var oldValue;
         if (event.data.action.assetsCache) {
             // Turns caching on or off (a string value of 'enable' turns it on, any other string turns it off)
+            oldValue = useAssetsCache;
             useAssetsCache = event.data.action.assetsCache === 'enable';
-            console.debug('[SW] Use of assetsCache was switched to: ' + event.data.action.assetsCache);
+            if (useAssetsCache !== oldValue) console.debug('[SW] Use of assetsCache was switched to: ' + event.data.action.assetsCache);
         }
         if (event.data.action.appCache) {
             // Enables or disables use of appCache
+            oldValue = useAppCache;
             useAppCache = event.data.action.appCache === 'enable';
-            console.debug('[SW] Use of appCache was switched to: ' + event.data.action.appCache);
+            if (useAppCache !== oldValue) console.debug('[SW] Use of appCache was switched to: ' + event.data.action.appCache);
         }
         if (event.data.action === 'getCacheNames') {
             event.ports[0].postMessage({ 'app': APP_CACHE, 'assets': ASSETS_CACHE });
--- a/service-worker.js
+++ b/service-worker.js
@@ -55,6 +55,14 @@ const APP_CACHE = 'kiwixjs-appCache-' + appVersion;
  */
 var useCache = true;
 
+/**
+ * A global Boolean that governs whether the APP_CACHE will be used
+ * This is an expert setting in Configuration
+ * @type {Boolean}
+ */
+ var appCache = true;
+
+
 /**  
  * A regular expression that matches the Content-Types of assets that may be stored in ASSETS_CACHE
  * Add any further Content-Types you wish to cache to the regexp, separated by '|'
@@ -236,6 +244,11 @@ self.addEventListener('message', function (event) {
             useCache = event.data.action.useCache === 'on';
             console.debug('[SW] Caching was turned ' + event.data.action.useCache);
         }
+        if (event.data.action.appCache) {
+            // Enables or disables use of appCache
+            appCache = event.data.action.appCache === 'enable';
+            console.debug('[SW] Use of appCache was switched to: ' + event.data.action.appCache);
+        }
         if (event.data.action === 'getCacheNames') {
             event.ports[0].postMessage({ 'app': APP_CACHE, 'assets': ASSETS_CACHE });
         }
@@ -328,7 +341,7 @@ function removeUrlParameters(url) {
  */
 function fromCache(cache, requestUrl) {
     // Prevents use of Cache API if user has disabled it
-    if (!useCache && cache === ASSETS_CACHE) return Promise.reject('disabled');
+    if (!appCache && cache === APP_CACHE || ! useCache && cache === ASSETS_CACHE) return Promise.reject('disabled');
     return caches.open(cache).then(function (cacheObj) {
         return cacheObj.match(requestUrl).then(function (matching) {
             if (!matching || matching.status === 404) {
@@ -349,7 +362,7 @@ function fromCache(cache, requestUrl) {
  */
 function updateCache(cache, request, response) {
     // Prevents use of Cache API if user has disabled it
-    if (!useCache && cache === ASSETS_CACHE) return Promise.resolve();
+    if (!appCache && cache === APP_CACHE || ! useCache && cache === ASSETS_CACHE) return Promise.resolve();
     return caches.open(cache).then(function (cacheObj) {
         console.debug('[SW] Adding ' + request.url + ' to ' + cache + '...');
         return cacheObj.put(request, response);
--- a/service-worker.js
+++ b/service-worker.js
@@ -239,10 +239,10 @@ self.addEventListener('message', function (event) {
             outgoingMessagePort = null;
             fetchCaptureEnabled = false;
         }
-        if (event.data.action.useCache) {
-            // Turns caching on or off (a string value of 'on' turns it on, any other string turns it off)
-            useAssetsCache = event.data.action.useCache === 'on';
-            console.debug('[SW] Caching was turned ' + event.data.action.useCache);
+        if (event.data.action.assetsCache) {
+            // Turns caching on or off (a string value of 'enable' turns it on, any other string turns it off)
+            useAssetsCache = event.data.action.assetsCache === 'enable';
+            console.debug('[SW] Use of assetsCache was switched to: ' + event.data.action.assetsCache);
         }
         if (event.data.action.appCache) {
             // Enables or disables use of appCache
--- a/service-worker.js
+++ b/service-worker.js
@@ -250,13 +250,13 @@ let fetchCaptureEnabled = false;
             // Turns caching on or off (a string value of 'enable' turns it on, any other string turns it off)
             oldValue = useAssetsCache;
             useAssetsCache = event.data.action.assetsCache === 'enable';
-            if (useAssetsCache !== oldValue) console.debug('[SW] Use of assetsCache was switched to: ' + event.data.action.assetsCache);
+            if (useAssetsCache !== oldValue) console.debug('[SW] Use of assetsCache was switched to: ' + useAssetsCache);
         }
         if (event.data.action.appCache) {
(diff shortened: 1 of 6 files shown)


Jaifroid:
If these settings or UI are considered confusing for basic users of the app (non-developers) we could hide them and have a small link at the bottom of the Expert settings panel that would say:

`Reset and dev options >`

This link would reveal the UI.

service-worker.js
```diff
@@ -231,10 +239,18 @@ self.addEventListener('message', function (event) {
...
-        if (event.data.action.useCache) {
-            // Turns caching on or off (a string value of 'on' turns it on, any other string turns it off)
-            useCache = event.data.action.useCache === 'on';
-            console.debug('[SW] Caching was turned ' + event.data.action.useCache);
+        var oldValue;
+        if (event.data.action.assetsCache) {
+            // Turns caching on or off (a string value of 'enable' turns it on, any other string turns it off)
+            oldValue = useAssetsCache;
+            useAssetsCache = event.data.action.assetsCache === 'enable';
+            if (useAssetsCache !== oldValue) console.debug('[SW] Use of assetsCache was switched to: ' + event.data.action.assetsCache);
```
mossroy, line 247: Maybe you should log `useAssetsCache` instead of `event.data.action.assetsCache`

service-worker.js
```diff
@@ -231,10 +239,18 @@ self.addEventListener('message', function (event) {
...
+        }
+        if (event.data.action.appCache) {
+            // Enables or disables use of appCache
+            oldValue = useAppCache;
+            useAppCache = event.data.action.appCache === 'enable';
+            if (useAppCache !== oldValue) console.debug('[SW] Use of appCache was switched to: ' + event.data.action.appCache);
```
mossroy, line 253: Same here

www/js/app.js
```diff
@@ -83,7 +83,9 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
     // Maximum number of article titles to return (range is 5 - 50, default 25)
     params['maxSearchResultsSize'] = settingsStore.getItem('maxSearchResultsSize') || 25;
     // A global parameter that turns caching on or off and deletes the cache (it defaults to true unless explicitly turned off in UI)
-    params['useCache'] = settingsStore.getItem('useCache') !== 'false';
+    params['assetsCache'] = settingsStore.getItem('assetsCache') !== 'false';
+    // A global parameter that disables use of the appCache (it defaults to true unless explicitly turned off in UI)
```
mossroy, line 87: The comment looks confusing to me. The one of assetsCache is more clear IMHO

www/js/lib/settingsStore.js
```diff
@@ -91,6 +92,120 @@ define([], function () {
     return type;
   }
 
+  /**
+   * Performs a full app reset, deleting all caches and settings
+   * Or, if a paramter is supplied, deletes or disables the object
```
mossroy, line 97: typo : parameter

www/js/lib/settingsStore.js
```diff
@@ -91,6 +92,120 @@ define([], function () {
...
+        // If the cookie key starts with the keyPrefix
+        if (~params.keyPrefix.indexOf(decodeURIComponent(cookieCrumb[0]))) {
+          cook = true;
+          key = cookieCrumb[1];
+          // This expiry date will cause the browser to delete the cookie on next page refresh
+          document.cookie = key + '=;expires=Thu, 21 Sep 1979 00:00:01 UTC;';
+        }
+        cookieCrumb = cookieKeys.exec(currentCookie);
+      }
+      if (cook) console.debug('All cookie keys were expiered...');
```
mossroy, line 120: typo : expired

www/js/lib/settingsStore.js
```diff
@@ -91,6 +92,120 @@ define([], function () {
...
+   * Performs a full app reset, deleting all caches and settings
+   * Or, if a paramter is supplied, deletes or disables the object
+   * @param {String} object Optional name of the object to disable or delete ('cookie', 'localStorage', 'cacheAPI')
+   */
+  function reset(object) {
+    // If no specific object was specified, we are doing a general reset, so ask user for confirmation
+    if (!object && !confirm('WARNING: This will reset the app to a freshly installed state, deleting all app caches and settings!')) return;
+    
+    // 1. Clear any cookie entries
+    if (!object || object === 'cookie') {
```
mossroy, line 105: The variable names in this code block are not very meaningful

mossroy (commented):
I agree with the purpose of the PR.
I did not test but had a quick look at the code (see my comments).

I'm wondering if we should not move the content injection mode choice outside of the "expert settings", now that we are close to switching it to SW by default (but it might be a different topic/PR)
And it could be renamed "Expert/developer settings" with your new content.

Regarding the "Bypass appCache" checkbox, I find the UI confusing. The fact that it's on the same line as the reset button tends to make me believe it is an option of the reset button.
And I think it should be hidden when in jQuery mode

Jaifroid:
@mossroy Thanks for the quick response. Some comments below:

> I'm wondering if we should not move the content injection mode choice outside of the "expert settings"

That's a good idea, and can be done as part of this PR. Any idea what to call the box for this? Suggestions:

* "Compatibility settings"
* "Compatibility mode"
* "Operation"
* "Mode of operation"

> Regarding the "Bypass appCache" checkbox, I find the UI confusing.

OK, I can see that. I can relocate to next line and only show that option if SW mode is enabled. The Reset button works in both modes, btw.

Jaifroid:
He, he, I had to fire up IE11 and run it from the file:// protocol to discover (remember) that we don't actually use the keyPrefix ('kiwixjs-') in front of legacy cookie keys, hence settings weren't being reset in IE11.

I've amended the code to fix this and also discovered another very minor incompatibility also fixed.

I was actually quite impressed that our code still runs fine from file protocol in IE11! That's quite some achievement in 2022...

Jaifroid:
Suggested new UI in screenshots below (left is what you see in jQuery mode, right in SW mode). I've also set the gh-pages implementation to this branch: https://kiwix.github.io/kiwix-js/www/index.html . If you had already visited this implementation, it won't immediately show the new code. It will self-update in SW mode, but only after exit and restart of app, if it has had time to cache the new code. Thereafter, the reset button should show. Alternatively, go to Dev Tools, Application tab (or Storage tab in FF), and delete the appCache, then reload.

image

Jaifroid:
Should I move the option "Permanently hide active content warning" into Expert settings too? I'm not sure we really want ordinary users deactivating that warning if we move to having SW mode as the default. We need to be able to warn them if they've switched to JQuery mode and forgotten about it.

Jaifroid:
Possible alternative UI arrangement (rough approximation, because this is done as a mockup in KJSW):

image

mossroy:
Thanks for the screen snapshots.
"Compatibility settings" looks good. Maybe we should remove the "requires secure context" : with the PWA version for Firefox, it should now be supported on all recent platforms. And I don't think the users will understand what we mean by "secure context".

For "Expert settings", I prefer your second ("alternative") proposal. The layout makes it more clear that these are separate options/actions
I agree with moving the "Permanently hide active content warning" option there

Jaifroid:
This is what UI looks like now (after latest commit):

image

mossroy:
Looks great!

Jaifroid:
Test implementation: https://kiwix.github.io/kiwix-js/ updated to latest code on this branch, for quick testing. App may need to self-update to version 3.3.7 (a dummy version number) and should notify when that version is ready (in Config) if app is being used. Anything lower, and it's not on the latest code.

www/js/app.js
```diff
@@ -431,27 +449,29 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
...
             resetCssCache();
             if ('caches' in window) caches.delete(ASSETS_CACHE);
             refreshCacheStatus();
         }
     });
+    var titleSearchRangeVal = document.getElementById('titleSearchRangeVal');
     document.getElementById('titleSearchRange').addEventListener('change', function(e) {
         settingsStore.setItem('maxSearchResultsSize', e.target.value, Infinity);
         params.maxSearchResultsSize = e.target.value;
+        titleSearchRangeVal.innerHTML = e.target.value;
```
mossroy, line 471: The code scanning warning can probably be solved easily here, even if there's no real security issue.
Using `.textContent` instead of `.innerHTML` is probably enough.

mossroy (approved):
I did not test a lot, but I'm sure you did.
OK for me

www/js/app.js
```diff
@@ -154,7 +154,7 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
     document.getElementById('hideActiveContentWarningCheck').checked = params.hideActiveContentWarning;
     document.getElementById('showUIAnimationsCheck').checked = params.showUIAnimations;
     document.getElementById('titleSearchRange').value = params.maxSearchResultsSize;
-    document.getElementById('titleSearchRangeVal').innerHTML = encodeURIComponent(params.maxSearchResultsSize);
+    document.getElementById('titleSearchRangeVal').textContent = encodeURIComponent(params.maxSearchResultsSize);
```
mossroy, line 157: I suppose it's now unnecessary to call `encodeURIComponent` here (we probably did that to avoid the same warning)

Jaifroid: OK, will check.

Jaifroid:
Will test on real IE11 on a Windows 10 machine before merging. Works OK in IE11 Mode on Windows 11, but it's not 100% the same.

Jaifroid:
Test on real IE11 successful.

Jaifroid:
Small "bug" (more of a feature!) discovered when testing reset in the Firefox extension in SW mode. Because the app uses the query string to set parameters between the local extension and the PWA, if the user presses the reset button while the query string still has several parameters in it, then the parameters are set again from the query string.

The solution will be to clear the query string EXCEPT for the `referrerExtensionURL`, which should be retained so that the user can go back to the extension from the UI. We will have to retain the `contentInjectionMode` too, as by definition the remote extension/PWA should be in SW mode.

I'll test a quick fix for that.

Jaifroid:
That issue is now fixed. All tests passing, including the following:

* Reset App button works fine in IE11
* Reset App in the PWA launched from Firefox extension will remain in SW mode and will retain access to local extension code if user switches back to jQuery mode
* Unchecking the Bypass App Cache checkbox will reboot the app to repopulate the AppCache, but all settings are retained;
* Works in Chromium extension

So I'll squash/merge this after midday GMT unless I hear any objections.

PR #804, opened 2022-01-20
Added bootstrap modal for alert and confirm popups:WIP #798

gaurav7019:
# Closes #798

### Changes
- Added the HTML code for a Bootstrap modal in the `www/index.html`
- Added `systemAlert()` function(returns a promise) in the `uiUtil.js` to call the bootstrap modal. It resolves with true if the Confirm button(only available if apt arg is passed) was clicked or false if the Close button was clicked
- Used(in `app.js`) the `systemAlert()` function to call an alert popup when a user tries to search without loading a ZIM archive file

### Flags
- In `settingsStore.js`, I wasn't able to use functions exposed by `uiUtil.js`. uiUtil is returning undefined even when I have added that in the dependency array in the define() at the beginning
- Using custom JS for closing the modal by setting the display property to none doesn't work as it leaves behind the backdrop of the modal. I can instead use custom JS code to both show and hide the modal which was throwing some errors earlier, will work on that
- The documentation of `systemAlert()` in the `uiUtil.js` is not updated yet, will do once approved

### Screenshots
Screenshots of the unit tests and the modal

unit tests

search alert

### Related Issues
- Issue #798

PR #805, opened 2022-01-20
Update changelog for v3.3.0

Jaifroid:
This is still work in progress.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ Then it was renamed Kiwix-html5 (and uses ZIM files), and then again was renamed
 Released on *TODO*
 
 * FEATURE: Provided a workaround to enable Service Worker mode in Firefox extensions (via a PWA)
-* NEW: App now defaults to Service Worker mode if accessed as a PWA
+* NEW: App now defaults to Service Worker mode if accessed directly as a PWA (does not apply to browser extensions)
 * NEW: Provided methods for resetting the app and bypassing the appCache from the UI
 * NEW: Support for non-http URLs in img and link tags (e.g. data: javascript:)
 * NEW: Added helpful tooltips to icons and settings in the app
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -9,15 +9,19 @@ Then it was renamed Kiwix-html5 (and uses ZIM files), and then again was renamed
 Released on *TODO*
 
 * FEATURE: Provided a workaround to enable Service Worker mode in Firefox extensions (via a PWA)
+* NEW: App now defaults to Service Worker mode if accessed as a PWA
```
mossroy, line 12: It would be better to give a more specific explanation here.
With the line above, one could conclude that we now default to SW mode in Firefox extensions (which is wrong)

```suggestion
* NEW: App now defaults to Service Worker mode if accessed directly as a PWA (does not apply to browser extensions)
```

Jaifroid: Thanks, I committed that change.

CHANGELOG.md
```diff
@@ -8,7 +8,22 @@ Then it was renamed Kiwix-html5 (and uses ZIM files), and then again was renamed
 
 Released on *TODO*
 
-Detailed changelog: TODO
+* FEATURE: Provided a workaround to enable Service Worker mode in Firefox extensions (via a PWA)
+* NEW: App now defaults to Service Worker mode if accessed directly as a PWA (does not apply to browser extensions)
+* NEW: Provided methods for resetting the app and bypassing the appCache from the UI
+* NEW: Support for non-http URLs in img and link tags (e.g. data: javascript:)
```
mossroy, line 14: I suppose it's only about jQuery mode? If so, it would be better to write it here

Jaifroid:
I've added changelog items for #804 , because I think it's ready. Obviously I won't merge this changelog until after merging #804 if it is indeed ready. Let me know if there are in principle any further items you'd like to add to this changelog, @mossroy .

Jaifroid:
As far as I'm concerned, this changelog is now complete.

mossroy (approved):
Looks good!
You can merge

Issue #802, opened 2022-01-15
Release 3.3.0


kelson42:
The 3.3 milestone is already quite interesting and has still a lot of work to do. I see work to do for at least 4 months. Last release is already almost 6 months ago. Release early, release often is good. I want to encourage you to reassess the milestone to 3.3.0.

- [x] Merge at least #792, #797, #800
- [x] Fix critical bug #806
- [ ] ~~Possibly merge #804~~
- [x] Update the changelog file
- [x] Further tests 
- [x] Rename existing milestones (increment their names), create a new 3.4 milestone and move open issues to this new 3.4
- [x] Create a tag on git
- [x] Generate unsigned extensions (should be automatically triggered on GitHub Actions by the tag creation)
- [x] Check that the PWA version is correctly set in moz-extension.kiwix.org (should be automatically triggered by tag creation) - had to do this manually
- [x] Upload the unsigned extensions to Stores:
  - [x] Firefox - https://addons.mozilla.org/fr/developers/addon/kiwix-offline/edit
  - [x] Chrome
  - [x] Edge - https://partner.microsoft.com/en-us/dashboard/microsoftedge/
  - [x] Ubuntu touch - https://open-store.io
- [x] Prepare the source code for version 3.4
- [x] When available, copy the signed extension binaries to https://download.kiwix.org/release/browsers/
- [x] Update the kiwix.github.io implementation - should no longer be required, but check that the implementation is synchronized automatically with master
- [ ] Publicize these new versions

Jaifroid:
@kelson42 - [I fixed tick boxes above]

I agree we are quite close. We don't need to do everything marked as v3.3, we can push those not complete and those that are not essential to v3.4. I'll add the current PRs we should merge to this.

mossroy:
I fixed the version numbers in the tick boxes.

I agree we're close to have enough changes to release a 3.3

Jaifroid:
See #805 for initial changelog.

mossroy:
I suggest that we try to do this release this week-end, if possible.
If the current open PRs are not merged, they will be for next release, never mind

mossroy:
I've done some tests on my Firefox OS device, with various ZIM files.
Apart from #814 , I did not find any issue.

Jaifroid:
[...]

Issue #790, opened 2022-01-04
Add downloadable fonts (from ZIM) to files processed in jQuery mode


Jaifroid:
While we said we would not optimize jQuery mode any further, this looks like a bugfix. Recent Gutenberg ZIMs seem to rely on font files donwloaded form the ZIM to produce the PDF / EPUB / HTML icons. Since we don't process these as we do images, there are now empty placeholders where the icons should be. See screenshot and note the corresponding "donwloadable font" errors in the console. This shouldn't be too hard to fix within our existing processing of images or possibly media. It would be a case of converting these URLs to blob URLs. This isn't technically a regression in our code, but it is a regression in usability of Gutenberg archives in jQuery mode.

![image](https://user-images.githubusercontent.com/4304337/148126804-e0d36b4d-5695-4220-a657-843c3ed58c15.png)

ankur12-1610:
@Jaifroid I would like to work on this issue, but I'm not sure how to proceed :sweat_smile:

kelson42:
I don't think anything has changed recently on this. We always have used font-awesome for these icons.

Jaifroid:
Well it's possible we've only just noticed an issue that has existed for a while, but I'm 99% sure that old Gutenberg ZIMs were able to show icons (they are quite prominent, and I have worked a lot on Gutenberg ZIMs supporting downloadable books before, so I would have noticed such a glaring issue). Now they can't display the icons (in Kiwix JS), And in any case, we don't have any code to extract these fonts from the ZIM in jQuery mode, which is what is needed here. So it's worth working on this despite the fact that it works natively in SW mode, as it is really just a small addition to the image download code, I think.

Jaifroid:
@ankur12-1610 I've assigned you -- thanks.

mossroy:
> While we said we would not optimize jQuery mode any further, this looks like a bugfix. Recent Gutenberg ZIMs seem to rely on font files donwloaded form the ZIM to produce the PDF / EPUB / HTML icons. Since we don't process these as we do images, there are now empty placeholders where the icons should be. See screenshot and note the corresponding "donwloadable font" errors in the console. This shouldn't be too hard to fix within our existing processing of images or possibly media. It would be a case of converting these URLs to blob URLs. This isn't technically a regression in our code, but it is a regression in usability of Gutenberg archives in jQuery mode.

Forgot to approve to work on this if you wish

ankur12-1610:
[...]

PR #807, opened 2022-01-22
Fix incorrect processing of titles with question marks and anchor parameters

### LINKED ISSUE #806: Title with question mark not handled correctly in Service Worker mode
Using the ZIM `gutenberg_it_all_2022-01.zim`, if I search for the title `Che cosa è l'amore?` in jQuery mode, I can open the pages corresponding to the book and the cover with no problem.

If I switch to SW mode and search for the same title, either from our search field, or by searching for the author Panzoni in the custom UI and then click on the title, I get an error in console: `Title A/Che cosa è l'amore not found in archive.`

See screenshot, which shows first the error from the custom UI, and then the same error from our own search.

I think we should fix this before releasing 3.3.

image

Jaifroid:
Fixes #806.

This initial fix needs very thorough testing.


Jaifroid:
Now added processing of anchor parameters on links in JQuery mode.

Tested on many titles with question marks and quotation marks, both single and double.

Tested on links with anchor attributes (e.g. ../An_article#Heading_to_jump_to) in both SW and JQuery modes.

service-worker.js
```diff
@@ -339,7 +340,11 @@ function fetchRequestFromZIM(fetchEvent) {
  * @returns {String} The same URL without its parameters and anchors
  */
 function removeUrlParameters(url) {
-    return url.replace(/([^?#]+)[?#].*$/, '$1');
+    // Remove any querystring
+    var strippedUrl = url.replace(/\?[^?]*$/, '');
+    // Remove any anchor parameters
+    strippedUrl = strippedUrl.replace(/#[^#]*$/, '');
```
Jaifroid, line 346: There is a potential weakness here. A valid encoded URI might have for example: 'My_friend&#39;s_house'. This regexp would strip everything after the #. Note that running decodeURI or decodeURIComponent does **not** decode that as an apostrophe. In practice, I don't believe our ZIMs are using this kind of encoding. Working around it is tricky because negative lookbehinds are not supported in any browsers except recent Chromium, so we can't easily search for a # that is not preceded by an ampersand. We could use the trick of reversing the string and using a negative lookahead (which is supported in all browsers). The logic is a bit mind-bending, though.

Jaifroid: OK, although in HTML5 almost anything can be an ID for an element, in practice JavaScript limits this to the following for backward compatibility with HTML4:

> ID and NAME tokens must begin with a letter ([A-Za-z]) and may be followed by any number of letters, digits ([0-9]), hyphens ("-"), underscores ("_"), colons (":"), and periods (".").

Source: https://stackoverflow.com/a/79022/9727685

In fact, `document.getElementById` cannot select an id that begins with a number, and colons and periods have to be escaped to use CSS selectors for obvious reasons.

In practice it should be safe to say that if we find a semicolon after the # then it is **not** an anchor and we shouldn't strip it, as this will allow us to distinguish any ampersand-encoded characters that use #.

mossroy: Maybe a safer approach could be to let the browser do that for us.
I did not test, but we might create a https://developer.mozilla.org/en-US/docs/Web/API/URL object from that string, then reset its `search` and `hash` properties. Or maybe even read its `pathname` property?

Jaifroid: The issue is IE11 not supporting the properties of a URL that we would need to access. We could do something similar to what we did with `uiUtil.deriveZimUrlFromRelativeUrl()` , which is to use the API if it's available and fall back to a regex if not (assuming the API works for our use case).

mossroy: Here we could safely use it, as we're inside a serviceworker (that will not be executed in IE11).

Jaifroid: OK, true. We also have the same funciton in uiUtil, (unfortunately we have to repeat it) so ideally we should use the same code which will, as you say, use the API in SW mode universally, even though it will have the fallback in it (and we should have a note in both implementations to say that the code must be kept synchronous). I can have a go at this.

mossroy (commented):
Congratulations for this PR.
I know these issues can be quite difficult to handle.

But I would suggest to try to use the URL object https://developer.mozilla.org/en-US/docs/Web/API/URL instead of regular expressions where/when it's possible (for sure in SW mode, maybe in jQuery mode too).
It would certainly be safer, and handle all cases.
For IE11 (that does not support this object properly), we could fallback to the regexp. And never mind if some edge cases are not properly supported for IE11 (let's not waste time on it)

service-worker.js
mossroy: I'm OK with that

Jaifroid:
I've been experimenting a bit with the URL option. It's not as clever as we might like to think. We're supposed to feed it a bare URL string, and it will return a parsed and encoded URL according to RFC3986. So I tried our problematic URL, and we get:

image

It's reproduced the error we were trying to avoid, and it believes `?.38140.htm` is a querystring (look at `search` property) rather than a question mark that needs to be encoded. Additionally, it has swallowed the relative part of the URL string `../`. It's not possible to feed this function just the pathname without the protocol:

image

So we have to encode the URL, and then decode it:

image

NB This constructor fails with the entity references that our `uiUtil.htmlEscapeChars()` produced. But I've abandoned that as outdated (and this is a good example of why -- those codes need themselves to be encoded before putting them in a URL, or else the ampersand and the hash are misconstrued).

In sum, I can work with this constructor, though it is not doing anything more sophisticated than finding everything after a `?` as a querystring, up to a `#` for an anchor, and we still have to take care not to encode accidentally any querystring or anchor.

mossroy:
I'm not surprised it's expecting an encoded URL (else it would have no way to see the difference between a question mark for the querystring and a question mark in the path/filename).
But it's what we already have in SW mode, I suppose.

Wouldn't be enough to do the following?

```
urlObject = new URL("http://localhost/" + url);
return urlObject.pathname;
```
(untested)

My concern is to make sure we don't miss edge cases like the one you found with encoded characters (ex: &#39; ). Are there any other ones we did not think of? Using the built-in object of the browser makes sure all are covered.

But if it's becoming too complicated, never mind.

NB : I don't mind if we keep the regexp for jQuery mode (for IE11 compatibility), and use URL object for SW mode

Jaifroid:
I agree, and I'm testing a solution along these lines.

Jaifroid:
OK, after many frustrating tests trying to create a universal function that could be used in the same way in SW mode and in jQuery mode, I came to the conclusion that the ONLY place that the `URL()` method can be used viably is in the Service Worker.  And the reason we can use it there is only because the Service Worker provides us with a fully qualified URL ('http://domain/path/to/article.html'), which is what is needed for the URL method to work with unless we can provide a base URL.

In jQuery mode, we are working with relative URLs and ZIM URLs, and it is unreliable and complicated to construct artificial fully qualified URLs, since we are inside an iframe which has a fixed `src` . I wrote a ridiculously complicated function to work around the issues in calculating a full URL, or in using a dummy + base URL. Then I realized that this would introduce a lot of overhead in long documents with hundreds if not thousands of hyperlinks, as we would have to do calculations and create URLs for every single one of them, and in IE11 it would be worse using the workaround of creating a DOM (like in `deriveZIMUrlFromRelativeUrl`). I was introducing complexity where the regexes are in fact extremely simple and elegant and are fully road-tested.

So I now agree with you that the most elegant solution is to use the URL function in the Service Worker, which is in fact very simple, and to use our road-tested regexes in jQuery mode. This is what I've done in the latest.

I've finessed the stripping of or conserving of the querystring in the SW as well. I had been stripping it for the APP_:CACHE, but this also stripped them for ZIM assets, which is incorrect, given that a querystring on the end of a JS file can be significant. They are now being correctly conserved (there is a test case in StackExchange ZIMs that use MathJax, where a MathJax JS file requires a significant querystring and so it could not be found in the cache before).

Jaifroid:
> Wouldn't be enough to do the following?
> 
> ```
> urlObject = new URL("http://localhost/" + url);
> return urlObject.pathname;
> ```
> 
> (untested)

Just to explain why this doesn't work in jQuery mode (or with a dummy URL): because we lose the path information if we add the relative URL on to this dummy. It gets stripped back to a bare ZIM URL in the best scenario, and sometimes loses the namespace in the worst scenario.

We in fact take advantage of the relative URL calculations to derive the ZIM URL once we've clicked on a link (but the dummy is added to the base URL, not to the asset URL). However, it is onerous to do it for every link in an article.

In SW, we have `event.request.url`, which is a fully qualified string (not an object), so we don't need any trickery to get an accurate URL object from this using the URL method.

www/js/lib/uiUtil.js
```diff
@@ -307,10 +309,16 @@ define(rqDef, function(settingsStore) {
     }
 
     /**
+     * DEV: This function is no longer used in the project and could be removed unless it is of historical interest.
```
mossroy, line 312: I think it would be better to remove it if it's unused.
We still have it in git history if needed

mossroy (approved):
Great!
And thanks for the unit test, too.

I did not test, but the code looks good

mossroy:
This looks good.
After merging this, maybe we could start the 3.3.0 release?

Jaifroid:
I just want to do some careful manual testing in both modes and and different browsers, and some inspecting of the caches produced in SW mode to be sure it's all working as intended, as the code dependency is a bit tricky, as we've seen from the number of errors we've had to correct over the years coming from non-obvious gotchas in URLs.. It shouldn't take too long. Then, yes, I agree we can continue with the checklist at #802. I'll also need to add some changelog entries for what this PR does. It fixes two or three things to do with URL parameters, including scrolling to URLs with anchors in jQuery mode.

service-worker.js
```diff
@@ -187,14 +187,20 @@ let fetchCaptureEnabled = false;
...
+self.addEventListener('fetch', function (event) {
     // Only cache GET requests
     if (event.request.method !== "GET") return;
-    // Remove any querystring before requesting from the cache
-    var rqUrl = event.request.url.replace(/\?[^?]+$/i, '');
+    var rqUrl = event.request.url;
+    var urlObject = new URL(rqUrl);
+    // Test the URL with querystring removed (hashes are not relevant in this context)
+    var searchParam = decodeURIComponent(urlObject.search);
+    var strippedUrl = event.request.url.replace(searchParam, '');
```
Jaifroid, line 197: Unclear logic here. It should be safer just to use `urlObject.pathname` like we do in the `fetchUrlFromZim` function below, since we're only using this to find out which Cache to use.

Also check assumption that Cache API caches encoded URIs, so we should only fully decode URI before requesting it from the ZIM, not before requesting it from the cache.

Jaifroid: This assumption must be correct because Cache API caches requests and responses. A proper request cannot have a bare question mark or hash anywhere in it except to indicate a querystring or an anchor target.

Jaifroid:
My testing has detected an issue with some fonts that are being requested in jQuery mode that are not being detected as ZIM assets and so are being trapped as if they were requests for the app's own files. I need to sort this out before merging this. It's not hugely serious but is causing a lot of notified failures in console.debug.

Jaifroid:
It is indeed better in terms of UX to cache a 404 response, as it prevents a slow Network request (slow to fail) when that asset is requested again. The symptom, when the app is offline, is that update of the UI is delayed and slow while the browser is waiting for Network. If we have cached the 404 response, then we retrieve the 404 much faster when the app is offline.

To be clear, this issue only affects the PWA if the user switches to jQuery mode, and the Service Worker tries to retrieve an unprocessed asset from the Network (unprocessed because we do not currently process ZIM assets in CSS files, for example).

Also, this will not affect the correct extraction of these assets in Service Worker mode, because in that mode the asset URL is prefixed with the ZIM archive's name. Caching 404s from the Network will also not affect the app's own code (APP_CACHE assets) because if one of those fails to fetch when the Service Worker is registered, then the caching of the entire app will throw and report an error (this is part of the PWA spec).

So, the changes in https://github.com/kiwix/kiwix-js/pull/807/commits/d88b523f36c47af8bceda39f7fc0a575453e96a3 appear safe to me. But let me know if you have concerns. And let me know if it's OK to merge (pending final tests).

Jaifroid:
Apologies, @mossroy, but I'm getting cold feet about caching 404 responses to work around an edge case. I'm worried about unintended consequences, and I note that the Cache API specification bars 404 responses from being cached when the `Cache.add()` method is used (we use `Cache.put()` because it is more permissive, as we are constructing artificial Responses).

It just feels wrong and is also out of scope of this PR, so I'm going to revert https://github.com/kiwix/kiwix-js/pull/807/commits/d88b523f36c47af8bceda39f7fc0a575453e96a3.

Jaifroid:
Interesting error exposed by the rebase of this PR. We are not dealing correctly with a colon in the domain + port, at least in jQuery mode. It might explain #790.

image

Jaifroid:
OK, this turned out to be nothing. It's just the way Chromium displays the failed network requests for these fonts, which are not processed by us at all.

So, I've rebased this on master and have tidied up a bit, and as far as I'm concerned, it's ready to squash and merge. Tested multiple times on Edge Chromium, Firefox and IE11, and as a Chromium extension, both for titles with question marks in and for hash anchors, which with this PR now work in both jQuery and SW modes.

@mossroy Let me know if you're happy for me to do squash/merge.

mossroy (approved):
You can merge

Issue #801, opened 2022-01-14
App doesn't cache its own code if opened only in jQuery mode from a secure server


Jaifroid:
This may only be for information, though a tweak or two could be made if it's considered an issue.

If a user visits https://moz-extension.kiwix.org/ independently in a browser, but never switches to Service Worker mode, then the app never caches its own code, so will not work offline. Once the user switches to SW mode, then the app will continue to work offline even if they switch back to jQuery mode.

In some senses this is academic because when we implement #196 this issue will go away. However, there is a possible interim solution which might be a way of "softly" introducing and testing #196. This would be to test whether the app is running from an https: server (or localhost) and switch to Service Worker automatically if it is.

This would be a half-way house to #196. It would not affect browser extensions (the main target), and would be a way of testing Service Worker as default. It would be pretty much a one-line PR:

`if (/^https:|\/localhost[:\/]/.test(window.location.href) && navigator.serviceWorker) setContentInjectionMode('serviceworker');`

PR #809, opened 2022-01-29
Default to SW mode when accessing app as PWA

### LINKED ISSUE #808: Make Service Worker the default if the app is being accessed as a PWA
This would be a step towards #196. It would be a very short PR (probably one-liner). It makes no sense that accessing the app as a PWA -- which requires Service Worker by definition -- opens the app in JQuery mode. So, the idea here is that if the app is being accessed from http(s): and Service Worker is available, it should open in SW mode by default (the user can still change back to JQuery).

This would not yet affect our extensions, which would open in JQuery mode by default.

This would also fix #801 implicitly. I'd recommend this being part of release 3.3.

Jaifroid:
Fixes #808 and #801. Basically a one-liner, but formatted as two for legibility 😊.
EDIT: It will be interesting to see what effect this has on the secure browser tests...

### DIFF
--- a/www/js/app.js
+++ b/www/js/app.js
@@ -93,7 +93,9 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
     // A parameter to access the URL of any extension that this app was launched from
     params['referrerExtensionURL'] = settingsStore.getItem('referrerExtensionURL');
     // A parameter to set the content injection mode ('jquery' or 'serviceworker') used by this app
-    params['contentInjectionMode'] = settingsStore.getItem('contentInjectionMode') || 'jquery'; // Defaults to jquery for now
+    params['contentInjectionMode'] = settingsStore.getItem('contentInjectionMode') || 
+        // Defaults to jquery in extensions, and serviceworker if accessing as a PWA
+        (/http/i.test(window.location.protocol) && isServiceWorkerAvailable()) ? 'serviceworker' : 'jquery';
 
     // An object to hold the current search and its state (allows cancellation of search across modules)
     appstate['search'] = {
--- a/www/js/app.js
+++ b/www/js/app.js
@@ -95,7 +95,7 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
     // A parameter to set the content injection mode ('jquery' or 'serviceworker') used by this app
     params['contentInjectionMode'] = settingsStore.getItem('contentInjectionMode') || 
         // Defaults to jquery in extensions, and serviceworker if accessing as a PWA
-        (/http/i.test(window.location.protocol) && isServiceWorkerAvailable()) ? 'serviceworker' : 'jquery';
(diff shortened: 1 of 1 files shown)


Jaifroid:
Interesting that all tests passed. I was expecting the secure browser tests to fail due to differences between jQuery and SW mode, but clearly the differences are immaterial. That's good.

www/js/app.js
```diff
@@ -93,7 +93,9 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
     // A parameter to access the URL of any extension that this app was launched from
     params['referrerExtensionURL'] = settingsStore.getItem('referrerExtensionURL');
     // A parameter to set the content injection mode ('jquery' or 'serviceworker') used by this app
-    params['contentInjectionMode'] = settingsStore.getItem('contentInjectionMode') || 'jquery'; // Defaults to jquery for now
+    params['contentInjectionMode'] = settingsStore.getItem('contentInjectionMode') || 
+        // Defaults to jquery in extensions, and serviceworker if accessing as a PWA
+        (/http/i.test(window.location.protocol) && isServiceWorkerAvailable()) ? 'serviceworker' : 'jquery';
```
mossroy, line 98: I would suggest to use a more specific regexp, like: 
```suggestion
        (/^https?$/i.test(window.location.protocol) && isServiceWorkerAvailable()) ? 'serviceworker' : 'jquery';
```
It should avoid matching a protocol like `this-is-not-http-protocol` (which does not exist, of course, it's an example)
NB : I did not test

mossroy (commented):
Good idea, even if directly accessing the PWA is not the main usage

mossroy:
> Interesting that all tests passed. I was expecting the secure browser tests to fail due to differences between jQuery and SW mode, but clearly the differences are immaterial. That's good.

That's very interesting.
Do we have a way to check which mode has been used by each browser from the UI tests logs?
Based on https://caniuse.com/mdn-api_serviceworker, it should use jQuery mode for Edge 15, and SW mode for all other ones (as IE11 does not run these tests at all)

Jaifroid:
> That's very interesting. Do we have a way to check which mode has been used by each browser from the UI tests logs? Based on https://caniuse.com/mdn-api_serviceworker, it should use jQuery mode for Edge 15, and SW mode for all other ones (as IE11 does not run these tests at all)

I think it might depend on whether we can see the console output in the tests snapshots. If they are running in a secure context and the browser is capable of supporting SW mode, they should launch in SW mode with this PR.

www/js/app.js
Jaifroid: OK, will do.

Jaifroid:
So I guess this is ready to merge. I've tested on Firefox & Edge Chromium from localhost. You have to reset the app to see what it defaults to when there is no `contentinjectionmode` key in the localStorage. In both browsers it defaults to SW mode. In IE11 of course it remains in jQuery mode after a reset.

mossroy (approved):
OK for me

PR #815, opened 2022-02-04
Fix warning on each start when ServiceWorker is not supported

### LINKED ISSUE #814: On some devices that do not support ServiceWorkers, a warning is displayed on every run
At least it's the case on my Firefox OS device.
Each time I start it, it says "The ServiceWorker API is not available on your device. Falling back to jQuery mode". It's true, but I don't see why it's trying to switch to SW mode.

window.location is "app://kiwix.org/www/index.html" on this device, and it's using Local Storage.

mossroy:
Fixes #814


Jaifroid (approved):
Thanks for debugging this! Pesky parentheses....

PR #818, opened 2022-02-05
UT icon and splash updated

### LINKED ISSUE #817: [UT] App icon doesn't follow UT design guidelines
The icon is black on black.
imatge

with no background
imatge

(MR in the way)

cibersheep:
- Added svg icon
- Edited UT/desktop file to use svg and white background on the splash

I think the official colors for the app are black on white, right?

Should fix #817


Jaifroid:
The reason the test has failed is because we need to exempt `kiwix_scalable.svg` from the files that are required by the PWA version, which is currently not relevant to Ubuntu touch. If you add this file to `$ListOfExemptions` in https://github.com/kiwix/kiwix-js/blob/master/scripts/Check-OfflineFilesList.ps1 , then the `test-offline-files` test should succeed.

Jaifroid:
> I think the official colors for the app are black on white, right?

Whatever works best for the platform in this case. Be aware that some users will be using the app in dark mode. If there is a lot of white and a small amount of black, it can be quite harsh or jarring if using an app at night.

cibersheep:
> The reason the test has failed is because we need to exempt `kiwix_scalable.svg` from the files that are required by the PWA version, which is currently not relevant to Ubuntu touch. If you add this file to `$ListOfExemptions` in https://github.com/kiwix/kiwix-js/blob/master/scripts/Check-OfflineFilesList.ps1 , then the `test-offline-files` test should succeed.

Oh, I can do that. Sorry


> Be aware that some users will be using the app in dark mode.

Well spotted. I can do some mockups and you can decide. I was going for «official look» but I also get flashed by white backgrounds :D

cibersheep:
I'll do the changes now, but in the meantime. Something like this?
imatge 
imatge imatge

kelson42:
This is the old logo, the logo has the circle nearer to the feets. See https://commons.wikimedia.org/wiki/File:Kiwix_logo_v3.svg

Jaifroid:
Personally I'd prefer the slightly darker background bottom left because of not getting blinded if opening the app at night (as mentioned above).

Regarding the new logo, is this the right one, @kelson42? - 

https://github.com/kiwix/kiwix-js-windows/blob/master/www/img/icons/favicon-270x270.png

Is there a better source with SVG for example? I think I grabbed that one form the kiwix.org web site.

cibersheep:
> This is the old logo, the logo has the circle nearer to the feets. See https://commons.wikimedia.org/wiki/File:Kiwix_logo_v3.svg

Oops. Sorry, I didn't check. I'll update it


> Personally I'd prefer the slightly darker background bottom left because of not getting blinded if opening the app at night (as mentioned above).

Then I'll go with darker grey. This is something we can change later :)

Jaifroid (approved):

mossroy:
I'll review when @cibersheep will have updated the icon

cibersheep:
> I'll review when @cibersheep will have updated the icon

Ok. I have:
- Updated the icon to the last version :)
- Added a splash image
- Tweaked the desktop file: add the dark grey to the splash, use the splash image and deleted the "translation" for enUS (I guess it was showing "kiwix.desktop" as the name of the app

mossroy (approved):
I could not test, but the icons look good and this PR is harmless

Jaifroid:
OK. @cibersheep are you happy for us to squash/merge now?

cibersheep:
Warp 9!

Issue #812, opened 2022-02-01
Function switchHomeKeyToFocusSearchBar blocks dialogue which gives advice to user on how to run app


Jaifroid:
If a user attempts to run Kiwix JS from the file system in any browser except IE11 and Edge Legacy, then the app should show a dialogue box or info that explains the problem. However, this is blocked since we introduced `switchHomeKeyToFocusSearchBar`, as that function attempts to access the iframe too early in the load process, producing the CORS error and leaving us with an unresponsive app.

This function should only be run once a ZIM has been loaded. Although this is an edge case, the error is currently rendering useless a block of code that is intended to provide good UX.

![image](https://user-images.githubusercontent.com/4304337/151941735-a077f577-1aeb-4911-984f-b2d78d4266ce.png)

PR #822, opened 2022-02-06
Fix race condition while loading decompressors

### LINKED ISSUE #820: Fix loading of decoders for Ubuntu touch
I'm testing on Pixel 3a latest dev:
Tested: kiwix-ubuntu-touch-3.3commit-b143e83.click
- Install click: ok
- API warning: ServiceWorker API available but not registered
- API error: Error loading ZSTD descompressor
- Load a zim file (~140 MB): ok
- Search finds results: ok
- Tapping on a result shows loading icon

log: https://dpaste.com/6HZUX2QED

_Originally posted by @cibersheep in https://github.com/kiwix/kiwix-js/issues/802#issuecomment-1030634882_

Jaifroid:
This may fix #820. It also contains a fix for #812 because that blocks the app completely when loading from file protocol if the iframe is inaccessible, and prevents us accessing the API panel or posting an explanatory message to the user.

I've only tested this in Edge accessing from the file system. This is as close a simulation as I can get to the errors shown in the Ubuntu touch log. The race condition was undoubtedly affecting the loading sequence on Ubuntu touch, because it failed to fall back to ASM only with one of the decompressors (ZSTD), which happened to be the one needed.

Of course something else may be blocking the Ubuntu touch app, but fixing these two issues should allow us to diagnose better, if there are other issues in the way of a solution.

### DIFF
--- a/www/js/app.js
+++ b/www/js/app.js
@@ -504,6 +504,16 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
     //switch on/off the feature to use Home Key to focus search bar
     function switchHomeKeyToFocusSearchBar() {
         var iframeContentWindow = document.getElementById('articleContent').contentWindow;
+        // Test whether iframe is accessible (because if not, we do not want to throw an error at this point, before we can tell the user what is wrong)
+        var isIframeAccessible = true;
+        try {
+            iframeContentWindow.removeEventListener('keydown', focusPrefixOnHomeKey);
+        }
+        catch (err) {
+            console.error('The iframe is probably not accessible', err);
+            isIframeAccessible = false;
+        }
+        if (!isIframeAccessible) return;
         // when the feature is in active state
         if (params.useHomeKeyToFocusSearchBar) {
             //Handle Home key press inside window(outside iframe) to focus #prefix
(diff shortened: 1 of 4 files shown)


mossroy (approved):
This looks good.
I tested on my Firefox OS device and did not see any change.
Of course, I could only test non-regression.

cibersheep:
I cannot find a click file :( sorry, I get quite confused :)

If you could publish a link, I can test it on my device

mossroy:
@cibersheep : a .click file should be generated tonight in https://download.kiwix.org/nightly/2022-02-07/ (if this PR is merged before: it creates nightly builds on master)

Jaifroid:
@cibersheep We need to merge this first to master, and then it would be generated as a nightly for testing. Unless we can force making the click app @mossroy, and do it from a branch?

Note that the release click file for v3.3 doesn't have this (possible) fix in it, so we'd be releasing a v3.3.1 for Ubuntu Touch if the fix works.

Jaifroid:
Tested on a ZSTD ZIM and on an XZDEC ZIM, both seem able to load, no regression on Edge and Firefox.  Also tested in IE11 mode (on Edge). No regression. Home key still focuses search bar if that option is on. I'll merge.

cibersheep:
> @cibersheep : a .click file should be generated tonight in https://download.kiwix.org/nightly/2022-02-07/ (if this PR is merged before: it creates nightly builds on master)

Aah, silly me. I understand. Thanks

cibersheep:
Tested `kiwix-ubuntu-touch-3.3commit-ba994b8`!
Loaded, searched and read. Whooohooo :)

log: https://dpaste.com/6ZDSNW7CA

Jaifroid:
Excellent news, thank you very much! So, once your PR is reviewed and merged, we'll be able to create a 3.3.1 tag, which will generate the new click.

PR #826, opened 2022-02-06
Increase nightwatch timeout

### LINKED ISSUE #737: Issues with Sauce Labs - testing
With almost every commit on almost any PR I make with Kiwix JS, I have to re-run the cross-browser end-to-end tests, because they fail first time. They will usually pass second time, occasionally third time. When I look, it's usually a socket reset that's the problem.

Is this likely to be an issue with Sauce Labs, or an issue with some race condition in Kiwix JS? In the past, with some PRs there has been a race condition that is clearly the fault of new code in Kiwix JS (though it's sometimes taken weeks to find it). However, most of the time the errors occur for no apparent reason.

mossroy:
We can re-run the CI on this branch to see if it might fix #737

### DIFF
--- a/nightwatch.js
+++ b/nightwatch.js
@@ -39,6 +39,11 @@ module.exports = {
       "silent": true,
       "username": "${SAUCE_USERNAME}",
       "access_key": "${SAUCE_ACCESS_KEY}",
+      request_timeout_options: {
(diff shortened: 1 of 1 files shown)


mossroy:
It did not work.

PR #827, opened 2022-02-06
Update manifest for UT

cibersheep:
- Name of the app in title in capital letters
- Updated description (It might have been too short)
- Update name hook (it may break updates if the app was installed before, it needs to be checked)

### DIFF
--- a/ubuntu_touch/manifest.json
+++ b/ubuntu_touch/manifest.json
@@ -1,13 +1,13 @@
 {
   "name": "kiwix",
-  "description": "Kiwix : offline Wikipedia reader",
+  "description": "Kiwix is an offline reader for online content like Wikipedia, Project Gutenberg, or TED Talks.",
   "architecture": "all",
   "maintainer": "Kiwix team <contact+ubuntutouch@kiwix.org>",
   "framework" : "ubuntu-sdk-16.04",
   "version" : "3.3-WIP",
-  "title": "kiwix",
+  "title": "Kiwix",
   "hooks": {
-    "ubports-little-webapp": {
+    "kiwix": {
       "apparmor": "kiwix.apparmor",
       "desktop":  "kiwix.desktop"
     }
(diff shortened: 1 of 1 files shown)


mossroy:
OK for me on these changes, if they are tested on UT

ubuntu_touch/manifest.json
```diff
@@ -1,13 +1,13 @@
 {
   "name": "kiwix",
-  "description": "Kiwix : offline Wikipedia reader",
+  "description": "Kiwix is an offline reader for online content like Wikipedia, Project Gutenberg, or TED Talks.",
```
Jaifroid, line 3: The use of "online" is a bit confusing. How about:

> Kiwix is an offline reader for content from Wikipedia, Project Gutenberg, TED Talks, Wikivoyage, Stackexchange, etc.

Jaifroid (approved):
Looks fine, other than possibly finessing the description if @mossroy agrees.

mossroy (approved):
I agree with the wording remark of @Jaifroid 
Else it's fine for me

ubuntu_touch/manifest.json
cibersheep: True. Thanks.
Changes pushed.

Jaifroid (approved):

Jaifroid:
Before we squash/merge, could I just check what you found out about this potentially breaking updates for those who have the app already? I mean, they will probably have uninstalled it by now because it's broken, but there's a small risk of people being left with an orphaned app they thing will never work again, and they don't try out the new one.

But this is more "your" platform than ours, so I trust your judgement on the type of users it has. I guess they're likely to be independent-minded people who will proactively check if an app they need isn't working....

cibersheep:
With this change it might be needed to delete the cache. Not a major problem, I would say.
I think, UT users are quite used to do that anyway, and yep, look for help :)

Issue #762, opened 2021-09-15
Automate publishing the Edge extension in the Microsoft Store


Jaifroid:
Microsoft is introducing a REST API to automate publishing Edge browser extensions. See https://techcommunity.microsoft.com/t5/articles/introducing-api-to-automate-publishing-and-updating-microsoft/m-p/2654592 . The estimated date for availability is December.

Jaifroid:
Instructions are now available on how to use this API. I believe I could create a GitHub Action that would update the extension. It would need to be triggered by the same code that creates the unsigned extension zip, or chained from it.

PR #837, opened 2022-02-12
Prepare source code for future version 3.4

mossroy:


mossroy:
I initially forgot to update app.js, which properly failed the build, as expected, with a meaningful message: thanks for that @Jaifroid !

Jaifroid (approved):
Looks good! Maybe we should try to automate updating these files somehow!

mossroy:
> Looks good! Maybe we should try to automate updating these files somehow!

Probably.
More generally, I think we should try to release more often, if we can, following the https://en.wikipedia.org/wiki/Release_early,_release_often philosophy.
It would push us to automate everything that can be automated, which would be good if one day we have a critical bug to solve quickly, and also avoids human errors. Publishing on stores might be complicated to automate, but if it's the only manual step that remains, it would already be great.
It would also show users that the project is alive

Jaifroid:
OK, happy with that. I have #762 for the Edge Store. Last time I checked (a week ago or so), the API was still in private testing. I imagine the other stores have API endpoints for upload and release.

PR #838, opened 2022-02-12
Upgrade webp-hero library and reuse the canvas

### LINKED ISSUE #835: Images are not displayed loading ZIMs with WebP format images in icecat 60.7
Issue reported by a user via email. I have been able to reproduce it in icecat running in WSL-Ubuntu (Windows 11). The images are being inserted as data URIs rather than as BLOBs, which indicates that this version is using the polyfill to read WebP.

image

mossroy:
Fixes #835


www/js/lib/uiUtil.js
```diff
@@ -51,15 +51,12 @@ define(rqDef, function(settingsStore) {
...
-                // DEV: WebpMachine.decode() returns a data: URI
-                // We callback before the node is set so that we don't incur slow DOM rewrites before processing more images
-                if (callback) callback(uri);
-                node.setAttribute(nodeAttribute, uri);
-            }).catch(function (err) {
-                console.error('There was an error decoding image in WebpMachine', err);
-                if (callback) callback();
-            });
+            var canvas = document.createElement("canvas");
+            webpMachine.decodeToCanvas(canvas, content);
```
Jaifroid, line 55: If there is a way to detect that the image has been decoded on the canvas, or loaded into it, we could Promisify this ourselves quite easily. I haven't tested anything, so only speculating at this stage.

Jaifroid: @mossroy I just tested refactoring this and treating `webpMachine.decodeToCanvas()` as a Promise. I tested it in IE11 (which does not support WebP), and it is working fine. It does appear to return a Promise, and looking at the source code, it uses an async function (here: https://github.com/chase-moskal/webp-hero/blob/master/source/webp-machine.ts#L43) which would normally be rendered by TypeScript as a Promise when bundled. Obviously the formatting is wonky because we're not removing the image nodes yet, but all the images appear to be there, and no error is thrown with the Promise.
image

Jaifroid: Code I used:
```
var canvas = document.createElement("canvas");
webpMachine.decodeToCanvas(canvas, content).then(function () {
    // TODO copy all the img attributes to the new canvas, and remove the img node afterwards
    node.parentNode.appendChild(canvas);
    // TODO we don't handle images sequentially
    if (callback) callback();
});
```

Jaifroid: Alternative to inserting the canvas in the document is to use canvasElement.toBlob() if it isn't also blacklisted by the anti-fingerprinting in IceCat. See https://caniuse.com/mdn-api_htmlcanvaselement_toblob . It has to have an ms- prefix for IE11 but is otherwise supported.

mossroy: Cool! Thanks @Jaifroid 
I could not test but you're certainly right about the Promise. I got confused by the typescript transpilation to javascript.
Regarding the `toBlob` API, it's worth testing but there are high chances it will be "blocked" too

chase-moskal: @Jaifroid during my recent work on `webp-hero`, i had found that icecat appears to be blocking both `canvas.toDataURL` and also `canvas.toBlob()`, which is why i started inserting the whole canvas elements into the dom.

what i hadn't tried, was [canvas.mozGetAsFile(name, type)](https://developer.mozilla.org/en-US/docs/Web/API/HTMLCanvasElement/mozGetAsFile) – it's a deprecated function, but it might be available in icecat, might be worth looking at.

Jaifroid: @chase-moskal Thank you! We can try that as this is only needed for old Mozilla browsers. I tried `canvas.toBlob()`, but as @mossroy suggested, it is (of course) blocked. In fact, any access to canvas properties appears to be blocked, which is a problem because we can't set the width/height, or at least I've so far failed to do so. I'm getting this kind of result (screenshot below), and if I try to set the width/height either with `canvas.width='12'0'` or` canvas.setAtrribute('width', '120')`, the images get blanked out.

image

Jaifroid: It looks like `canvas.mozGetAsFile` is also locked down 😞. The resulting File objects can be converted to blobs, but they are empty. I guess this isn't surprising.

image

www/js/lib/uiUtil.js
```diff
@@ -51,15 +51,12 @@ define(rqDef, function(settingsStore) {
...
+            // TODO copy all the img attributes to the new canvas, and remove the img node afterwards
+            node.parentNode.appendChild(canvas);
+            // TODO we don't handle images sequentially
+            if (callback) callback();
```
Jaifroid, line 59: @mossroy  In case it's useful, here's my code so far as a suggestion. Alternatives I've tried that I think don't work are commented out. This works in IceCat, but I've failed to be able to set the canvas size correctly (see commented out code for what I tried).
```suggestion
            var canvas = document.createElement("canvas");
            webpMachine.decodeToCanvas(canvas, content).then(function () {
                if (canvas.msToBlob) {
                    insertBlob(canvas.msToBlob());
                } else if (canvas.mozGetAsFile) {
                    
                    // ** DEV: This is mozGetAsFile method - but results in blank images **
                    // var f = canvas.mozGetAsFile('f.png');
                    // insertBlob(f);

                    // ** DEV: This is Insert canvas method - but if we set the height or width, the imgaes get blanked **
                    if (callback) callback(); // Calling back as soon as possible speeds up extraction
                    // canvas.setAttribute('width', node.getAttribute('width'));
                    // canvas.setAttribute('height', node.getAttribute('height'));
                    node.parentNode.replaceChild(canvas, node);

                } else {
                    canvas.toBlob(function (imgBlob) {
                        insertBlob(imgBlob);
                    });
                }
            }).catch(function (err) {
                console.error('Error converting WebP image!', err);
            });
```

chase-moskal: @Jaifroid i also had initial troubles setting the canvas dimensions – but i found success with the following technique https://github.com/chase-moskal/webp-hero/blob/f75a7593a1dd1271ef9b075d9c1611bc9c2466ea/source/webp-machine.ts#L153

you can see this in action in the live webp-hero demo: https://webp-hero.chasemoskal.com/?force?useCanvasElements

a friend of mine pointed out that canvas elements stop scrolling from working while the user hovers over a canvas -- i'm intending to add the css rule `pointer-events: none;` to mitigate this.

i can expose the webp-hero helper functions as statics on WebpMachine in the next version of webp-hero, i should be able to get to that in a few hours :+1:

Jaifroid: That's incredibly helpful @chase-moskal . I was actually thinking to check your source code as you obviously need to set dimensions too when polyfilling the whole document. We'll try the style trick!

Jaifroid: @chase-moskal That works! Thank you! Note that I had to add the 'px' explicitly or else the images were still the wrong size, as I think a simple number (even as a string) is not accepted by the `style.width/height` attributes. I'm also using the `node.parentNode.replaceChild(newNode, oldNode)` method to replace the old node with the canvas, because of the need to support IE11 in our code still.
```
canvas.style.width = node.getAttribute('width') + 'px';
canvas.style.height = node.getAttribute('height') + 'px';
node.parentNode.replaceChild(canvas, node);
```

image

Jaifroid: @mossroy I updated the suggestion above with latest working code.

chase-moskal: @Jaifroid @mossroy 

:tada: please see [this comment](https://github.com/kiwix/kiwix-js/issues/835#issuecomment-1038000222) in the issue

mossroy: Many thanks @chase-moskal 
@Jaifroid : feel free to commit your work-in-progress on this branch/PR and continue it (if you are willing to).

Jaifroid: OK, I've pushed my changes. I haven't yet switched to latest Web-Hero to use the new static function. If you want to have a go at that, please do, as I won't have much more time today.

www/js/lib/uiUtil.js
```diff
@@ -46,30 +46,48 @@ define(rqDef, function(settingsStore) {
...
+                    node.parentNode.replaceChild(canvas, node);
+
+                } else {
+                    canvas.toBlob(function (imgBlob) {
+                        insertBlob(imgBlob);
+                    });
+                }
             }).catch(function (err) {
-                console.error('There was an error decoding image in WebpMachine', err);
-                if (callback) callback();
```
Jaifroid, line 61: Should revert this to previous catch code, it was more meaningful, plus callback is missing.

www/js/lib/uiUtil.js
```diff
@@ -46,30 +46,49 @@ define(rqDef, function(settingsStore) {
...
             // However, the loadImagesJQuery() function in app.js is sequential (it waits for a callback
             // before processing another image) so we do not need to queue WebP images here
-            webpMachine.decode(content).then(function (uri) {
-                // DEV: WebpMachine.decode() returns a data: URI
-                // We callback before the node is set so that we don't incur slow DOM rewrites before processing more images
-                if (callback) callback(uri);
-                node.setAttribute(nodeAttribute, uri);
+            var canvas = document.createElement('canvas');
+            webpMachine.decodeToCanvas(canvas, content).then(function () {
+                if (canvas.msToBlob) {
```
mossroy, line 64: Is it necessary to distinguish these 3 cases?
We might simply do `webpMachine.replaceImageWithCanvas(node, canvas);` in all cases? (if it's working well), and call the callback (if set) of course.
Do you see a side effect of replacing images by canvas on all browsers that do not support webp?
It would simplify the code, and might be a bit faster (from what I understood, it skips the conversion to/from base64)

Jaifroid: I kept it in there because it was a working method for IE11 and I was hoping to fine an analogous Firefox-specific one for IceCat.

The only reason to keep the IE11 method, and possibly the generic one, is that the one targeted at IceCat is very much a workaround: we have to guess at which attributes to copy from the `` tag, and it's not very generic. For example, do we/can we copy alt text? What about a ZIM in the future that has image dimensions in `em` and not in points, etc.?

Of course it's only a tiny number of browsers that don't support WebP images, so maybe we're not sacrificing much if we use a single function. Having said that, the IE11 code is only a couple of lines. It would be worth seeing if there is an appreciable difference in IE11. Inserting a node into the DOM and creating a blob probably take the same amount of time. I also need to test on Edge Classic / UWP, as this impacts on the UWP app (on Windows Mobile, WebP isn't supported either). I suspect the `msToBlob` method works on Edge Classic.

I don't feel strongly about this, unless there is a significant performance loss.

Jaifroid: Plus there is the scrolling issue with inserted canvases: it certainly affects IceCat. I haven't tested on IE11. If the mouse is hovering over an image, it's not possible to scroll with mouse wheel or trackpad. It makes me think we should only use this method for cases that absolutely need it.

mossroy: OK, that makes sense.
The scrolling issue, and the risk to badly copy image attributes are enough to keep the original method for browsers that properly support it.

Problem is how to detect that.
On IceCat 60.7, there is no error. The anti-fingerprinting feature simply alters the canvas reading: that's hard to detect.
Maybe we should simply add a setting to use canvas or not (disabled by default).
This is probably a very rare case: we're talking about old browsers (that do not support webp) with a bad anti-fingerprinting feature. I say "bad" because the problem does not occur on latest Firefox with [CanvasBlocker](https://addons.mozilla.org/en-US/firefox/addon/canvasblocker/) extension (and with webp manually disabled)

Jaifroid: The nearest I could get to targeting it at IceCat was to use `if (canvas.mozGetAsFile)`. This is a deprecated function and any modern Firefox won't even go near this code because it supports WebP natively.

So this is likely to trap: only a browser that does not support WebP and is Mozilla-based (realistically the only browser type anyone would try to mod for privacy purposes, I think). It's not perfect, but it does apply to IceCat.

It might be worth checking if this accidentally traps Firefox OS, as presumably that's not so desirable. I can't remember if FFOS supports WebP natively. Maybe there's a way to get FFOS to run the native `canvas.toBlob()`, but I realize we're getting quite close to browser sniffing here, which we don't want to do.

mossroy: I'll test on Firefox OS when we'll have something working.
It's based on Gecko 44, so it does not support webp, see https://caniuse.com/webp
According to https://caniuse.com/mdn-api_htmlcanvaselement_mozgetasfile, it supports `mozGetAsFile`, so would probably be trapped by your test...
This test would trap any browser based on Gecko >=4 and <=65 (or <=73 and with webp manually disabled).
I think it's unfortunately too broad.

FFOS should support `canvas.toBlob()`, according to https://caniuse.com/mdn-api_htmlcanvaselement_toblob , and webp-hero already works well on it, with the current kiwix-js.

mossroy: There would probably be a way to properly detect that: take a small known webp image binary, make webp-hero convert it, and compare the result with the expected one.
If it's different, it means we have to fallback to keep the canvas.
This could be tested once, and displayed in the API settings

Jaifroid: Interesting, I like the idea. We currently do something similar already with WebHero to test whether the browser understands WebP (so we can load the polyfill conditionally, as it is quite heavy in terms of MB). Doing so in the same place would be easy.

Jaifroid:
@mossroy I checked the possibility of using the string comparison technique by using this simple statement to compare the strings produced:

`var canvasData = document.createElement('canvas').toDataURL('image/png');`

This actually produces a very long dataURI. Unfortunately, the dataURI produced by Edge Chromium and that produced by IE11 are different, and neither have any antifingerprtinting software installed. Then I did a bit of reading and found that these differences are in fact the basis of browser fingerprinting. Each browser and graphics rendering engine will produce subtly different data when reading/converting the pixels from a canvas. This is how tracking sites try to identify you uniquely. So I don't think simple string comparison will work. Maybe your idea of an option in the UI to turn this on is best after all? IceCat users in any case have to whitelist Kiwix JS so that LibreJS doesn't block the app.

mossroy:
It seems to me that LibreJS does not have to be disabled when running kiwix-js as an extension.

You're certainly right about canvas and anti-fingerprinting.
On the other hand, what IceCat 60.7 produces is a base64 string that results in a corrupted (or white) image, while the other browers produce a base64 string that can be turned into a true image. There might be a way to detect that?

mossroy:
For the record, I tested the `decodeToCanvas` way on my Firefox OS device, and it works fine too on it.

chase-moskal:
i think it would be a good idea to write a function like `detectCanvasReadingSupport()`, which would return true if the browser is able to read data from a canvas, thus informing us whether or not we need to use either `decodeToCanvas` or `decode`.

if we write that function, i would like to use it as the WebpMachine default value for `useCanvasElements`, which would then allow webp-hero to work out-of-the-box on icecat, without needing a custom option that could have drawbacks in other browsers.

i'll be too busy for the next day or two, but afterwards, if i find you folks haven't written such a function, i probably will write one, because i want the improved default behavior for webp-hero :)

:beers:

Jaifroid:
Thank you @chase-moskal! We're trying to puzzle out how this can be detected, because browser anti-fingerprinting techniques are quite diverse, and there's a bit of an arms race between those trying to fingerprint devices and those that are trying to hide the anti-fingerprinting (because anti-fingerprinting can itself be part of a unique fingerprint for a browser!). I tried extracting a dataURI from a canvas, and both Chromium and IE11 give different results just for a blank canvas. @mossroy has suggested it might be possible to test if the image is corrupted with invalid data (which seems to blank out images in IceCat). I haven't yet found anyone offering a reliable method. There are some web sites that will test this for you, but the ones I found don't give their source.

One typical one is: https://webbrowsertools.com/canvas-fingerprint/ (WARNING: it will calculate and display a fingerprint for you!)

Another, less informative, is  https://browserleaks.com/canvas.

For the top one, IceCat shows the following (I tried looking at the source for the page, but the scripts are both obfuscated and very long):

image

Jaifroid:
Quote from that page. Maybe we can use one of these techniques:

> There are many methods to identify a browser. The canvas fingerprinting is a popular and simple one which is used frequently. Note that even if you have the fingerprint protection enabled, still, most likely, it is possible to detect that there is a protection method active. In this page, there are two methods used for protection detection.
> 1. Detect fixed canvas manipulation: in this technique, a fixed image is printed on the canvas area and then the hash code is generated. If there is no protection in place, a determined value for the hash code is expected. If not, then the browser uses the first protection method.
> 2. In this method, the hash code is generated twice once before the DOM is ready and once after it is fully loaded on the exact same image. If hash codes differ, then the fingerprinting is protected. Note that it is possible to identify protection and use an alternative method if the ID that is generated by this method is unreliable.

Jaifroid:
Just to add that I only tested a blank canvas, not an actual known image (method 1 above). It strikes me that a blank canvas element may well have different dimensions in different browsers, and that may be why I got two different results from IE11 and Chromium.

But I'm getting the feeling we risk complicating our code quite a lot for an edge case in JQuery mode. Basic support should be sufficient: just use the canvas-image substitution by default for any browser that does not support WebP and does not have `msToBlob()` method, and give an option to turn that off if it causes difficulties. This would very easy to code (current code that we have, with a test for the UI option).

mossroy:
I'll be AFK on the 3 following days on my side.
I tend to think like @Jaifroid that a setting might be enough for such a rare edge case, if it's too complicated to test in a reliable way.

mossroy:
@Jaifroid feel free to commit the changes you are suggesting on this branch if you have more time than me. I'm OK with them.

Jaifroid:
I've worked on an implementation of what we agreed. The UI looks like screenshot below. It's a bit wordy. Suggestions for simplifying would be welcome.

Some notes on the (still draft) implementation:

* It doesn't really do anything clever to find out if the workaround is needed. It just does two basic tests, and if one of these passes, the implementation is not used and hidden. The vast majority of browsers won't use the implementation, but IceCat and (likely) FFOS will by default. Tests can be seen [here](https://github.com/kiwix/kiwix-js/blob/d20ff1b9118d25cb0fa3f357ee1aae761508816b/www/js/lib/uiUtil.js#L103).
* Since classic WebP machine produces data URIs, and FFOS, IE11 and Windows Mobile webview are all happy with dataURIs, it struck me that we should probably standardize on dataURIs for images for all browsers. It has these advantages:
  + We avoid an extra step in creating a BLOB in memory and assigning it a URI;
  + We do not have to have a callback to destroy the blob after the image loads;
  + Incidentally this greatly improves the user's experience of right-clicking and saving images #694 (NB only images are serialized, not videos);
  + There is an extremely fast, native, cross-browser way of doing this (see links provided in code comments);
  + Memory is not in practice an issue: old low-memory devices like FFOS and Windows Mobile will use Data URIs anyway because of the WebP polyfill, and if they can cope, modern browsers are more than capable of coping. There isn't much in it between creating a BLOB, keeping it in memory, attaching an event listener, and simply serializing the data to a dataURI;
  + However, I've kept the old implementation commented out, should we wish to revert.
* The UI is entirely hidden from the user unless the browser determines it **may** be needed. Additionally, if user has selected Service Worker mode, the UI is hidden (it's irrelevant in that mode).
* The workaround is on by default if it may be needed, but user can turn it off.

image

www/js/lib/uiUtil.js
```diff
@@ -51,26 +51,74 @@ define(rqDef, function(settingsStore) {
...
+     * extract from a canvas element (as an anti-fingerprinting strategy), but we know that in certain situations it is not needed.
+     * It will never be used in Service Worker mode, but we still need to determine it in case the user switches to jQuery mode.
+     * @returns {Boolean} A deterministic value to indicate whether the workaround may needed 
+     */
+    function determineCanvasElementsWorkaround() {
+        // Determines setting automatically
+        var useCanvasElements = ! webpMachine ? false : // 1. Feature should always be off if browser supports WebP natively
+            HTMLCanvasElement.prototype.msToBlob ? false : // 2. IE11 and UWP webview (Edge Legacy) should never use this feature
+            true; // 3. Otherwise, default to using feature
+        // Show or hide the imageHanldingModeDiv accordingly
```
Jaifroid, line 106: Typo Hanlding -> Handling

www/index.html
```diff
@@ -551,6 +551,23 @@ <h3>Compatibility settings</h3>
...
+                            <div class="card card-warning">
+                                <div class="card-header">Image handling</div>
+                                <div class="card-body">
+                                    <p>Work around browser's anti-fingerprinting measures by substituting WebP images with canvas elements
+                                        (setting is determined semi-automatically, but may be disabled if not needed):
+                                    </p>
+                                    <div class="checkbox">
+                                        <label title="For browsers that do not have native support for WebP and that implement anti-fingerprinting measures, we cannot extract the image data from the canvas elements produced by the polyfill, so we use the canvas as substitute for the image. If not needed, disabling this may improve performance." >
+                                            <input type="checkbox" name="useCanvasElements" id="useCanvasElementsCheck">
+                                            <strong>App determines if WebP workaround is needed</strong> (we recommend disabling this if not needed)
```
Jaifroid, line 565: Alternative checkbox text:

> Use WebP workaround (we recommend disabling if you find it is not needed)

Jaifroid:
That's great! I'll simplify this PR to make use of the new options in the polyfill.

www/js/lib/uiUtil.js
```diff
@@ -43,34 +43,87 @@ define(rqDef, function(settingsStore) {
...
      */
     function feedNodeWithBlob(node, nodeAttribute, content, mimeType, callback) {
         // Decode WebP data if the browser does not support WebP and the mimeType is webp
         if (webpMachine && /image\/webp/i.test(mimeType)) {
+            // If we're dealing with a dataURI, first convert to Uint8Array
+            if (/^data:/i.test(content)) {
+                // Below uses Mozilla-recommended alternative to atob() : see https://developer.mozilla.org/en-US/docs/Glossary/Base64
+                content = util.dataURItoUint8Array(content);
+                // Or we could use webpHero's utility: it uses deprecated atob() but probably OK for image data
+                // content = webpHero.convertDataURIToBinary(content);
```
Jaifroid, line 56: There is a choice here between using the Mozilla-recommended alternative to deprecated atob() (which I've included in util.js, in place of a deprecated function that was there), or using the wepHero conversion utility. The webpHero version does use atob(), but as it will never be dealing with UTF8 data (it only deals with WebP images), this is probably OK.

Jaifroid:
Tested on IceCat 67.0 on WSL Ubuntu, on IE11 (IE mode), and on Edge Chromium.

In IceCat, the workaround is exposed in the UI, and user can disable it. On re-enabling it, the user is prompted by IceCat as to whether they wish to allow extraction from the canvas. If user says yes, then our workaround is no longer needed (after reloading the page with Ctrl-R, or after attempting to switch content injection mode, since detection is re-run when user attempts to switch).

On IE11, webpHero correctly detects that it can extract data from the canvas, and so the UI is not exposed and the workaround is not used.

In Edge Chromium, WebP is natively supported, and so the UI is not exposed.

www/js/app.js
```diff
@@ -416,6 +419,16 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
         // Do the necessary to enable or disable the Service Worker
         setContentInjectionMode(this.value);
     });
+    document.getElementById('useCanvasElementsCheck').addEventListener('change', function () {
+        if (this.checked) {
+            settingsStore.removeItem('useCanvasElements');
```
mossroy, line 424: I don't understand why you remove this setting here, instead of setting it to true.

www/js/app.js
```diff
@@ -416,6 +419,16 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
...
+            params.useCanvasElements = uiUtil.determineCanvasElementsWorkaround();
```
mossroy, line 425: The responsibilities of this function `uiUtil.determineCanvasElementsWorkaround` is not clear to me.
It seems to me that the function sets `params.useCanvasElements`, but also returns its value, with is set again here.

www/js/app.js
```diff
@@ -416,6 +419,16 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
...
+            this.checked = params.useCanvasElements;
```
mossroy, line 426: So it means that there are cases where we would refuse that the user checks this? If so, we should tell the user why

www/js/lib/uiUtil.js
```diff
@@ -43,34 +43,87 @@ define(rqDef, function(settingsStore) {
...
+
+    /**
+     * Determines whether the Canvas Elements Workaround for decoding WebP images is needed, and sets UI accordingly
+     * Note that the workaround will never be used in Service Worker mode, but we still need to determine it in case the user switches to jQuery mode
+     * @returns {Boolean} A value to indicate whether the workaround may needed 
+     */
+    function determineCanvasElementsWorkaround() {
+        var userPreference = settingsStore.getItem('useCanvasElements') !== 'false';
+        // Determines setting automatically
+        var useCanvasElements = webpMachine && webpHero && !webpHero.detectCanvasReadingSupport();
```
mossroy, line 109: This variable name looks misleading to me: it stores the capability of the browser

www/js/lib/uiUtil.js
```diff
@@ -43,34 +43,87 @@ define(rqDef, function(settingsStore) {
...
+        var useCanvasElementsCheck = document.getElementById('useCanvasElementsCheck');
+        if (! useCanvasElements) {
+            imgHandlingModeDiv.style.display = 'none';
+            useCanvasElementsCheck.checked = false;
+        } else if (params.contentInjectionMode === 'jquery') {
+            imgHandlingModeDiv.style.display = 'block';
+            useCanvasElementsCheck.checked = userPreference;
+        }
+        // The setting has no meaning in SW mode, so we hide the UI for it
+        if (params.contentInjectionMode === 'serviceworker') {
```
mossroy, line 121: It might be better to test on  `!== 'jquery'`, in case of we add more modes in the future.
And maybe these tests could refactored to be more clear:
```
if !== 'jquery' {
  hide the UI
} else {
  if browser capable of canvas reading {
    hide the UI
  } else {
    show the UI
  }
}
```

www/js/lib/uiUtil.js
```diff
@@ -43,34 +43,87 @@ define(rqDef, function(settingsStore) {
...
+            // node.setAttribute(nodeAttribute, url);
+        }
+    }
+
+    /**
+     * Determines whether the Canvas Elements Workaround for decoding WebP images is needed, and sets UI accordingly
+     * Note that the workaround will never be used in Service Worker mode, but we still need to determine it in case the user switches to jQuery mode
+     * @returns {Boolean} A value to indicate whether the workaround may needed 
+     */
+    function determineCanvasElementsWorkaround() {
```
mossroy, line 106: I find it difficult to understand this function, because it handles many things at the same time.
Maybe it could be split between handling the UI (only showing/hiding the div) and setting the param/settingsStore

www/js/lib/util.js
```diff
@@ -249,37 +249,63 @@ define([], function() {
...
+    }
 
-        var blob = new Blob(byteArrays, {type: contentType});
-        return blob;
+    /**
+     Converts a dataURI to Uint8Array
+     * @param {String} dataURI The data URI to convert
+     * @returns {Uint8Array} A Uint8Array with the converted buffer
+     */
+     function dataURItoUint8Array(dataURI) {
```
mossroy, line 297: If unused, this function could be removed (and the 2 other ones above, too)

mossroy (commented):
Congrats for finding out the anti-fingerprinting setting of IceCat 60.7.
None of us had seen it so far, so I don't expect regular users to see it can be configured.
I agree with you that webpHero should be upgraded anyway.
I'm still in favor of finishing and merging this PR, as it should allow to workaround anti-fingerprinting features on other browsers, and make kiwix-js work out-of-the-box on them.

I tested on Firefox OS: it properly detects its capabilities, converts webp images to base64-encoded png, and does not show the setting (as expected).
I also tested on Firefox 97 (including after manually disabling webp support on it) and on IceCat 67 (in a Windows 10 VM): it works well.

www/js/lib/uiUtil.js
```diff
@@ -43,34 +43,87 @@ define(rqDef, function(settingsStore) {
...
-            if (callback) callback(url);
-            node.addEventListener('load', function () {
-                URL.revokeObjectURL(url);
-            });
-            node.setAttribute(nodeAttribute, url);
+            if (callback) callback(); // Calling back as soon as possible speeds up extraction
+            // In browsers that support WebP natively, or for non-WebP images, we can simply convert the Uint8Array to a data URI
+            // DEV: we use FileReader method because btoa fails on utf8 strings (in SVGs, for example)
+            // See https://developer.mozilla.org/en-US/docs/Web/API/WindowBase64/Base64_encoding_and_decoding#The_Unicode_Problem
+            // This native browser method is very fast: see https://stackoverflow.com/a/66046176/9727685
```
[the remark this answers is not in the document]
Jaifroid: Sorry I should have been clearer. The revoking function actually concerns a Blob URL: [URL.revokeObjectURL()](https://developer.mozilla.org/en-US/docs/Web/API/URL/revokeObjectURL). Because we don't create a URL, there is nothing to revoke. A Blob itself is just a version of File and works like any other Object. There is no method for revoking an Object because it will be destroyed when the closure it was created inside ends.

Regarding the use of the FileReader method, I switched to using this in Kiwix JS Windows after I was getting errors with SVGs that could not be 64-bit encoded with `btoa()` (because SVGs can contain UTF8-encoded strings). Since we need to decode SVGs (the function here will run on any type of image), we either use this native browser method, or else we use the [Mozilla rewrite of `btoa()`](https://developer.mozilla.org/en-US/docs/Web/API/WindowBase64/Base64_encoding_and_decoding#The_Unicode_Problem).

Would you like me to add a timer and test both methods? The Mozilla method is a lot of code, but it would go in `util.js`. I haven't had any problems using the FileReader method in KJSW (it is an option currently that I've had maybe for 2 years or so).

www/js/app.js
Jaifroid: Because we are not allowing user to force it on. The app (or rather, WebpHero) will decide if the option is needed. The user can only disable its use (in which case we store `false` as the parameter value), or else return the setting to auto. I could make this clearer in the UI. In fact, the user will never have a chance to force this setting on (even if permitted) because the app will not even show this UI if it determines that the workaround is not needed).

www/js/app.js
Jaifroid: Yes you are right. I changed it recently so that the function sets params.useCanvasElements so that we didn't end up running it twice on every app load. And I forgot to remove the assignment here. This should just run the function, not set it to the result.

www/js/app.js
Jaifroid: Yes, user can only turn this off if the app turned it on. Not vice versa.

www/js/lib/uiUtil.js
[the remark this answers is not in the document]
mossroy: Thanks for the explanation.
If you already tested it on KJSW, it's fine: we can keep it

www/js/lib/util.js
Jaifroid: OK. I can remove these.

www/js/lib/uiUtil.js
Jaifroid: I simplified, but in a slightly different way.

www/js/lib/uiUtil.js
Jaifroid: I've greatly simplified this function now and added clear comments for the things it does. The structure was confused before. Let me know if you think it still needs splitting in two.

www/js/app.js
```diff
@@ -416,6 +419,17 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
...
     });
+    document.getElementById('useCanvasElementsCheck').addEventListener('change', function () {
+        if (this.checked) {
+            // User can only *disable* this auto-determined setting, not force it on, so we do not store a value of true
+            settingsStore.removeItem('useCanvasElements');
+            uiUtil.determineCanvasElementsWorkaround();
+            this.checked = params.useCanvasElementsForWepTranscoding;
+        } else {
+            params.useCanvasElementsForWepTranscoding = false;
+            settingsStore.setItem('useCanvasElements', false, Infinity);
```
Jaifroid, line 430: Use same name in settingsStore

www/js/lib/util.js
```diff
@@ -318,14 +241,11 @@ define([], function() {
...
         endsWith: endsWith,
         readIntegerFrom4Bytes: readIntegerFrom4Bytes,
-        readIntegerFrom2Bytes : readIntegerFrom2Bytes,
-        readFloatFrom4Bytes : readFloatFrom4Bytes,
-        uint8ArrayToHex : uint8ArrayToHex,
-        uint8ArrayToBase64 : uint8ArrayToBase64,
-        readFileSlice : readFileSlice,
+        readIntegerFrom2Bytes: readIntegerFrom2Bytes,
+        readFloatFrom4Bytes: readFloatFrom4Bytes,
+        uint8ArrayToHex: uint8ArrayToHex,
```
Jaifroid, line 246: A few of above functions are not used in the app any more. It's probably best to determine what to remove in a separate PR. I'll create an issue for it.

www/js/lib/util.js
```diff
@@ -318,14 +241,11 @@ define([], function() {
...
+        readFileSlice: readFileSlice,
         binarySearch: binarySearch,
-        b64toBlob: b64toBlob,
-        uintToString: uintToString,
```
Jaifroid, line 328: I removed those unused functions that have to do with converting between uint(8)Arrays and strings, as they are obsolete.

www/js/lib/util.js
Jaifroid: See #843.

www/js/lib/uiUtil.js
mossroy: It's more clear, but the header should mention it also sets the `params`.

www/js/app.js
```diff
@@ -96,6 +96,9 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
     params['contentInjectionMode'] = settingsStore.getItem('contentInjectionMode') || 
         // Defaults to jquery in extensions, and serviceworker if accessing as a PWA
         ((/^https?:$/i.test(window.location.protocol) && isServiceWorkerAvailable()) ? 'serviceworker' : 'jquery');
+    // A parameter to circumvent anti-fingerprinting technology in browsers that do not support WebP natively by substituting images
+    // directly with the canvas elements produced by the WebP polyfill [kiwix-js #835]. NB This is only currently used in jQuery mode.
+    params['useCanvasElementsForWepTranscoding']; // Value is determined in uiUtil.determineCanvasElementsWorkaround(), called when setting the content injection mode
```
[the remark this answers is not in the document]
Jaifroid: I must be going blind! Thanks!

www/js/lib/uiUtil.js
Jaifroid: Note added to header.

Jaifroid:
Yes, please go ahead. It's best to test thoroughly on the latest codebase. No problem about my resetting my local copy of the PR.

mossroy:
I've rebased on master.
Tested on Firefox OS: no regression
Tested on Chromium 98 and Firefox 97 (with and without webp disabled), both in jQuery and SW modes: no regression

mossroy:
Tested on IceCat 60.7 (on Windows 10) as a browser extension: it works out-of-the-box, with canvas for images.
If I uncheck the workaround setting, images are missing again. But they appear again if I ask IceCat to allow to "extract canvas data"
So, to me, it's ready to be merged

Jaifroid:
OK, I'll do some tests on IE11 and Edge Legacy first.

Jaifroid:
I set a branch to this PR and then set the gh-pages implementation to the branch so that I could test on Edge Legacy on Windows Mobile. Tests on Edge Legacy and on IE11 are fine, I don't see any regression.

On IceCat 67.0 on WSL-Ubuntu (Windows 11), I checked that videos are unaffected. I could load a Khan Academy, select a lesson, and play the video. I checked that the video is still inserted as a BLOB URL (which is what we want -- in any case it never used `feedNodeWithBlob()`, we insert videos independently as BLOB URLs).

I found a minor issue on IceCat 67.0. When viewing images inserted via the canvas while dark mode is on, the images fail to re-invert. See screenshot. I think this is a simple edit to the stylesheet for dark mode to include the canvas element as a source of images that should be re-inverted. Shall I make this change, or would you rather it were a different PR?

I have one more minor suggestion I'll put in the code in a moment.

image

www/js/lib/uiUtil.js
```diff
@@ -86,34 +86,73 @@ define(rqDef, function(settingsStore) {
      * @param {String} nodeAttribute The attribute to set to the URI
      * @param {Uint8Array} content The binary content to convert to a URI
      * @param {String} mimeType The MIME type of the content
-     * @param {Function} callback An optional function to call with the URI
+     * @param {Function} callback An optional function to call to start processing the next item
      */
     function feedNodeWithBlob(node, nodeAttribute, content, mimeType, callback) {
```
Jaifroid, line 91: The name of this function is now inaccurate, as it no longer inserts BLOBs. Should it be changed to `feedNodeWithDataURI`?

Jaifroid (approved):
All good for me and testing finished with just those two suggestions:

1. Add or amend stylesheet rule to invert canvas images in dark mode
2. Change name of feedNodeWithBlob if we will no longer use it for BLOBs.

www/js/lib/uiUtil.js
mossroy: You're right. I'm doing it

mossroy:
I agree with your suggestions.
I'm doing the renaming.
Could you do the stylesheet rule?

Jaifroid:
> I agree with your suggestions. I'm doing the renaming. Could you do the stylesheet rule?

Yes, will do.

mossroy:
Renaming done

Jaifroid:
I added canvas elements, but need to test.

Jaifroid:
Seems to work (screenshot). Note there are some limitations: because we don't copy all the attributes of `` over to `<canvas>`, the style that implements wikimedia workarounds won't be any different from the generic inversion style. Personally I think we've done enough to support IceCat, and don't think we should add any more code to support such edge cases.

image

mossroy:
I agree.
Can I merge?

Jaifroid:
Fine by me!

PR #842, opened 2022-02-19
Fix 2 codefactor warnings and remove whitespaces at end of lines of app.js

mossroy:
Removed unused variable messageChannel, and a weird character in a comment

### DIFF
--- a/www/js/app.js
+++ b/www/js/app.js
@@ -1,22 +1,22 @@
 /**
  * app.js : User Interface implementation
  * This file handles the interaction between the application and the user
- * 
+ *
  * Copyright 2013-2014 Mossroy and contributors
  * License GPL v3:
- * 
+ *
  * This file is part of Kiwix.
- * 
+ *
  * Kiwix is free software: you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
  * the Free Software Foundation, either version 3 of the License, or
  * (at your option) any later version.
- * 
+ *
  * Kiwix is distributed in the hope that it will be useful,
  * but WITHOUT ANY WARRANTY; without even the implied warranty of
  * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
  * GNU General Public License for more details.
- * 
+ *
  * You should have received a copy of the GNU General Public License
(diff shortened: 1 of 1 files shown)


www/js/app.js
```diff
@@ -677,7 +677,6 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
                 // If this is the first time we are initiating the SW, allow Promises to complete by delaying potential reload till next tick
                 delay = 0;
             }
-            messageChannel = tmpMessageChannel;
```
Jaifroid, line 680: I'm not sure I understand why this has been removed. Is it because we don't actually need to store this message channel?

Jaifroid (approved):
Thanks for this cleanup! All looks good to me. I didn't test the effect of removing the messageChannel / tmpMessageChannel, but I had noticed previously that messageChannel was marked by my editor as being unused, so it's good to remove that anomaly.

www/js/app.js
mossroy: In fact, the global variable `messageChannel` was defined, assigned here, but never used.
I suppose we were keeping a single messageChannel instance at some point in the past, and afterwards we changed our mind.
So I removed the variable declaration, and assignment here.

Jaifroid: OK. At some point (not now) it might be good to check whether the implementation still needs the Service Worker to be kept alive. The Service Worker spec is specifically designed to put the SW to sleep when not being used, and it should fire up again immediately it is invoked again. It's possible, in fact, that an older implementation tried to keep the messageChannel alive (hence we have a global messageChannel assignment here), which should not be necessary. Each use of the SW can define a new messageChannel. I'm only speculating here, nothing tested!

mossroy: I doubt we could simply remove this keepalive implementation.
The SW can be removed and re-created by the browser, by design from the beginning.
The problem is that, when it's re-created, it does not have the messageChannel variable to ask app.js for the content. We send this messageChannel through the "init" message.

When I faced this issue in https://github.com/kiwix/kiwix-js/issues/145, I did not find a better solution.
But I'd be very glad to have a better one.

Jaifroid: OK, we can investigate another time. Fine to merge this if you've done some testing on SW mode.

Jaifroid:
Sorry, commented too late!

PR #848, opened 2022-03-24
Trigger webapp restart

rgaudin:
**Note**: This app's hosting has **already** been migrated to our new (k8s-based) hosting. We can discuss different ways to offer this feature but as of now, your app is not reloading automatically on image pushes.

Manually trigger a *restart* on Kiwix k8s after a successful image push.
Previous behavior was to let Docker Hub trigger a restart on (former) hosting via a webhook:
- we don't want to rely on Docker Hub anymore
- hosting platform changed

### DIFF
--- a/.github/workflows/publish-extension.yaml
+++ b/.github/workflows/publish-extension.yaml
@@ -49,17 +49,17 @@ jobs:
       # Publish to docker only if explicitly requested or we are releasing
       - name: Build and push to docker
         if: github.event.inputs.target == 'docker' || github.event_name == 'release'
-        uses: openzim/docker-publish-action@v6
+        uses: openzim/docker-publish-action@v8
         with:
           image-name: kiwix/kiwix-moz-extension
           credentials: |
-            DOCKERIO_USERNAME=${{ secrets.DOCKERHUB_USERNAME }}
-            DOCKERIO_TOKEN=${{ secrets.DOCKERHUB_PASSWORD }}
+            GHCRIO_USERNAME=${{ secrets.GHCR_USERNAME }}
+            GHCRIO_TOKEN=${{ secrets.GHCR_TOKEN }}
           tag-pattern: /^v?([0-9.]+)$/
           latest-on-tag: true
           dockerfile: docker/dockerfile-moz-extension.pwa
           restrict-to: kiwix/kiwix-js
-          registries: docker.io
+          registries: ghcr.io
           manual-tag: ${{ github.event.inputs.version }}
       # Restart live webapp only if we pushed an image to registry
       - name: Restart live webapp
--- a/.github/workflows/publish-extension.yaml
+++ b/.github/workflows/publish-extension.yaml
@@ -61,6 +61,14 @@ jobs:
           restrict-to: kiwix/kiwix-js
           registries: docker.io
           manual-tag: ${{ github.event.inputs.version }}
+      # Restart live webapp only if we pushed an image to registry
+      - name: Restart live webapp
+        if: github.event.inputs.target == 'docker' || github.event_name == 'release'
+        uses: actions-hub/kubectl@master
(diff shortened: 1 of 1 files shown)


kelson42:
@Jaifroid Could we just have the Docker image in the ghcr.io image repository. This is only "private" (for Kiwix) and no need to have it "public".

Jaifroid:
Sorry for slow reply, I'm travelling and in a rural location in Latin America with very intermittent Internet connection (great for testing Kiwix!). I need to understand the implications of the change. No need for the *image* to be public, but the sites (moz-extension.kiwix.org and pwa.kiwix.org on KJSW) need to be available to the public.

rgaudin:
This has no effect on the webapps themselves which will obviously be kept public!

The differene is that we would stop pushing the image to docker.io (Docker Hub). We'd only push them to ghcr.io.
We'll keep the images *public* so anyone would still be able to do `docker run ghcr.io/kiwix/kiwix-pwa` for instance but `docker run docker.io/kiwix/kiwix-pwa` (or its shortcut  `docker run kiwix/kiwix-pwa`) won't work.

rgaudin:
Reasoning behind this is we'd like to limit what we publish on Docker Hub to only *public* images intended for others to use.
We keep non-private but kiwix-usage-only images on ghcr.io.
We are not using private images anywhere at the moment.
Docker Hub has been trying numerous business model changes in the last couple years and the least we *depend* on t, the better.

Jaifroid:
OK, thanks for explanation, @rgaudin. This is looking good to me. I can't see any issues with the workflow, though it's difficult to test thoroughly until we have  a new release here. On KJSW I can test with the upcoming release of WikiMed, but that will have to wait till I have stable Internet..

Jaifroid (approved):
Thank you for this PR. @mossroy should check it too.

.github/workflows/publish-extension.yaml
```diff
@@ -61,6 +61,14 @@ jobs:
           restrict-to: kiwix/kiwix-js
           registries: docker.io
```
mossroy, line 62: Shouldn't this line be changed to ghcr.io, following what @kelson42 was asking for?

rgaudin: Yes, @kelson42's comment followed the PR. Let me change it and add the credentials as well

.github/workflows/publish-extension.yaml
```diff
@@ -49,18 +49,26 @@ jobs:
...
-            DOCKERIO_USERNAME=${{ secrets.DOCKERHUB_USERNAME }}
-            DOCKERIO_TOKEN=${{ secrets.DOCKERHUB_PASSWORD }}
+            GHCRIO_USERNAME=${{ secrets.GHCR_USERNAME }}
+            GHCRIO_TOKEN=${{ secrets.GHCR_TOKEN }}
           tag-pattern: /^v?([0-9.]+)$/
           latest-on-tag: true
           dockerfile: docker/dockerfile-moz-extension.pwa
           restrict-to: kiwix/kiwix-js
-          registries: docker.io
+          registries: ghcr.io
```
mossroy, line 62: Isn't there a need to change that in https://github.com/kiwix/k8s/blob/main/pwa/moz-extension.yaml#L19 too?

mossroy (approved):
Looks good.
But it's hard to test without releasing a new version of kiwix-js

.github/workflows/publish-extension.yaml
rgaudin: Indeed but I'm gonna have to manually push both image to ghcr first.

rgaudin:
✅  I've manually pushed both images to ghcr.io (only last version tag and corresponding latest), updated k8s to use them and deleted the *repo* on Docker Hub.

Issue #753, opened 2021-08-22
Restrict what an iframe can do, for security reasons


mossroy:
This is a follow-up of #751 

In case of a ZIM file with bad content in it, or a ZIM file that fetches bad content from the Internet, it could raise security issues.
Our browser extension does not ask for any permission in its manifest, but some nasty javascript could alter the content, send the articles read to an external server, break/alter our UI etc.

We should try to enforce that we only can execute code coming from the ZIM file (not from the Internet) and/or that the executed code is restricted to the iframe (no access to the UI).

Using [the sandbox attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/iframe#attr-sandbox) might be enough. But it could also have bad consequences on the ability to open external links (see discussion in #404)

Jaifroid:
This specific post summarizes my research on this: https://github.com/kiwix/kiwix-js/issues/404#issuecomment-861280543.

To summarize the summary: we can block external resources from running by injecting a `<meta>` tag defining a CSP into the article before loading it into the iframe. To be clear, this alters the article. It's not a practical problem, but it could be a "philosophical" problem: should we alter documents for security reasons? Can we implicitly trust every ZIM published by Kiwix.org? Could a hacker insert content onto a site's pages which then gets incorporated "automatically" into a ZIM archive? What about ZIMs that users make with ZimIt (these don't currently work in Kiwix JS, I think)?

If we also want to block **navigation** to external resources, we would have to add a CSP to `index.html` (our own), but this produces a terrible UX IMHO. This is less of a security concern for us, I think.

danielzgtg:
I no longer think it's possible to enforce sandbox restrictions on kiwix-js. I failed to do `Object.freeze(window.location)` and other things. To the browser, the kiwix-js controls look like just another script on the webpage. There's not really any practical way to tell the browser that the kiwix-js controls are more privileged and can control the zim scripts.

There are ways, but they have drawbacks:
- Static analysis (a.k.a modifying the HTML before loading). Malicious zims could always obfuscate their content. One way I did so was using base64, but there are infinitely many other ways to scramble text. Therefore this would either be equivalent to solving the halting problem, or it would have to block good scripts.
- Embed another Javascript engine using WASM. This would create a hard dependency on WASM, which old browsers don't have, and it will be slightly slow
- Add a Javascript interpreter written in Javascript to emulate all the scripts. This will be very slow.

[...]

PR #862, opened 2022-05-30
Open external links in new tabs, and warn the user

### LINKED ISSUE #404: In ServiceWorker mode, the external links are opened inside the iframe
Some websites refuse to be opened inside an iframe, which leaves a blank page, and raises an error in the console like : 
Refused to display 'https://www.nytimes.com/1984/01/19/us/carter-leaves-hospital.html' in a frame because it set 'X-Frame-Options' to 'sameorigin'.

The external links should be opened in a new tab (like in jQuery mode).

Using a `<base target="_blank" />` would probably work, but I suppose it would also apply to internal links, which is not what we want.
I'd like to avoid parsing the DOM or using regexp on the HTML content, but don't see a better way for now...

mossroy:
Fixes #404

This is just a refresh of the work of Jaifroid in branch https://github.com/kiwix/kiwix-js/tree/Check-for-external-links-and-warn

### DIFF
--- a/www/js/app.js
+++ b/www/js/app.js
@@ -90,6 +90,8 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
     params['appTheme'] = settingsStore.getItem('appTheme') || 'light'; // Currently implemented: light|dark|dark_invert|dark_mwInvert|auto|auto_invert|auto_mwInvert|
     // A global parameter to turn on/off the use of Keyboard HOME Key to focus search bar
     params['useHomeKeyToFocusSearchBar'] = settingsStore.getItem('useHomeKeyToFocusSearchBar') === 'true';
+    // A global parameter to turn on/off opening external links in new tab (for ServiceWorker mode)
+    params['openExternalLinksInNewTabs'] = settingsStore.getItem('openExternalLinksInNewTabs') ? settingsStore.getItem('openExternalLinksInNewTabs') === 'true' : true;
     // A parameter to access the URL of any extension that this app was launched from
     params['referrerExtensionURL'] = settingsStore.getItem('referrerExtensionURL');
     // A parameter to set the content injection mode ('jquery' or 'serviceworker') used by this app
@@ -162,6 +164,7 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
     document.getElementById('titleSearchRangeVal').textContent = params.maxSearchResultsSize;
     document.getElementById('appThemeSelect').value = params.appTheme;
     document.getElementById('useHomeKeyToFocusSearchBarCheck').checked = params.useHomeKeyToFocusSearchBar;
+    document.getElementById('openExternalLinksInNewTabsCheck').checked = params.openExternalLinksInNewTabs;
     switchHomeKeyToFocusSearchBar();
     document.getElementById('bypassAppCacheCheck').checked = !params.appCache;
     document.getElementById('appVersion').innerHTML = 'Kiwix ' + params.appVersion;
@@ -477,6 +480,10 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
         params.useHomeKeyToFocusSearchBar = this.checked ? true : false;
         settingsStore.setItem('useHomeKeyToFocusSearchBar', params.useHomeKeyToFocusSearchBar, Infinity);
         switchHomeKeyToFocusSearchBar();
+    });
+    $('input:checkbox[name=openExternalLinksInNewTabs]').on('change', function () {
+        params.openExternalLinksInNewTabs = this.checked ? true : false;
+        settingsStore.setItem('openExternalLinksInNewTabs', params.openExternalLinksInNewTabs, Infinity);
     });
     document.getElementById('appThemeSelect').addEventListener('change', function (e) {
         params.appTheme = e.target.value;
@@ -1456,6 +1463,22 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
                     // Configure home key press to focus #prefix only if the feature is in active state
                     if (params.useHomeKeyToFocusSearchBar)
                         iframeArticleContent.contentWindow.addEventListener('keydown', focusPrefixOnHomeKey);
+                    if (params.openExternalLinksInNewTabs) {
+                        // Add event listener to iframe window to check for links to external resources
+                        iframeArticleContent.contentWindow.addEventListener('click', function (event) {
+                            // Find the closest enclosing A tag (if any)
+                            var clickedAnchor = uiUtil.closestAnchorEnclosingElement(event.target);
+                            if (clickedAnchor) {
+                                var href = clickedAnchor.getAttribute('href');
+                                // We assume that, if an absolute http(s) link is hardcoded inside an HTML string,
+                                // it means it's a link to an external website.
+                                // We also do it for ftp even if it's not supported any more by recent browsers...
+                                if (/^(?:http|ftp)/i.test(href)) {
+                                    uiUtil.warnAndOpenExternalLinkInNewTab(event, clickedAnchor);
+                                }
+                            }
+                        });
+                    }
                     // Reset UI when the article is unloaded
                     iframeArticleContent.contentWindow.onunload = function () {
                         // remove eventListener to avoid memory leaks
@@ -1561,7 +1584,7 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
 
     // A string to hold any anchor parameter in clicked ZIM URLs (as we must strip these to find the article in the ZIM)
     var anchorParameter;
-
+    
     /**
      * Display the the given HTML article in the web page,
      * and convert links to javascript calls
@@ -1699,10 +1722,14 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
                 } else if (anchorTarget) {
                     // It's a local anchor link : remove escapedUrl if any (see above)
                     anchor.setAttribute('href', '#' + anchorTarget[1]);
-                } else if (anchor.protocol !== currentProtocol ||
-                    anchor.host !== currentHost) {
+                } else if ((anchor.protocol !== currentProtocol ||
+                    anchor.host !== currentHost) && params.openExternalLinksInNewTabs) {
                     // It's an external URL : we should open it in a new tab
-                    anchor.target = '_blank';
+                    anchor.addEventListener('click', function(event) {
+                        // Find the closest enclosing A tag
+                        var clickedAnchor = uiUtil.closestAnchorEnclosingElement(event.target);
+                        uiUtil.warnAndOpenExternalLinkInNewTab(event, clickedAnchor);
+                    });
                 } else {
                     // It's a link to an article or file in the ZIM
                     var uriComponent = uiUtil.removeUrlParameters(href);
--- a/www/js/lib/uiUtil.js
+++ b/www/js/lib/uiUtil.js
@@ -590,6 +590,59 @@ define(rqDef, function(settingsStore) {
 
     // If global variable webpMachine is true (set in init.js), then we need to initialize the WebP Polyfill
     if (webpMachine) webpMachine = new webpHero.WebpMachine({useCanvasElements: true});
+    
+    /**
+     * Warn the user that he/she clicked on an external link, and open it in a new tab
+     * 
+     * @param {Event} event the click event (on an anchor) to handle
+     * @param {Element} clickedAnchor the DOM anchor that has been clicked (optional, defaults to event.target)
+     */
+    function warnAndOpenExternalLinkInNewTab(event, clickedAnchor) {
+        event.preventDefault();
+        event.stopPropagation();
+        if (!clickedAnchor) clickedAnchor = event.target;
+        var target = clickedAnchor.target;
+        var message = '<p>Do you want to open this external link?';
+        if (!target || target === '_blank') {
+            message += ' (in a new tab)';
+        }
+        message += '</p><p style="word-break:break-all;">' + clickedAnchor.href + '</p>';
+        systemAlert(message, 'Opening external link', true).then(function (response) {
+            if (response) {
+                if (!target)
+                    target = '_blank';
+                window.open(clickedAnchor.href, target);
+            }
+        });
+    }
+    
+    /**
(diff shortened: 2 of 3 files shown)


www/js/app.js
```diff
@@ -1456,6 +1456,22 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
                     // Configure home key press to focus #prefix only if the feature is in active state
                     if (params.useHomeKeyToFocusSearchBar)
                         iframeArticleContent.contentWindow.addEventListener('keydown', focusPrefixOnHomeKey);
+                    // Add event listener to iframe window to check for links to external resources
+                    iframeArticleContent.contentWindow.addEventListener('click', function (e) {
+                        if (e.target.tagName === 'A') {
+                            var href = e.target.getAttribute('href');
+                            // We assume that, if an absolute http(s) link is hardcoded inside an HTML string,
+                            // it means it's a link to an external website
+                            if (/^http/i.test(href)) {
```
Jaifroid, line 1465: @mossroy Might it be worth including at least `^(?:http|ftp|mailto)` in this test? I can't think of others we are likely to encounter... except, maybe, `gopher`  🤣

mossroy: Well, I think we only need to handle the protocols that would otherwise be opened in the browser iframe.
`ftp` is not supported by recent browsers any more. But we could add it for older browsers anyway
Regarding `mailto`, `tel` or `geo` for example, browsers that know them will in already open another app. So we don't need to trap them.
I'll add the ftp protocol.
I think the other ones can be safely ignored :-)

mossroy:
I still need to add a setting to optionally disable this behavior, as we decided in #404
Will do that tomorrow

www/js/app.js
```diff
@@ -1456,6 +1456,27 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
...
+                    // Add event listener to iframe window to check for links to external resources
+                    iframeArticleContent.contentWindow.addEventListener('click', function (e) {
+                        if (e.target.tagName === 'A') {
+                            var href = e.target.getAttribute('href');
+                            // We assume that, if an absolute http(s) link is hardcoded inside an HTML string,
+                            // it means it's a link to an external website.
+                            // We also do it for ftp even if it's not supported any more by recent browsers...
+                            if (/^(?:http|ftp)/i.test(href)) {
+                                var target = e.target.getAttribute('target');
+                                // We don't override the default browser behavior if there's already a target attribute
```
Jaifroid, line 1468: Actually, I'm not sure about this. If I ask to be warned before the browser opens a link on the Internet (albeit in a new tab), I would expect to be warned uniformly in all cases, not only if the link happens not to have a target. So I think the warning should happen anyway, just don't set the target if it is already set. What do you think @mossroy ?

I have tested the functionality, and it works well.

mossroy: Based on https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#attr-target , the target can be used to communicate between tabs/windows/frames/iframes etc
I thought it might be safer to not do anything in such cases.
But you're right: we're talking about the case of an external target URL, for which it should be better to warn the user anyway.
I'll change accordingly

mossroy:
I've added the UI option to disable this behavior, in the "display" settings.
But I'm wondering if I should move it to the "expert" settings (and invert the logic, to call it "disableOpenExternalLinksInNewTabs", unchecked by default). It's what we initially intended in #404

mossroy:
The warning is already on by default. I'll had the comment in the dialog box to let the user know he/she can change this behavior.

My questioning is if it should be in "display" settings or in "expert" settings.
If we move it in "expert" settings, it's probably better to invert the setting: checked meaning "disable the opening in new tab" (it's currently the contrary: checked means "open in new tab")

Jaifroid:
Ah, so it's really just about how discoverable this is, and whether we want to discourage people from turning it off. I'm not sure it's really and "Expert Setting". I think of those as settings for the more technically minded, and I don't know if this is the case.

Jaifroid:
Well, they are closely related, so it's OK!

My personal feeling is that it's not an expert setting. I also think that it would be trivial to add the warning also in jQuery mode, for uniformity and for the same privacy reasons. But that's up to you!

mossroy:
> I also think that it would be trivial to add the warning also in jQuery mode, for uniformity and for the same privacy reasons. But that's up to you!

You're right, I'm trying to do that

mossroy:
I've implemented the same behavior for jQuery mode: it was indeed straightforward.

To me, this PR is ready for review.
I did not implement the CSP/sandbox discussed in https://github.com/kiwix/kiwix-js/issues/404#issuecomment-1141447090 : I read your comments and see it's not trivial. I see it as a separate issue (#753) that would be in a different PR

Jaifroid:
I've been doing a quick bit of testing. This link fails, and opens in the iframe instead. Any idea why? It's in the English Wikipedia article on poetry.

image

www/js/app.js
```diff
@@ -1561,6 +1582,29 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
...
+    function warnAndOpenExternalLinkInNewTab(event) {
+        event.preventDefault();
+        event.stopPropagation();
+        var target = event.target.target;
+        var message = 'Do you want to open this external link?';
+        if (!target || target === '_blank') {
+            message += ' (in a new tab)';
+        }
+        message += '<br />' + event.target.href + '<br />(this behavior can be disabled in settings if you know what you do)';
+        uiUtil.systemAlert(message, 'Opening external link', true).then(function (response) {
```
Jaifroid, line 1600: Instead of `<br />`, I would try wrapping each part of this message in `<p>` tags. At the moment it looks crowded and a bit ugly. The URL should be on its own line, with some vertical whitespace between the URL and the surrounding messages.

www/js/app.js
```diff
@@ -1561,6 +1582,29 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
...
+     */
+    function warnAndOpenExternalLinkInNewTab(event) {
+        event.preventDefault();
+        event.stopPropagation();
+        var target = event.target.target;
+        var message = 'Do you want to open this external link?';
+        if (!target || target === '_blank') {
+            message += ' (in a new tab)';
+        }
+        message += '<br />' + event.target.href + '<br />(this behavior can be disabled in settings if you know what you do)';
```
Jaifroid, line 1599: Slight correction to English, the verb needs to use continuous tense:

`this behavior can be disabled in settings if you know what you are doing`

Or, because it's a bit wordy, it could be simplified to `this warning can be disabled in settings`.

www/js/app.js
```diff
@@ -1456,6 +1463,20 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
                     // Configure home key press to focus #prefix only if the feature is in active state
                     if (params.useHomeKeyToFocusSearchBar)
                         iframeArticleContent.contentWindow.addEventListener('keydown', focusPrefixOnHomeKey);
+                    if (params.openExternalLinksInNewTabs) {
+                        // Add event listener to iframe window to check for links to external resources
+                        iframeArticleContent.contentWindow.addEventListener('click', function (e) {
```
Jaifroid, line 1468: As mentioned in another comment, it looks like we need some logic to detect a click on an element that contains an anchor element. I think it can be caught by specifying the bubbling phase without having to add lots of new logic, but that's always been a bit obscure to me and it needs some experimentation. This seems to affect some links in Wikipedia citations, for example. If we can't catch it in a simple way, simply inspecting the innerHTML (or other content) of the trapped element for an anchor target should not be hard.

I think I've faced this issue before somewhere, and there was a way to get, using DOM methods, the final target of a click....

Jaifroid (changes requested):
It looks good, but I've only tested briefly, and haven't yet tested in IE11 or IE mode with jQuery, only in Chromium.

www/js/app.js
Jaifroid: Btw, "behavior" is US spelling. In UK, it would be "behaviour". But use what you prefer, I'm not nationalistic!!
Or just use "warning".

www/js/app.js
mossroy: I agree, will use `<p>` tags

www/js/app.js
mossroy: In fact, the setting not only affects the warning, but also the fact that it opens in a new tab.
And maybe we should not encourage that, as it can have bad side-effects.
So maybe I should simply remove this sentence

mossroy:
> This link fails, and opens in the iframe instead. Any idea why?

I managed to reproduce. Will investigate

mossroy:
In the ZIM file where I reproduced the issue, it comes from the fact that there is a `<cite>` tag inside the `<a>` tag.
So `event.target` is the `cite` tag, and its `tagName` is CITE instead of A
In this particular case, testing `event.parentElement.target` would work.
To work in all cases, I suppose we need to browse all parent elements to find the closest `<a>` tag (if any)

Jaifroid:
Ah, so I think there is a difference between event.target and event.currentTarget (entirely from memory, I haven't checked). Maybe you can check?

Jaifroid:
See https://developer.mozilla.org/en-US/docs/Web/API/Event/currentTarget

Jaifroid:
Come to think of it event.currentTarget will probably be the iframe, right?

Jaifroid:
Great! Do you know about `Element.closest()`? https://developer.mozilla.org/en-US/docs/Web/API/Element/closest. I presume you didn't use because not supported by IE11 (though there is a polyfill, it's overkill for this use case).

mossroy:
> Do you know about `Element.closest()`?

Actually, I did not, thanks a lot for the pointer! Our support of old browsers tend to give me bad habits to do things by hand...

I think it would be a good idea to use that instead of my custom code, and add the polyfill for IE.
It's much more generic

mossroy:
The code is much simpler with `closest()`, thanks for the hint!
I've included the poyfill, but could not test with IE (or any other browser that would need the polyfill)

mossroy:
The test-offline-files github actions did very well its job to warn me I had to add the new file in the precache list of the serviceworker :+1:

Jaifroid:
> I've included the poyfill, but could not test with IE (or any other browser that would need the polyfill)

I should be able to test with IE11.

Jaifroid:
While testing I came across this case:

image

As you can see, because the URL doesn't have any space/break at which to wrap, it bleeds outside of the dialogue box. I think there might be a way to make it wrap regardless of whitespace. But I don't know without checking.

Jaifroid:
Not working in IE11, unfortunately. I'll try to determine why. I didn't get the dialogue box, it just showed me this.

image

Jaifroid:
Looks like the polyfill didn't load correctly.

image

Jaifroid:
Polyfill is loaded. It's in the Element.prototype, but somehow hasn't attached itself to event.target. When pausing debugger on the click event (jQuery mode) I did some checks in console:

image

www/js/app.js
```diff
@@ -1699,10 +1722,14 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
...
-                } else if (anchor.protocol !== currentProtocol ||
-                    anchor.host !== currentHost) {
+                } else if ((anchor.protocol !== currentProtocol ||
+                    anchor.host !== currentHost) && params.openExternalLinksInNewTabs) {
                     // It's an external URL : we should open it in a new tab
-                    anchor.target = '_blank';
+                    anchor.addEventListener('click', function(event) {
+                        // Find the closest enclosing A tag
+                        var clickedAnchor = event.target.closest('a');
+                        uiUtil.warnAndOpenExternalLinkInNewTab(event, clickedAnchor);
```
Jaifroid, line 1731:
```suggestion
                        // DEV: We use jQuery here to support IE11, which doesn't have Element.closest()
                        var clickedAnchor = $(event.target).closest('a');
                        clickedAnchor = clickedAnchor[0] || clickedAnchor;
                        uiUtil.warnAndOpenExternalLinkInNewTab(event, clickedAnchor);
```

mossroy:
Many thanks for your tests
I'm reluctant to use jQuery here, as we would like to remove it in #367 

Maybe a compromise could be to revert to last commit and, in `uiUtil.closestAnchorEnclosingElement` function, test if `closest` is supported. Use it if it's supported, else use the code I had written

Jaifroid:
OK. I'm really puzzled by why the Polyfill wouldn't run in IE11. It must be something to do with Require isolating it, though the Element prototype is patched by the Polyfill.

I think using your handcrafted solution if closest isn't available is a good idea if you would like to avoid jQuery here.

mossroy:
I just pushed it.
There are still 2 things to implement: the text wrap you mentioned, and also consider supporting the same feature with `<area>` tags (because it's what we already do in jQuery mode)

mossroy:
It's pushed.
But no rush for the review (I'll be AFK for the next 24 hours)

www/js/lib/uiUtil.js
```diff
@@ -578,6 +578,58 @@ define(rqDef, function(settingsStore) {
...
+     * 
+     * @param {Element} element
+     * @returns {Element} closest enclosing anchor tag (if any)
+     */
+    function closestAnchorEnclosingElement(element) {
+        if (Element.prototype.closest) {
+            // Recent browsers support that natively. See https://developer.mozilla.org/en-US/docs/Web/API/Element/closest
+            return element.closest('a,area');
+        }
+        else {
```
Jaifroid, line 619: I think we generally use the `} else {` style (I realize we're not consistent). Up to you.

www/js/lib/uiUtil.js
```diff
@@ -578,6 +578,58 @@ define(rqDef, function(settingsStore) {
...
+            // For other browsers, notably IE, we do that by hand (we did not manage to make polyfills work on IE11)
+            var currentElement = element;
+            while (currentElement.tagName !== 'A' && currentElement.tagName !== 'AREA') {
+                // If there is no parent Element, we did not find any enclosing A tag
+                if (!currentElement.parentElement)
+                    return;
+                // Else we try the next parent Element
+                currentElement = currentElement.parentElement;
```
Jaifroid, line 627: I think this would be easier to read set out like this (I know the effect is the same):
```suggestion
                if (!currentElement.parentElement) {
                    return;
                } else {
                    // Else we try the next parent Element
                    currentElement = currentElement.parentElement;
                }
```

Jaifroid:
PS Very useful PR, that makes behaviour consistent between Kiwix JS and other Kiwix apps. I'll also add it downstream when merged.

www/js/lib/uiUtil.js
mossroy: You're right

PR #864, opened 2022-05-31
Handle Enter key in Bootstrap modal dialogs

### LINKED ISSUE #863: Bootstrap dialog boxes should handle a default action with Enter key
Currently, when a bootstrap dialog box is displayed, hitting Enter key has no effect.

I think it should trigger the default action: OK when it's only an info, Confirm when it's a confirmation request.

mossroy:
Fixes #863

### DIFF
--- a/www/js/lib/uiUtil.js
+++ b/www/js/lib/uiUtil.js
@@ -73,16 +73,15 @@ define(rqDef, function(settingsStore) {
                     resolve(false);
                 }
             });
-            // Enter key should accept default action
-            $("#alertModal").keydown(function(e){
+            document.getElementById('alertModal').addEventListener('keyup', function (e) {
                 if (/Enter/.test(e.key)){
                     // We need to focus before clicking the button, because the handler above is based on document.activeElement
                     if (isConfirm) {
-                        $("#approveConfirm").focus();
-                        $("#approveConfirm").click();
+                        document.getElementById('approveConfirm').focus();
+                        document.getElementById('approveConfirm').click();
                     } else {
-                        $("#closeMessage").focus();
-                        $("#closeMessage").click();
+                        document.getElementById('closeMessage').focus();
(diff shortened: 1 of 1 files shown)


www/js/lib/uiUtil.js
```diff
@@ -73,6 +73,19 @@ define(rqDef, function(settingsStore) {
                     resolve(false);
                 }
             });
+            // Enter key should accept default action
+            $("#alertModal").keydown(function(e){
```
Jaifroid, line 77: I wonder if keyup is more usual, i.e. you expect the event to occur after you have released the key? Main reason to use keydown is if you're trying to get in first before other event listeners.

I notice you have used jQuery here. Is this to match the style of similar keydown blocks we already have?

In line with #367, I now try to use only native DOM methods. But there's an argument for matching coding style of nearby blocks.

Otherwhise:
```
document.getElementById('alertModal').addEventListener('keyup', function (e) {
    ...
}
```

Jaifroid (approved):
I tested, and it works. I may be wrong about keyup vs keydown.

PS Tested in Chromium (Edge) and IE11 mode.

www/js/lib/uiUtil.js
mossroy: Many thanks for this feedback.
The choice of keyup/keydown/keypress event was not really thought about, on my side. No problem to switch to keyup.

I used jQuery to match the style of the code above it, but I fully agree to switch to native DOM methods. I will change the code

Jaifroid: > The choice of keyup/keydown/keypress event was not really thought about, on my side. No problem to switch to keyup.

Use whatever seems most natural to you. I'm not very convinced by my own reasoning last night!

Jaifroid:
I tested the native DOM code in IE11. Working fine. I think it can be merged.

PR #868, opened 2022-06-04
Send requested byte range from SW, with correct headers

### LINKED ISSUE #531: Seeking in videos does not always work in SW mode in Chromium extension
With kiwix-js 2.5.0 or latest code, and with Chromium 75.

It fails with "Video Error" in the console, and the video is stuck with a spinner on it.
I did not notice this issue in older versions of Chromium.
I think we should try to fix this before releasing a version 2.6 (#525)

mossroy:
Fixes #531

### DIFF
--- a/service-worker.js
+++ b/service-worker.js
@@ -316,8 +316,18 @@ function fetchUrlFromZIM(urlObject, range) {
                 if (contentLength) headers.set('Content-Length', contentLength);
                 if (contentType) headers.set('Content-Type', contentType);
                 
+                // Test if the content is a video or audio file. In this case, Chrome & Edge need us to support ranges.
+                // See kiwix-js #519 and openzim/zimwriterfs #113 for why we test for invalid types like "mp4" or "webm" (without "video/")
+                // The full list of types produced by zimwriterfs is in https://github.com/openzim/zimwriterfs/blob/master/src/tools.cpp
+                if (contentLength >= 1 && /^(video|audio)|(^|\/)(mp4|webm|og[gmv]|mpeg)$/i.test(contentType)) {
+                    headers.set('Accept-Ranges', 'bytes');
+                }
+                
                 var slicedData = msgPortEvent.data.content;
                 if (range) {
+                    // The browser asks for a range of bytes (usually for a video or audio stream)
+                    // In this case, we honor this request, and send a partial content with the requested range
+                    // It's currently not optimal, as we in fact read all the content from the ZIM file
                     let partsOfRangeHeader = regexpByteRangeHeader.exec(range);
                     var begin = partsOfRangeHeader[1];
                     var end = partsOfRangeHeader[2];
@@ -331,14 +341,6 @@ function fetchUrlFromZIM(urlObject, range) {
                     console.debug('data content: first byte is ' + view[0] + ', last byte (offset ' + (end-begin) + ') is ' + view[end-begin], slicedData);
                 }
                 
-                // Test if the content is a video or audio file
-                // See kiwix-js #519 and openzim/zimwriterfs #113 for why we test for invalid types like "mp4" or "webm" (without "video/")
-                // The full list of types produced by zimwriterfs is in https://github.com/openzim/zimwriterfs/blob/master/src/tools.cpp
-                if (contentLength >= 1 && /^(video|audio)|(^|\/)(mp4|webm|og[gmv]|mpeg)$/i.test(contentType)) {
-                    // In case of a video (at least), Chrome and Edge need these HTTP headers or else seeking doesn't work
-                    // (even if we always send all the video content, not the requested range, until the backend supports it)
-                    headers.set('Accept-Ranges', 'bytes');
-                }
                 var responseInit = {
                     // HTTP status is usually 200, but has to bee 206 when a partial content (range) is sent
                     status: range? 206 : 200,
--- a/service-worker.js
+++ b/service-worker.js
@@ -94,7 +94,7 @@ const regexpZIMUrlWithNamespace = /(?:^|\/)([^/]+\/)([-ABCIJMUVWX])\/(.+)/;
  * 
  * @type {RegExp}
  */
-const regexpByteRangeHeader = /^[ \t]*bytes=(\d+)-/;
+const regexpByteRangeHeader = /^\s*bytes=(\d+)-/;
 
 /**
  * The list of files that the app needs in order to run entirely from offline code
@@ -199,9 +199,6 @@ let fetchCaptureEnabled = false;
  * Intercept selected Fetch requests from the browser window
  */
 self.addEventListener('fetch', function (event) {
-    if (event.request.url.endsWith('video.webm')) {
-        console.debug('SW is requested for the video', event.request, ...event.request.headers);
-    }
     // Only cache GET requests
     if (event.request.method !== "GET") return;
     var rqUrl = event.request.url;
@@ -294,7 +291,7 @@ self.addEventListener('fetch', function (event) {
  * Handles URLs that need to be extracted from the ZIM archive
  * 
  * @param {URL} urlObject The URL object to be processed for extraction from the ZIM
- * @param {String} range optional byte range string
+ * @param {String} range Optional byte range string
  * @returns {Promise<Response>} A Promise for the Response, or rejects with the invalid message port data
  */
 function fetchUrlFromZIM(urlObject, range) {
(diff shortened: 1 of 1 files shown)


mossroy:
@Jaifroid : this is still a draft but it seems to solve the issue on my Chromium.
Could you test with the browser/environments where you reproduced the issue?

Jaifroid:
That's good news! Great sleuthing. I'll do a test of this on Electron as soon as I can.

mossroy:
It seems to me that it started working when I fixed the response HTTP status code.
So maybe only a part of this PR is really needed.
In particular, I don't know if it's wise to send only the requested bytes: as we in fact always read all the data from the ZIM files, it might make us read it many times, while we could have sent it completely to the browser once for all.
I'll test

mossroy:
No, it was not only a matter of http status code.
I'm implementing what could be a compromise, where I honor only the beginning of the requested range: if the browsers asks for range x to y, I'll send the bytes x to the end of the stream (with HTTP code 206)
Hopefully it will avoid the browser from asking the following chunks (because it already has them)

Jaifroid:
OK, interesting. At a very low level, our FileReaders (at least for compressed BLOBs in the archive) decompress data as a stream. When I implemented ZSTD decoding, a lot of the challenge was in understanding how the decoder extracts just enough of the data stream to decompress a few bytes, then loops back and asks for more bytes only if needed, stopping at the end of the requested BLOB. This means that we only decompress just enough of the data stream to get the BLOB we are looking for, even though a compressed cluster may contain several BLOBs.

But I expect video BLOBs are stored without ZSTD compression, like images.

mossroy:
Apart from the debug logs, that I will have to remove, this PR is ready for test/review (but no rush)

Jaifroid:
I've tested this in Electron, by incorporating this code into the Service Worker for KJSW (on [a branch](https://github.com/kiwix/kiwix-js-windows/tree/Test-video-range-fix)). I've only tested on Windows so far.

It's a great improvement on the original behaviour. 😀

Skipping forward particularly works very well. Occasionally, skipping backwards, the video will stop, and I have to press play again, but it mostly will resume from where I had skipped back to. Occasionally, it will show a reload icon and ask me to reload the video, after which it works fine again (but from the beginning). Is this your experience?

By clicking backwards and forwards too many times without waiting much between clicks, eventually the video crashes. I assume this is just me overwhelming the back-end.

In any case, I would definitely consider the issue to be fixed, and I wouldn't expect absolute ability to jump around in the stream, given that it is not cached in memory.

@mossroy Would you like me to test other frameworks and other OS? If so, I would do a nightly build based on the branch and test the packages.

Jaifroid:
Just a thought: if we want truly random access to the video stream, would it be worth adding a simple video cache to the Service Worker (like the memory-based assetsCache we have?). Just a Map. There would have to be logic to clear it (to free up memory) once the SW receives a request for a new video OR a new article. The downside is increased memory use by the app, but it would be no more than we already use in jQuery mode (since making a BLOB URL must be the same kind of memory requirement).

mossroy:
Thanks for the test!
On my Chromium browsers, it always works without issue, forward and backward. I saw no crash.
I sometimes have the video paused, but it's when my click is a bit above the progress bar, and clicking on the video makes it pause.

Which ZIM file are you using to test?
I've tested with the ones on which I could most easily reproduce the issue before: https://download.kiwix.org/zim/videos/voa_learning_english-word-of-the-day_2021-12.zim and ted_en_global_issues_2018-07.zim

Jaifroid:
> Which ZIM file are you using to test?

I tested with: `ted_en_global_issues_2021-08.zim` and `ted_en_playlist-10-days-of-positive-thinking_2021-01.zim`.

There could be a different issue with Electron apps, to do with the non-standard use of the file:// protocol. It would be a better test for me to try with NWJS, because this uses a standard extension and standard extension URLs.

I may also be causing the "stop" by the issue you point out (of clicking too far above the line).

I'll test on NWJS.

Jaifroid:
Good news: it works very well in NWJS, and it's also working well in Electron 👍. The issue with the video stopping was indeed what you identified: because I hadn't changed the video to full screen, it was quite hard to hit the video timeline perfectly. When testing in full screen, I can be sure to click exactly on the line, and then the video doesn't stop, and I can jump around, forwards and backwards, in both frameworks.

There's one possible improvement. If you try "dragging" the video pip in the timeline backwards and forwards, seeking that way is very slow. On phones and tablets this is the main way people move video forward with their finger. Now these ZIMs are not well optimized for this use-case, because it's an old video.js that doesn't seem optimized for touch (video timeline is too narrow to touch properly). But you can sort of simulate it by dragging the pip with the mouse.

It seems clear that when we do this, the SW is (I imagine) processing multiple requests for closely adjacent bits of the file, and it has to cancel those requests and re-request potentially multiple times a second.

Would some kind of rate-limiting be possible? If not, or if it's too complicated, don't worry. Dragging does work, it's just much slower than simply clicking.

To see how it "should" work, try dragging in jQuery mode. Because the video is in memory, it drags forwards and backwards at an acceptable pace, even showing some frames along the way.

Improving dragging could be separate issue. You've already made a very big improvement here!

service-worker.js
```diff
@@ -85,6 +85,17 @@ var regexpExcludedURLSchema = /^(?:file|chrome-extension|example-extension):/i;
...
+/**
+ * Pattern to parse the first offset of a "range" request header
+ * NB: this only reads the first offset of the first byte range, where the spec allows several ranges, and several units.
+ * See https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Range
+ * But, in our case, we send a header to tell the browser we only accept the bytes unit.
+ * I did not see multiple ranges asked by a browser.
+ * 
+ * @type {RegExp}
+ */
+const regexpByteRangeHeader = /^[ \t]*bytes=(\d+)-/;
```
Jaifroid, line 97: With `[ \t]`, are you attempting to match whitespace? If so, you could just use `\s` (not in square brackets. `\s` matches any whitespace, including spaces, tabs, newlines (which you won't have here of course). NB I haven't tested.

service-worker.js
```diff
@@ -279,9 +294,10 @@ self.addEventListener('fetch', function (event) {
  * Handles URLs that need to be extracted from the ZIM archive
  * 
  * @param {URL} urlObject The URL object to be processed for extraction from the ZIM
+ * @param {String} range optional byte range string
```
Jaifroid, line 297: Just for clarity, I'd make "Optional" with capital letter, as it separates the variable name from the description visually.

service-worker.js
```diff
@@ -303,22 +319,43 @@ function fetchUrlFromZIM(urlObject) {
...
-                    // (even if we always send all the video content, not the requested range, until the backend supports it)
                     headers.set('Accept-Ranges', 'bytes');
-                    headers.set('Content-Range', 'bytes 0-' + (contentLength - 1) + '/' + contentLength);
                 }
+                
+                var slicedData = msgPortEvent.data.content;
+                if (range) {
+                    // The browser asks for a range of bytes (usually for a video or audio stream)
+                    // In this case, we partially honor the request: if it asks for offsets x to y,
+                    // we send a partial content starting at x offset, till the end of the data (ignoring y offset)
```
Jaifroid, line 334: "partial contents" better than "a partial content".

service-worker.js
```diff
@@ -303,22 +319,43 @@ function fetchUrlFromZIM(urlObject) {
...
+                    // So it's probably better to send all we have: hopefully it will avoid some subsequent requests of
+                    // the browser to get the following chunks (which would trigger some other complete reads in the ZIM file)
+                    // This might be improved in the future with the libzim wasm backend, that should be able to handle ranges.
+                    let partsOfRangeHeader = regexpByteRangeHeader.exec(range);
+                    let begin = partsOfRangeHeader[1];
+                    let end = contentLength-1;
+                    slicedData = slicedData.slice(begin);
+                    
+                    headers.set('Content-Range', 'bytes ' + begin + '-' + end + '/' + contentLength);
+                    headers.set('Content-Length', end-begin+1);
```
Jaifroid, line 345: Maybe add whitespace for readability: `end - begin + 1`

service-worker.js
```diff
@@ -303,22 +319,43 @@ function fetchUrlFromZIM(urlObject) {
...
+                    console.debug('video is read from backend and sent with headers', ...headers);
+                    let view = new Uint8Array(slicedData);
+                    console.debug('data content: first byte is ' + view[0] + ', last byte (offset ' + (end-begin) + ') is ' + view[end-begin], slicedData);
+                }
+                
                 var responseInit = {
-                    status: 200,
+                    // HTTP status is usually 200, but has to bee 206 when a partial content (range) is sent
```
Jaifroid, line 352: I think you can use singular but without "a", so "when partial content (range) is sent". It's because "content" is not a countable noun. You can't say "one content".

service-worker.js
```diff
@@ -303,22 +319,43 @@ function fetchUrlFromZIM(urlObject) {
...
+                    status: range? 206 : 200,
```
Jaifroid, line 353: For clarity: `range ? 206`

Jaifroid (approved):
Very nice PR!

service-worker.js
mossroy: You're right, I'll change.
I tend to have this habit because `\s` is not supported in all regular expression implementations. But it is in javascript so let's use it

service-worker.js
mossroy: I agree

service-worker.js
mossroy: OK

service-worker.js
mossroy: OK

service-worker.js
mossroy: OK

mossroy:
I've commited the changes you requested, and also removed the debug logs.

I also think that the dragging slowness can be considered a separate improvement.
To me, the right way to do that would be to have a backend that supports reading only a chunk of data. Our custom javascript can currently not, but libzim does: https://libzim.readthedocs.io/en/stable/api/classzim_1_1Item.html#_CPPv4NK3zim4Item7getDataE11offset_type9size_type
So, in theory, this should be possible with libzim wasm, simply by passing the offset/size parameters to `getData()`. It has to be tested, of course

PR #874, opened 2022-06-13
UIutil changes

me-aashish:
tested working of the site by replacing all the .innerHTML with .textContent


Jaifroid:
With this PR; the HTML formatting of dialogue boxes will be lost. As this is one of the advantages of using bootstrap, I think this is not the right way to do it. I'm closing, but please re-open if you have a solution.

PR #878, opened 2022-08-04
Switch to PWA in Firefox extension when SW API is unavailable

### LINKED ISSUE #877: Firefox extension does not switch to the PWA any more since Firefox 103.0
On several devices where I have kiwix-js 3.4.0 Firefox extension installed, it no longer switches to the PWA when trying to use the ServiceWorker mode.
It says "The ServiceWorker API is not available on your device. Falling back to JQuery mode".

I suspect that there has been a change in Firefox behavior, in version 103.0.
In previous version of Firefox, the ServiceWorker API was available in extensions (`'serviceWorker' in navigator` returned `true`), but failed to register. And we trapped this registraion exception to propose a switch to the PWA (when inside a Firefox extension).

In my current version (103.0 64 bits on Ubuntu), `'serviceWorker' in navigator` now returns `false` in the extension, so kiwix-js does not propose to use the PWA, and fallbacks to jQuery mode.

I do not see any mention of this change in https://developer.mozilla.org/en-US/docs/Mozilla/Firefox/Releases/103 so am not 100% sure

mossroy:
This is necessary for Firefox >=103, where the ServiceWorker API has been removed in WebExtensions.
See https://hg.mozilla.org/integration/autoland/rev/3a2907ad88e8

Fixes #877

### DIFF
--- a/www/js/app.js
+++ b/www/js/app.js
@@ -780,76 +780,85 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
                 caches.delete(ASSETS_CACHE);
             }
         } else if (value === 'serviceworker') {
-            if (!isServiceWorkerAvailable()) {
-                uiUtil.systemAlert('The ServiceWorker API is not available on your device. Falling back to JQuery mode', 'ServiceWorker API not available').then(function () {
-                    setContentInjectionMode('jquery');
-                });
-                return;
-            }
-            if (!isMessageChannelAvailable()) {
-                uiUtil.systemAlert('The MessageChannel API is not available on your device. Falling back to JQuery mode', 'MessageChannel API not available').then(function () {
-                    setContentInjectionMode('jquery');
-                });
-                return;
-            }
             var protocol = window.location.protocol;
-            if (!isServiceWorkerReady()) {
-                $('#serviceWorkerStatus').html("ServiceWorker API available : trying to register it...");
-                if (navigator.serviceWorker.controller) {
-                    console.log("Active service worker found, no need to register");
-                    serviceWorkerRegistration = true;
-                    // Remove any jQuery hooks from a previous jQuery session
-                    $('#articleContent').contents().remove();
-                    // Create the MessageChannel and send 'init'
-                    initOrKeepAliveServiceWorker();
-                    refreshAPIStatus();
-                } else {
-                    navigator.serviceWorker.register('../service-worker.js').then(function (reg) {
-                        // The ServiceWorker is registered
-                        serviceWorkerRegistration = reg;
-                        // We need to wait for the ServiceWorker to be activated
-                        // before sending the first init message
-                        var serviceWorker = reg.installing || reg.waiting || reg.active;
-                        serviceWorker.addEventListener('statechange', function(statechangeevent) {
-                            if (statechangeevent.target.state === 'activated') {
-                                // Remove any jQuery hooks from a previous jQuery session
-                                $('#articleContent').contents().remove();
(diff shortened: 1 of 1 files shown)


mossroy:
Switched this PR to draft as it's not working properly for now

mossroy:
Now tested on Firefox 103.0 on Ubuntu

Jaifroid:
Thanks for this PR. I'll test the UX (by installing as a developer-mode extension).

I wonder if there is another way of preventing users of old Firefox editions (I mean really old ones that don't support Service Worker at all) from being invited to switch to SW mode? Is this a case where testing explicitly for (Firefox < 103 AND SW unavailable) is warranted? Or could we test for a proxy API that will assure us that we are in a recent Firefox?

mossroy:
After fighting to disable auto-update of Firefox, I managed to test with version 52.9.0ESR.
When asking for SW mode, it suggests to use the PWA. If I accept, it opens the PWA, which "The ServiceWorker API is not available on your device" and fallsback to jQuery mode (but stays inside the PWA).
When you close the extension and open it again, it says "The last attempt to launch the PWA appears to have failed. Do you wish to try again?". If you confirm, it switches again to the PWA (same behavior). If you cancel, it stays in local jQuery mode.

So it's not that bad.

If we want to test the browser version, it should only be a "hint", like adding in the systemAlert that it will probably not work because this browser probably does not support ServiceWorkers

Jaifroid:
OK, that's good. I think it's enough for now. We might have to introduce something once we switch to SW mode as default (assuming we will apply that default to the Firefox extension version).

I've tested on:

* Firefox 103 (portable) on Windows 11 in a temporary extension. Works fine.
* Edge for Windows 11 (running from localhost). It starts automatically in SW mode (running from localhost).
* IE11 on Windows 11 (yes, there's a trick to be able to launch IE11 still on Windows 11!!), running from localhost. It rightly tells me that SW mode won't work when I try to switch, and falls back to jQuery mode.
* I haven't tested on old Edge legacy, but it's irrelevant now, and I think the behaviour will just mirror IE11 except that in some desktop versions of Edge legacy SW mode works. In that case it will act the same way as Edge/Chromium.

So I don't think I need to do any more tests. I'll take a look at the code.

www/js/app.js
```diff
@@ -780,76 +780,86 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
...
-                                $('#articleContent').contents().remove();
-                                // Create the MessageChannel and send the 'init' message to the ServiceWorker
+            // Since Firefox 103, the ServiceWorker API is not available any more in Webextensions. See https://hg.mozilla.org/integration/autoland/rev/3a2907ad88e8 and https://bugzilla.mozilla.org/show_bug.cgi?id=1593931
+            // Previously, the API was available, but failed to register (which we could trap a fews lines below).
+            // So we now need to suggest a switch to the PWA if we are inside a Firefox Extension and the ServiceWorker API is unavailable.
+            // Even if some older firefox versions do not support ServiceWorkers at all (versions 42, 43, 45ESR, 52ESR, 60ESR and 68ESR, based on https://caniuse.com/serviceworkers). In this case, the PWA will not work either.
+            if (protocol === 'moz-extension:' && !isServiceWorkerAvailable()) {
+                launchMozillaExtensionServiceWorker();
+            }
+            else {
```
Jaifroid, line 791: Use same coding style as surrounding code 😉
```suggestion
            } else {
```

Jaifroid (approved):
Good catch!

Just one small coding-style suggestion below. When you're happy you've done enough testing, feel free to squash/merge.

PR #882, opened 2022-08-13
Exempt more content from active content warning

### LINKED ISSUE #746: Kiwix-js detects "active content" on some new ZIM files, in jQuery mode
If you open http://tmp.kiwix.org/alcohol_meta_stackexchange_com_2021-08.zim in jQuery mode, it displays the "active content" warning popup.
Maybe we should consider white-listing this kind of ZIM files, as they mostly work properly, even in jQuery mode?

It should be decided when they're finalized (it seems they're not yet)

Jaifroid:
Fixes display of unnecessary active content warning with mdWiki ZIMs. I think this more generally fixes #746, but not necessarily directly as a result of this PR... We first posted #746 because of new Type 1 Stackexchange ZIMs throwing the active content warning. But I just tested the master branch against `beer.stackexchange.com_en_all_2022-08.zim` (which I've verified is a Type 1 ZIM), and it doesn't throw any warnings. So perhaps those ZIM types no longer trigger the warning?


mossroy (approved):
I did not test but this looks good

Jaifroid:
It's been quite widely treated on KJSW as I've been using this regex for a few months.

mossroy:
Sure, so you can merge

Issue #856, opened 2022-04-14
Update logo on Firefox and Windows stores


Popolechien:
These two ([here](https://addons.mozilla.org/en-US/firefox/addon/kiwix-offline/) and [here](https://www.microsoft.com/fr-ch/p/kiwix-js/9p8slz4j979j?rtc=1&activetab=pivot:overviewtab)) are still using the old logo. 

The updated one has been around for nearly two years now:
https://commons.wikimedia.org/wiki/File:Kiwix_logo_v3.svg

Jaifroid:
Oops, sorry! I have been meaning to update the Windows Store one, but other things always ended up having higher priority. I agree this is getting long in the tooth...

mossroy:
I've updated it on Firefox store: https://addons.mozilla.org/en-US/firefox/addon/kiwix-offline/

Jaifroid:
I'll update the Microsoft Store for the next release, early next month.

mossroy:
@Jaifroid : it seems to me that you updated the logo in the Microsoft Store, so I'm closing this issue.
Feel free to reopen it if necessary

Jaifroid:
Yes, thanks -- updated vis v2.0.0 release.

PR #884, opened 2022-08-23
Icon dark mode visible #881

### LINKED ISSUE #881: Kiwix Icon not really visible with Firefox Dark-mode
See my screenshot
image

Probably a white background should be forced.

blakecooper:
Fixes #881

The original icon (black with no outline) was not clearly visible in browsers with dark mode enabled. All icons in img/ and img/icons/ have been replaced with versions with a white outline from https://github.com/kiwix/kiwix-js-windows/tree/master/www/img.


blakecooper:
@kiwix/lieutenants

mossroy (changes requested):
Thanks for this PR.
It probably works, but I think it needs some changes:

- the result is not the same for all dimensions of the file. They all have the white outline, but some do not exactly correspond to the original image. You can see that with the distance between the feet and the ball: there should be no gap between them in the latest version of this logo (see latest update in #856)
- it might be better to keep both versions of each file: we want to switch to the outlined ones for the browser extension icon (handled in manifest.json file), but it's probably better to keep the original ones elsewhere (inside the app, for example). It would let us have the choice in the future, but would cost around 100 kB more in each package. I don't think it's a big issue, what do you think @Jaifroid ?
- if you do add an outlined version of each png file that are not used inside the app (bullet point above), you'll also have to list all these new files in scripts/Check-OfflineFilesList.ps1, because they do not need to be pre-cached by the PWA

blakecooper:
Thank you @mossroy, I can go through and make sure the outlined icons match the placement of the previous ones. I'll wait for @Jaifroid to weigh in before adding the original files to the app and updating the script.

Jaifroid:
I agree we only need to change the small icon.Is there any point keeping any of the deprecated icons? It  would be better to generate new sizes from a high resolution version - - I think I have one.

mossroy:
> I agree we only need to change the small icon.Is there any point keeping any of the deprecated icons? It would be better to generate new sizes from a high resolution version - - I think I have one.

These images are not only used as browser extension icons. Some are also used inside the app itself, or as icons for other platforms (Ubuntu Touch etc). But, well, maybe it's not a problem after all.

Jaifroid:
FWIW, if you look here: https://github.com/kiwix/kiwix-js-windows/tree/master/images, you'll see some high-res versions that could be used to generate the different sizes with border (probably most useful is `2000px-Kiwix_logo_v3_with_border.png`). However, for this PR I would say that it's enough to fix those that will be used as the icon in a file system / OS (perhaps those referenced directly in our manifests as file icons). There are online services that will turn a PNG into all the required icon sizes for various OSs or for manifest.json or other manifests. I'm sorry I don't have one to recommend right now.

image

blakecooper:
Thank you for your help and guidance! I used the file @Jaifroid recommended to generate the icons in all the sizes in the manifest. They should all be standardized to the correct image now.

mossroy:
It seems to me that some icons are blurry, while some other ones are not.
For example kiwix-128 looks sharp (OK), where kiwix-120 looks blurry. And most small ones look blurry too.
It's noticeable when you see them side by side in https://github.com/kiwix/kiwix-js/pull/884/files for example.

If I resize the high-res image with Gimp for example, to 120x120, it looks sharp:
image
Which tool did you use for the resizing?

I tested the browser extension on my firefox: compare the icon on the left and the one on the right in my browser:
image
For these small icons, it might also come from the fact that the actual drawing is a bit smaller inside the image. A workaround could be to crop a bit the high-res image before resizing it.

NB: Kiwix_icon_transparent_600x600.png is still the wrong image

blakecooper:
Hmm, I'm not sure what happened there. I used GIMP to scale the original 2000x2000 image to the required icon sizes, but maybe I accidentally scaled the already-scaled images for a few of them? I just re-generated the icons, taking care to always scale down from the original image for each, and made a new commit. The side-by-side images in GitHub look to my eye to be improved, but please let me know if you're still seeing the issue and I can try cropping it first. I also updated the 600x600 image. Thanks for your patience, and apologies for the back and forth!

blakecooper:
Looking into this, it seems the bird is smaller because of the extra space taken up by the outline. I can go through and make the bird the same size as before, but it looks like it would require clipping the outline at the edge of the image.

Here is a before and after image (using the 256x256 icon as an example):
example

Hopefully the image illustrates the problem better than I can by describing: the outline, when the bird is sized as before, will go beyond the canvas on the left and right sides. One option would be to make the outline thinner; I would be happy to do that, I would just need to be pointed toward the original .svg file so I could adjust in Inkscape. If there's a better approach, let me know and I'll be happy to follow up.

mossroy:
I agree with you @blakecooper
Is the file www/img/icons/kiwix-splash-scalable.svg (from the git repo) enough for you to create this thinner outline?

blakecooper:
It should be, let me try it out this weekend (I'm at work and don't have my personal laptop with me today) and get back to you, hopefully with some versions with a thinner outline.

blakecooper:
Sorry for the delay, my semester just started and I've had a lot going on. I created a version of the icon with a thinner outline, and tried to scale the icon to roughly the same size as before. How does these look? The new icon is on the left, and the previous icon is on the right:

Light mode:
Screenshot from 2022-09-11 14-35-02

Dark mode:
Screenshot from 2022-09-11 14-33-16

If this version seems to meet the requirements, let me know and I'll stage the files and do another pull request. Thanks for your patience!

mossroy:
No problem for the delay: thanks for not giving up.
This looks OK to me, even if it's hard to make up my mind without seeing all the files and testing it.

So please proceed.

NB: you probably don't need to create another PR: just update this one

blakecooper:
Thanks again, I added the new icons in each size and updated the 600x600 image. Let me know how it looks when you get a chance.

mossroy:
In real size, this is the result on my computer in light mode:
image

and in dark mode:
image

I think it's a good compromise: in dark mode, it's much more visible than before, and in light mode it's just slightly more blurry.

mossroy (approved):
This looks good to me.
The files are much heavier than before: 251kB vs 72kB. It's no surprise that the outline has a "cost" here, but I'm still surprised by the difference. Could you check there's no trap you fell into?
I saw there were metadata (at least software and date of creation), but these 2 small strings are not significant in terms of size.

blakecooper:
Yes, I’ll check on the file size issue and get back to you.-BlakeOn Sep 13, 2022, at 7:32 AM, Mossroy ***@***.***> wrote:﻿
@mossroy approved this pull request.

This looks good to me.
The files are much heavier than before: 251kB vs 72kB. It's no surprise that the outline has a "cost" here, but I'm still surprised by the difference. Could you check there's no trap you fell into?
I saw there were metadata (at least software and date of creation), but these 2 small strings are not significant in terms of size.

blakecooper:
I was able to reduce the icon file sizes pretty significantly. First of all, the resolution was set to 300 ppi when I was exporting (my mistake--thanks again for your patience with me, I swear I'm usually more diligent in my work when time allows), so I scaled that back down to 72 ppi like the originals. I also deleted most of the metadata.

That being said, the files still seem to be 1.5-2x the size of the originals. Like you said, the outline itself will have some cost in terms of storage, but if anybody has any other ideas they want to explore to reduce these I'm happy to look into it.

PR #885, opened 2022-08-25
Detect Zimit archive type and warn user

### LINKED ISSUE #883: Provide a warning if user tries to open Zimit-based archive
Kiwix Desktop currently provides a warning if a user tries to open a Zimit-based archive, It occurred to me we should do something similar in this app (and also ensure the app doesn't just hang when trying to open such a file). I have experience in detecting a Zimit archive reliably.

Jaifroid:
Fixes #883. I have included some information for the user about how to open Zimit archive types. Clearly this would need to be kept up to date, but I think it's better to be helpful to users to avoid the kind of frustrating exchange you see in this [reddit comment](https://www.reddit.com/r/Kiwix/comments/wj8ivr/comment/ijwwyfs/?utm_source=share&utm_medium=web2x&context=3).

### DIFF
--- a/www/js/lib/zimArchive.js
+++ b/www/js/lib/zimArchive.js
@@ -140,7 +140,7 @@ define(['zimfile', 'zimDirEntry', 'util', 'utf8'],
 
     /**
      * Detects whether the supplied archive is a Zimit-style archive or an OpenZIM archive and
-     * sets a file.type property accordingly; also returns the detected type. Extends ZIMFile.
+     * sets a _file.zimType property accordingly; also returns the detected type. Extends ZIMFile.
      * @returns {String} Either 'zimit' for a Zimit archive, or 'open' for an OpenZIM archive
      */
      ZIMArchive.prototype.setZimType = function () {
@@ -150,7 +150,7 @@ define(['zimfile', 'zimDirEntry', 'util', 'utf8'],
             this._file.mimeTypes.forEach(function (v) {
                 if (/warc-headers/i.test(v)) fileType = 'zimit';
             });
-            this._file.type = fileType;
+            this._file.zimType = fileType;
             console.debug('Archive type set to: ' + fileType);
         } else {
             console.error('ZIMArchive is not ready! Cannot set ZIM type.');
--- a/www/js/lib/zimArchive.js
+++ b/www/js/lib/zimArchive.js
@@ -76,6 +76,8 @@ define(['zimfile', 'zimDirEntry', 'util', 'utf8'],
                         countName: 'articleCount'
                     }
                 ]);
+                // Set the archive file type ('open' or 'zimit')
+                that.setZimType();
                 // DEV: Currently, extended listings are only used for title (=article) listings when the user searches
                 // for an article or uses the Random button, by which time the listings will have been extracted.
                 // If, in the future, listings are used in a more time-critical manner, consider forcing a wait before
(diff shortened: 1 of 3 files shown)


Jaifroid:
Tested and working in Edge Chromium and Internet Explorer 11.

Jaifroid:
I tested against the two August Zimit archives in the zimit directory:

* birds-of-ladakh_en_all_2022-08.zim
* internet-encyclopedia-philosophy_en_all_2022-08.zim

Jaifroid:
Tested and working in Firefox 103.0.2.

www/js/lib/zimfile.js
```diff
@@ -101,7 +102,8 @@ define(['xzdec_wrapper', 'zstddec_wrapper', 'util', 'utf8', 'zimDirEntry', 'file
      * @property {Integer} mimeListPos Position of the MIME type list (also header size)
      * @property {Integer} mainPage Main page or 0xffffffff if no main page
      * @property {Integer} layoutPage Layout page or 0xffffffffff if no layout page
-     * @property {Map} mimeTypes The ZIM file's MIME type table rendered as a Map (calculated entry)
+     * @property {String} type Extended property: either 'open' for OpenZIM file type, or 'zimit' for the warc2zim file type used by Zimit (set in zimArchive.js)
```
mossroy, line 105: Regarding the naming of this variable, `type` looks a bit too broad to me. And, technically, Zimit files also comply to OpenZIM spec.
Maybe we could simply add a boolean variable `isZimit` instead of `type` ? And set it to true or false in `setZimType`

www/js/lib/zimArchive.js
```diff
@@ -135,7 +137,27 @@ define(['zimfile', 'zimDirEntry', 'util', 'utf8'],
...
+     * Detects whether the supplied archive is a Zimit-style archive or an OpenZIM archive and
+     * sets a file.type property accordingly; also returns the detected type. Extends ZIMFile.
+     * @returns {String} Either 'zimit' for a Zimit archive, or 'open' for an OpenZIM archive
+     */
+     ZIMArchive.prototype.setZimType = function () {
+        var fileType = null;
+        if (this.isReady()) {
+            fileType = 'open';
+            this._file.mimeTypes.forEach(function (v) {
+                if (/warc-headers/i.test(v)) fileType = 'zimit';
```
mossroy, line 151: It's very minor but check if you want to match `warc-headers` anywhere in the mimeType string, or exactly this string. In this case it would be safer to use:
```suggestion
                if (/^warc-headers$/i.test(v)) fileType = 'zimit';
```

mossroy (approved):
I did not test, but it looks good.

www/js/lib/zimArchive.js
Jaifroid: I think it is best to keep this broad because any appearance of `warc-headers` in a MIME type indicates we have a ZIM that was created with warc2zim. The full string is currently `application/warc-headers`, but there's no particular reason why `text/warc-headers` (for example) would be invalid, since the headers are in text format. But we might in future have `application/warc-headers-encoded` (there's no guarantee of stability, as there is no published spec), so making the test too specific could limit the ability to detect a Zimit ZIM in this case.

www/js/lib/zimfile.js
Jaifroid: Thanks for the suggestion. My only doubt is whether there might be other types in the future that we might want to use this property for  -- but if so it would be easy enough to change. Regarding whether Zimit files comply with the OpenZIM specification, I would say that they are probably intended to comply, but that they don't for all the reasons listed in https://github.com/openzim/warc2zim/issues/86. However, on balance, they have enough resemblance to the OpenZIM spec to be a special case of that spec currently undocumented.

In any case, I'm happy to change this to a Boolean.

www/js/lib/zimArchive.js
mossroy: Sure, no problem

www/js/lib/zimfile.js
mossroy: OK no problem, you know the topic much better than I do.
You can keep it as it is if you whish

Jaifroid: I've changed the name of this property to `zimType`, which is more specific than `type`, but I've left it as a string in case we end up with some more types in the future.

Jaifroid:
I re-ran tests after latest code changes. Everything seems to be working, including in IE11. I'll merge this afternoon unless I hear to the contrary.

Issue #865, opened 2022-06-02
Fix scrapers to remove inline javascript in ZIM files


mossroy:
As discussed in #789, browser extensions now enforce some CSP that block inline javascript.
And it might happen in some other contexts, too.

Here is a list of affected scrapers that should be fixed (most critical first):

- [x] https://github.com/openzim/gutenberg/issues/145 : main page unusable, other pages seem to work well
- [x] https://github.com/openzim/nautilus/issues/34 : main mage unusable, other pages seem to work well
- [x] https://github.com/openzim/phet/issues/134 : main page unusable, other pages seem to work well
- [x] https://github.com/openzim/phet/pull/152
- [x] https://github.com/openzim/youtube/issues/154 : "only" affects the drop-down list of the home page, and loading of the webp-polyfill
- [x] https://github.com/openzim/openedx/issues/168: a lot of inline javascript blocks + some "onclick" attributes
- [x] https://github.com/openzim/sotoki/issues/249 : several inline javascript blocks + some "onerror" attributes on some images
- [x] https://github.com/openzim/mwoffliner/issues/1578 : minor. It seems to block webp-polyfill related code
- [x] https://github.com/openzim/wikihow/issues/108 : minor. It blocks some video-related javascript
- [x] https://github.com/openzim/ted/issues/128 : minor. It seems to block webp-polyfill related code
- [x] https://github.com/openzim/kolibri2zim/issues/39 : minor. It seems to block webp-polyfill related code

NB: please tick the items in the list only when it has been tested with Kiwix-js with a public ZIM file

Jaifroid:
It looks like we might have got as far as we can with this issue. It has become less critical now that we are forced (due to Manifest V3) to use the PWA workaround for the Chromium extension. However, "conceptually" it would be nice to fix the remaining three scrapers. @kelson42 Do you think this issue should now be closed (but leave the unresolved three issues listed above still open)?

kelson42:
Still something to finish, now officially on the plate of @benoit74 ;)

benoit74:
`kolibri` is now done with https://github.com/openzim/kolibri/pull/43, will probably be released very soon with https://github.com/openzim/kolibri/pull/44, please tick the box (I don't have sufficient rights ;o))

Jaifroid:
@benoit74 Wonderful, many thanks! Box duly ticked!

benoit74:
@Jaifroid : I forgot to mention it, you can tick `openedx` as well

[...]

Issue #789, opened 2022-01-04
Allow inline scripts in extensions or provide workaround


Jaifroid:
At least in the Chromium extension, our declared CSP currently restricts inline script execution. Due to the increasing (and unfortunate) use of inline scripts in ZIMs, this is causing a large number of exceptions to appear in console. ~~So far, they don't significantly affect usability~~ at least in Wikimedia ZIMs. **EDIT: Usability is affected in non-Wikimedia ZIMs: see next comment.** However, as the message suggests (see screenshot) we may be able to run them by declaring `unsafe-inline` in the CSP (in the manifest). Of course, this would depend on whether the Stores allow apps to be submitted with this CSP now.

Additionally, ss #752.

![image](https://user-images.githubusercontent.com/4304337/148057909-3d45f699-1258-496c-88ae-8ee2b236056b.png)

Jaifroid:
Investigating this a little further, unfortunately the CSP prevents loading the proprietary UI of Gutenberg ZIMs in the Chromium extension. Compare the PWA from Kiwix JS Windows on left, and the Chromium extension on the right in screenshot below. This makes this issue rather more urgent than I thought.

![image](https://user-images.githubusercontent.com/4304337/148059704-204e7ed1-7436-4a97-bb73-624c5d7881f4.png)

mossroy:
It also breaks Phet, which I reported here : https://github.com/openzim/phet/issues/134

Jaifroid:
I guess this will break any ZIM UI that relies on ReactJS too, given extensive use of script as markup in React. See https://javascript.plainenglish.io/creating-a-chrome-extension-with-react-d92db20550cb#481f about creating SHA keys in order to run React, with precisely our problem.

As creating SHA keys for every inline script in every ZIM will be impossible for us, we will have to try the `unsafe-inline` option, like we had to do with `unsafe-eval` in order to load WASMs conditionally. The latter didn't block submission to the Store, so maybe we can get away with the former too, and as we know the source of our content, it doesn't seem to be risky.

mossroy:
The issue on Gutenberg should be reported on https://github.com/openzim/gutenberg/issues : from what I saw, it should be easy to fix for future ZIMs.

I don't think it breaks all ZIM files that use React, but it does break many existing ZIM files (because the scrapers add inline javascript)

I agree with you that, if it's technically possible, we should allow inline javascript in our extension. At least to be able to open existing or old ZIM files.
[...]

PR #889, opened 2022-09-23
Detect more ZIMs for active content warning

Jaifroid:
New nautilus-based ZIMs don't trigger the active content warning in jQuery mode. This adds detection. There is also an incorrect instruction in the warning to "type a space". This doesn't do anything in Kiwix JS (in Kiwix JS Windows, typing a space opens a special Archive Index UI, and I think this instruction is a left-over from that).

We could make it so that typing a space displays the title index starting from the first entry within the current UI, but that would be a different PR.

### DIFF
--- a/www/index.html
+++ b/www/index.html
@@ -662,7 +662,7 @@
                     <div id="activeContent" style="display:none;" class="kiwix-alert alert alert-warning alert-dismissible fade show">
                         <button type="button" class="close" data-hide="alert">&times;</button>
                         <strong>Unable to display active content:</strong> This ZIM is not fully supported in jQuery mode.<br />
-                        Content may be available by searching above (type a space or a letter of the alphabet), or else 
+                        Content may be available by searching above (type a letter of the alphabet), or else 
                         <a id="swModeLink" href="#contentInjectionModeDiv" class="alert-link">switch to Service Worker mode</a>
                         if your platform supports it. &nbsp;[<a id="stop" href="#expertSettingsDiv" class="alert-link">Permanently hide</a>]
                     </div>
--- a/www/js/app.js
+++ b/www/js/app.js
@@ -1580,12 +1580,12 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
     // quote marks (') in the URL.
     var regexpTagsWithZimUrl = /(<(?:img|script|link|track)\b[^>]*?\s)(?:src|href)(\s*=\s*(["']))(?![a-z][a-z0-9+.-]+:)(.+?)(?=\3|\?|#)/ig;
     // Regex below tests the html of an article for active content [kiwix-js #466]
-    // It inspects every <script> block in the html and matches in the following cases: 1) the script loads a UI application called app.js
-    // or init.js; 2) the script block has inline content that does not contain "importScript()", "toggleOpenSection" or an "articleId"
-    // assignment (these strings are used widely in our fully supported wikimedia ZIMs, so they are excluded); 3) the script block is not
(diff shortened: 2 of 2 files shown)


mossroy (approved):
Looks godd

Jaifroid:
I'm wondering if we need to check some of the other ZIM types: we were to some extent relying on the fact that some of these ZIMs had inline script to detect active content, and now some of those have been addressed (not all yet). I should perhaps check the ZIM types that have been fixed, in case we need to add to this PR.

mossroy:
You're right.
Hopefully there would be an easy way to detect that

Jaifroid:
OK, here is a list of ZIMs that should be tested (mirroring the list in #865):

- [ ] Gutenberg (still uses inline JS, but would be good to future-proof);
- [ ] Nautilus;
- [ ] PhET (still uses inline JS);
- [ ] YouTube;
- [ ] OpenEdx (still uses inline JS);
- [ ] Sotoki;
- [ ] Wikihow;
- [ ] Ted Talks;
- [ ] Kolibri (still uses inline JS).

Jaifroid:
To support detecting Gutenberg in the future, once inline JS is removed, we could detect this script: `languages.js`, or more generically `tools.js`or `l109.js`. I guess the more generic the better, so that one detection might work for more than one ZIM type? In that case `l109.js' might be best? Though I have seen `l18n` used too (not in this ZIM).

Jaifroid:
There are just two left to test: OpenEdx and Kolibri. Unfortunately, I have no idea which ZIMs actually represent these ZIM types! Any ideas @mossroy, @kelson42? It's not crucial to test them, as these are scrapers that haven't been updated yet, so they will still show the active content warning. However, we could future proof them, if they are sitll active projects, by including them in this PR.

mossroy:
OpenEdx generates some moocs, see https://github.com/kiwix/kiwix-js/issues/789#issuecomment-1059085846 and https://github.com/openzim/openedx/issues/168
Kolibri generates some Khan Academy videos, see https://github.com/openzim/kolibri/issues/39

mossroy:
It's true that we should not rely on inline javascript detection any more (as we're making scrapers get rid of it).
This can not be checked in a very reliable way, but what you do looks like a good compromise to me : detect some javascript libraries that would "imply" some dynamic content (i.e the content would not be readable without executing them)

Jaifroid:
OK thanks -- I should have followed up the PRs.

Khan Academy (specifically the one you linked) contains `app.js`, so we're covered already.

The OpenEdx one is a curious case: it actually runs better in jQuery mode than in Service Worker mode. All assets are browsable in jQuery mode, but in SW mode, the assets  "break out" of the iframe, and so destroy the Kiwix JS reader. As it looks like this scraper won't get fixed (developer(s) have abandoned it??), I'll exempt this ZIM (via a unique JS it contains) so that we don't put up a misleading message to jQuery users. We might want to investigate if we can prevent the breakout from happening in SW mode, but only if it can be done in a generic way, which I doubt.

kelson42:
Both scrapers are not abandoned, but openedx is pretty recent an kolibri still a work in progress.

Jaifroid:
Ah, sorry, it was rgaudin saying "changes of a fix are low" for one of them that made me think this! Very glad to hear they are in progress.

Jaifroid:
OK, I've finished testing all ZIMs. There is a bit of new code you should look at, @mossroy , because in order to exempt ZIMs containing the `mooc.js` script, I need to do a separate search of the HTML. This is because you can't prove a negative in a positive regular expression search. The active content regex stops searching as soon as it has matched, so it might never see that the html further down loads `mooc.js`. Hence an extra search is needed for that. The search is only run if active content was found, so it won't impact on load times (and regexes are very fast anyway).

mossroy (approved):
I did not test, but it looks good to me.

PR #893, opened 2022-10-06
Rename nightly packages to show date instead of revsion ID

Jaifroid:
This was requested on https://github.com/kiwix/k8s/issues/50. It will provide the benefit that we will have permalinks to the latest nightly release, and is being implemented Kiwix-wide.

My Linux scripting skills are basic, but I think the included code should rename files like this:

```
build/kiwix-firefox-unsigned-extension-3.6commit-9a6e6f7.zip -->
build/kiwix-firefox-unsigned-extension_2022-10-06.zip
```

If you agree, I can test by dispatching the workflow on this branch.

### DIFF
--- a/scripts/create_all_packages.sh
+++ b/scripts/create_all_packages.sh
@@ -84,7 +84,7 @@ if [ "${DRYRUN}zz" == "zz" ]; then
         # It's a nightly build, so rename files to include the date and remove extraneous info so that permalinks can be generated
         for file in build/*; do
             target=$(sed -E "s/-[0-9.]+commit[^.]+/_$CURRENT_DATE/" <<<"$file")
-            mv "$file" "$target"
+            mv --backup=numbered "$file" "$target"
         done
     fi
     # Upload the files on master.download.kiwix.org
--- a/scripts/create_all_packages.sh
+++ b/scripts/create_all_packages.sh
@@ -84,7 +84,7 @@ if [ "${DRYRUN}zz" == "zz" ]; then
         # It's a nightly build, so rename files to include the date and remove extraneous info so that permalinks can be generated
         for file in build/*; do
             target=$(sed -E "s/-[0-9.]+commit[^.]+/_$CURRENT_DATE/" <<<"$file")
-            mv --backup=numbered "$file" "$target"
+            mv "$file" "$target"
         done
     fi
     # Upload the files on master.download.kiwix.org
--- a/scripts/create_all_packages.sh
+++ b/scripts/create_all_packages.sh
@@ -80,11 +80,11 @@ if [ "${DRYRUN}zz" == "zz" ]; then
     # Change permissions on source files to match those expected by the server
     chmod 644 build/*
     CURRENT_DATE=$(date +'%Y-%m-%d')
-    if [ "${TAG}zz" == "zz" ]; then
+    if [ ! "${CRON_LAUNCHED}zz" == "zz" ]; then
         # It's a nightly build, so rename files to include the date and remove extraneous info so that permalinks can be generated
         for file in build/*; do
             target=$(sed -E "s/-[0-9.]+commit[^.]+/_$CURRENT_DATE/" <<<"$file")
-            mv --backup=numbered "$file" "$target"
+            mv "$file" "$target"
         done
     fi
     # Upload the files on master.download.kiwix.org
--- a/scripts/create_all_packages.sh
+++ b/scripts/create_all_packages.sh
@@ -37,7 +37,7 @@ else
     # and we have to comply with their version string : https://developer.mozilla.org/en-US/docs/Mozilla/Toolkit_version_format
     # So we need to replace every number of the commit id by another string (with 32 cars max)
     # We are allowed only a few special caracters : +*.-_ so we prefered to use capital letters
-    # (hopping this string is case-sensitive)
+    # (hoping this string is case-sensitive)
     COMMIT_ID_FOR_MOZILLA_MANIFEST=$(echo $COMMIT_ID | tr '[0123456789]' '[ABCDEFGHIJ]')
     VERSION_FOR_MOZILLA_MANIFEST="${MAJOR_NUMERIC_VERSION}commit${COMMIT_ID_FOR_MOZILLA_MANIFEST}"
     echo "Packaging version $VERSION"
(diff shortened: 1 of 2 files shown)


Jaifroid:
Hmm, seems to be a problem on the SauceLabs side. I've rerun the tests three times and get the same error. Will try again this afternoon/evening.

image

scripts/create_all_packages.sh
```diff
@@ -80,6 +80,11 @@ if [ "${DRYRUN}zz" == "zz" ]; then
     # Change permissions on source files to match those expected by the server
     chmod 644 build/*
     CURRENT_DATE=$(date +'%Y-%m-%d')
+    # Rename files to include the date and remove extraneous info so that permalinks can be generated
+    for file in build/*; do
+        target=$(sed -E "s/-[0-9.]+commit[^.]+/_$CURRENT_DATE/" <<<"$file")
+        mv "$file" "$target"
```
mossroy, line 86: It might be a good idea to handle the case of `$target` already existing.
It's true that our nightly build runs once a day normally. But we might want/need to run it several times.
Using the `--backup` option of `mv` command-line might be enough for this rare use-case

scripts/create_all_packages.sh
```diff
@@ -80,6 +80,11 @@ if [ "${DRYRUN}zz" == "zz" ]; then
     # Change permissions on source files to match those expected by the server
     chmod 644 build/*
     CURRENT_DATE=$(date +'%Y-%m-%d')
+    # Rename files to include the date and remove extraneous info so that permalinks can be generated
+    for file in build/*; do
+        target=$(sed -E "s/-[0-9.]+commit[^.]+/_$CURRENT_DATE/" <<<"$file")
```
mossroy, line 85: When we create a tag (for a released version), we go through the same code. We have a different file name, that will not match this regexp.
It will generate an error with `mv` because `$file` and `$target` would be the same file.
It should be simple to just not to do the `mv` in this case.

mossroy (changes requested):
It looks good, but filenames coming from a tag should be handled properly.

scripts/create_all_packages.sh
Jaifroid: Thanks for the suggestion! In my version of Linux (Ubuntu on WSL), in the bash terminal, a `mv` command overwrites silently (without error) any existing target, unless I add the `-no-clobber` option. Wouldn't it be enough that it's overwritten, or do you want explicitly to keep any builds made when re-running the script? If you think we should keep a backup of any overwritten versions, then I'll add `--backup` as you suggest.

scripts/create_all_packages.sh
Jaifroid: OK, thanks. Good idea.

scripts/create_all_packages.sh
Jaifroid: I've added backup option.

Jaifroid:
I just did a test run on this branch and it seems to work:

image

Jaifroid:
However, I've realized that adding the backup option to the `mv` command won't do what we thought it would. The files are copied to the server via `scp`, not with the `mv` command. That's only used to rename them before uploading. Any copies on the server will be overwritten unless we rename them.

Jaifroid:
Actually, it's sftp, not scp, and according to the [man page](https://manpages.ubuntu.com/manpages/kinetic/en/man1/sftp.1.html)  there is no option to rename instead of overwriting.

Jaifroid:
Unrelated: Sauce Labs seems definitievly broken with our tests. It has not passed once even with multiple pushes and restarts. It seems one version of Firefox refuses to load. Could be deprecated (that particular version)?

Jaifroid:
Sorry, I'm being dense. I forgot this script is run locally, and in that context `mv --backup` does indeed makes sense. I've put it back.

mossroy:
I'm having another idea: we might instead distinguish a nightly build from other builds by passing a command-line argument when executed by github actions.
If it's a nightly build, we do the renaming, And it's not a problem to overwrite an existing file with the same name
If it's not, we don't rename. It can be the case when releasing (on a tag), or any manual launch we might do

mossroy:
Regarding SauceLabs, they don't seem to have deprecated any version of Firefox, based on https://saucelabs.com/platform/supported-browsers-devices

Jaifroid:
> I'm having another idea: we might distinguish a nightly build from other builds by passing a command-line argument when executed by github actions. If it's a nightly build, we do the renaming, And it's not a problem to overwrite an existing file with the same name If it's not, we don't rename. It can be the case when releasing (on a tag), or any manual launch we might do

If you know how to implement your idea quickly, feel free to contribute to this PR. Otherwise, I'll tackle it at the weekend!

Jaifroid:
> Regarding SauceLabs, they don't seem to have deprecated any version of Firefox, based on https://saucelabs.com/platform/supported-browsers-devices

Hmm, then we need to investigate, because it looks completely broken rather than the intermittent failure we'er used to dealing with.

Jaifroid:
> by passing a command-line argument when executed by github actions

I believe there's an API for that: a variable is set, which tells you what type of launch it is. There certainly is for chron-launched jobs. Not sure for jobs dispatched from the GitHub UI or from the API (as opposed to local running of script). I think it's possible to test that you're in a GitHub environment, which might be enough here.

Jaifroid:
Are the changes in https://github.com/kiwix/kiwix-js/pull/893/commits/ba9dbd1f958ace6ec3355faa24456beb76e366b6 roughly what you were thinking of? This would only do the renaming if the job was launched as a CRON job. It would mean that we could clearly distinguish the packages produced by setting a tag from those produced by a nightly build, and the former wouldn't overwrite the latter. On the other hand, we wouldn't be able to produce a nightly build manually (by workflow dispatch). If needed (I'm not sure it is), then that can be done by altering the workflow to add a user-selectable "nightly" parameter to the dispatch UI on GitHub (I did this for a different workflow on KJSW).

scripts/create_all_packages.sh
```diff
@@ -80,6 +80,13 @@ if [ "${DRYRUN}zz" == "zz" ]; then
     # Change permissions on source files to match those expected by the server
     chmod 644 build/*
     CURRENT_DATE=$(date +'%Y-%m-%d')
+    if [ ! "${CRON_LAUNCHED}zz" == "zz" ]; then
```
mossroy, line 83: It certainly works, but I don't find this line easy to understand.

It's my fault, it's the code style I had used in this script. I'm not proud of that: I just hope I had a good reason, at that time, like execution on very old versions of bash because of Travis?

Anyway, there are more explicit ways to test if a variable exists and is not empty.
Try with:
```suggestion
    if [ -n "${CRON_LAUNCHED}" ]; then
```
(the same syntax could be used above, and maybe in other scripts)
Feel free to move that in a separate issue if you wish

mossroy (approved):
Yes, it's what I meant

Jaifroid:
Might as well update the whole file now. From reading around, I gather `-n` means the variable is non-zero lenght (i.e. it isn't null), and `-z` means it is zero-length (null). So I've update the tests accordingly, and will test.

Jaifroid:
So, I've added a UI interface (on GitHub) so that the user can choose between a "nightly" release or a normal (untagged) release. In first screenshot you see the UI, and in second you see the result of running it first with an input of "release" (actually it could be anything) -- these are the five files at the bottom of the list --, and second with an input of "nightly", which produces the top five files. So it looks like it's all working, though we won't know for sure till I squash/merge and we have had a real (not simulated) nightly run.

image

image

mossroy:
Cool! You can squash/merge for me

PR #906, opened 2022-10-24
Update README.md

### LINKED ISSUE #905: Extension no longer works in Firefox 45 ESR - seems to be CSP issue
I was testing #903 in a portable version of Firefox 45 ESR (on Windows), and was puzzled that the extension version I was testing did not seem to have access to the iframe. Accessing the PWA directly works with no problem (in jQuery mode only), but in the extension, access to the iframe is blocked.

I thought it might be because I was testing an unsigned extension, so I installed the signed version from the Mozilla add-ons web site, but this shows the same issue. Typical console.log errors are shown below. This doesn't affect the Firefox OS simulator. It must have something to do with what is considered a secure origin in that version of Firefox, but I'm very surprised we didn't catch this before.

While we shouldn't spend long on this old version, we do claim officially to support Firefox >= 45, so if we can fix (through CSP?) we should.

image

image

Jaifroid:
This textual PR (no code) closes #905 by updating the base Firefox version to >=52, and also modifying some other minor details in the readme.

Additionally, the new default status of Service Worker mode is added to the readme.


Jaifroid:
I have one query for @mossroy. We currently list Firefox OS as a supported platform, which is true in the sense that we test on it (and I aim to continue testing on the simulator). But I wonder if it is time to move it to the "deprecated" section, given that the app in Firefox OS only runs in a now-deprecated mode (JQuery). I don't feel strongly about this, but it woudl probably be logical.

mossroy:
Sure, you can clearly move it to "deprecated".

PR #911, opened 2022-11-09
Changelog for v3.6.0

Jaifroid:
Proposed changelog for v3.6.0.

I have not included the fact that I had to disable end-to-end Unit tests temporarily while we look for a new solution, as it is not a "feature" of the released client.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ Released on *TODO*
 
 * FEATURE: ServiceWorker Mode is now the default, and compatible clients upgrade automatically to this mode
 * FEATURE: On first run after update, the user is informed of the ServiceWorker Mode upgrade status (upgraded or incompatible)
+* COMPATIBILITY: Minimum Firefox version has been raised to >=52, due to lack of full Extension APIs in earlier versions 
 * UPDATE: Detection of active content updated for compatibility with more no-namespace ZIM archives
 * UPDATE: Nightly packages on the download server now include the date in their filenames
 * FIX: Kiwix icon now has an outline so that it is visible against dark OS backgrounds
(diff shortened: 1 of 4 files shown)


mossroy (approved):
Looks good
Maybe it would be worth mentioning the update on minimum Firefox version: as you wish

Jaifroid:
Thanks! Yes, that's a good idea, will do.

mossroy (approved):

mossroy:
Be careful, you commited the removal of "-WIP" suffixes, which might cause issues if you create the tag now

Jaifroid:
Ah, OK, thanks. I'll see what happens, and correct as necessary. Saw this too late....

Jaifroid:
Script appears to have handled this gracefully, but I'm checking. I'm not seeing errors in the Build action, and packages created on download.kiwix.org:

image

mossroy:
Removing these "-WIP" suffixes is one of the steps of the package creation. It is done on several files, not only the ones you changed.
But, as it's currently implemented (using "sed" commands), it probably just silently ignores if this change had already been done.
So it might in fact have worked as expected (but needs to be checked)

Jaifroid:
I'm checking the code - I can see that -WIP is hard-coded in the script, so the version may not have been correctly substituted in all places where it should be for all packages. It probably worked for browser extensions, but I think thee could be issues with the manifests of other packages, so I should probably run the action again after reverting the commit.

This hard-coding of -WIP is probably something that I should try to change in the script at some point, so that it can handle other tag formats. I'll create an issue for that.

Jaifroid:
OK, I've determined that:

* Substitution worked correctly when the tag was created because in those places where I had already replaced the -WIP, the tag was not altered, and where the -WIP was still in place, it was changed;
* We do not handle passing a version number when running the build script from the GitHub actions Workflow Dispatch page. We can build release versions (rather than nightly), but they will be built with a commitid and uploaded to nightly. This should be corrected to allow setting a version explicitly in the Workflow Dispatch.

Bottom line is that the original packages built when I created the tag (and which are still in nightly) have the required numeric-only version number in the correct places.

mossroy:
I saw a (very minor) issue: in the Firefox extension, when switching between the SW and jQuery mode, the version displayed at the top left switches between 3.6 (coming from the commit in this PR) and 3.6.0 (coming from the tag name).

Issue #916, opened 2022-11-14
Remove jQuery hide and show statements for elements selected by ID


Jaifroid:
This is a sub-issue of #367. The vast majority of our usage of jQuery is in the form of simple hide/show statements based on DOM elements selected by ID.

There are two ways of doing this:

1. Simply substitute the jQuery #selector with `document.getElementById('selector').style.display = ''` or `'none'` respectively (this can be done with some carefully crafted regexes as a mass search and replace);
2. Perhaps more elegantly, have a section at the start of app.js that defines variables for each DOM element like `var #selector = document.getElementById('selector');` and then substitute the jQuery selector with `#selector.style.display = ''` or `'none'`.

The disadvantage of 2 is that it is a slower process, and a similar selector block would be needed for each file that uses jQuery methods on the DOM (e.g. `app.js`, `uiUtli.js`). But it would reduce the overall amount of code in the end. Gain compared to 1 wouldn't be noticeable in any of our supported browsers.

Jaifroid:
Fixed in #917.

PR #917, opened 2022-11-14
Replace jQuery for hide and show statements selected by ID

Jaifroid:
This is initial work on #916.

So far, it only deals with simple substitution of jQuery statements with their native DOM equivalents for hide and show selected by ID in `app.js`.


mossroy (approved):
I did not test but it seems to be the good approach to me (bonus point for the regexp ;-) )

Jaifroid:
Because jQuery handles errors (e.g. if an element doesn't exist but we try to set its style property), we need to be sure to test any transient elements that we are hiding or showing if there is any possibility of their being removed from the DOM or not present (yet) in the DOM. There should only be a very small number of these (possibly only the one pushed in https://github.com/kiwix/kiwix-js/pull/917/commits/da80bcf972d42193890fca2fe6bb2185d1065d6d). Needs careful checking.

Jaifroid:
I tested on Firefox OS simulator and IE11 (as well as Edge Chromium), as these are most likely to be affected by removal of jQuery. No problems found.

Issue #928, opened 2022-11-25
Add a Directory Listing object to the Zim Archive metadata for Xapian indices


Jaifroid:
This is a sub-issue of #769. In order to detect whether we need to load libzim for Xapian search, we need to be aware of the existence of a Xapian listing. We currently have `titlePtrPos` and `articlePtrPos` as attributes of the ZIM file object for the Directory Listings that they point to. We should add an appropriate attribute pointing to the Xapian data, which could be simply the pathname, that would only be present if the loaded ZIM archive has a Xapian index, and would be null if there is no Xapian index.

PR #932, opened 2022-11-26
Add fulltext directory listing to archive metadata

Jaifroid:
Completes #928.


Jaifroid:
Tested in Chromium, Firefox and IE11. Also tested on Kiwix JS Windows. It detects the presence of a fullTextIndex and addst the pointer to the ZIM metadata, as well as the size of the index. If these metadata are not available, it lists the path to the index (if it exists).

Jaifroid (commented):
Tested.

Issue #931, opened 2022-11-25
Integrate array of paths returned by Xapian full-text search into the standard title search results


Jaifroid:
This is a sub-issue of #769. Once Xapian search has been launched, we will need to deal with the returned array of paths. Currently, the implemented search in the JS port of libzim only returns paths of articles that contain the searched terms, not snippets or article titles. We already have an algorithm to take the paths, read the Directory Entries, and extract the article titles (which are not always the same as the article path). However, it may be faster to get the titles from the libzim port rather than process the paths the old-fashioned way, if we can expose a way to do that.

This issue should include integrating the search results into the list of articles already found by title search, which should include elimination of duplicates and prioritization. Search results with the searched term in the article title should be prioritized over full-text search that doesn't have the term in the title.

Jaifroid:
Fixed in #935

PR #935, opened 2022-11-27
Integrate full-text search results into title search results

### LINKED ISSUE #930: Launch full-text Xapian search when user undertakes a title search under certain conditions
This is a sub-issue of #769. The main condition would be that the libzim worker has been initialized (#929). However, we could set other conditions, such as initiating the search only if title search has returned fewer than `params.maxSearchResultsSize` in x seconds (i.e. use a timeout and check how many results have been returned with title search).

Jaifroid:
Fixes #930 and #931.


Jaifroid:
Most of the kinks are ironed out, and this appears to be working pretty well now, both in A-namespace ZIMs, and in C-namespace ZIMs (including Zimit archives, although the content of found Zimit articles is poorly formatted in Kiwix JS of course, because only the HTML is loaded with no CSS of JS).

Jaifroid:
Example: Full English Wikipedia with full-text search (A-namespace):

image

Jaifroid:
Beer Stack Exchange (OpenZim Type 1 ZIM with C-namespace):

image

Jaifroid:
And, for fun, Mozilla Development Network, a Zimit-style ZIM (C-namespace):

image

Jaifroid:
Needs thorough testing on older browsers and for null searches before merging.

Jaifroid:
Tests:

* Chromium for Windows - works fine
* Firefox latest for Windows - works fine
* Firefox on Ubuntu - works REALLY SLOWLY, but this may be to do with filesystem emulation on WSL
* Edge Legacy (also UWP app) - BROKEN. The WASM won't compile due to invalid opcodes. Possible solution may be to compile with pthreads disabled -- see https://github.com/emscripten-core/emscripten/issues/13152#issuecomment-753226011
* Chromium on Android - DOESN'T WORK (probably a similar issue with compiling for too high a target)
* Firefox on Android - works fine (but Firefox on Android isn't usable for large archives because it loads ZIM archives into memory before accessing them)

Jaifroid:
Now added better API reporting, and updated the binaries, including source code.

image

Jaifroid:
Tests need fixing.

Jaifroid:
Tests:

* Firefox OS - degrades gracefully to title-only search
* IE11 - same
* Firefox Developer 109 - shows FT index results
* Edge Chromium - same
* Firefox Extension - shows FT index results
* Chromium Extension - shows FT index results

PR #941, opened 2022-12-18
Initiate full-text search earlier

### LINKED ISSUE #940: Launch full-text search in parallel with title search
Although full-text search turns out to be pretty fast, there is room for some more optimization that should be an easy win.

Since it is on a separate browser thread, there is no need to wait for title search to complete before launching the request for full-text search in the Worker. The Worker might as well be getting on with this earlier on. The main thing will be to wait till title-search has completed before *reporting* rather than to wait for it to complete before initiating.

Jaifroid:
Fixes #940.

It would be worth seeing if we can report interim results while waiting for dirEntry lookup promises to resolve.

### DIFF
--- a/www/js/lib/zimArchive.js
+++ b/www/js/lib/zimArchive.js
@@ -252,7 +252,7 @@ define(['zimfile', 'zimDirEntry', 'util', 'uiUtil', 'utf8'],
         var dirEntries = [];
         search.scanCount = 0;
         // Launch a full-text search if possible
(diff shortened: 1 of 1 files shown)


Jaifroid:
I tested reporting each dirEntry promise as soon as it resolves, but it doesn't make enough of a difference and complicates the code. It might be possible to use the [Search API](https://libzim.readthedocs.io/en/stable/6to7.html#searching) to pull fewer results at a time in a more progressive way, but that would be for another PR, perhaps one that includes optional Suggestions.

PR #949, opened 2023-01-02
Remove hard coding of WIP in build script

### LINKED ISSUE #913: Remove hard-coding of -WIP in the build script
`create_all_packages.sh` only works properly if the source code contains a tag formated as x.x.x-WIP in several places (manifests and version numbers). It should not be too hard to make the substutions more agnostic, so that they write the passed version/tag in the right places without needing to find `-WIP`. This would be easier for future developers, and also allow for future changes in tag format as necessary without rewriting the script.

Jaifroid:
Fixes #913.


Jaifroid:
Instead of hard-coding a version number in the script, this gets the values from app.js.

PR #950, opened 2023-01-03
Handle setting version number explicitly in release workflow

### LINKED ISSUE #914: Handle setting a version number explicitly when building the app using Workflow Dispatch
`Release.yml` currently allows the user to set the build type as "nightly" or "release" when dispatching the workflow from the GitHub online UI. But it doesn't allow the user to set a version number explicitly (equivalent to passing a tag to the build script). This would not be difficult to add to the workflow.

Jaifroid:
Fixes #914. This has the dry-run switch added for testing. Remove before merge.

### DIFF
--- a/.github/workflows/Release.yml
+++ b/.github/workflows/Release.yml
@@ -18,6 +18,9 @@ on:
         description: Do you wish to build release or nightly?
         required: false
         default: 'nightly'
+      version:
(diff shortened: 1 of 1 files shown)


Jaifroid:
Works as expected. Now removed dryrun switches.

PR #951, opened 2023-01-03
Publish some installation packages to release

### LINKED ISSUE #915: Add/amend script to publish some installation packages directly to release on download.kiwix.org
So that we do not need to copy manually packages that are not Store-signed to the release folder, this could easily be done as part of the build/publish workflow. Currently only the Ubuntu touch and Firefox OS packages would be included. The Chrome/Edge and Firefox browser-extension packages need to be certified by the respective Stores first.

Jaifroid:
Fixes #915. I've also added a switch to simulate a CRON_LAUNCHED job. The script is now more verbose, explaining what it's doing and instructing the developer on location of unsigned packages.

### DIFF
--- a/scripts/create_all_packages.sh
+++ b/scripts/create_all_packages.sh
@@ -4,9 +4,10 @@ echo -e "\nBASEDIR is $BASEDIR"
 cd "$BASEDIR"
 
 # Reading arguments
-while getopts tdv: option; do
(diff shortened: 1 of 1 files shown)


Jaifroid:
Everything seems to work fine with the local simulation. But the proof will be in the pudding once we do the actual release.

PR #962, opened 2023-02-02
Updated jquery .on event listeners to their native JS equivalent

Abhishek-1337:
Updated jquery .on event listeners to native JS code. Let me know if there's anything I forgot before requesting pull from main branch.

### DIFF
--- a/www/js/app.js
+++ b/www/js/app.js
@@ -234,9 +234,10 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
     $(window).resize(resizeIFrame);
 
     // Define behavior of HTML elements
+    var prefixElement = document.getElementById('prefix');
     var searchArticlesFocused = false;
-    $('#searchArticles').on('click', function() {
-        var prefix = document.getElementById('prefix').value;
+    document.getElementById('searchArticles').addEventListener('click', function() {
+        var prefix = prefixElement.value;
         // Do not initiate the same search if it is already in progress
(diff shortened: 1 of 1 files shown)


Jaifroid (changes requested):
Thank you for this PR, @Abhishek-1337. I did a quick manual test and ran into this error when searching. Unfortunately, only full-text search runs, and the title search is blocked due to this error. When getting elements by querySelector, if those elements are volatile and could return null, then you need to test first before adding an event listener. In most cases you can rely on the element being present, but sometimes, like with search, it's possible that null will be returned, so you need to test first. This is something JQuery did automatically, but when we remove it, we have to do it manually.

image

Jaifroid:
Once you've fixed this, please test thoroughly as outlined in https://github.com/kiwix/kiwix-js/blob/main/CONTRIBUTING.md, and confirm which tests you have done. Don't forget to test thoroughly in both jQuery and Service Worker modes.

Abhishek-1337:
@Jaifroid I am having trouble reproducing this. Can you guide me through this, I just want to check why is it happening?

Jaifroid:
@Abhishek-1337 It happens during search if title search doesn't produce any result (because the term searched for isn't in the title list). This means that the query returns an empty result. You have to test for this, and not attempt to add an event listener if there is nothing to add the listeners to! Does that help?

Abhishek-1337:
@Jaifroid Got it. I was trying too hard to find it but its the obsevation that is useful. Thanks for clarification.

Jaifroid:
There doesn't seem to be any progress on this PR. I'm closing, but feel free to request re-opening.

PR #963, opened 2023-02-05
Use latest GitHub workflow actions

kelson42:
This was recommended to avoid warnings in the CI because old actions workflow rely on Node.js 12 which has been meanwhile deprecated


Jaifroid (approved):
Looks good, thanks!

PR #967, opened 2023-02-22
Add option to disable drag-and-drop

### LINKED ISSUE #966: Need an option to disable drag-and-drop
I need an option to disable drag-and-drop. That feature keeps getting activated on my computer when I already have a zim open and am reading it. It causes me to lose my place and distracts me when I have to find where I was.

On my computer I have the following problems:
- '\0' is typed into browser windows when tabbing to switch from Konsole
- Alt-tab is sometime interpreted as starting a drag-and-drop
- Drag-and-drop/paste sometimes breaks on Discord/GitHub/Brightspace with ~~`Mozilla/5.0 (X11; Ubuntu; Linux x86_64; rv:109.0) Gecko/20100101 Firefox/109.0`~~ `Mozilla/5.0 (X11; Ubuntu; Linux x86_64; rv:109.0) Gecko/20100101 Firefox/110.0` until I detach the tab into a new window
- Copying text on another computer paired over KDE Connect causes kiwix-js drag-and-drop to activate
- I like to click on text when reading, and accidentally dragging it sends me to the configure page
- Copying text from the zim succeeds, but immediately sends me to the configure page
- Locking-then-unlocking the screen causes the same problem
- These problems don't occur when I have my browser window switched to another tab

It doesn't happen on a fresh browser windows, but I have found that the following reproduces it reliably:
1. Open another desktop window such as Konsole
2. Open Firefox (on `kwin-wayland 4:5.27.0-0ubuntu1~ubuntu22.10~ppa1`, and the problem does not occur if `WAYLAND_DISPLAY` is unset)
3. Go to https://en.wikipedia.org/wiki/Main_Page
4. Select "Welcome"
5. Drag it anywhere
6. Open kiwix-js in Firefox
7. Open a zim
8. See that we are on the "Home" tab
9. Press Alt-tab twice
10. See that we are on the "Configure" tab

There are a lot of things currently broken with my setup, but other websites work. Therefore, I would like an option to disable drag-and-drop so that kiwix-js would work like other websites.

Jaifroid:
Fixes #966.

### DIFF
--- a/www/index.html
+++ b/www/index.html
@@ -626,6 +626,13 @@
                                         experienced users)
                                     </label>
                                 </div>
+                                <div class="checkbox">
+                                    <label title="In some browsers on some platforms, drag-and-drop may malfunction and make it difficult to select text, and other operations. Disable it here if it is causing issues." >
+                                        <input type="checkbox" name="disableDragAndDrop"
+                                            id="disableDragAndDropCheck">
+                                        <strong>Disable drag-and-drop</strong> (in case it is causing anomalies)
+                                    </label>
+                                </div>
                                 <div class="checkbox" id="bypassAppCacheDiv">
                                     <label
                                         title="WARNING: Leaving this checked will prevent offline usage of the PWA. Setting will clear all existing Cache API caches, but assetsCache will be used unless also disabled above. For testing new code with the PWA.">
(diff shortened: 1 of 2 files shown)


Jaifroid:
Appears to work reliably. Tested in Chromium and IE11, and OP tested in Firefox on Ubuntu.

Issue #969, opened 2023-02-28
Adding `http-server` run script


ankur12-1610:
Instead of running `http-server` command just add it to scripts, for eg:
![Screenshot from 2023-02-28 21-12-21](https://user-images.githubusercontent.com/76884959/221940991-461c2842-71f7-4ccd-a4a8-feb5f643708b.png)

or `npx http-server .`  will work as well.

Jaifroid:
@ankur12-1610 This seems like a good idea, thanks! You'd also have to add http-server as a developer dependency.

ankur12-1610:
On it!

satyampsoni:
Hey @ankur12-1610. Have you completed it?
I would like to work on it too.

PR #971, opened 2023-03-02
Added `http-server` script

ankur12-1610:
Fixes: #969

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -13,8 +13,7 @@ Please follow these guidelines when contributing:
   supported via a polyfill;
 - Do not prettify code you are not working on;
 - You must test your code yourself before asking for review, like this:
-  - set up a local Web server (we recommend Node/NPM's [http-server](https://www.npmjs.com/package/http-server));
-  - serve the root directory of the repository (e.g. `http-server .`);
+  - set up a local Web server (we recommend Node/NPM's [http-server](https://www.npmjs.com/package/http-server)) by running `npm run serve` in the root of this repo;
   - in a browser, navigate to the URL of the main `index.html` (e.g. http://localhost:8080/www/index.html);
   - manually test your fix in at least Firefox and Chromium (Edge or Chrome), ideally also in IE11 or in "IE Mode" in Edge;
   - be sure to test your fix in both "JQuery" mode and "Service Worker" mode (see Configuration);
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -13,7 +13,7 @@ Please follow these guidelines when contributing:
   supported via a polyfill;
 - Do not prettify code you are not working on;
 - You must test your code yourself before asking for review, like this:
-  - set up a local Web server (we recommend Node/NPM's [http-server](https://www.npmjs.com/package/http-server)) by running `npm run serve` in the root of this repo;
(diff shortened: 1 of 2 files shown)


ankur12-1610:
@Jaifroid kindly review :)

Also, we might need to update the documentation as well: [this](https://github.com/ankur12-1610/kiwix-js/blob/http-server/CONTRIBUTING.md#:~:text=serve%20the%20root%20directory%20of%20the%20repository%20(e.g.%20http%2Dserver%20.)%3B) is the link of the same

ankur12-1610:
> @ankur12-1610 Thank you for this PR. I tested by running `npm install --save-dev` and then npm start, and the server does indeed start. However, I already have http-server installed globally. Do you know if this would work when it is not installed? I would have thought we'd need to have it listed as a developer dependency in the json, but I haven't tested that scenario. It could be tested in Windows Sandbox or another clean testbed.
> 
> Regarding your suggestion to update the documentation, I think that would be part of this PR, but should be done once the above question is tested.
> 
> Finally, I'm not sure `npm start` is the best choice because it is the standard script for launching an Electron app, and Electron is enabled downstream in kiwix-js-windows repository (which depends on this repository).
> 
> Perhaps `serve` would be better, and then the start command would be `npm run serve`. What do you think?

Hey @Jaifroid, I've tested it in MacOS and it works perfectly fine in it. I'll it in try other operating systems as well.

Regarding the dependency, we already have included `http-server` in package.json.
> Screenshot from 2023-03-06 02-26-24

Also, updated the PR now it is `npm run serve` and I've updated the docs as well.

The only final step left is verification :)

CONTRIBUTING.md
```diff
@@ -13,8 +13,7 @@ Please follow these guidelines when contributing:
   supported via a polyfill;
 - Do not prettify code you are not working on;
 - You must test your code yourself before asking for review, like this:
-  - set up a local Web server (we recommend Node/NPM's [http-server](https://www.npmjs.com/package/http-server));
-  - serve the root directory of the repository (e.g. `http-server .`);
+  - set up a local Web server (we recommend Node/NPM's [http-server](https://www.npmjs.com/package/http-server)) by running `npm run serve` in the root of this repo;
```
Jaifroid, line 16:
```suggestion
  - set up a local Web server: we recommend Node/NPM's [http-server](https://www.npmjs.com/package/http-server)), which you can launch by running `npm run serve` in the root of this repository;
```

Jaifroid (approved):
@ankur12-1610 Thank you, this looks good. I tested launching the server also, and it seems to work fine.

Do you know whether http-server gets installed if a user hasn't installed it globally? You had originally mentioned perhaps needing to run it with npx. I wasn't able to test this because I have it installed globally.

I have one suggestions below. You can use the GitHub buttons to commit the suggestion if you think it's OK.

ankur12-1610:
> @ankur12-1610 Thank you, this looks good. I tested launching the server also, and it seems to work fine.
> 
> Do you know whether http-server gets installed if a user hasn't installed it globally? You had originally mentioned perhaps needing to run it with npx. I wasn't able to test this because I have it installed globally.
> 
> I have one suggestions below. You can use the GitHub buttons to commit the suggestion if you think it's OK.

Yeah I tested it without installing http-server globally in a new environment, npm install does the work for the user so using npx isn't necessary.

Sure I'll do it👍

Jaifroid:
Great! So I think this can be squashed and merged. Just confirm that it's ready and I'll press the button (I assume you don't have access to the Squash and merge button on this page).

Issue #972, opened 2023-03-03
Recent wiktionary archives break the iframe and destroy the app's controls


danielzgtg:
*EDIT by @Jaifroid: Discussion of issue should be in https://github.com/openzim/mwoffliner/issues/1803. This issue is to consider implementing a local patch.*

Kiwix-js no longer works with `wiktionary_en_all_maxi_2023-02.zim`. It was working fine with `wiktionary_en_all_maxi_2022-09.zim`. Now it just redirects to some other URL for the main page that doesn't have the search bar.

Editing the URL manually doesn't work around this. It goes to "404 Not Found nginx." when I replace `Wiktionary%3AOffline` with `wiki`. I also shows "404 Not Found" when I press the back button. My browser keeps on showing the page icon as the browser's loading animation for a long time before it switches to no icon at all. Pressing back at the original page goes back to the proper app with the search bar, but then it immediately redirects again. I need to close the tab and open another one to get something usable.

This affects ServiceWorker mode only. There is no redirect problem in JQuery mode and I am able to open articles. However, I do not like JQuery because it doesn't support my browser's CSS :visited, bfcache, nor Dark Reader extensions. This does not affect Android either.

I think the problem is that kiwix-js ServiceWorker mode is failing to intercept the redirect, while the other modes/apps were able to. Here is a video of the problem:

https://user-images.githubusercontent.com/25646384/222708434-9b531e0b-018b-48fe-a2c9-a0f4f543ac3a.mp4

Jaifroid:
@danielzgtg Thank you for the error report. I'll investigate!

Jaifroid:
Oops, I can't transfer the issue across orgs (from Kiwix to OpenZIM), so I'll just copy the issues.

Jaifroid:
Closing in favour of https://github.com/openzim/mwoffliner/issues/1803

danielzgtg:
Here is a Tampermonkey userscript to work around this:

```js
// ==UserScript==
// @name         Kiwix fix rogue redirect
// @namespace    http://tampermonkey.net/
// @version      0.1
// @description  try to take over the world!
// @author       You
// @match        https://moz-extension.kiwix.org/*
// ==/UserScript==

(function() {
    'use strict';
    const iframe = document.querySelector('iframe');
    iframe && (iframe.sandbox = "allow-same-origin allow-scripts");
})();
```

Jaifroid:
[...]

Issue #924, opened 2022-11-16
Downloading epubs from Gutenberg broken in IE11


Jaifroid:
I think I fixed the same issue in KJSWL a couple of months ago, and I thought it was an issue with the code there, but it turns out to affect this code too. This has nothing to do with the removal of jQuery. I tested this both on that branch and on master, with the same effect.

Jaifroid:
Probably related to https://github.com/openzim/gutenberg/issues/181. The epub MIME type is incorrect in version 1 Gutenberg ZIMs.

Jaifroid:
Unfortunately, this regression has returned in the latest Gutenberg ZIM types. Moreover, #987 exacerbates the issue because in these ZIMs it's impossible even to see the download button due to the fonts not rendering.

PR #976, opened 2023-03-09
Add sandbox to iframe

Jaifroid:
This addresses (but does not fix fully) #753 and #972. A full fix will require a separate PR implementing the Content Security Policy sandbox directive delivered via HTTP headers through the Service Worker.

### DIFF
--- a/www/index.html
+++ b/www/index.html
@@ -6,7 +6,7 @@
         <title>Kiwix</title>
         <meta name="description" content="Offline Wikipedia reader">
         <meta name="viewport" content="width=device-width, initial-scale=1.0">
-        <meta http-equiv="Content-Security-Policy" content="default-src 'self' data: https://download.kiwix.org https://master.download.kiwix.org https://moz-extension.kiwix.org https://kiwix.github.io 'unsafe-inline' 'unsafe-eval'; frame-src 'self' moz-extension:; object-src 'none';">
+        <meta http-equiv="Content-Security-Policy" content="default-src 'self' data: https://download.kiwix.org https://master.download.kiwix.org https://moz-extension.kiwix.org https://kiwix.github.io 'unsafe-inline' 'unsafe-eval'; frame-src 'self' moz-extension: chrome-extension:; object-src 'none';">
         <meta name="referrer" content="none">
         <!--
         Kiwix (offline Wikipedia reader) - HTML5/Javascript version
--- a/www/index.html
+++ b/www/index.html
@@ -7,6 +7,7 @@
         <meta name="description" content="Offline Wikipedia reader">
         <meta name="viewport" content="width=device-width, initial-scale=1.0">
         <meta http-equiv="Content-Security-Policy" content="default-src 'self' https://download.kiwix.org https://master.download.kiwix.org 'unsafe-inline' 'unsafe-eval'; img-src 'self' data:; frame-src 'self'; object-src 'none';">
+        <meta name="referrer" content="none">
         <!--
         Kiwix (offline Wikipedia reader) - HTML5/Javascript version
         Home page : https://www.kiwix.org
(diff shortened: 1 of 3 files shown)


Jaifroid:
This successfully fixes #972. See screenshot below:

image

Jaifroid:
Works on Wikipedia in both jQuery and SW modes. Works in PhET, doesn't prevent any scripts running.

Needs testing in extension context.

Jaifroid:
Unfortunately, the sandbox prevents Gutenberg PDFs from loading in the iframe. Strange, even though it's the same origin, the browser perceives it as cross-origin and blocks it.

Jaifroid:
Added code to handle PDF download, due to conflict between sandbox and browser-provided PDF viewer. This simply has to be opened in a new tab.

www/index.html
```diff
@@ -686,7 +687,7 @@ <h3>Expert settings</h3>
                         if your platform supports it. &nbsp;[<a id="stop" href="#expertSettingsDiv" class="alert-link">Permanently hide</a>]
                     </div>
                 </div>
-                <iframe id="articleContent" class="articleIFrame" src="article.html"></iframe>
+                <iframe id="articleContent" class="articleIFrame" src="article.html" referrerpolicy="same-origin" sandbox="allow-same-origin allow-scripts allow-modals allow-forms allow-popups allow-downloads"></iframe>
```
danielzgtg, line 690: Which zim requires `referrerpolicy="same-origin"`? As compared to `no-referrer`

Jaifroid: I was thinking of this as a way to block any referrer calls outside of the app's context / domain, but you're right that no-referrer might be better. I can't think of any situation in which we need to allow referral. Thank you, I'll test.

Jaifroid:
I'm going to set the development PWA (kiwix.github.io/kiwix-js/www/index.html) to this branch so I can test these security settings in SW mode in a Firefox temporary extension.

Jaifroid:
I had to unblock communication from the local extension to the moz-extension PWA. Now it works OK going that way, but I also need to unblock messaging back from the moz-extension PWA to the local extension, otherwise on each launch of the extension it believes it failed to launch last time (this is because the confirmation of successful launch that is sent back to the local extension effectively blocked by the CSP). Good think I tested this process manually (it hadn't occurred to me as a potential gotcha)!

Jaifroid:
Now working correctly in Mozilla extension, and two-way communication re-enabled between the PWA and the local extension.

Jaifroid:
When downloading content in IE11, e.g. an epub from a Gutenberg ZIM, we get "Navigation to the webpage was cancelled". I guess this is because we're attempting to navigate to a blob: URL of the epub, which is supposed to trigger the download, but is being blocked by the CSP. However, see #924, which indicates this was already broken, so may not be the CSP. Worth seeing if it's something simple, though.

image

Jaifroid:
OK, I've determined the IE11 issue above is not related in any way to this PR (I used exactly the same sandbox and CSP on the code at KJSWL, and download from Gutenberg ZIMs works fine there).

Jaifroid:
Tests:

- [ ] IE11 - working fine (except unrelated issue above) from file:// URL and from localhost
- [ ] Firefox extension - working fine, including transition from local to remote extension and back
- [ ] Chrome extension
- [ ] Firefox OS
- [ ] Edge Chromium
- [ ] Edge Legacy
- [ ] Safari on iOS
- [ ] Safari on macOS
- [ ] Old Icecat

Jaifroid:
Broken in chrome-extension. Strange, because the app is running from chrome-extension: URI, so should be allowed as same-origin:

```
app.js:1587 
        
       Uncaught DOMException: Blocked a frame with origin "chrome-extension://damiliajfhgmjjkakekojccjaabdgppl" from accessing a cross-origin frame.
    at iframeArticleContent.onload (chrome-extension://damiliajfhgmjjkakekojccjaabdgppl/www/js/app.js:1587:60)
```

Jaifroid:
Explicitly allowing `chrome-extension:` protocol as an iframe src fixes the issue, even though it should be same-origin:

* Working in Chrome 49 (this version doesn't support Service Workers, and doesn't recognize some of the sandbox attributes, but loading of articles isn't blocked);
* Working in Chromium 110 as an unpacked extension.

Jaifroid:
Seems ready to merge.

PR #978, opened 2023-03-12
Fix missing full text search cancellation

### LINKED ISSUE #977: Search results list fails to close when typing quickly
The search results sometimes don't close after navigating to the page. This occurs mainly when performing actions quickly.

# Steps

1. Open a zim
2. Type an article into the search box. Do so quickly enough so that the results don't pop up before you finish typing
3. Spam press the enter key
4. Repeat steps 2-3 until the bug occurs

I am running locally off of fdb4238c5f4e17fdc53dd23b2a93cbc3e6283ad8, but this occurs on the deployed kiwix-js 3.7.2 too.

# Workaround

`header#top:has(>form#formArticleSearch>div.container>div.row>span.col-10>span.input-group>input#prefix:placeholder-shown)+article.view-content>div.container#articleListWithHeader { display: none !important }` should work on Chrome. Sadly, I'm on Firefox which hasn't enabled the `:has` selector because they haven't implemented this selector's support for pseudo-classes.

Unfocusing the input box by clicking somewhere else on the page might work. It's not very convenient because I want to use applications with my keyboard only when I need to work efficiently.

# Expected Behavior

The search results list should close after navigating to each page. It should be closed whenever the search box is empty.

# Actual Behavior

The bug does not occur all the time. In the video below, it happens after 1:00.

https://user-images.githubusercontent.com/25646384/224547454-c48c63e6-a5a2-42c9-9d18-b0011b11e65a.mp4

https://user-images.githubusercontent.com/25646384/224549100-6f65fa96-c78b-483e-9760-ad219f67b96f.mp4

danielzgtg:
This fixes the corruption of the `search.status` variable due to an illegal state transition `cancelled -> complete` in the full text search callback. Before it was causing the search results list to show around 10% of the time when I press enter quickly, and 100% if I cause the full text search to be delayed artificially. After this PR, the search results list cancellation bug for the full text search occurs 0% of the time.

Closes #977

### DIFF
--- a/www/js/lib/zimArchive.js
+++ b/www/js/lib/zimArchive.js
@@ -253,6 +253,10 @@ define(['zimfile', 'zimDirEntry', 'util', 'uiUtil', 'utf8'],
         search.scanCount = 0;
         // Launch a full-text search if possible
         if (LZ) that.findDirEntriesFromFullTextSearch(search, dirEntries).then(function (fullTextDirEntries) {
+            // If user initiated a new search, cancel this one
+            // In particular, do not set the search status back to 'complete'
+            // as that would cause outdated results to unexpectedly pop up
+            if (search.status === 'cancelled') return callback([], search);
             dirEntries = fullTextDirEntries;
             search.status = 'complete';
             callback(dirEntries, search);


Jaifroid (approved):
Excellent, many thanks for this PR! Let me know if you're happy for me to press the squash/merge button.

danielzgtg:
Ready!

Issue #755, opened 2021-08-22
Migrate to manifest v3 for browser extensions


mossroy:
Chrome introduced a version 3 of the extensions manifest format : https://developer.chrome.com/docs/extensions/mv3/intro/
Edge is following, as it's now based on Chromium : https://blogs.windows.com/msedgedev/2020/10/14/extension-manifest-chromium-edge/
Firefox will implement it too, with some differences (because some features of v3 are controversial), see https://blog.mozilla.org/addons/2021/05/27/manifest-v3-update/

It's available since Chromium 88, and will be available in Firefox at the end of 2021.
The manifest v2 (that we currently use) is still supported everywhere. It will eventually be deprecated and removed, but we have some time (I did not read any deadline yet)

Anyway, we will have to switch to it, maybe in 2022.

Base on what I read, and the fact that we are an offline extension with no permissions requested, I suppose the switch should be easy.
But it raises the question of backward compatibility : an extension with manifest v3 will probably not run on "old" versions of Firefox and Chromium that only support v2. If the switch is as easy as expected, and if we believe it's necessary, we might consider generating 2 versions of our extensions : one with a manifest v3, and one with a manifest v2 (with the same code)

Jaifroid:
We should probably test v3, but stick with v2 for as long as it is allowed, in order to support older versions of browsers for as long as possible.

Jaifroid:
There is now a timeline in Chrome for sunset of manifest v2: https://developer.chrome.com/docs/extensions/mv3/mv2-sunset/. To summarize, by June 2023 "All existing Manifest V2 items with visibility set to Public at that time will have their visibility changed to Unlisted", and by January 2024 "Chrome Web Store stops accepting updates to existing Manifest V2 extensions".

Mozilla is adopting manifest v3, but it seems it is going to keep many aspects of manifest v2 due to developer discontent with the limitations of the new manifest. I noted that their first Firefox Developer version of manifest v3 won't actually support Service Workers as the background script, so it looks like they are taking a staged approach: "In Firefox, we have decided to support Event Pages in MV3, and our developer preview will not include Service Workers (we’re continuing to work on supporting these for a future release)" - https://blog.mozilla.org/addons/2022/05/18/manifest-v3-in-firefox-recap-next-steps/.

[...]

PR #984, opened 2023-04-16
Experiments with extension manifest v3 for Chromium

Jaifroid:
This PR is for demonstrating the code used to run a Chromium extension with the webextension manifest v3. See #755. The code is experimental. Not for merging.


Jaifroid:
As I have incorporated some innovations and bugfixes from here into master, I have had to rebase and force-push this branch.

We are at the stage where we can now fully support Chromium extensions using MV3, with the remote PWA workaround we have already been using for Firefox. 

The problem will be that we will have to keep supporting MV2 in Firefox because FF hasn't yet enabled the background Service Worker in MV3. But even then, it will only be for the latest browser, so we will have to keep an MV2 version, arguably also for Chromium browsers, as only Chromium >88, or possibly >=90 is supported.

Jaifroid:
Ready to be merged, pending decision on #1017.

Jaifroid:
Code now builds nightly versions of both MV2 and MV3 extensions for Chromium, while building the Firefox extension only as MV2 for now. Until Firefox starts to support Service Workers as background scripts, there is no point trying to change it to MV3 (and especially as there is no shutdown date for MV2 on FF as compared to Chrome).

PR #989, opened 2023-04-21
Fall back to MIME type when downloading doc in jQuery mode

### LINKED ISSUE #858: When downloading a PDF, check the mime type as well as the extension of the file
We currently check for downloadable files by relying on the extension of the file. However, the Zimit port in Kiwix JS Windows has revealed a flaw (I think): we should also check the mimetype of the dirEntry. We can have PDF files that are just links to data, not apparently a file at all (at least this is revealed in the cheatography.com Zimit ZIM). Both SW mode and jQuery mode seem to fail in this situation.

This issue needs more investigation, but if my supposition is correct, then a fix should improve Kiwix JS more broadly, not just Zimit support. I have opened https://github.com/kiwix/kiwix-js-windows/issues/252 to investigate.

Jaifroid:
Fixes #858 for jQuery mode. This does not touch ServiceWorker mode, as this should be handled natively in that mode. We cannot rely only on the filename extension, but must also consider the MIME type especially when there is no extension.

### DIFF
--- a/www/js/app.js
+++ b/www/js/app.js
@@ -2116,12 +2116,13 @@ define(['jquery', 'zimArchiveLoader', 'uiUtil', 'settingsStore','abstractFilesys
     function goToArticle(path, download, contentType) {
         document.getElementById('searchingArticles').style.display = '';
         selectedArchive.getDirEntryByPath(path).then(function(dirEntry) {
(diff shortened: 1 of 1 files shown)


Jaifroid:
Also fixes #924.

PR #1000, opened 2023-05-05
Add PWA ServiceWorker mode for Chromium extension

### LINKED ISSUE #992: Consider adding the option to use the PWA in SW mode inside the Chrome extension (as prerequisite for Zimit support)
Due to the continued restriction on inline JS, which is not alleviated by [MV3 (Cheomium extension manifest v3) ](https://github.com/kiwix/kiwix-js/issues/755), I am considering using the same mechanism in the Chromium extension as we use in the Firefox extension for ServiceWorker mode, i.e. send the user to the offline-first moz-extension PWA.

Although we can fix the use of inline JS in the openZIM-produced scrapers (see #865), we will never be able to ban use of inline JS in Zimit ZIMs, as they come from "wild" and unpredictable sources.

As a prerequisite for supporting Zimit (upcoming Hackathon), it seems necessary to have more flexibility. This means that the extension, like the Mozilla extension, would effectively be a "shell" that redirects the user to a Progressive Web App, which is much more flexible than an extension.

Although I have #986, which could provide that flexibility, it's a major re-engineering effort, and I'm not sure it'll provide the flexibility even then.

@kelson42 Any advice?

Jaifroid:
Fixes #992.

### DIFF
--- a/www/index.html
+++ b/www/index.html
@@ -584,14 +584,14 @@
                                     <label title="This mode requires that the browser or framework be capable of installing a Service Worker. It works by intercepting the browser or framework's Fetch calls and supplying the requested content from the ZIM.">
                                         <input type="radio" name="contentInjectionMode" value="serviceworker"
                                             id="serviceworkerModeRadio">
-                                        <strong>ServiceWorker</strong> (default, supports archives with dynamic content and inline JavaScript)
+                                        <strong>ServiceWorker</strong> (<i>recommended</i>, fully supports dynamic content and inline JavaScript)
                                     </label>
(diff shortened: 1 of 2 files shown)


Jaifroid:
This PR takes only the ServiceWorkerLocal code from #984, not the Manifest v3 code (which we are not ready to merge). This PR is necessary preparation for enabling Zimit support and for the eventual shift to Manifest v3 for modern browsers.

Jaifroid:
I've just noticed this is PR number 1000! That's kind of... scary. 🤔

Issue #1014, opened 2023-06-12
Fix many stylistic issues identified by ESLint


Jaifroid:
I have added an ESLint configuration to the repository.

It would be worth having some dedicated PRs to address the main styling issues it identifies.

PR #1016, opened 2023-06-12
Fix some ESLint issues in app.js

Jaifroid:


Jaifroid:
Contributes towards fixing #1014.

Issue #907, opened 2022-11-04
Add cross-browser end-to-end testing with BrowserStack


Jaifroid:
As a result of #894 and #895, end-to-end tests with SauceLabs have been disabled. These need to be replaced with a more modern configuration. While the longer-term aims are outlined in https://github.com/kiwix/kiwix-js/pull/895#issuecomment-1303104400, in the short term I aim to add an Action to run our end-to-end tests on BrowserStack, given that they have very thorough documentation of how to set this up. This would address the long-standing issues in #737 as well.

Krinkle:
Fwiw, I'm quite happy with BrowserStack in several projects as well, numerically more often than Sauce even.

If you experience flakiness that seems specific to their hardware for some reason, that's worth considering I'd say. It's also possible that it's something else however. I'm not saying this to defend a vendor, but rather to potentially save you time.

I'll comment further at https://github.com/kiwix/kiwix-js/issues/737

Jaifroid:
Now that I have been able to re-enable Unit testing of the source app with GitHub Actions and TestCafé in the ES6 modules branch, it should be relatively easy to begin porting our end-to-end UI tests to run on the bundled version of the app on BrowserStack. See #1025.

Jaifroid:
Closed in favour of #1041.

Issue #879, opened 2022-08-04
Test current kiwix-js with very old Firefox/Chrome browsers


mossroy:
We claim to support Firefox >=45 and Chrome >=58, and our CI tests a scenario on these versions.

However, when working on #878, I did not always manage to run kiwix-js properly on old versions Firefox.
For example, on Firefox 45.0.2 (on Ubuntu 16.04), it does not manage to switch to the PWA for the ServiceWorker mode (maybe we use some unsupported syntax there).
I also see errors in the console regarding communication between the iframe and its surrounding document.

This should be investigated (with Chromium/Chrome 58, too) to have a status.

If it's easy to fix, let's fix these problems. Else we could consider adjusting our support policy.

NB: I regularly test on Firefox OS (which uses Gecko 44). As Firefox OS does not support ServiceWorkers, it's normal that I did not see the issue with the PWA switch. However, I do not see the errors regarding iframe communication on Firefox OS either...

Jaifroid:
Interesting. I guess we didn't test enough on these early versions. Although I have a vague recollection of some remarks on a very early Firefox having non-stable support for Service Workers? Or was that when we were testing IceCat versions?

If there's a quick syntax fix, I'm all for doing that. Communication between PWA and the extension could indeed depend on it having been enabled by a specific Firefox version. It might have been more locked down in the "distant" past. It would need debugging.

456535688765:
can you assign this issues to me

Jaifroid:
@456535688765 Happy to do so!

Jaifroid:
I guess the outcome of this issue would be a PR changing our documentation, as the automated tests are not currently functioning well. However, if you have experience setting up automated tests, then that could be a possibility (we would use Browser Stack in that case). If you will undertake manual testing, that's fine too.

Jaifroid:
It's currently assigned. @456535688765, any news? if the assigned person is no longer interested, I can reassign.

PR #1025, opened 2023-06-18
Use native es6 modules instead of require js

Jaifroid:
Initial working version of native ES6 module. No build process yet.


Jaifroid:
Some tests re-enabled.

Jaifroid:
Build configuration (with rollup and babel) now added.
Also vite dev and production server (but not fully functional yet).

Jaifroid:
This is getting close. Just a few more things to do:

* Expose the built app in gh-pages in the dist folder
* Update documentation for contributors and devs
* Create an issue to adapt our UI tests to run on Browser Stack

I'm pleased that we can now run QUnit tests (in modern browsers) using testcafe on GitHub Actions. These tests run in the unbuilt version of the app, i.e., using ES6 modules. They are not suitable for running in the built version of the app, as they only use small parts of the code which cannot be isolated in the built app. This means that Unit tests will only work in modern browsers.

However, UI tests will be able to run in the built app on all supported browsers.

Jaifroid:
I reduced the release of fontawesome to a version supported by IE11, to avoid some nasty pollyfills that weren't working properly.

There remains one issue: IE11 isn't correctly using webpHero polyfill for WebP images, so they are not being transformed.

I know this is solvable, because it works with the nearly identical setup in Kiwix JS Windows.

Jaifroid:
Fixed loading of webP images in IE11. However, Firefox <= 52 (ESR edition) is bugging out with a "file not found" error on attempting to load any article even in mwOffliner ZIMs.

Jaifroid:
But it works fine in the Firefox OS simulator, which runs on Firefox 58.

Jaifroid:
After careful testing, it's clear that the issue with Firefox compatibility is not with this PR, but a longstanding one as stated in #879. I have now tested carefully, and have updated the Compatibility section of the Readme to make clear that support for Firefox in SW mode is FF >= 56 (non-ESR version). We also support at least 52-56 and 58 ESR in jQuery mode only (hence support for Firefox OS). This info has been added.

Jaifroid:
Last things:

- [x] Expose the built app in gh-pages in the dist folder
- [x] Update documentation for contributors and devs
- [x] Create an issue to adapt our UI tests to run on Browser Stack - see #907
- [x] Fix IE11 issue with fontawesome
- [x] Fix IE11 not using webpHero correctly
- [x] Fix file not found error loading any article in Firefox <= 52
- [x] Build the production version and upload that one to the docker container
- [ ] Ensure the build scripts use the distribution version
- [ ] Fix regression with useCanvasElementsWorkaround caused by new way of loading webpHero
- [ ] Add new packages to acknowledgements

Jaifroid:
Everything appears to be working. Proof will be in the daily pudding. 😉. Merging now.

PR #1035, opened 2023-07-25
Convert jQuery .on event listeners to their native JS equivalents

RG7279805:


www/js/app.js
```diff
@@ -328,56 +328,56 @@ $('#prefix').on('keydown', function (e) {
...
+document.getElementById('btnRescanDeviceStorage').addEventListener('click', function () {
     searchForArchivesInStorage();
 });
 // Bottom bar :
-$('#btnBack').on('click', function () {
+document.getElementById('btnBack').addEventListener('click', function () {
     history.back();
     return false;
 });
 $('#btnForward').on('click', function () {
```
RG7279805, line 365:
```js
document.getElementById('btnHomeBottom').addEventListener('click', function () {
    document.getElementById('btnHome').click();
    return false;
});

```

I converted this jquery event into the above vanilla JS not sure why that is not working even though other events are working if you can suggest something that would be good else i will look into this on my own

www/js/app.js
```diff
@@ -468,9 +468,11 @@ $('#btnAbout').on('click', function () {
     setTimeout(resizeIFrame, 400);
     return false;
 });
-$('input:radio[name=contentInjectionMode]').on('change', function () {
-    // Do the necessary to enable or disable the Service Worker
-    setContentInjectionMode(this.value);
+document.querySelectorAll('input[name="contentInjectionMode"][type="checkbox"]').forEach(function (element) {
```
Jaifroid, line 471: You will need to check that this transpiles correctly for use with IE11, which doesn't support `.forEach` on DOM Elements. Hopefully Babel will handle it, but please check the compiled code by running `npm build-src` and looking in the `bundle.js` code.

www/js/app.js
```diff
@@ -512,22 +514,30 @@ document.getElementById('disableDragAndDropCheck').addEventListener('change', fu
...
     });
 });
-$('input:checkbox[name=hideActiveContentWarning]').on('change', function () {
-    params.hideActiveContentWarning = !!this.checked;
-    settingsStore.setItem('hideActiveContentWarning', params.hideActiveContentWarning, Infinity);
-});
-$('input:checkbox[name=showUIAnimations]').on('change', function () {
-    params.showUIAnimations = !!this.checked;
-    settingsStore.setItem('showUIAnimations', params.showUIAnimations, Infinity);
+document.querySelectorAll('input[type="checkbox"][name=hideActiveContentWarning]').forEach(function (element) {
```
Jaifroid, line 517: Ditto (and all similar examples below)

www/js/app.js
```diff
@@ -267,17 +267,17 @@ $('#searchArticles').on('click', function () {
...
 });
-$('#searchArticles').on('mousedown', function () {
+document.getElementById('searchArticles').addEventListener('mousedown', function () {
     // We set the flag so that the blur event of #prefix can know that the searchArticles button has been clicked
     searchArticlesFocused = true;
 });
-$('#formArticleSearch').on('submit', function () {
+document.getElementById('formArticleSearch').addEventListener('submit', function () {
     document.getElementById('searchArticles').click();
     return false;
```
Jaifroid, line 276: `return false` is jQuery syntax. Please use the vanilla JS equivalent, of omit if not needed.

www/js/app.js
```diff
@@ -328,56 +328,56 @@ $('#prefix').on('keydown', function (e) {
...
 
-$('#btnRescanDeviceStorage').on('click', function () {
+document.getElementById('btnRescanDeviceStorage').addEventListener('click', function () {
     searchForArchivesInStorage();
 });
 // Bottom bar :
-$('#btnBack').on('click', function () {
+document.getElementById('btnBack').addEventListener('click', function () {
     history.back();
     return false;
```
Jaifroid, line 363: Ditto about `return false`

www/js/app.js
```diff
@@ -328,56 +328,56 @@ $('#prefix').on('keydown', function (e) {
...
 });
 $('#btnForward').on('click', function () {
     history.forward();
     return false;
```
Jaifroid, line 367: Check all the `return false` statements below and either convert or remove as needed.

Jaifroid (changes requested):
Thank you very much for this PR! It's looking good, but there are some minor details to deal with as in individual comments below. I didn't test. When you ask for review again, please describe how you've tested and on which browsers.

www/js/app.js
RG7279805: I am sorry i pasted wrong snippet here

```js

document.getElementById('btnForward').addEventListener('click', function () {
    history.forward();
});
```
This is the snippet which was replaced but the event is not fired

www/js/app.js
RG7279805: I checked compiled code but it does contain `.forEach` which is not supported by IE11

RG7279805: I have solved this issue [here](https://github.com/RG7279805/kiwix-js/blob/forEach-loop-solution/www/js/app.js#L514) If you think it seems ok I will merge it

Jaifroid: Although the compiled code contains `.forEach`, core-js probably injected a polyfill that would allow it to work on IE11 in any case, so what's really needed is to test the distribution version on IE Mode in Edge. I don't believe you can use `.forEach` on the util object, because that would cause you to iterate over every property of that object!

The way to solve it without relying on a polyfill is by assigning the selected nodes to a variable, and then using `Array.prototype.slice.call(variable).forEach(...)`. Search in the code of `app.js` for this pattern and copy the pattern.

www/js/app.js
Jaifroid: You probably need to test by loading a ZIM, navigating to a few articles, then go back in history, then go forward again. It should then fire. It won't go forward in history if you haven't first gone back.

RG7279805:
I have completed testing on these browsers 

- [x] Web testing
    - [x] Manual Chrome (service and jquery)
    - [x] Manual IE11 (jquery)
    - [x] Manual Edge (service and jquery)
- [x] As Extension
    - [x] Manual Chrome (service and jquery)
    - [x] Manual Firefox (service and jquery)

www/js/app.js
RG7279805: I have tested it on Edge (IE11 mode) and its working fine

www/js/app.js
```diff
@@ -328,57 +328,60 @@ $('#prefix').on('keydown', function (e) {
     }
 });
 // Search for titles as user types characters
-$('#prefix').on('keyup', function (e) {
+const prefixElement = document.getElementById('prefix')
```
Jaifroid, line 331: This line should be moved up to around 279, so that you can use `prefixElement` also there instead of `document.getElementById('prefix')` that appears on line 280.

www/js/app.js
```diff
@@ -414,9 +417,8 @@ $('#btnHome').on('click', function () {
     }
     // Use a timeout of 400ms because uiUtil.applyAnimationToSection uses a timeout of 300ms
     setTimeout(resizeIFrame, 400);
-    return false;
 });
-$('#btnConfigure').on('click', function () {
+document.getElementById('btnConfigure').addEventListener('click', function () {
```
Jaifroid, line 421: Like with `btnHome` you here need to add an `event.preventDefault();` and add `event` in the function parameters.

www/js/app.js
```diff
@@ -328,57 +328,60 @@ $('#prefix').on('keydown', function (e) {
...
     var articleContent = document.getElementById('articleContent');
     articleContent.contentWindow.scrollTo({ top: 0, behavior: 'smooth' });
-    // We return true, so that the link to #top is still triggered (useful in the About section)
-    return true;
 });
 // Top menu :
-$('#btnHome').on('click', function () {
+document.getElementById('btnHome').addEventListener('click', function (event) {
     // Highlight the selected section in the navbar
+    event.preventDefault()
```
Jaifroid, line 384: Missing semicolon after this command.

www/js/app.js
```diff
@@ -441,9 +443,8 @@ $('#btnConfigure').on('click', function () {
     uiUtil.checkUpdateStatus(appstate);
     // Use a timeout of 400ms because uiUtil.applyAnimationToSection uses a timeout of 300ms
     setTimeout(resizeIFrame, 400);
-    return false;
 });
-$('#btnAbout').on('click', function () {
+document.getElementById('btnAbout').addEventListener('click', function () {
```
Jaifroid, line 447: Ditto about `event.preventDefault();`.

Jaifroid (changes requested):
It's looking good! Some very minor small changes in comments, and please also update the branch with main to apply the latest changes. These include some end-to-end tests that it would be useful to run against this PR.

Jaifroid:
You should see and "Update branch" button below that can merge the latest changes from main into this branch automatically.

Jaifroid:
Thank you, @RG7279805. I think in this case, the changes suggested were about minimizing the code and being more efficient, and they are directly relevant to the PR. The general rule of thumb is not to change things that are not relevant to the PR unless they are minor/small improvements in the code you are working on. But you are free to do the implementation that you consider to be the best and most concise/efficient, and the reviewer will of course tell you if you have made an irrelevant or unwanted change. If in doubt, you can of course make the suggestion in a comment and seek guidance. Of your options, I would say go for the first if you feel confident that it is relevant to the PR and is an improvement. Go for the second if you are in doubt. I don't think you should do the third option, because opening a new branch is not a good idea when you're working on a PR. You can always revert a commit to your existing branch/PR if necessary. It's much easier for a reviewer to follow a single branch/PR rather than having to jump to a new one and have to start reviewing all over again.

I see the e2e tests (which include a test on IE11) are working well with your PR. I'll proceed to do some manual tests to double-check your tests. Thank you for testing. I'll get back to you asap, as it is probably ready to merge soon.

Jaifroid (changes requested):
@RG7279805 Unfortunately, there is a problem: the code for switching between jQuery and Service Worker mode is not working. In Internet Explorer, it should give a warning that SW mode is unavailable. In Chrome/Edge, switching to jQuery mode does nothing (you can see that the API panel at the bottom does not change as it should do

It seems that the change event is not properly registered for the controls. Could you take a look and debug?

Jaifroid:
This is the message you should see when attempting to switch to SW mode in IE11 or IE Mode (Edge). I've just checked, and it's working fine on main, so it is definitely an issue with the jQuery conversion on this branch.

image

Jaifroid:
Also, remember that after making changes, you need to rebuild the dist directory before testing on IE11, using `npm run build-src`, or else you can use `npm run preview` which will rebuild and open up a server.

Jaifroid:
OK, thanks @RG7279805. If you could update the branch again, I have incorporated a specific e2e test for the jQuery and ServiceWorker modes. It's not an in-depth test of functionality, but it could pick up this kind of issue. I would be interested to see how the test performs on your branch. In the meantime, I'll continue my manual testing.

Jaifroid (approved):
Great, thanks! I've completed manual tests, and all is working fine, even on an old Firefox OS with Firefox 45. And the automated tests with the SW / JQuery test are all passing fine. So this is ready for merge.

Jaifroid:
Let me know if you're happy for this to be merged now.

RG7279805:
Sure @Jaifroid, And Thanks again for being my mentor

Jaifroid:
You're welcome, and thank you for the very useful PR.

PR #1039, opened 2023-08-02
Add end-to-end tests with Selenium

### LINKED ISSUE #1038: Write some end-to-end tests using Selenium
I've determined that Selenium is the most viable testing framework for us, because it still has support for old browsers such as IE11, but it can undertake these tests in IE Mode under modern Edge. This gives us a good path forward.

Jaifroid:
Will close #1038.


Jaifroid:
Now ready to merge.

PR #1040, opened 2023-08-04
Change text into icon

Rbcoder1:
In PR1 I Only Change the text into icon and  move the random button into navbar properly 

Screenshot 2023-08-04 155217


Jaifroid (changes requested):
Tests are failing because you added files without adding them to the build chain.

However, you must not re-introduce `require.js`. We are now using ES6 modules and you need to use ES6 module syntax (`import x from y`).

Jaifroid:
Tests are still failing because you haven't added a new file here to the Service Worker implementation. Please see: https://github.com/kiwix/kiwix-js/actions/runs/5775729221/job/15674364839?pr=1040.

Rbcoder1:
Yes I solve it

Jaifroid:
OK! Let me know if you need any help/explanation. Please also update the branch (see button below).

Rbcoder1 (commented):
all changes review and updated

Jaifroid:
Thanks for updating the branch. You still need to fix the cause of the test failure before I review fully Just click on "Details" next to the test failure for `test-files-windows` to see the cause. You can ignore the `tests-windows` failure for now, as I believe this is just a passing failure that will probably correct itself when next run.

Rbcoder1 (commented):
All changes made

service-worker.js
```diff
@@ -108,6 +108,7 @@ const precacheFiles = [
     'www/css/kiwixJS_invert.css',
     'www/css/kiwixJS_mwInvert.css',
     'www/css/transition.css',
+    'www/img/icons/kiwix-blue-32.png',
```
Jaifroid, line 111: Please place this entry in alphabetical order in the list. `kiwix-blue...` should come after `kiwix-60...`.

www/index.html
```diff
@@ -90,9 +90,12 @@ <h5 class="modal-title" id="modalLabel">Modal title goes here</h5>
...
                     <div class="collapse navbar-collapse" id="collapsibleNavbar">
                         <ul class="navbar-nav">
-                            <li id="liHomeNav" class="nav-item"><button type="button" class="btn btn-light" id="btnHome">Home</button></li>
-                            <li id="liConfigureNav" class="nav-item"><button type="button" class="btn btn-light" id="btnConfigure">Configure</button></li>
-                            <li id="liAboutNav" class="nav-item"><button type="button" class="btn btn-light" id="btnAbout">About</button></li>
+                            <li id="liHomeNav" class="nav-item"><button type="button" class="btn btn-light" id="btnHome"><i class="fa fa-home"></i></button></li>
+                            <li id="liConfigureNav" class="nav-item"><button type="button" class="btn btn-light" id="btnConfigure"><i class="fa fa-wrench"></i></button></li>
+                            <li id="liAboutNav" class="nav-item"><button type="button" class="btn btn-light" id="btnAbout"><i class="fa fa-info"></i></button></li>
+                            <li><span class="btn-group">
```
Jaifroid, line 96: Why have you placed this button inside a `btn-group` span? Is there a reason for that? I can't see how this button is different from the others in the same row.

Jaifroid (changes requested):
Thank you for this initial PR. Some minor changes to the code suggested in comments.

I think the PR needs some work before it is complete, as follows:

* You should remove the Home icon and just use the new Kiwix icon as the home button;
* You should make the buttons fully functional with the intended behaviour of being able to click them "on" and "off" (like in Kiwix JS Windows). In other words, if I click Config, it will open Config. If I then click Config again, it will close Config and go back Home;
* The buttons need to be styled so that they indicate when they are pressed. Only Config and About should remain pressed while the user is viewing Config or About. Home and Random should remain pressed only until the relevant article is loaded;
* When the user is viewing an article and they press Home, there should be no sliding transition. It should merely open the landing page. However, the slide transition should be maintained when switching between Home, Config and About;
* I think the buttons will all need to be right-aligned rather than left-aligned, to make room for the search bar in the next PR. It might be better to do that now, or it could be done in the next PR.

Please note, that because the app will look very odd after this PR, it won't be possible to merge it once approved. Instead, you should start the branch for PR 2 based on this branch once this PR is approved. I think it should be possible to merge after PR 2.

Jaifroid:
I think you can do the "depressed button" effect by using the Bootstrap `btn btn-light` class on the images. Or there is probably another Bootstrap solution if that one doesn't work.

Rbcoder1:
Ok I will try it
Thank you

Rbcoder1:
**You should make the buttons fully functional with the intended behaviour of being able to click them "on" and "off" (like in Kiwix JS Windows). In other words, if I click Config, it will open Config. If I then click Config again, it will close Config and go back Home**

for this 
How can i go back to home if i click again on same button

in pwa i see they use some function like lastVisitedPage or many other.

Jaifroid:
> for this How can i go back to home if i click again on same button
> 
> in pwa i see they use some function like lastVisitedPage or many other.

Actually, we have an issue for that here: #127. Additionally, there is already code in the app that implements this, but it is not attached to unclicking the buttones, it is currently only used in the link `[ Show article with applied theme ]` (see [screenshot here](https://github.com/kiwix/kiwix-js/pull/582#issuecomment-570810945)) after the user changes theme. So, you could take a look at what happens when the user clicks on that link.

This could be a separate (and much easier) PR: making it so the currently loaded article (if any) is displayed if the Configuration or About links are "unclicked". The Home button itself should still go to the landing page.

Rbcoder1:
> > for this How can i go back to home if i click again on same button
> > in pwa i see they use some function like lastVisitedPage or many other.
> 
> Actually, we have an issue for that here: #127. Additionally, there is already code in the app that implements this, but it is not attached to unclicking the buttons, it is currently only used in the link `[ Show article with applied theme ]` (see [screenshot here](https://github.com/kiwix/kiwix-js/pull/582#issuecomment-570810945)) after the user changes theme. So, you could take a look at what happens when the user clicks on that link.
> 
> This could be a separate (and much easier) PR: making it so the currently loaded article (if any) is displayed if the Configuration or About links are "unclicked". The Home button itself should still go to the landing page.

So can I implement it into this PR or Not

Jaifroid:
@Rbcoder1 Yes, you can, but if you would like to have a PR that we can merge to main sooner rather than later, then tackling #127 first as a separate PR would be a good candidate.

As I mentioned earlier, when this (current) PR is complete, we would not be able to merge it to main because the app would be left with a half-finished UI, and we would at least need a further PR to make the UX consistent.

Rbcoder1:
Ok so I pause this PR here and make seperate PR for #127 to to reopen last page.

And then come back to this PR ok.

Jaifroid:
Closing for now. Can be re-opened if necessary.

PR #1047, opened 2023-08-12
Set up e2e tests on BrowserStack

### LINKED ISSUE #1041: Enable end-to-end testing on a range of platforms using Browser Stack
Since some basic e2e testing is now enabled with Selenium + Mocha for latest Firefox, Edge, Chrome and IE11 (via Edge's IE Mode) on GitHub actions (once #1038 is merged), it should be relatively easy to add some older Firefox and Chromium browsers, possibly also Safari, via Browser Stack.

Jaifroid:
Closes #1041.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -56,7 +56,7 @@ jobs:
       - name: Test a full build of the app
         run: npm run build
 
-      - name: End-to-end tests (Linux)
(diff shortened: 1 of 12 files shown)


Jaifroid:
The repeated failure is because we have to upload the archives files or else do it via XHR. Since BrowserStack only allows five files to be uploaded per session, we'll need to do this via XHR.

PR #1057, opened 2023-08-16
Moved configuration code to init.js

Daksh119:
Moved some code from `app.js` to `init.js` and added some `settingsStore' functions.
I Wwant my changes to be reviewed if I moved more or less code, hence the WIP title.
Thank You.


Jaifroid:
@Daksh119 That seems like enough code for a first PR. There are a number of other parameters defined elsewhere in app.js (without explicit `params` object being used) that could be converted to this style and moved to init.js, but I think the point of this PR is to establish the principle and move the obvious parameters, which is what you've done.

Note that one test (linux) is failing because it incorporates a check for a the `appVersion` parameter, which it expects to find in `app.js`, but which you've now (correctly) moved to `init.js`. So you need to change the test, which you can find in `./scripts/test_duplicate_values.sh`.

I'm not sure of the reason for the test failure in IE11 (tests-windows), but it might be transitory.

www/js/init.js
```diff
@@ -33,8 +33,176 @@
...
+// A parameter to disable drag-and-drop
+params['disableDragAndDrop'] = getSetting('disableDragAndDrop') === 'true';
+// A parameter to access the URL of any extension that this app was launched from
+params['referrerExtensionURL'] = getSetting('referrerExtensionURL');
+// A parameter to keep track of the fact that the user has been informed of the switch to SW mode by default
+params['defaultModeChangeAlertDisplayed'] = getSetting('defaultModeChangeAlertDisplayed');
+// A parameter to set the content injection mode ('jquery' or 'serviceworker') used by this app
+params['contentInjectionMode'] = getSetting('contentInjectionMode') ||
+    // Defaults to serviceworker mode when the API is available
+    (isServiceWorkerAvailable() ? 'serviceworker' : 'jquery');
```
Jaifroid, line 72: This method isn't available in init.js. You need to change it to something like `('serviceWorker' in navigator)`. Ditto with `isMessageChannelAvailable`. You can check the code of these methods in app.js, but keep it simple here, just use the basic `x in y` syntax.

Daksh119: There are two more methods that are available in app.js but not in init.js namely `setContentInjectionMode` and `switchHomeKeyToFocusSearchBar` but they have quite a good amount of code. Shall I copy them from app.js or move these function calls in app.js only.

www/js/init.js
```diff
@@ -33,8 +33,176 @@
...
+} else if (params.contentInjectionMode === 'serviceworker') {
+    // User is already in SW mode, so we will never need to display the upgrade alert
+    params.defaultModeChangeAlertDisplayed = true;
+    setSetting('defaultModeChangeAlertDisplayed', true);
+}
+if (!/^chrome-extension:/i.test(window.location.protocol)) {
+    document.getElementById('serviceWorkerLocal').style.display = 'none';
+    document.getElementById('serviceWorkerLocalDescription').style.display = 'none';
+}
+setContentInjectionMode(params.contentInjectionMode);
```
Jaifroid, line 157: Regarding your query, just move this line back to app.js. You copied a bit too much in this case, it doesn't need to be in init.js.

www/js/init.js
```diff
@@ -33,8 +33,176 @@
...
+document.getElementById('titleSearchRange').value = params.maxSearchResultsSize;
+document.getElementById('titleSearchRangeVal').textContent = params.maxSearchResultsSize;
+document.getElementById('appThemeSelect').value = params.appTheme;
+document.getElementById('useHomeKeyToFocusSearchBarCheck').checked = params.useHomeKeyToFocusSearchBar;
+document.getElementById('openExternalLinksInNewTabsCheck').checked = params.openExternalLinksInNewTabs;
+switchHomeKeyToFocusSearchBar();
+document.getElementById('bypassAppCacheCheck').checked = !params.appCache;
+document.getElementById('appVersion').textContent = 'Kiwix ' + params.appVersion;
+// We check here if we have to warn the user that we switched to ServiceWorkerMode
+// This is only needed if the ServiceWorker mode is available, or we are in an Extension that supports Service Workers
```
Jaifroid, line 141: ACtually, I **think** you can probably move all of this from the line above to the comment below back to the beginning of app.js, but remember to change `setSetting` back to the settingsStore version once you move it back.

Jaifroid: To be clear, I mean move from line 140 to line 157 back.

Jaifroid:
This test result explains why it's still failing the file check:

https://github.com/kiwix/kiwix-js/actions/runs/5893557867/job/15985385103?pr=1057 

Basically, you have another value to change in the Linux script from `app.js` to `init.js`.

Jaifroid:
It's a good idea to do your own manual tests each time you make code alterations before pushing to this PR, as it would save you a lot of development time to solve problems before automated tests fail. You need to test your code changes in both jQuery and ServiceWorker modes, with a real ZIM archive, and you need to check that parameters are saved between sessions.

Jaifroid:
Please look at the Tests section of CONTRIBUTING.md for further info.

Daksh119:
Sure...I'll definitely do that from now on, will save a lot of time. Thanks for all the suggestions.

Daksh119:
Hey @Jaifroid I ran the IE11 tests after defining the keyPrefix before the getSettings function but its showing the same errors. I think the `serviceworker` mode is not available in IE and therefore it shows a modal overlay to switch to jQuery mode which doesn't allow further interactions.

Jaifroid:
The modal overlay **should** be accepted by the test (there is a search for it and it should click on the button). I'll see if I can debug and get back to you.

Daksh119:
@Jaifroid I did the same changes in tests as you did in PR #1059 and they are getting passed locally

Jaifroid:
OK, but I'm just finalizing (hopefully!) that PR, so I think there will be a few more changes to merge. Let's see. I should merge in about 5 mins or so when CodeQL tests finish.

Jaifroid:
To resolve the conflict, just copy the whole file from main into your branch, given that your branch is not touching these tests.

Jaifroid:
OK, I had to run Windows test twice (probably because the VMs are very busy with overnight runs in USA), but all OK.

Daksh119:
Yes , got me worried there for a second.

scripts/test_duplicate_values.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-# This bash script tests whether app.js and service-worker.js have the same value for appVersion and for ASSETS_CACHE.
+# This bash script tests whether init.js and service-worker.js have the same value for appVersion and also whether app.js and service-worker.js have the same value for ASSETS_CACHE.
```
Jaifroid, line 3: I think this line could be made more concise now:

```suggestion
# This bash script tests the variables appVersion and ASSETS_CACHE to enrure they are correctly set in app.js, init.js and/or service-worker.js
```

scripts/test_duplicate_values.sh
```diff
@@ -9,13 +9,13 @@ REPO_DIR="$(dirname "$SCRIPT_DIR")"
 # Check values in files
 cd $REPO_DIR
 SW_VERSION="$(grep 'appVersion\s=' service-worker.js | sed -E "s/[^[:digit:]]+([^\"']+).*/\1/")"
-APP_VERSION="$(grep 'params\[.appVersion' www/js/app.js | sed -E "s/[^[:digit:]]+([^\"']+).*/\1/")"
+APP_VERSION="$(grep 'params\[.appVersion' www/js/init.js | sed -E "s/[^[:digit:]]+([^\"']+).*/\1/")"
 echo "service-worker.js : $SW_VERSION"
-echo "app.js            : $APP_VERSION"
+echo "init.js            : $APP_VERSION"
```
Jaifroid, line 14: You need to remove one space after `init.js` so that the colons align in the display.

www/js/init.js
```diff
@@ -33,17 +33,168 @@
...
+document.getElementById('showUIAnimationsCheck').checked = params.showUIAnimations;
+document.getElementById('titleSearchRange').value = params.maxSearchResultsSize;
+document.getElementById('titleSearchRangeVal').textContent = params.maxSearchResultsSize;
+document.getElementById('appThemeSelect').value = params.appTheme;
+document.getElementById('useHomeKeyToFocusSearchBarCheck').checked = params.useHomeKeyToFocusSearchBar;
+document.getElementById('openExternalLinksInNewTabsCheck').checked = params.openExternalLinksInNewTabs;
+document.getElementById('bypassAppCacheCheck').checked = !params.appCache;
+document.getElementById('appVersion').textContent = 'Kiwix ' + params.appVersion;
+
+var keyPrefix = params.keyPrefix;
```
Jaifroid, line 140: This line doesn't do what you think it does. The function `getSetting()`  is called before `params.keyPrefix` is defined. Although variables like this are "hoisted" (you can define them anywhere, even after a function that uses them), unfortunately that is not the case with the keys of the `params` object. Hence, you must define `params.keyPrefix` towards the very top of the code, at least before the first time that `getSetting()` is called.

You can test if your code is working by running `npm run preview`. Turn Bypass Appcache on and disable slide transition. Then press Ctrl-R and you'll see that the slide transition is turned back on if your code isn't working properly (i.e. the stored values are not being applied on startup).

www/js/init.js
```diff
@@ -33,17 +33,168 @@
...
+
+function getSetting (name) {
+    var result;
+    if (params.storeType === 'cookie') {
+        var regexp = new RegExp('(?:^|;)\\s*' + name + '=([^;]+)(?:;|$)');
+        result = document.cookie.match(regexp);
+        result = result && result.length > 1 ? decodeURIComponent(result[1]) : null;
+    } else if (params.storeType === 'local_storage') {
+        // Use localStorage instead
+        result = localStorage.getItem(keyPrefix + name);
```
Jaifroid, line 150: If you correct the code as I have suggested, then you can get rid of the `keyPrefix` definition and just use `params.keyPrefix` here and below.

Jaifroid (changes requested):
Thank you for this PR. It's nearly ready. Take a look at my suggestions and comments for some small changes.

www/js/init.js
Jaifroid: Just to demonstrate, I paused on a line that uses `keyPrefix` when rebooting the app, and you'll see from the screenshot below that it is `undefined`:

image

Jaifroid:
Still not applying settings correctly at startup.

www/js/init.js
```diff
@@ -33,8 +33,157 @@
...
+// A parameter to set the content injection mode ('jquery' or 'serviceworker') used by this app
+params['contentInjectionMode'] = getSetting('contentInjectionMode') ||
+    // Defaults to serviceworker mode when the API is available
+    (('serviceWorker' in navigator) ? 'serviceworker' : 'jquery');
+// A parameter to circumvent anti-fingerprinting technology in browsers that do not support WebP natively by substituting images
+// directly with the canvas elements produced by the WebP polyfill [kiwix-js #835]. NB This is only currently used in jQuery mode.
+params['useCanvasElementsForWebpTranscoding'] = null; // Value is determined in uiUtil.determineCanvasElementsWorkaround(), called when setting the content injection mode
 // The key prefix used by the settingsStore.js (see comment there for explanation), but we also need it below
-params['keyPrefix'] = 'kiwixjs-'
+params['keyPrefix'] = 'kiwixjs-';
```
Jaifroid, line 77: You need to move this line (and comment) to BEFORE the first ***use*** of `getSetting()`. It's not enough to have it before the function. It has to be before the function is first called by the code above.

www/js/init.js
```diff
@@ -33,8 +33,157 @@
...
+document.getElementById('disableDragAndDropCheck').checked = params.disableDragAndDrop;
+document.getElementById('showUIAnimationsCheck').checked = params.showUIAnimations;
+document.getElementById('titleSearchRange').value = params.maxSearchResultsSize;
+document.getElementById('titleSearchRangeVal').textContent = params.maxSearchResultsSize;
+document.getElementById('appThemeSelect').value = params.appTheme;
+document.getElementById('useHomeKeyToFocusSearchBarCheck').checked = params.useHomeKeyToFocusSearchBar;
+document.getElementById('openExternalLinksInNewTabsCheck').checked = params.openExternalLinksInNewTabs;
+document.getElementById('bypassAppCacheCheck').checked = !params.appCache;
+document.getElementById('appVersion').textContent = 'Kiwix ' + params.appVersion;
+
```
Jaifroid, line 139: I suggest adding a comment here:

```suggestion

// This is a simplified version of code in settingsStore, because that module is not available in init.js
```

www/js/init.js
```diff
@@ -33,8 +33,157 @@
...
+        var regexp = new RegExp('(?:^|;)\\s*' + name + '=([^;]+)(?:;|$)');
+        result = document.cookie.match(regexp);
+        result = result && result.length > 1 ? decodeURIComponent(result[1]) : null;
+    } else if (params.storeType === 'local_storage') {
+        // Use localStorage instead
+        result = localStorage.getItem(params.keyPrefix + name);
+    }
+    return result === null || result === 'undefined' ? null : result === 'true' ? true : result === 'false' ? false : result;
+}
+
```
Jaifroid, line 152: And a similar comment here.

Jaifroid:
Thanks for moving `keyPrefix` parameter. That is now in the right place.

But... did you test that all the settings now work? I tested the UI animations setting, and it's not working.

The reason is that the simplified `getSettings()` function is subtly different form the full settingsStore function. To aid in getting parameters, this function returns an actual `false` if a setting is set to false in localStorage or cookie, and doesn't return the string 'false' like the settingsStore function does. This means we can (and must) now write some of these getSetting statements in a different way, e.g.:

`params['showUIAnimations'] = getSetting('showUIAnimations') !== false;`

Think about the logic here, and why I have simplified this to  `!== false`. Clue: we want the default tol be `true` if the value has never been set (if it's  null or blank). It will only be `false` if in fact the value returned is `false`.

There are a few others you'll need to go through and rewrite if necessary.

Daksh119:
My bad sorry I forgot to run the tests again, my mind is pretty occupied elsewhere today thats the reason for the faults, I'll rectify the mistakes once I get comfortable again. Thanks once again for reviewing my PR for so long.

Jaifroid:
> My bad sorry I forgot to run the tests again, my mind is pretty occupied elsewhere today thats the reason for the faults, I'll rectify the mistakes once I get comfortable again. Thanks once again for reviewing my PR for so long.

No worries, take your time.

Jaifroid:
See failure here (you'll need to update the appVersion to latest in both init.js and serviceworker file):

https://github.com/kiwix/kiwix-js/actions/runs/5966600090/job/16187360801?pr=1057

Jaifroid:
Actually, you only need to update ini.js.

Jaifroid:
Automatic tests are passung. Do you need to do any more manual tests!?

Daksh119:
No I think everything is fine. 
Please let me know if I'm missing any more errors.

Jaifroid:
OK, I'll test!

Jaifroid:
Manual tests seem good too!

scripts/test_duplicate_values.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-# This bash script tests whether app.js and service-worker.js have the same value for appVersion and for ASSETS_CACHE.
+# This bash script tests the variables appVersion and ASSETS_CACHE to enrure they are correctly set in app.js, init.js and/or service-worker.js
```
Jaifroid, line 3: Typo: `enrure` --> `ensure` (probably my typo...)

Jaifroid (approved):
Just one tiny typo. Let me know when you're ready for this to be squashed/merged.

Daksh119:
YES...All good

Jaifroid:
Great, just running the last test run.

Jaifroid:
Thank you, @Daksh119 for this useful PR. Now squashed and merged.

Daksh119:
My pleasure @Jaifroid 
Thank You for bearing with me and helping me throughout.

PR #1061, opened 2023-08-19
Add internationalization framework

### LINKED ISSUE #1060: Translate the UI into French and Spanish with `i18next.js`
This would be a milestone towards the very old issue #13. Maybe this is achievable by the Kiwix JS 10-year anniversary? At least I think Spanish is achievable (because I can do it myself).

It's a bit ambitious: the anniversary is 23rd August.... 🤔 And we wouldn't be able to get a release out by that date as well.

Jaifroid:
Will eventually close #1060.

This PR adds the Banana dependencies and a few sample strings in English and Spanish.

Currently working in Chromium only (due to import assertion syntax), and not compile-able with rollup / babel (yet).


Jaifroid:
It seems impossible to use Banana with ES6 modules. So I'm now experimenting with i18next.js.

Jaifroid:
OK, I've done French and Spanish for the whole app... Pretty much finished if all tests pass.

Issue #1063, opened 2023-08-22
Adapt BrowserStack e2e test runners so that they run the non-legacy ZIM test


Jaifroid:
This could be undertaken after the completion of #1062.

RG7279805:
@Jaifroid Sure, no problem

PR #1064, opened 2023-08-22
Folder structure redone for `tests`

### LINKED ISSUE #1062: Folder structure for test and proper file segregation
As I have already mentioned in #1049 we should go on and restructure the folder and file structure
You can assign this issue to me and I will start working + please drop any suggestions

RG7279805:
closes #1062


RG7279805:
I have left out `tests/unit` as it is. Do you think i should do any refactor over there I mean there is a single and single file with test
If you think it should be done i could come up with some structure as well

Jaifroid:
Happy for you to propose something, or just structure the e2e tests, as you wish!

RG7279805:
@Jaifroid I think its good enough to merge now

tests/foo-zstd.zim
Jaifroid, line 1: This file should be maintained in the repository, because it is the basic "reference file" for the ZIM format with ZSTD compression. Even if we don't use it currently, please put it in its own folder like the other ZIMs.

Jaifroid (approved):
It looks good, and much more organized, thank you! There's one small change (see comment).

Jaifroid (approved):

Jaifroid:
Many thanks for this PR. Squashing/merging now.

Jaifroid:
@RG7279805 Unfortunately, it looks like this PR broke the BrowserStack implementation, and of course that didn't get tested in your testing of it, because BrowserStack tests can't run with PRs from forked repositories. I'm re-running the tests, but we might have to revert. I'm also getting failures on my internationalization branch that appear to be caused by the PR.

Jaifroid:
Hmm, definitely failing, check out:

https://github.com/kiwix/kiwix-js/actions/runs/5951851073/job/16149678973 and https://github.com/kiwix/kiwix-js/actions/runs/5953037189/job/16146376792?pr=1061.

I tried to fix the paths for the legacy edge runner on my branch, but it still fails.

Jaifroid:
So, I've reverted for now, but your PR is saved at this commit:

https://github.com/kiwix/kiwix-js/commit/632f01929155967e3eb0d711e7120f5b6686a3bb

It probably needs some adjustment, but I can't seem to work out what at the moment, and you will have a clearer idea of what changed.

RG7279805:
I think then I should create a PR fixing #1062 , #1063 and also making sure everything runs smoothly

PR #1070, opened 2023-08-24
Trying to CI timeout issue

### LINKED ISSUE #1069: Selenium tests can be flaky, especially with slow-running VMs
Probably a timeout issue, but I quite often have to re-run tests that subsequently pass on the second try. Clearly some lingering race condition. @RG7279805 I wonder if restructuring the legacy test so that it no longer uses the `describe` wrapper might increase reliability, especially given that it seems to be the legacy test that falls over more often than your modern ZIM test? Any thoughts on that?

RG7279805:
trying to fix #1069

I am creating this PR to fix the CI issue. It might take multiple attempts so please bear with me

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -34,6 +34,7 @@ jobs:
     # We want to run on external PRs, but not on our own internal PRs as they'll be run by the push to the branch
     if: github.event_name == 'push' || github.event.pull_request.head.repo.full_name != github.repository
     runs-on: ubuntu-20.04
+    timeout-minutes: 15
     steps:
       # Clone the repo and checkout the commit for which the workflow was triggered
       - uses: actions/checkout@v3
@@ -113,6 +114,7 @@ jobs:
   tests-windows:
     if: github.event_name == 'push' || github.event.pull_request.head.repo.full_name != github.repository
     runs-on: windows-latest
+    timeout-minutes: 15
     steps:
       # Clone the repo and checkout the commit for which the workflow was triggered
       - uses: actions/checkout@v3
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -34,7 +34,7 @@ jobs:
     # We want to run on external PRs, but not on our own internal PRs as they'll be run by the push to the branch
     if: github.event_name == 'push' || github.event.pull_request.head.repo.full_name != github.repository
     runs-on: ubuntu-20.04
-    timeout-minutes: 15
+    timeout-minutes: 30
     steps:
       # Clone the repo and checkout the commit for which the workflow was triggered
       - uses: actions/checkout@v3
@@ -114,7 +114,7 @@ jobs:
   tests-windows:
     if: github.event_name == 'push' || github.event.pull_request.head.repo.full_name != github.repository
     runs-on: windows-latest
-    timeout-minutes: 15
+    timeout-minutes: 30
     steps:
       # Clone the repo and checkout the commit for which the workflow was triggered
       - uses: actions/checkout@v3
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -56,16 +56,21 @@ jobs:
       - name: Test a full build of the app
         run: npm run build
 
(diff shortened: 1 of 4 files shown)


RG7279805:
@Jaifroid can you check if this kind of Stuck CI pipeline is being billed to your organization
https://github.com/kiwix/kiwix-js/actions/runs/5976046146/job/16213149963
If so we need to take a measure to kill them after some time

Jaifroid:
Thanks, I just cancelled it. We don't get billed because we're open source, but for all sorts of reasons, including ecological, we should find a way to cancel them if they run, day, double the expected time.

RG7279805:
@Jaifroid for now i have added a time limit of 15 mins

RG7279805:
@Jaifroid This should fix the CI issue I ran CI 20-30 times and fixed any problem. After a lot of trials, I am getting consistent successful runs 12 times in a row (1 timeout error for CI). To be honest I am not 100% sure that this PR will completely fix all the issues on CI.Do you mind having a look and maybe suggesting something 

One thing I noticed is test always fail on windows

.github/workflows/CI.yml
```diff
@@ -56,16 +57,21 @@ jobs:
       - name: Test a full build of the app
         run: npm run build
 
-      - name: End-to-end tests on GitHub (Linux)
+      - name: End-to-end tests on Chrome (Linux)
```
RG7279805, line 60: It's not required to run tests on each browser individually. If you think it is not good I can revert `CI.yml`

Jaifroid:
OK, many thanks! I'll take a look properly tomorrow. One thing I'd like to do would be to make a parallel dummy branch set to the same last commit as this PR, which in my experience usually gets the tests to run on BrowserStack as well as on GitHub (due to the secrets then being available).

Jaifroid:
In any case, it's certainly looking a lot more stable, which is great!

Jaifroid:
So I managed to get BrowserStack tests to run on this PR (by pushing a dummy PR set to the same commit on another branch), and they're running fine and smoothly using the newly restructured legacy test, which is great! You can review them here: https://github.com/kiwix/kiwix-js/actions/runs/5989335303/job/16245414394?pr=1070.

The Windows test (which doesn't run on BrowserStack, only on GitHub) timed out because it got stuck on the Gutenberg test on Firefox. I let it run to test your timeout, which worked very well, so thank you for that too. I'll now re-run the failed test to be sure it was transitory.

Jaifroid:
It might be worth updating the Gutenberg ZIM to the latest release in this PR:

https://download.kiwix.org/zim/gutenberg/gutenberg_ro_all_2023-08.zim.

This is because there have been significant backend changes. I don't think it will affect the tests at all, but we might want to introduce (later) a test for support of dynamic JS in the Chromium Browser Extension version. This version has a very strict CSP, and it would be good to be running it on a ZIM that complies with the CSP (which the August versions of Gutenberg ZIMs now do).

.github/workflows/CI.yml
Jaifroid: I think it's fine to have more granularity. It does seem more structured like that, so keep this change.

Jaifroid (approved):
Another excellent PR, many thanks indeed! I tested locally as well as with BrowserStack as per comment above.

The only suggestions I have are:

* Change to the latest Gutenberg RO archive (August 2023) as mentioned in another comment above;
* When a runner runs both the legacy and the modern tests, it shows the two messages below in the test console. It would be great if each message could add the name of the test which could appear as something like `Running legacy-ray_charles.ese.spec.js tests on: chrome 116.0.5845.96` (and same for non-legacy). If it's not too hard to do!

image

Jaifroid:
To get the script filename, you can probably just use `const scriptName = path.basename(__filename);`

Jaifroid:
Thanks for making those change. BrowserStack tests are passing as well. Let me know when it is ready to squash/merge!

RG7279805:
I think its all good we can merge

Jaifroid:
Really useful work -- will make a big difference to the development experience here. Thank you! Merged.

RG7279805:
Happy to help you guys

Issue #1072, opened 2023-08-24
Some translations missing: at least the active content warning panel


Jaifroid:
As this isn't a standard dialogue box, it got missed. There may be one or two other Bootstrap-based inline notices that need to be addressed as well.

Issue #701, opened 2021-02-06
Regression: Animate transition between app pages breaks links that jump or scroll to elements on another page


Jaifroid:
As per title. One workaround is to wait for the transition to complete (around 500 - 600ms), but this introduces a race and might not work on slow browsers or slow environments. The best solution would be to prevent the transition from interfering, or to make its methods aware of local anchors.

Jaifroid:
To be clear, this issue refers to links from one of the app's pages to another inside the app. It does not refer to jumping to sections within an article.

RG7279805:
I am back after a small break and start working on this again. Here is what I think could work 

PROPOSAL: 
- Add a global variable to manage if the animation is active or not (`let animationActive = true/false`) 
- We can disable the animations by default
- cleanup the current animation code or maybe just replace it with something easy

Jaifroid:
Global variables are attached to the `params` object and are set in `init.js` (now). There should already be one relating to this setting.

Jaifroid:
Changing `params['showUIAnimations'] = getSetting('showUIAnimations') !== false;` to ` === true;` should change the default: it will always be false unless explicitly set to true in localStorage.

PR #1078, opened 2023-08-28
Review French translations and propagate changes to other languages

### LINKED ISSUE #1071: Review the French translations
While I am pretty confident about the Spanish translations, as it's a language I work with regularly, I'm aware the French translations are clunky in a number of places, but I don't have the fluency and technical vocab to correct them. I admit to having some help from DeepL...

@kelson42 @Popolechien, while it's essential to use translatewiki for other languages in the future, in the interests of making an "anniversary" v4.0.0 release sooner rather than later with French and Spanish, would one of our native French speakers be able to read through the app and send me corrections of anything obviously bad or incomprehensible or horribly clunky? They wouldn't need to edit the actual translation strings. Just copying the visible text on the "Configrurer" and "Informations" pages into a Word / OpenOffice document and marking up would be extremely helpful, or even just making a list in this issue if there are not too many problems.

It would probably take someone 30 mins max that way. I could offer to reciprocate for Spanish (or for English).

Jaifroid:
Fixes #1071.


Jaifroid:
Also need to fix #1072.

Jaifroid:
Added workaround for some internal links for #701. But it still needs a proper fix.

Jaifroid:
I have got as far as I can with this fiddly business, but I think the whole UI is improved as well as many of the translations in both French and Spanish (and streamlined the English too). A number of UI features added or repaired as well.

PR #1082, opened 2023-08-31
Update build scripts to use version from init.js

### LINKED ISSUE #1081: Update build scripts to use version number from `init.js`
Several build scripts still reference (or update) the location of the version number as being in `app.js,` whereas #1057 moved that location to `init.js`. So we need to update them.

Jaifroid:
Fixes #1081.


Jaifroid:
Tested altered scripts manually, but the packaging script can only be fully tested as part of release process.

PR #1089, opened 2023-08-31
Restore IE Mode e2e testing in Windows

### LINKED ISSUE #1088: Restore testing in IE11 Mode in the Windows tests
In a recent PR, the IE Mode test got swapped out for a Chrome test in error. We don't need to run Chrome in Windows, because Edge is the same thing, and Chrome is run in the Linux tests. But what we can uniquely run in Windows tests is the IE mode of Edge, and this is useful because it tests code for various regressions.

I'll restore this.

Jaifroid:
Fixes #1088. This test got accidentally removed in a recent PR. Please see explanation in #1088 as to why I don't believe it's worth running Chrome (as opposed to Edge) tests specifically in the Windows VM. Testing is already quite slow, so we should focus on tests that clearly add value.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -144,21 +144,21 @@ jobs:
           }
           exit $AggregateExitCode
 
-      - name: End-to-end tests on Chrome(Windows)
-        env:
-          GITHUB_ACTION: ${{ github.event_name }}
-        run: npm run test-e2e-chrome
-
-      - name: End-to-end tests on Edge(Windows)
+      - name: End-to-end tests on Edge Chromium (Windows)
         env:
           GITHUB_ACTION: ${{ github.event_name }}
         run: npm run test-e2e-edge
 
-      - name: End-to-end tests on Firefox(Windows)
+      - name: End-to-end tests on Firefox (Windows)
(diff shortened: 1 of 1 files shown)


RG7279805 (approved):

RG7279805:
**@Jaifroid, I like your idea of removing the Chrome Windows Tests, as all tests were taking around 8 minutes to complete. Continuing this discussion, I believe there's an opportunity to explore an even more effective solution. Currently, the tests are categorized by OS. Perhaps, we could consider a different categorization approach. Presently, Windows tests involve about 3 to 4 steps, whereas Linux tests encompass around 10 to 12 steps. Let's brainstorm for potential improvements.**

Jaifroid:
Great, thank you! The main aim here was to reintroduce the IE11 test (IE Mode), and so as not to extend the Windows tests even more, it seemed best to remove the Chrome test which is really redundant on Windows if we are also testing in Edge. I agree we should look for ways of making tests more efficient if possible. They run significantly faster on Linux than on Windows (I suspect because the Windows VMs on GitHub just don't have so many runners), but we do need some Windows testing. I'll merge now.

PR #1100, opened 2023-09-06
Ensure dockerfile serves jsonp with correct MIME type

### LINKED ISSUE #1099: Translation strings delivered with wrong MIME type at least on GitHub server - check if also on GHCR container
The MIME type for `.jsonp` should be `application/javascript`, but according to Firefox (`The script from “https://kiwix.github.io/kiwix-js/i18n/en.jsonp” was loaded even though its MIME type (“application/octet-stream”) is not a valid JavaScript MIME type`) the MIME type is being delivered as `application/octet-stream`.

This could cause problems in the future on some browsers.

We could either change the strings to `.jsonp.js`, or else (better) set the correct MIME type at least for the docker build.

Jaifroid:
Should partially fix #1099 for the GHCR container.

### DIFF
--- a/docker/dockerfile-browser-extension.pwa
+++ b/docker/dockerfile-browser-extension.pwa
@@ -2,6 +2,7 @@ FROM nginx:latest
 
 EXPOSE 80
 
+RUN echo "application/javascript jsonp;" >> /etc/nginx/mime.types
 RUN mkdir /usr/share/nginx/html/current/
 COPY ./docker/index.nginx.html /usr/share/nginx/html/index.html
 COPY ./dist/manifest.json /usr/share/nginx/html/current/


Jaifroid:
From web searches, it appears it is not possible to change the MIME type of files on gh-pages, though it is possible to contribute a changed MIME type at https://github.com/jshttp/mime-db#adding-custom-media-types. However, the process looks pretty complicated, and if it every breaks on gh-pages, we can simply rename the files to `.jsonp.js`.

Jaifroid:
There's no real way to know if this one-liner will work until we try it on the live container as part of a release, so will need to be vigilant at release time. @kelson42 As you originally set up our dockerfile and are fluent in Linux, could you take a look at this one-liner?

Jaifroid:
We just have to test this empirically.

Issue #114, opened 2015-05-04
Add a way to easily download ZIM archives


mossroy:
Like it's currently done in Kiwix.
We should take extra care of files whith size >4GB, which need to be splitted to fit on FAT32 filesystems (which are still very common in smartphones/tablets)

mossroy:
There would be at least 2 ways to do that :
- implement an HTTP download manager (like recently added on Android) : with pause/resume etc
- use a bittorrent client like https://github.com/feross/webtorrent (thanks @peter-x ). The main issue is that it uses a slightly different protocol, that does not seem to be handled by the Kiwix bittorrent seeder for now

Jaifroid:
Although it doesn't include a download manager, on kiwix-js-windows I have implemented an AJAX-based repository browser. A link in configuration (with a warning that it will use an Internet connection) opens up a panel with the formatted contents of the index at http://download.kiwix.org/zim/ . Can browser forwards and backwards in repo.

Clicking on a .zim link opens a list of download sources with a warning about file size. Currently it warns (in red) if the file is larger than 200MB, but I could easily enough add a warning about files larger than 4GB as well.

Clicking on one of the sources will open a browser window and offer to download and save the file (except for the mirrorservice.org bug, so I've disabled that link). Download of large files is tested and works fine on Windows 10 Mobile and Windows 10 UWP (desktop/tablet), and the file is saved to Downloads, or wherever the user has chosen to save downloads in Windows settings. Downloads work in the background.

I don't know what effect this would have on other Mobile systems (i.e., using browser to manage downloads). I'll happily backport if this solution is of interest. Note that @sharun-s has a different (and elegant) solution, which is a list of links to certain popular files, based on pre-downloaded JSON data, so it works offline. However, since a connection is needed to download the file, I'm not sure there is an advantage to keeping the list of repos offline. 

Screenshots below (after the panel is opened -- until then the panel is hidden).

![image](https://user-images.githubusercontent.com/4304337/29916469-cb35d968-8e36-11e7-91f6-1015bec5533c.png)

And after clicking on a large ZIM archive:

![image](https://user-images.githubusercontent.com/4304337/29917029-87607a5c-8e38-11e7-9a6c-21af2e6b9d1f.png)

mossroy:
This looks cool, but I'm not sure it's the best technical way to do that.
[...]

PR #1102, opened 2023-09-07
Transition animation cleanup and page animation turned off by default 

### LINKED ISSUE #78: Simplify the source code for hiding/showing of HTML elements

### LINKED ISSUE #1101: Code cleanup for Tab transition animation
While working on issue #846 Adding a new tab seems more complex than expected code seems really confusing and there is a lot of repetition for the same function
check out 
 https://github.com/kiwix/kiwix-js/blob/main/www/js/lib/uiUtil.js#L490

RG7279805:
closes #1101
closes #78

### DIFF
--- a/www/js/app.js
+++ b/www/js/app.js
@@ -351,7 +351,7 @@ document.getElementById('btnHome').addEventListener('click', function (event) {
     document.getElementById('liAboutNav').setAttribute('class', '');
     $('.navbar-collapse').collapse('hide');
     // Show the selected content in the page
-    uiUtil.removeAnimationClasses();
+    // uiUtil.removeAnimationClasses();
     if (params.showUIAnimations) {
        uiUtil.applyAnimationToSection('home');
     } else {
@@ -390,7 +390,7 @@ document.getElementById('btnConfigure').addEventListener('click', function (even
     document.getElementById('liAboutNav').setAttribute('class', '');
     $('.navbar-collapse').collapse('hide');
     // Show the selected content in the page
-    uiUtil.removeAnimationClasses();
+    // uiUtil.removeAnimationClasses();
     if (params.showUIAnimations) {
         uiUtil.applyAnimationToSection('config');
     } else {
@@ -417,7 +417,7 @@ document.getElementById('btnAbout').addEventListener('click', function (event) {
     document.getElementById('liAboutNav').setAttribute('class', 'active');
     $('.navbar-collapse').collapse('hide');
     // Show the selected content in the page
-    uiUtil.removeAnimationClasses();
+    // uiUtil.removeAnimationClasses();
     if (params.showUIAnimations) {
         uiUtil.applyAnimationToSection('about');
     } else {
--- a/www/js/app.js
+++ b/www/js/app.js
@@ -381,7 +381,6 @@ document.getElementById('btnConfigure').addEventListener('click', function (even
     document.getElementById('liAboutNav').setAttribute('class', '');
     $('.navbar-collapse').collapse('hide');
     // Show the selected content in the page
-    // uiUtil.removeAnimationClasses();
 
     uiUtil.tabTransitionToSection('config', params.showUIAnimations);
 
@@ -398,8 +397,8 @@ document.getElementById('btnAbout').addEventListener('click', function (event) {
     document.getElementById('liConfigureNav').setAttribute('class', '');
     document.getElementById('liAboutNav').setAttribute('class', 'active');
     $('.navbar-collapse').collapse('hide');
+
     // Show the selected content in the page
-    // uiUtil.removeAnimationClasses();
     uiUtil.tabTransitionToSection('about', params.showUIAnimations);
 
     // Use a timeout of 400ms because uiUtil.applyAnimationToSection uses a timeout of 300ms
--- a/www/js/app.js
+++ b/www/js/app.js
@@ -400,19 +400,8 @@ document.getElementById('btnAbout').addEventListener('click', function (event) {
     $('.navbar-collapse').collapse('hide');
     // Show the selected content in the page
     // uiUtil.removeAnimationClasses();
-    if (params.showUIAnimations) {
-        uiUtil.tabTransitionToSection('about');
-    } else {
-        document.getElementById('about').style.display = '';
-        document.getElementById('configuration').style.display = 'none';
-        document.getElementById('articleContent').style.display = 'none';
-    }
-    document.getElementById('navigationButtons').style.display = 'none';
-    document.getElementById('formArticleSearch').style.display = 'none';
-    document.getElementById('welcomeText').style.display = 'none';
-    document.getElementById('articleListWithHeader').style.display = 'none';
-    document.getElementById('searchingArticles').style.display = 'none';
-    document.querySelector('.kiwix-alert').style.display = 'none';
+    uiUtil.tabTransitionToSection('about', params.showUIAnimations);
+
     // Use a timeout of 400ms because uiUtil.applyAnimationToSection uses a timeout of 300ms
     setTimeout(resizeIFrame, 400);
 });
--- a/www/js/app.js
+++ b/www/js/app.js
@@ -351,17 +351,8 @@ document.getElementById('btnHome').addEventListener('click', function (event) {
     document.getElementById('liAboutNav').setAttribute('class', '');
     $('.navbar-collapse').collapse('hide');
     // Show the selected content in the page
-    // uiUtil.removeAnimationClasses();
-    if (params.showUIAnimations) {
-       uiUtil.applyAnimationToSection('home');
-    } else {
-        document.getElementById('articleContent').style.display = '';
-        document.getElementById('about').style.display = 'none';
-        document.getElementById('configuration').style.display = 'none';
-    }
-    document.getElementById('navigationButtons').style.display = '';
-    document.getElementById('formArticleSearch').style.display = '';
-    document.getElementById('welcomeText').style.display = '';
+    uiUtil.tabTransitionToSection('home', params.showUIAnimations);
+
     // Give the focus to the search field, and clean up the page contents
     document.getElementById('prefix').value = '';
     document.getElementById('prefix').focus();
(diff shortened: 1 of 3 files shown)


Jaifroid:
Should also close #78 and #114.

Jaifroid:
I was thinking it might be better to separate this into two PRs, one to clean up transition animation, and one to do the library. Generally best not to mix too many issues (even if related) in a PR!

I'm slightly worried there won't be room in the UI for yet another button / menu item, especially after #523 is completed. Why not follow the model of KJSW and have a button in Config that opens a pane in which the library is framed? Or is layout an issue?

RG7279805:
> I was thinking it might be better to separate this into two PRs, one to clean up transition animation, and one to do the library. Generally best not to mix too many issues (even if related) in a PR!

Sure, Do you think that slide animation is required we could simply a lot of code by a fade animation 

> I'm slightly worried there won't be room in the UI for yet another button / menu item, especially after #523 is completed. Why not follow the model of KJSW and have a button in Config that opens a pane in which the library is framed? Or is layout an issue?

Thats one way to fix this issue,do you think we can can add a button on home screen (with welcome text) that will show library?

Jaifroid:
> Sure, Do you think that slide animation is required we could simply a lot of code by a fade animation

Personally, I dislike the slide animation (it makes me feel dizzy!), and I turn it off, but a previous main developer of Kiwix had asked for it and a trainee dev spent some time coding it, so I think we kind of have to maintain it. I think I would like it to be off by default if you agree (that's easy enough to do probably in the params setup in `init.js`). It also causes problems when it is on because it interrupts actions like jumping to an anchor (see #701), which we have to fix. (You don't have to do that in your PRs unless you see an obvious way to do it, it could be fixed separately.)

> Thats one way to fix this issue,do you think we can can add a button on home screen (with welcome text) that will show library?

I'm open to any reasonable solution, and even the separate tab approach, I just don't think we'll have space (take a look at http://pwa.kiwix.org and reduce the screen to mobile-style dimensions) when we fix #523. A link in the Welcome Text that opens the library in a new page of the app (not a new tab of the browser) would be fine, so long as a button or link is also in Configuration.

A previous dev felt it was redundant in this app to provide an in-app library, because we could simply provide a link (as we do) to https://library.kiwix.org, and it doesn't add a huge amount to have the library in-app. But my feeling is that it's disorientating to the average user to be sent out of the app, unless they really choose to do so, and it's better to have an in-app solution even if it's only wrapping an iframe around https://library.kiwix.org. When I coded KJSW (the PWA), library.kiwix.org did not exist, so I wrote a lot of code processing the HTTP responses coming from download.kiwix.org to provide language filters and subject filters, but fortunately, now, all this is done for us by library.kiwix.org, and I don't think we need much if any processing of its output. It also avoids us having to deal with CORS issues.

www/js/init.js
```diff
@@ -49,7 +49,7 @@ params['storeType'] = getBestAvailableStorageAPI();
 // The key prefix used by the settingsStore.js (see comment there for explanation), but we also need it below
 params['keyPrefix'] = 'kiwixjs-';
 params['hideActiveContentWarning'] = getSetting('hideActiveContentWarning') === true;
-params['showUIAnimations'] = getSetting('showUIAnimations') !== false;
+params['showUIAnimations'] = getSetting('showUIAnimations') !== true;
```
Jaifroid, line 52: This needs to be ` === true;`. The logic is that if the user has set it explicitly to `true`, then this will return `true`. But if it's any other value (`null`, '', `undefined`, `false`, 'false'), it will be `false`.

Jaifroid (changes requested):
Thanks for the PR. It looks promising, but I found some glitches. I built a bundled version and ran it in IE11, and this is what I got:

Slide_animation_IE11

Also the UI Animation option isn't turned off by default when you reset the app. I explain why in a code comment below.

I'll do a code review when you've fixed the glitches and checked it all works smoothly!

RG7279805:
@Jaifroid Do  you mind telling me how you are running tests on IE as i dont have IE installed on windows 11 neither i can manage to run it. It seems like you are on lower windows version. I tried running it on IE mode (edge) but that doesnt support dev tools so i cant debug

Jaifroid:
I use IE11 simply as a proxy for the type of issue that might be encountered in older browsers. If it works there, it'll probably work anywhere.

RG7279805:
I tried to recreate this issue but could encounter this glitch at all

https://github.com/kiwix/kiwix-js/assets/47270636/3541b8ba-bba7-4ac4-86df-fb36fbdbaca0

Jaifroid:
I'm afraid I can reproduce this in IE Mode. Please note I have not loaded a ZIM, so this might make a difference. Also, I am not switching fast between tabs, I am simply switching back from About to Configuration, and pause, and the glitch shows.

Slide_animation_IE_Mode

RG7279805:
Not sure what to do now should I close the PR and leave it for now? I have tested this multiple times and still doesn't get this glitch
https://github.com/kiwix/kiwix-js/assets/47270636/4f5b7558-d518-4baa-b284-63c62e9b0195

RG7279805:
I am sorry but I think I didn't understand what you were trying to say
I mean on `upstream/main` aren't we on version 3.10.2?
image

Jaifroid:
Yes, but we were not testing `upstream/main`, you asked me to test this branch. I can only test code you've committed to this branch, and at the time you asked me to review it was on 3.10.0. You've just pushed some updates which will have changed that. My point is only that we need to be testing the exact same code!

RG7279805:
Ok, I finally encountered this issue now on 3.10.2 i can finally try to fix the issue

Jaifroid:
Yes, I'm still getting the same issue with the code you've just pushed (I hard reset my branch to it, since you force-pushed it). I'm glad you now see it -- hopefully you can debug!

RG7279805:
Sorry for pinging again but could you review again?

Jaifroid:
> Sorry for pinging again but could you review again?

No problem -- please do explicitly ask when you want a review, I had no idea you were waiting for one!

Jaifroid:
I confirm that the IE11 glitch is now fixed! That's great!

However, I found one slight glitch involving the animation when clicking the "Show article with applied theme" link (after changing from light to dark or vice versa). The animation no longer works with that link (on the main branch, it does work). So the link is no longer interfacing correctly with the new code. Also, the welcome text is erroneously displayed after clicking on that link. Again, this doesn't happen on main. It's easier to show than to explain:

Glitch with show applied theme

www/js/lib/uiUtil.js
```diff
@@ -465,80 +465,159 @@ function isElementInView (el, fully) {
...
-            setTimeout(function () {
-                document.getElementById('about').style.display = 'none';
-            }, 300);
+function tabTransitionToSection (toSection, isAnimationRequired = false) {
+    // all the references of the sections/tabs
+    const config = document.getElementById('configuration');
+    const about = document.getElementById('about');
+    const home = document.getElementById('articleContent');
+
+    const isZimLoaded = true // get from settings or something
```
RG7279805, line 580: Could you tell me how I check if a Zim file is loaded
do I directly use DOM and check if any file is uploaded `document.getElementById("archiveFiles").files.length`
or is there any other variable (I didn't found anything similar)

RG7279805:
@Jaifroid Can you take a quick look

Jaifroid:
Apologies, was caught up in wall-to-wall meetings. Will look no.

www/js/lib/uiUtil.js
Jaifroid: There is the object `selectedArchive` which represents a loaded ZIM archive. It has a method `isReady` (or something like that).

Jaifroid:
Although the erroneous message is fixed if the user has transition animation turned on, it's still there if they're turned off. ☹️

image

RG7279805:
Sorry but I am still working on this and didn't know how to take Zim file state so there will be one or two more commits

Jaifroid:
OK, no problem!

RG7279805:
@Jaifroid could you try once more now it has been fixed

Jaifroid:
Yes! It now looks like it's fixed. Do you want me to review the code?

RG7279805:
Yes it seems its all good for a review

www/js/lib/uiUtil.js
```diff
@@ -465,80 +465,157 @@ function isElementInView (el, fully) {
...
+ */
+function fromSection () {
+    const isConfigPageVisible = !$('#configuration').is(':hidden');
+    const isAboutPageVisible = !$('#about').is(':hidden');
+    const isArticlePageVisible = !$('#articleContent').is(':hidden');
+
+    if (isConfigPageVisible) return 'config';
+    else if (isAboutPageVisible) return 'about';
+    else if (isArticlePageVisible) return 'home';
+}
```
Jaifroid, line 566: Best to add a single blank line between functions, for legibility.

www/js/app.js
```diff
@@ -390,19 +381,10 @@ document.getElementById('btnConfigure').addEventListener('click', function (even
...
-        document.getElementById('about').style.display = 'none';
-        document.getElementById('configuration').style.display = '';
-        document.getElementById('articleContent').style.display = 'none';
-    }
-    document.getElementById('navigationButtons').style.display = 'none';
-    document.getElementById('formArticleSearch').style.display = 'none';
-    document.getElementById('welcomeText').style.display = 'none';
-    document.getElementById('searchingArticles').style.display = 'none';
-    document.querySelector('.kiwix-alert').style.display = 'none';
+    // uiUtil.removeAnimationClasses();
```
Jaifroid, line 384: Remove redundant comment

www/js/app.js
```diff
@@ -417,20 +399,9 @@ document.getElementById('btnAbout').addEventListener('click', function (event) {
...
-        document.getElementById('configuration').style.display = 'none';
-        document.getElementById('articleContent').style.display = 'none';
-    }
-    document.getElementById('navigationButtons').style.display = 'none';
-    document.getElementById('formArticleSearch').style.display = 'none';
-    document.getElementById('welcomeText').style.display = 'none';
-    document.getElementById('articleListWithHeader').style.display = 'none';
-    document.getElementById('searchingArticles').style.display = 'none';
-    document.querySelector('.kiwix-alert').style.display = 'none';
+    // uiUtil.removeAnimationClasses();
```
Jaifroid, line 402: Remove redundant comment

www/js/lib/uiUtil.js
```diff
@@ -465,80 +465,157 @@ function isElementInView (el, fully) {
         return rect.top < window.innerHeight && rect.bottom > 0 && rect.left < window.innerWidth && rect.right > 0;
     }
 }
```
Jaifroid, line 467: Best to add single blank line between functions for legibility

www/js/lib/uiUtil.js
```diff
@@ -465,80 +465,157 @@ function isElementInView (el, fully) {
...
+
+/**
+ * Returns the name of the section which is currently visible
+ * @returns {String} The name of the section which is currently visible
+ */
+function fromSection () {
+    const isConfigPageVisible = !$('#configuration').is(':hidden');
+    const isAboutPageVisible = !$('#about').is(':hidden');
+    const isArticlePageVisible = !$('#articleContent').is(':hidden');
+
```
Jaifroid, line 562: Maybe close up this gap, I'm not sure it adds any legibility.

Jaifroid (approved):
Great, many thanks for this useful PR. It cleans up a lot of code!
I did a quick test also in IE11, and it seems to be working fine and as intended (after building).
There are some very minor code formatting issues in comments below.
When you've completed everything to your satisfaction, feel free to squash and merge.

PR #1103, opened 2023-09-09
Optimize Windows testing by starting single server instance

Jaifroid:
No need to start and stop server three times.


Jaifroid:
In CI on Windows, unlike Linux, it appears that any background tasks get stopped and removed between steps in a job. This certainly appears to happen with a background `http-server` instance. It's not worth trying to work around this.

PR #1106, opened 2023-09-09
Fix active content warning for PhET ZIMs

### LINKED ISSUE #1104: Loading a PhET archive in jQuery mode does not trigger the Active Content Warning
At least for `phet_es_2023-01.zim`. This is an issue for the Firefox Browser Extension, which thereby doesn't warn users to switch to jQuery mode. It's a regression, I believe and should be fixed before releasing v3.10.0.

Jaifroid:
Fixes #1104.


Jaifroid:
Working in Chromium and Firefox Browser extensions.

PR #1108, opened 2023-09-09
Prevent redundant loading of WebP polyfill from ZIM

### LINKED ISSUE #1107: For browsers that support Service Worker but do not support WebP, block loading of WebP polyfill by ZIM
Some ZIMs now include a copy of the WebP polyfill, whereas we provide our own implementation in this app. For the small number of browsers that do not support WebP but do support ServiceWorker mode, this is a problem, because we get two (potentially conflicting) implementations of the WebP polyfill loaded.

I tested this on the BrowserStack version of Edge Legacy 18 (Edge Legacy 18 should support WebP, but the BrowserStack version seems to be an early build that doesn't, and is detected as unsupported by the WebP test). Images briefly show, but then are decoded again and disappear. The workaround is to switch to jQuery mode. However, this is not very satisfactory, and we should never load the WebP polyfill twice.

Jaifroid:
Fixes #1107.

### DIFF
--- a/service-worker.js
+++ b/service-worker.js
@@ -332,6 +332,13 @@ function fetchUrlFromZIM (urlObject, range) {
 
         var titleWithNameSpace = nameSpace + '/' + title;
 
+        // Prevent loading of WebP polyfill from ZIM because we already have a WebP decoder in the reader
+        if (/webp(?:handler|hero).*\.js$/i.test(titleWithNameSpace)) {
+            console.warn('[SW] Prevented redundant load of WebP polyfill from ZIM');
+            resolve(new Response(''));
+            return;
+        }
+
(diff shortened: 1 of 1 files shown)


Jaifroid:
So, this won't actually work without further processing, because in SW mode we use the implementation in the ZIM rather than our own version. Unfortunately, it is broken for Edge Legacy 18 at least.

Jaifroid:
Internal polyfill code works in Firefox 64 (which doesn't support WebP, but does support Service Worker). So it seems this issue is specific to Edge Legacy, which no-one is likely to be stuck on with no alternative, and the case of the Edge Legacy webview is ahandled fully by KJSW. So, not worth spending more time on this.

Issue #1120, opened 2023-09-12
Switch to Manifest V3 for Firefox


Jaifroid:
Mozilla is now accepting Manifest V3 extensions in its Store. However, unfortunately for us, the Mozilla devs are dragging their feet on the switch to (or option for) using Service Workers in place of the MV2 background script. This means that there is no technical advantage, or any real change other than changing some manifest keys, as far as we are concerned, in MV3. There is still no support for local Service Workers.

The question is, nevertheless, whether we should move with the times and update our Mozilla extension to MV3. We have already done this for the Chromium extension (which we were forced to do, since Google and Microsoft stopped accepting updates to MV2 extensions).

Pros:

* We keep the extension up-to-date with the latest standards;
* We will be ready for any shutdown of MV2 by Mozilla;
* We will be able to switch to a SW-based model easily as-and-when Mozilla enables SWs in extensions;
* We have already switched for Chromium browsers.

Cons:

* There is no technical benefit or improvement currently in adopting MV3 for Firefox due to the current limited implementation;
* No Firefox browser < 109 will be able to update or install an extension via the Store; such users will have to sideload an MV2 signed extension;
* We will have to maintain two versions of the Firefox extension, as we currently have to do for Chromium;
* We might lose users of older versions of Firefox if they perceive it to be too difficult to update their extensions manually.

My gut feeling is that we should probably not switch until we are forced to, or until there is a tangible benefit (e.g. local Service Workers). @kelson42 any thoughts / advice?

Jaifroid:
A test PR is in #1121.

Jaifroid:
An alternative to full switching might be to enable building of MV3 extensions and MV2 extensions for Firefox (including nightly), but we continue to upload the MV2 version to the Store until required to switch.

Jaifroid:
Now building both MV2 and MV3 packages for FF. But we should continue to publish the MV2 package to the Store and look into whether it is possible to sign the MV3 package separately. When we decide to switch to MV3, it will be imperative to provide a signed MV2 package that can be installed on older devices. Until we can do that, we should not switch because it would cut out a potentially large user base.

PR #1121, opened 2023-09-12
Switch to Manifest V3 for Firefox

Jaifroid:
This is for testing the MV3 extension in Firefox, as discussed in #1120.

To test, rename `manifest.json` to `manifest.cr.v3.json`, and then rename `manifest.fx.v3.json` to `manifest.json`. Then install a temporary Firefox extension in the usual way.

In sum: it works just like the old extension. It still requires using the remote PWA to run Service Workes as they are still completely disabled in Firefox MV3 extensions.

### DIFF
--- a/backgroundscript.js
+++ b/backgroundscript.js
@@ -32,8 +32,8 @@ if (typeof browser !== 'undefined') {
     // Chromium/Chrome
     genericBrowser = chrome;
 }
-
-genericBrowser.action.onClicked.addListener(function () {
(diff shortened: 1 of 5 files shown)


Jaifroid:
Hmm, it looks like we'd need two versions of backgroundscript.js given the API change...
It would be better to wait until Mozilla supports Service Workers as background scripts....

Jaifroid:
This now builds useable MV2 and MV3 versions for Firefox.

PR #1123, opened 2023-09-13
Launch spinner as soon as possible

### LINKED ISSUE #1122: In ServiceWorker mode, when clicking on a hyperlink, the spinner is displayed too late
This is normally not noticeable, but under some conditions when used offline in some browsers, it becomes obvious: there is a noticeable delay between clicking a link and the spinner showing. It is long enough to be disconcerting, or to give a sense of sluggishness to the UI.

I believe there is a solution which does not involve altering the DOM from the message channel code. Since we now trap clicks on the iframe in order to inspect whether a link is an external link or not, that same code block should invoke the spinner if it determines that the link is a ZIM link. This should restore responsiveness to the UI.

Jaifroid:
Fixes #1122.

We could also display the article title being extracted using this method.


Jaifroid:
Any further work on this should be done in #1126.

Issue #1032, opened 2023-07-08
Enable directory picking with the HTMLInputElement: webkitdirectory property


Jaifroid:
This might best be done as part of the #656, as the fallback for directory picking. Without the File System Access API, it would only be a very small win, because the user would still have to pick the directory containing ZIM archives every time they open the app. It would only be faster when selecting a different ZIM during the same session.

Jaifroid:
As mentioned, this is really a sub-issue of #656, and it may be easier to work on it holistically, because it's nearly the same underlying code for the File System Access API, to which supporting `webkitdirectory` is a kind of footnote (but an important one, because it works in Firefox, in Edge Legacy, and a number of other places where FSA doesn't work). So I'll add some comments in #656.

Jaifroid:
@Rishabhg71

Issue #656, opened 2020-10-07
Use the "File System Access API" to be able to "remember" last ZIM file


mossroy:
It is a popular demand, and was not technically possible (except on some specific devices that have specific APIs : Firefox OS and Windows Mobile).

This new API should allow to re-open the last ZIM file after closing/reopening kiwix-js.
It's very recent, and not supported by a lot of browsers (currently only Chrome >=86). Firefox is unfortunately not actively working on it for now. See https://caniuse.com/native-filesystem-api

It should be tested if it also works inside a browser extension

Jaifroid:
It works if the site is part of the Origin Trial (I did sign up for that for kiwix.github.io and have provided some feedback to Google). See the implementation in Kiwix JS Windows here (also some screenshots):

https://github.com/kiwix/kiwix-js-windows/issues/94 

I discuss some limitations in that issue, but generally it has a native feel (a few more prompts than native). It works best in a PWA when the user installs the PWA, otherwise the user would need to visit the website each time, while a PWA is an offline version of that.

mossroy:
Excellent!
Sorry, I had forgotten that you already worked on it. That's really cool.
Do you know if it works in a browser extension?

Jaifroid:
Hmm, I'm not sure (I never implemented an extension version of KJSW because I feel it's properly covered by Kiwix JS, and it would be confusing to have two different extensions built on essentially the same codebase). My understanding is that the API requires `https://` and that the `chrome-extension://` protocol used by Chrome/Edge (since Mozilla doesn't support the API at all yet) does not qualify. However, I never tested that understanding. I *think* this is an API designed to work particularly with PWAs. It is not difficult (given that we have a Service Worker) to turn Kiwix JS into a PWA -- it's just a matter of getting the app to cache its own code using the Cache API we already use in SW, and then deciding on the update strategy. But that's #388, which I think would need to be implemented here before this issue.

Jaifroid:
The Origin Trial token for kiwix.github.io seems to have finished or the token has expired (at least on Edge Chromium, for which it was working fine until very recently). :-( 
I'll see if I can renew it for continued testing.

Jaifroid:
I've now updated KJSW to use the finalized version of the API. Quite a few things had changed from the trial. I can also confirm that  for a completely fresh install of both Chrome and Edge in Windows sandbox, the Native Filesystem API is used out-of-the-box in the PWA without the user having set any flags.

[...]

PR #1127, opened 2023-09-20
Adding browser for zim library

### LINKED ISSUE #114: Add a way to easily download ZIM archives

RG7279805:
closes #114

https://github.com/kiwix/kiwix-js/assets/47270636/44329478-57f7-4bbe-9e39-01588a49a0a2

This kind of setup seems best to me if a easily visible if he want to open library + a new user don't have to look anywhere

### DIFF
--- a/www/index.html
+++ b/www/index.html
@@ -440,7 +440,9 @@
                     <br/>
                     <br/>
                 </div>
-
+                <div id='library' style="display: none; color: white;">
+                    Hello there for library
+                </div>
                 <div id="configuration" style="display: none;">
                     <div class="container">
                         <h2 data-i18n="configure-title">Configuration</h2>
@@ -472,6 +474,10 @@
                                     accept="application/octet-stream,.zim,.zimaa,.zimab,.zimac,.zimad,.zimae,.zimaf,.zimag,.zimah,.zimai,.zimaj,.zimak,.zimal,.zimam,.ziman,.zimao,.zimap,.zimaq,.zimar,.zimas,.zimat,.zimau,.zimav,.zimaw,.zimax,.zimay,.zimaz,.zimba,.zimbb,.zimbc,.zimbd,.zimbe,.zimbf,.zimbg,.zimbh,.zimbi,.zimbj,.zimbk,.zimbl,.zimbm,.zimbn,.zimbo,.zimbp,.zimbq,.zimbr,.zimbs,.zimbt,.zimbu,.zimbv,.zimbw,.zimbx,.zimby,.zimbz" />
                                 <span data-i18n="home-btn-fileselect">Select ZIM file(s)</span>
                             </label>
+                            <button class="btn btn-light custom-file-upload" id="libraryBtn">
+                                Browse ZIM Library(s)
+                                <!-- <button data-i18n="config-btn-library" id="libraryBtn"></button> -->
+                            </button>
                             <br />
                             <strong id="jqueryCompatibility" data-i18n="configure-static-content">Only ZIMs with static content (e.g. Wiki-style) are supported in JQuery mode.<br /></strong>
                             <span data-i18n="configure-supportedarchives">For information on ZIM compatibility, see</span> <a href="#usage" data-i18n="configure-about-usage-link" class="aboutLinks">About (Usage)</a>.<br />
--- a/www/js/app.js
+++ b/www/js/app.js
@@ -1296,6 +1296,12 @@ function handleFileDrop (packet) {
     document.getElementById('archiveFiles').value = null;
 }
 
+document.getElementById('libraryBtn').addEventListener('click', function (e) {
+    console.log('libraryBtn clicked');
+    e.preventDefault();
+    uiUtil.tabTransitionToSection('library', params.showUIAnimations);
+});
+
 // Add event listener to link which allows user to show file selectors
 document.getElementById('selectorsDisplayLink').addEventListener('click', function (e) {
     e.preventDefault();
--- a/www/js/lib/uiUtil.js
+++ b/www/js/lib/uiUtil.js
@@ -560,7 +560,9 @@ function fromSection () {
     const isConfigPageVisible = !$('#configuration').is(':hidden');
     const isAboutPageVisible = !$('#about').is(':hidden');
     const isArticlePageVisible = !$('#articleContent').is(':hidden');
+    const isLibraryPageVisible = !$('#library').is(':hidden');
     if (isConfigPageVisible) return 'config';
+    if (isLibraryPageVisible) return 'library';
     else if (isAboutPageVisible) return 'about';
     else if (isArticlePageVisible) return 'home';
 }
@@ -576,6 +578,7 @@ function tabTransitionToSection (toSection, isAnimationRequired = false) {
     // all the references of the sections/tabs
     const config = document.getElementById('configuration');
     const about = document.getElementById('about');
+    const library = document.getElementById('library');
     const home = document.getElementById('articleContent');
 
     // references of extra elements that are in UI but not tabs
@@ -594,14 +597,24 @@ function tabTransitionToSection (toSection, isAnimationRequired = false) {
         if (toSection === 'home') {
             if (from === 'config') slideToRight(home, config);
             if (from === 'about') slideToRight(home, about);
+            if (from === 'library') slideToRight(home, library);
+
             showElements(extraNavBtns, extraArticleSearch, extraWelcomeText, extraKiwixAlert);
         } else if (toSection === 'config') {
             if (from === 'about') slideToRight(config, about);
+            if (from === 'library') slideToRight(config, library);
             if (from === 'home') slideToLeft(config, home);
+
             hideElements(extraNavBtns, extraArticleSearch, extraWelcomeText, extraSearchingArticles, extraKiwixAlert);
         } else if (toSection === 'about') {
+            if (from === 'library') slideToRight(about, library);
             if (from === 'home') slideToLeft(about, home);
             if (from === 'config') slideToLeft(about, config);
+
+            hideElements(extraNavBtns, extraArticleSearch, extraWelcomeText, extraSearchingArticles, extraKiwixAlert);
+        } else if (toSection === 'library') {
+            // it will be always coming from config page
+            slideToLeft(library, config);
             hideElements(extraNavBtns, extraArticleSearch, extraWelcomeText, extraSearchingArticles, extraKiwixAlert);
         }
     } else {


Jaifroid:
Yes, this looks good to me! It **might** be a good idea for a new tab to appear at the top when the user presses the button, if that's possible (i.e. add one programmatically, that would disappear when leaving the page), but that's a detail that could be left till after implementation.

RG7279805:
@Jaifroid I don't think it's required at least not right now until this #523  gets completed 
or do i add it and should be removed later

Jaifroid:
You're right, no point adding it till #523, unless it's really clear that's going to take a very long time.

Rishabhg71:
@Jaifroid could you take a quick look I have added a library with working app theme 
one question how exactly a person can download a zim file (cant find download button)

Jaifroid:
It's looking good!

Regarding download, on each tile there is a small, narrow strip that says "Download" -- easy to miss. You have to click it, and not anywhere else on the tile, which is a bit counterintuitive IMHO, but it is what it is. It'a an accessibility nightmare, but not our problem directly...

I also find it problematic that the filename of the archive is only visible (in some browsers) by hovering over the tile and looking at any link preview the browser might provide. It really should be on each tile, and I've argued for this, but for some reason I really cannot fathom there is resistance to the idea... It's really difficult to distinguish amongst the many different Wikipedia offerings for this reason. Again, not our problem, directly, and patching the incoming code would create a maintenance issue.

Jaifroid:
Two small things:

* the button for browsing the library should just be in singular (it's only one library, even if it has many "books" in it);
* the button is also misaligned compared to the button next to it, by a few pixels. You might need to use Bootstrap columns (within a row) to fix that. Or it might be fixable with small style adjustments or applying bootstrap styles more carefully, or putting each button inside a div...

Jaifroid:
Ah, and clicking the download button leads to content being blocked by the sandbox, see below. To fix that, you'll have to add an onclick event to the iframe and check what's being clicked, redirect it to top-level scope so the download occurs outside the iframe. This is the same trick we use to download external links from inside a ZIM. Look for that code and if possible re-use it.

image

Jaifroid:
Look for code `params.openExternalLinksInNewTabs` round about 1601 in app.js to see how it's handled for the other iframe.

Hopefully the server permits CORS, so you can inspect the clicked link. If it's blocked, we'll need to ask the library.kiwix.org devs to add a CORS permissive header (I had to ask for this in the case of download.kiwix.org).

Jaifroid:
Final thing before I forget is that library.kiwix.org doesn't support IE11 or Edge Legacy. So I suggest when the user clicks the link, you should sniff for a function that only those two browsers support -- I use `if ('MSBlobBuilder' in window)` -- and use `download.kiwix.org` as the iframe source instead of `library.kiwix.org`.

I suggest defining a parameter in `init.js`: `params.downloadLibrary = 'https://library.kiwix.org` and another one `params.altDownloadLibrary = 'https://download.kiwix.org`, and use those params in your code (sorry, I haven't looked at your code yet, so don't know if you already define the variables in a similar way).

Rishabhg71:
I don't think we can add an event listener to any iframe content due to the same origin policy
the reason it's working on `articleContent` is because their origin is the same
we have to set a connection in the library using `window.parent.postMessage()` or maybe you could provide any idea?

image

Jaifroid:
Yes, that's why I said we might have to ask library.kiwix.org to set a CORS-permissive header. It looks like it.

Rishabhg71:
Can you tell me what exactly i have to do now?
Should i wait until library work is done

Jaifroid:
Need to check what header it is returning currently, and compare to the header returned by download.kiwix.org. You should be able to check this in the network tab of browser. Check for the CORS headers in download.kiwix.org.

It's possible even CORS permissions won't fix it, because in my code for download.kiwix.org, I get the raw HTML and then manipulate that before injecting it into a DIV. I was hoping we could avoid that with library.kiwix.org... You could try switching to `https://download.kiwix.org/zim/` just for a test to see whether you can access the links in JS. I think you should be able to. If you can, then this is fixable by asking for the CORS headers... I'll let you know how to do that when we've got the info (what CORS headers to set, etc.).

Rishabhg71:
Okay it fixed it, but do you mind telling me how can I get all the mirrors

Rishabhg71:
`mirror-sites-in.mblibrary.info` by adding this to the CSP meta tag under `frame-src`
if it's managed automatically then maybe adding `*.mblibrary.info` should work

Jaifroid:
OK, well done! So it was our sandbox rather than server-side lack of CORS policy. Well spotted.

Jaifroid:
On mirror source, I guess you'd have to experiment to see what works, it's not very transparent exactly how it works.

Rishabhg71:
I think you can review now

Jaifroid:
I'll need a bit of time -- I've got some very urgent deadlines...

Rishabhg71:
> I'll need a bit of time -- I've got some very urgent deadlines...

ok no worries

> On mirror source, I guess you'd have to experiment to see what works, it's not very transparent exactly how it works.

I have pushed a solution and I am 90% confident this will be it. I have allowed every subdomain for the mirror domain 
`*.mblibrary.info` so unless it's under the same domain I don't it its gonna break

Jaifroid:
Just quickly (without reviewing properly), I notice you haven't internationalized the new button text. Please add:

Spanish: "Biblioteca ZIM"
French: "Bibliothèque ZIM"

(I don't think we need the word for "browse" in each case, as the buttons will become too big).

Look at how it's done for the adjacent button (data-i18n attribute), and then add the appropriate strings in `en.jsonp.js`, `es.jsonp.js` and `fr.jsonp.js`.

Then test buttons in all three languages and on narrow screen sizes to ensure responsiveness.

Thanks!

Rishabhg71:
image

The button just moves to the next line and it seems good to me or do I move it inline

Jaifroid:
That's good -- it's what I'd expect for responsive behaviour. I'll try to look at this soon.

Jaifroid:
I did a quick download test and got the following (blocked):

image

Jaifroid:
If we have to list every possible mirror, then I don't think it's going to be a feasible solution... Did you test the CORS headers for library.kiwix.org?

Jaifroid:
What about turning off the sandboxing completely for this iframe? You can check where sandboxing was introduced by looking at my (close) PR here that first introduced sandboxing: https://github.com/kiwix/kiwix-js/pull/976 .

Jaifroid:
Also, did you try adding library.kiwix.org to the CSP defined in the meta http-equiv content of index.html?

Jaifroid:
Finally, as I [mentioned above](https://github.com/kiwix/kiwix-js/pull/1127#issuecomment-1728945479), for Edge Legacy and IE11, you need to load download.kiwix.org into the iframe, instead of library.kiwix.org, since the latter doesn't support old browsers. Ideally you need to test library.kiwix.org on old firefox and chrome that we claim to support, too, using BrowserStack live, to see if any of those need the alternative site. I suggested some code above on how to detect those two browsers using features rather than browser sniffing.

This is what I get when opening the library in IE11:

image

www/index.html
```diff
@@ -6,7 +6,8 @@
         <title>Kiwix</title>
         <meta name="description" content="Offline Wikipedia reader">
         <meta name="viewport" content="width=device-width, initial-scale=1.0">
-        <meta http-equiv="Content-Security-Policy" content="default-src 'self' data: https://download.kiwix.org https://master.download.kiwix.org https://browser-extension.kiwix.org https://kiwix.github.io 'unsafe-inline' 'unsafe-eval'; frame-src 'self' moz-extension: chrome-extension:; object-src 'none';">
+        <meta http-equiv="Content-Security-Policy" content="default-src 'self' data: https://download.kiwix.org https://master.download.kiwix.org https://browser-extension.kiwix.org https://kiwix.github.io 'unsafe-inline' 'unsafe-eval'; frame-src 'self' moz-extension: chrome-extension: https://library.kiwix.org/ download.kiwix.org *.mblibrary.info; object-src 'none';">
```
Rishabhg71, line 9: already tried `https://library.kiwix.org/` and `library.kiwix.org` but i will retry

Jaifroid: I've been thinking that you might be able to turn this off completely, programmatically, when you load the library. See https://stackoverflow.com/questions/18982228/how-to-add-meta-tag-in-javascript . It seems it can be manipulated.

Rishabhg71:
> I did a quick download test and got the following (blocked):
image

I will try to allow every mirror by some other method

> If we have to list every possible mirror, then I don't think it's going to be a feasible solution... Did you test the CORS headers for library.kiwix.org?

Not sure but maybe `*` this could work

>  Also, did you try adding library.kiwix.org to the CSP defined in the meta http-equiv content of index.html?

 Yes I did and it is already there too (with `http://`)

> Finally, as I https://github.com/kiwix/kiwix-js/pull/1127#issuecomment-1728945479, for Edge Legacy and IE11, you need to load download.kiwix.org/zim/ into the iframe, instead of library.kiwix.org, since the latter doesn't support old browsers. Ideally, you need to test library.kiwix.org on old Firefox and Chrome that we claim to support, too, using BrowserStack live, to see if any of those need the alternative site. I suggested some code above on how to detect those two browsers using features rather than browser sniffing.

I will start working on this

www/index.html
Jaifroid: But try the `*` solution first!

Rishabhg71:
@Jaifroid While working on the IE issue I encountered some kind of bug or something

When loading the app on the Edge IE mode/IE 11 it seems like the app is broken
tried both `npm run serve` and `npm run build` -> `npx http-server`
image
image

After checking the console there is no error over there but when I checked the server logs via `npx http-server`
I am getting this 404 on every refresh 
```shell
[2023-09-23T13:08:32.427Z]  "GET /www/js/app.js" "Mozilla/5.0 (Windows NT 10.0; WOW64; Trident/7.0; rv:11.0) like Gecko"
[2023-09-23T13:08:32.429Z]  "GET /www/js/webpHeroBundle_0.0.2.js" "Mozilla/5.0 (Windows NT 10.0; WOW64; Trident/7.0; rv:11.0) like Gecko"
[2023-09-23T13:08:32.430Z]  "GET /www/js/webpHeroBundle_0.0.2.js" Error (404): "Not found"
[2023-09-23T13:08:32.440Z]  "GET /www/img/icons/kiwix-60.png" "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/117.0.0.0 Safari/537.36 Edg/117.0.2045.36"
[2023-09-23T13:08:32.450Z]  "GET /www/img/icons/kiwix-120.png" "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/117.0.0.0 Safari/537.36 Edg/117.0.2045.36"
[2023-09-23T13:08:32.458Z]  "GET /www/img/icons/kiwix-256.png" "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/117.0.0.0 Safari/537.36 Edg/117.0.2045.36"
```
so I tried to fix this issue by changing this [line](https://github.com/kiwix/kiwix-js/blob/add-zim-library/www/js/init.js#L223C13-L223C13) but then I am getting an error in the console and don't know what to do after that
image



The even weirder thing is that when I run the test via `npm run test-e2e-iemode` it seems to run all good and page load is fine

Jaifroid:
Those browsers don't support ES6 modules, so you have to build the app and then access it from the dist directory.

So you have to change the address (in the browser) to, e.g. `http://localhost:8080/dist/`.

Jaifroid:
Or just use `npm run preview` which builds the app and then serves the `dist` automatically.

Rishabhg71:
> Finally, as I [mentioned above](https://github.com/kiwix/kiwix-js/pull/1127#issuecomment-1728945479), for Edge Legacy and IE11, you need to load `download.kiwix.org/zim/` into the iframe, instead of library.kiwix.org, since the latter doesn't support old browsers. Ideally you need to test library.kiwix.org on old firefox and chrome that we claim to support, too, using BrowserStack live, to see if any of those need the alternative site. I suggested some code above on how to detect those two browsers using features rather than browser sniffing.
> 
> EDIT: We need to find out what is failing in those browsers that are not supported, and test support in the browser for it instead of relying on browser quirks or browser sniffing.
> 
> This is what I get when opening the library in IE11:


I have tried to use `iframe.onerror = function (){}` but it's not invoked when `library.kiwix.com` is unable to load properly and when I tried to load iframe (`library.kiwix.com`) on older browsers like Chrome 58 and Chrome 60 both of them didn't work
but at least on Edge and IE11 using `if ('MSBlobBuilder' in window)` is loading `download.kiwix.com` properly
Is there any other way to check if the iframe is loaded properly or do I just check the browser version and set `libraryUrl`?
@Jaifroid

Jaifroid:
MSBlobBuilder method is useful if you determine that IE11 and Edge Legacy are the only browsers that cannot display library.kiwix.org, but I think there might be others that we claim to support. NB you don't have to test this in your implementation, you can simply test whether these browsers (and old Firefox / Chrome instances) can open the web page library.kiwix.org on BrowserStack -- you have full access to the address bar of the browsers, and also to their DEV tools. Check in DEV tools what it is (if anything) that is throwing an error, probably some unsupported method. If you can identify one that is common, then you can test for that in your implementation, and simply load the appropriate site (library... or download...) rather than load it first and try to catch an error. I hope that makes sense!

Jaifroid:
> do I just check the browser version and set `libraryUrl`

Yes, this is what I mean -- test for whatever tells you that the browser won't support library.kiwix.org, then set the URL as appropriate (before loading anything into the iframe).

Rishabhg71:
Or maybe if i reask this question 
My task to to test which browser version are not supported and load dowload.kiwix on those version
Example: i know chrome 60 is not supported so anything below version 60 should load download.kiwix

If yes wouldn't it be bad that we directly hard code a version number?

Jaifroid:
> so my goal is to find a method that could be causing the problem and then just check for the method

I'm just suggesting you try to find if there is a JS method that the browsers that fail do not support. I'm not saying that you should test for this method after you load the library. I'm saying you test whether the browser supports it before you load. For example, say it was the `Fetch` API that is not supported, you simply test whether the currently running browser supports the Fetch API (something like `if ('Fetch' in window) { // Supported } else { // Not supported }`. If it's supported, load library.kiwix.org. If unsupported, load download.kiwix.org.

In general, for JS apps it is much better to use feature detection (what I'm suggesting) and NOT browser sniffing. You don't know every possible browser a user might be using to run the app. But if you're using Feature detection, then you don't care which version is being run, you just use the fallback if the feature isn't supported.

Jaifroid:
This might be useful: https://developer.mozilla.org/en-US/docs/Learn/Tools_and_testing/Cross_browser_testing/Feature_detection

Rishabhg71:
Ohh, OK I finally understood, and sorry for it taking me time 🧠

Rishabhg71:
I feel this is now ready for another review

Jaifroid:
I confirm it's working well in IE11 and in modern Chrome -- downloads no longer seem to be blocked (in either browser). Well done! I haven't had a chance to test more widely. I'll take a look at the code.

www/index.html
```diff
@@ -6,7 +6,8 @@
         <title>Kiwix</title>
         <meta name="description" content="Offline Wikipedia reader">
         <meta name="viewport" content="width=device-width, initial-scale=1.0">
-        <meta http-equiv="Content-Security-Policy" content="default-src 'self' data: https://download.kiwix.org https://master.download.kiwix.org https://browser-extension.kiwix.org https://kiwix.github.io 'unsafe-inline' 'unsafe-eval'; frame-src 'self' moz-extension: chrome-extension:; object-src 'none';">
+        <meta http-equiv="Content-Security-Policy" content="default-src 'self' data: https://download.kiwix.org https://master.download.kiwix.org https://browser-extension.kiwix.org https://kiwix.github.io 'unsafe-inline' 'unsafe-eval'; frame-src 'self' moz-extension: chrome-extension: *; object-src 'none';">
```
Jaifroid, line 9: Hmm, adding '*'  to frame-src is like not having any Content Security Policy on the iframe at all, and becasue this will also apply to the iframe that displays ZIM contents, it's not acceptable to keep this here permanently. We had to introduced the CSP on the iframe for a specific security issue, and we can't allow code within a malicious ZIM arbitrarily to set the iframe to anything it wants. That way, the contents of the iframe can be spoofed, and the user will have no idea.

I think you have to find a way to add and remove this as needed. It definitely must be removed if a user is browsing a ZIM. Or, more easily, add it before opening the library, and remove it when the user returns to any of the other tabs.

www/js/app.js
```diff
@@ -153,11 +153,37 @@ function resizeIFrame () {
         }, 100);
     }
 }
+
+/**
+ * Resize the IFrame height, so that it fills the whole available height in the window
+ */
+function resizeLibrary () {
```
Jaifroid, line 160: Hmm, this function seems to be a direct copy of `resizeIframe()`, with some modifications. This is not an efficient way to repurpose that code. You should instead call the exisitng `resizeIframe()` funciton and if you need to run specific code for the library within it, you should pass a string like `library` (or better, the id of the iframe to resize) to it, so you can detect we're resizing the library. That should be pretty easy to do.

www/js/app.js
```diff
@@ -153,11 +153,37 @@ function resizeIFrame () {
...
+        }, 100);
+    }
+}
 document.addEventListener('DOMContentLoaded', function () {
     getDefaultLanguageAndTranslateApp();
     resizeIFrame();
+    resizeLibrary();
 });
 window.addEventListener('resize', resizeIFrame);
+window.addEventListener('resize', resizeLibrary);
```
Jaifroid, line 183: Never add two listeners for the same event. Instead, add a function to the event listener that branches the code, or runs `resizeIframe()` twice, once with the id of the ZIM iframe, and once with the id of the library iframe.

www/js/init.js
```diff
@@ -29,7 +29,40 @@
...
+ * @property {boolean} openExternalLinksInNewTabs - A global parameter to turn on/off opening external links in new tab (for ServiceWorker mode).
+ * @property {string} overrideBrowserLanguage - A global language override.
+ * @property {boolean} disableDragAndDrop - A parameter to disable drag-and-drop.
+ * @property {string} referrerExtensionURL - A parameter to access the URL of any extension that this app was launched from.
+ * @property {boolean} defaultModeChangeAlertDisplayed - A parameter to keep track of the fact that the user has been informed of the switch to SW mode by default.
+ * @property {string} contentInjectionMode - A parameter to set the content injection mode ('jquery' or 'serviceworker') used by this app.
+ * @property {boolean} useCanvasElementsForWebpTranscoding - A parameter to circumvent anti-fingerprinting technology in browsers that do not support WebP natively by substituting images directly with the canvas elements produced by the WebP polyfill.
+ * @property {string} libraryUrl - The URL of the Kiwix library or Download page.
+ * @property {DecompressorAPI} decompressorAPI - The URL of the Kiwix library.
```
Jaifroid, line 52: Looks good! I like this kind of documentation that a good IDE can pick up.

www/js/init.js
```diff
@@ -119,6 +153,15 @@ params['useCanvasElementsForWebpTranscoding'] = null; // Value is determined in
             if (kiwixComm) kiwixComm.remove();
         }, 3000);
     }
+
+    // old browsers dont support syntax for optional catch binding (https://caniuse.com/mdn-javascript_statements_try_catch_optional_catch_binding)
+    // this makes library.kiwix.org not load up properly making us fallback on download.kiwix.org for the iframe
```
Jaifroid, line 158: Great that you found this!

Jaifroid (changes requested):
There's lots of great code here, well done! There are a few niggly things I think need doing as outlined in review comments. In particular, programmatically manipulating the CSP could be challenging. I wonder if there is a way to set the frame-src specifically for the library iframe, as that would solve it without having to add or remove the * programmatically, so that should be checked first. I think the other minor things are easy but let me know if you have any queries.

But overall, it's looking very good and will be a great improvement to the app.

www/index.html
```diff
@@ -440,7 +441,9 @@ <h3 id="other" data-i18n="about-other-platforms">Other platforms/versions</h3>
                     <br/>
                     <br/>
                 </div>
-
+                <div id='library' style="display: none; color: white;">
+                    <iframe frameborder="0" width="100%" id="libraryIframe"></iframe>
```
Jaifroid, line 445: I think the ID of this iframe should be `libraryContent` to mirror the other iframe's ID `articleContent`.

www/js/lib/zimfile.js
```diff
@@ -54,14 +54,6 @@ if (!String.prototype.startsWith) {
...
 }
 
-/**
- * A global variable to track the assembler machine type and the last used decompressor (for reporting to the API panel)
- * This is populated in the Emscripten wrappers
- * @type {Object}
- * @property {String} assemblerMachineType The assembler machine type supported and/or loaded by this app: 'ASM' or 'WASM'
- * @property {String} decompressorLastUsed The decompressor that was last used to decode a compressed cluster (currently 'XZ' or 'ZSTD')
- * @property {String} errorStatus A description of any detected error in loading a decompressor
- */
```
Jaifroid, line 64: I see you've moved the typedef to init.js, which probably makes sense (though you've given it the wrong description there). However, please leave the explanation of this object property (lines 57-59), and refer the dev to the object definition in init.js. Also change "A global variable" to "A global parameter object to track".

www/index.html
Jaifroid: It may also be possible to load a local html document with its own separate CSP into the library iframe, and have a nested iframe in that document that loads https://library.kiwix.org or https://download.kiwix.org. Because it's local, you would be able to access the content of the "outer" iframe to set the frames `src` property programmatically.

i18n/en.jsonp.js
```diff
@@ -11,6 +11,7 @@ document.localeJson = {
       "home-btn-random": "Random",
       "home-btn-back": "Back",
       "home-btn-fileselect": "Select ZIM file(s)",
+      "home-btn-library": "Browse ZIM Library",
```
Jaifroid, line 14: Mea culpa, and it's my mistake, but I've just realized that neither the fileselect nor the library buttons are on the Home page. They are on the Confiuration page, so both of these should really be `configure-btn-library` and `configure-btn-fileselect`, and they should be moved directly beneath `configure-about-usage-link` below. Of course, you'd need to do the same for the other language files, and rename the `data-i18n` attribute of both buttons in `index.html`.

www/js/init.js
```diff
@@ -119,6 +153,15 @@ params['useCanvasElementsForWebpTranscoding'] = null; // Value is determined in
...
+    try {
+        // eslint-disable-next-line no-eval
+        eval('try {} catch {}');
+    } catch (error) {
+        params['libraryUrl'] = 'https://download.kiwix.org/zim/';
```
Jaifroid, line 163: For developer visibility, it would be a good idea not to hard-code the address of download.kiwix.org/zim/ here, but to assign it to another parameter like `params.altLibraryUrl` which you define above as `params['altLibraryUrl']` beneath the main libraryUrl definition. Then you can move this test code to a better place than init.js, e.g. into the main function that opens the library.

www/js/init.js
Rishabhg71: did a line copy and left it there by mistake

Rishabhg71:
After trying both of the options given above these are my findings 

- option 1: updating CSP on the go via `document.querySelector('meta[name="csp"]').setAttribute('content', '*')`
    The reason this is not working is (Not confirmed) whenever we load the page for the first time browser remembers our policy hence making policy changes doesn't actually affect the page itself (DOM updated but still getting content blocked)

- Option 2: Using nested iframes seems like this is a good idea but I don't know why every time I try to load `kiwix.library.org` in an iframe in the new page `www/library.html` the page keeps on loading with anything in frame (tried changing CSP)

I feel like option 2 could be implemented but I ask you if it's possible for you to create a new page and just load the library into a iframe. I will take care from there

@Jaifroid

Jaifroid:
I'm up to my eyeballs in exams, and can't experiment right now, unfortunately. You are setting the address with `https://library.kiwix.org` (i.e. you're not omitting the `https://` part...)? Sorry for naive question, probably, but wanted to double check browser isn't looking for a local page that gets routed through the Service Worker or something....

Also, make sure the new page is listed in the Service Worker. Other than that, I have no clue, really. Start with a really simple test case and build it up.

Rishabhg71:
Ok i will try my best then

Rishabhg71:
@Jaifroid 

>Option 2: Using nested iframes seems like this is a good idea but I don't know why every time I try to load kiwix.library.org in an iframe in the new page www/library.html the page keeps on loading with anything in a frame (tried changing CSP)

UPDATE: After hours of headache I finally know the root cause of this issue. When I created the `library.html` I also appended a script tag at the end to resize the iframe. That script tag was caused by the infinite load for the page. refer to this Stackoverflow post
it is similar to my situation https://stackoverflow.com/questions/38204722/infinite-loading-with-script-tag-inside-iframe

I can finally go on and try to fix the real issue in hand (CSP)

Jaifroid:
Ah, thanks for the update! Yes, sometimes it can be so tricky when you spend hours hunting for the cause of a bug, and it eludes you. Well done for finding it finally.

Just so you're aware, there's a good reason always to keep scripts and HTML separate: Chromium extensions automatically block (with no workaround) inline scripts, so they have to be run as attached scripts instead. However, since you're loading a local piece of HTML into the iframe, you have full access to it from the app's code, and there is no need to have local scripting in the iframe. It's the way we manipulate the contents of the ZIM iframe: we can do anything to the HTML in there, add event listeners to links, etc Obviously you won't be able to touch the content of the nested iframe, if you're still planning on nesting.

An alternative to nesting (but it could be a lot more code) is to use XMLHttpRequest to get the HTML, and add it to the local page you load in the iframe. Then you wouldn't have any CORS issues. But it's not such a simple solution, and there's a lot of dynamic stuff that might not run correctly that way.

Rishabhg71:
I feel like I may have celebrated a little early as the infinite load problem still persists. I think let's leave this CSP mess and maybe try out other solutions 
One solution that I think could be best is i directly edit the code for library.kiwix.org and maybe add a proper solution over there (if yes could tell which repo has that code?)

>An alternative to nesting (but it could be a lot more code) is to use XMLHttpRequest to get the HTML and add it to the local page you load in the iframe. Then you wouldn't have any CORS issues. But it's not such a simple solution, and there's a lot of dynamic stuff that might not run correctly that way

Could be done but I will put this as a last resort

Rishabhg71:
Finally, it's fixed and the download is working too. I have used nested iframes to load the page and allow downloads from any URL
But it created an issue now clicking on any link from `download.kiwix.org` (IE mode) opens the file in the browser instead of downloading

> An alternative to nesting (but it could be a lot more code) is to use XMLHttpRequest to get the HTML, and add it to the local page you load in the iframe. Then you wouldn't have any CORS issues. But it's not such a simple solution, and there's a lot of dynamic stuff that might not run correctly that way.

Do I try to load `download.kiwix.org` by this method? @Jaifroid

Jaifroid:
> Finally, it's fixed and the download is working too.

Hooray! I'll take a look!

> But it created an issue now clicking on any link from download.kiwix.org (IE mode) opens the file in the browser instead of downloading.

That sounds fixable, and probably not with XMLHttpRequest, which is quite a complicated way of doing it, though there is some code that already does this over at KJSWL, so not out of the question. But let's see if it can be fixed more simply first. I'll try it out and see if I have any ideas.

Jaifroid:
Just so you're aware, I get these unhandled exceptions on startup. I'll carry on reviewing, but it would be good to handle this in the resizeApp function.

image

Jaifroid:
OK, and when I click the Browser Library button with no ZIM loaded, I get the exception below, and the library doesn't show (completely blank page). I think this needs fixing before I carry on reviewing. Please be sure to test very thoroughly, with and without ZIMs loaded, in both ServiceWorker and JQuery modes!

image

Rishabhg71:
image
weird as it's working fine here and my working tree is clean

Jaifroid:
Strange, I just double-checked, made sure I'm completely up-to-date, did a full app reset, etc., and I still get the error. See video below.

Loading_errors

Rishabhg71:
I can assure you something is wrong on your end 
In logs, i can see there is a message saying library.html not found (404) but you can clearly see that that i have just added this file here
https://github.com/kiwix/kiwix-js/pull/1127/files#diff-89bc8b1d46100e91e15abb3172b75324bc49201cd40da88fffffa4b368386bfe

Jaifroid:
Hmm, I can see that file in my FS, but the browser says the server returned 404. Very odd. I'll investigate.

www/index.html
```diff
@@ -440,7 +440,9 @@ <h3 id="other" data-i18n="about-other-platforms">Other platforms/versions</h3>
                     <br/>
                     <br/>
                 </div>
-
+                <div id='library' style="display: none; color: white;">
+                    <iframe width="100%" id="libraryContent" src="/www/library.html"></iframe>
```
Jaifroid, line 444: I think the problem is that you're loading this with an absolute URL (starting with `/www`), but we can't always be sure what that is going to serve. You should use a URL that is relative to the document loading the iframe, i.e. use `library.html`, and test that this works. This is probably the issue. I'm not serving from the same base as you are. I'm serving from my `Repos` directory, not from inside the `kiwix-js` directory. So, the file can't be found at `Repos/www/library.html`. Always use fully relative references, because you can't control where the base directory is in relation to your content!

Rishabhg71:
> That sounds fixable, and probably not with XMLHttpRequest, which is quite a complicated way of doing it, though there is some code that already does this over at KJSWL, so not out of the question. But let's see if it can be fixed more simply first. I'll try it out and see if I have any ideas.

Any update?

Jaifroid:
Oh, I didn't know you were waiting for another look. I don't get notified of every push to a PR.

Jaifroid:
Hi, did you test this in a Chromium extension? The old library (download.kiwix.org) loads in that context, instead of the new one. This is undoubtedly because you're using `eval` to test for browser support, and `eval` is blocked in Chromium extensions. On the other hand, if it's blocked in the test code, it's probably blocked in the library itself too. If that's the case, then your code is working as expected. However, if the library would actually work in a Chromium extension, then the code might need finessing. Are they actually using `eval` in the library.kiwix.org code? Seems like poor coding! Even our bundler flags it as strongly discouraged:

image

Jaifroid:
OK, I just tested changing the altLibraryUrl to library.kiwix.org, and the library works fine in a Chromium extension, so I think that indicates you need to find a different test, not using eval.

Jaifroid:
Please also test in a Firefox MV3 extension, as I suspect `eval` may be forbidden in that context too (though I think `eval` is OK, but discouraged, in an MV2 extension).

Jaifroid:
Interestingly, downloading from the old `download.kiwix.org` seems to work fine in the Chromium extension, so it's only in IE11 (and possibly Edge Legacy) that downloads are failing.

Jaifroid:
Regarding IE11, what happens if, instead of loading `https://download.kiwix.org/zim/` into the nested iframe inside `library.html`, you instead load it into the top-level iframe (inside `index.html`)? Do this by changing the `src` programmatically. Since `download.kiwix.org` is already in the global CSP, it's possible this could work, depending on how download links are handled. If the browser sees them all as coming from download.kiwix.org, they might not be blocked. If they redirect explicitly to the mirror, then we have the same issue before and this solution won't work. It's worth a try, though.

Rishabhg71:
> Regarding IE11, what happens if, instead of loading https://download.kiwix.org/zim/ into the nested iframe inside library.html, you instead load it into the top-level iframe (inside index.html)? Do this by changing the src programmatically. Since download.kiwix.org is already in the global CSP, it's possible this could work, depending on how download links are handled. If the browser sees them all as coming from download.kiwix.org, they might not be blocked. If they redirect explicitly to the mirror, then we have the same issue before and this solution won't work. It's worth a try, though.

Just tried this method and it seems like we are again on square one. If I try to set src (`download.kiwix.org`) directly into the parent iframe CSP blocks all the mirrors 

@Jaifroid

Jaifroid:
OK, so it looks like we're stuck with the XHR method. Look at the function `readRemoteArchive()` in `app.js`. Change this to something more generic like `readRemoteUrl`, and maybe move it to `uiUtil.js` (optional, but see #507 about why we need to reduce the size of `app.js` if possible). Note that you don't want to get a BLOB out of this, so you'll have to make the `responseType` a parameter -- you can probably get away with not setting it for the `download.kiwix.org` use case, and only setting it to `blob` when that parameter is present in the function call.

The aim is to get the HTML from `https://download.kiwix.org/zim/` and set the contents of the (outer) iframe to that HTML. In other words, you do something like `libraryIframe.innerHtml = html` (where `html` is the HTML received from the XHR request. Then you'll need to trap the clicks on any URLs for subfolders: if the user has clicked on a subfolder, then you need to get the HTML from the new URL using `readRemoteUrl`, and load that into the iframe. However, if the user clicks on a ZIM link (recognizable because the `<a href="..."` property will end in `.zim`), then you simply want to launch that URL in a new browser tab, i.e.  (outside of our CSP). You should be able simply to set `target="_blank"` in the ZIM URL to achieve this.

It sounds a bit complicated, but I don't think it need be.

Rishabhg71:
Understood i will go on and try to prototype it real quick

Jaifroid:
By the way, you can use the method for trapping external URLs to do the link processing: look for `params.openExternalLinksInNewTabs` round about 1601 in `app.js` to see how it's handled for the other iframe. This is simpler than adding an event listener to every anchor in the iframe, as it just traps clicks on the iframe.

Jaifroid:
BTW, if you're not familiar with injecting HTML into an iframe, take a look at `app.js` round about 1821. You'll see you need to inject it into `iframe.contentDocument.documentElement`, not direclty into the iframe itself. And you need to test that `contentDocument` exists before finding `documentElement`.

Rishabhg71:
> Hi, did you test this in a Chromium extension? The old library (download.kiwix.org) loads in that context, instead of the new one. This is undoubtedly because you're using eval to test for browser support, and eval is blocked in Chromium extensions. On the other hand, if it's blocked in the test code, it's probably blocked in the library itself too. If that's the case, then your code is working as expected. However, if the library would actually work in a Chromium extension, then the code might need finessing. Are they actually using eval in the library.kiwix.org code? Seems like poor coding! Even our bundler flags it as strongly discouraged:

Yes, Chromium doesn't support `unsafe-eval` but it seems to be the only way to check if a browser supports `library.kiwix.org`
I am not checking for the execution of `eval` but rather checking if the browser supports [optional catch binding ](https://caniuse.com/mdn-javascript_statements_try_catch_optional_catch_binding)
Any browser that doesn't support `optional catch binding` can't run `library.kiwix.org`

I am trying to find any other way to check if the library works on which browser but I am unsure if there is any other way at all
as if `optional catch binding` is not supported `library.kiwix.org` will crash  

@Jaifroid

Rishabhg71:
For now, I am trying to find any other way to test `optional catch binding` (I have low hopes for this ) If this works then it's all good else I will go with a mix of options 2 and 3

Jaifroid:
I tried this:

```
try {
    // eslint-disable-next-line no-new-func
    var tryCatch = new Function('try {} catch {}');
    tryCatch();
    nestedFrame.setAttribute('src', params.libraryUrl);
} catch (error) {
    nestedFrame.setAttribute('src', params.altLibraryUrl);
}
```

It still doesn't work in Chromium extension (because new-func is considered the same as eval, and is blocked), but at least rollup.js doesn't protest....

And we can perhaps put up with Chromium loading the old library in local extensions (since we direct users to use the browser-extension.kiwix.org version in any case). Or you could test for Chromium >=66 and let it through...

Jaifroid:
Found this workaround for `eval` in MV3 extensions: https://medium.com/geekculture/how-to-use-eval-in-a-v3-chrome-extension-f21ca8c2160c . Since you're loading the library html file in any case, you could perhaps sandbox it and use the suggested approach to have an eval function inside it, which you access via postmessage. It's a bit hacky, but could work. Of course putting the library html in a sandbox would complicate setting the nested iframe's URL, which would also have to be done by postmessage... And you'd have to test that putting it all in a sandbox doesn't then cause issues with the download. Not sure if all this is worth it, but it's one possibility.

www/library.html
```diff
@@ -0,0 +1,15 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta charset="utf-8" />
+    <title>Placeholder for injecting an article into the iframe</title>
+    <meta
+      http-equiv="Content-Security-Policy"
+      content="default-src 'self' data: * 'unsafe-inline' 'unsafe-eval'; frame-src 'self' moz-extension: chrome-extension: *; object-src 'none';"
+    />
+    <script src="./js/library.js" ></script>
```
Rishabhg71, line 10: It seems like when we build the actual app all the files are moved from `js/lib/*` to `js/*` which seems to cause a problem where when we `npm run preview` I get a 404 since there is no file at `js/library.js`

Could you tell me how to change this path only during compile time from `js/lib/library.js` -> `js/library.js`

I have already tried `replace()` from `rollup.config.js` but can't make it work @Jaifroid

Jaifroid:
> * Add an extra iframe hidden all the time and only used to check if `eval` works

I imagine this could be one solution to avoid sandboxing the iframe you really want to work with. It shouldn't have much overhead, but maybe only if the method below doesn't work, or if sandboxing makes things far too difficult otherwise.

> * Use `library.html` for everything from loading `library.kiwix.org`/`download.kiwix.org` and handling `eval` but we lose any to interact with the parent (except `window.postMessage()`)
> 
> I like the second method and trying to implement it right now

I agree with you that this would be the more elegant solution, but the first one could be used if it seems impractical.

www/library.html
Jaifroid: All JS files should be bundled into bundle.js in rollup.config.js. I think you only need to add an import statement at the top of a file that uses it.

Jaifroid: Ah sorry, I didn't look at the code, and I think this script file is only used inside library.html. In that case, you don't need to (can't) turn it into a module or bundle it, you can instead put it in the same folder as library.html and ensure both are copied by rollup. In fact if both are in www, I think they should be copied automatically, and you don't need to touch the rollup config.

Jaifroid: If you don't like putting it in the same folder (i.e. in `www`, not in a subfolder), then you would have to do a search-and-replace inside the rollup config, which would search for the path `./js/lib/library.js` and replace it to `./js/library.js`, when you copy `library.html`. There are a few examples of doing such a search-and-replace in rollup config (a simple javascript `.replace` function can do it).

Note that you have to do it in two places, in the PRODUCTION (minified) buiild, and also in the normal (unminified) build. There are two code blocks in rollup config, one for production and the other for normal. There is a lot of duplicated code in the rollup config, which I'm not at all proud of, but it kind of built up like this from a small amount, and then mushroomed. I aim to find a way to share repeated code when I have a chance.

rollup.config.js
```diff
@@ -40,7 +40,7 @@ const config = {
         }),
         copy({
             targets: [{
-                src: ['www/js/lib/*dec-wasm.wasm', 'www/js/lib/libzim-asm.js', 'www/js/lib/libzim-wasm.*', 'www/js/lib/webpHeroBundle*',
+                src: ['www/js/lib/*dec-wasm.wasm', 'www/js/lib/library.js', 'www/js/lib/libzim-asm.js', 'www/js/lib/libzim-wasm.*', 'www/js/lib/webpHeroBundle*',
```
Jaifroid, line 43: If you place the file directly in `www`, you won't need this addition. If you decide to keep it in `lib`, then as well as copying it, you will need to do a search-and-replace inside `library.html` to change the path, because we don't keep `lib`.

Rishabhg71:
Again I have faced a problem that is nearly impossible to fix/overcome without a weird hack
When I implemented sandboxing for the Chrome extension `library.kiwix.org` failed to load properly since it needed cookies to load properly. here is an error traceback
```
Uncaught DOMException: Failed to read the 'cookie' property from 'Document': The document is sandboxed and lacks the 'allow-same-origin' flag.
    at getCookie (https://library.kiwix.org/skin/index.js?cacheid=07b06fca:73:37)
    at https://library.kiwix.org/skin/index.js?cacheid=07b06fca:24:19
    at https://library.kiwix.org/skin/index.js?cacheid=07b06fca:594:3
```

I am writing a summary of what I have tried and measures to fix the issue but every time i get back to square one after some time there will be a platform that will not be supported

|  Issue | Fix  | Why didnt work | Browsers affected
| :------------: | :------------: | :------------: | :------------: |
| CSP blocking mirror downloads | Using `*` as CSP |  Not a safe option at all | All browsers|
| CSP is not safe so making it safer | Using nested iframes where the inner one will be looser CSP  |  Nested frames don't start downloading in some browsers (opens as text)(only `download.kiwix`) | IE and legacy browsers |
| Finding a method to check which browsers support `library` | Using `try {} catch {}` |  Eval not allowed in Chrome manifest v3 | Chrome extension (PWA) |
| Eval not allowed | Sandboxing the iframe and setting |  Cant control the nested frame DOM due to sandbox | chrome extension (PWA) |
| No control over Inner DOM | Using `window.postMessage()` to share data | Raise issue with the cookies (seems like cookies are not allowed in sandbox due to using `eval` + `allow-same-origin` is not supported by manifest v3)  | chrome extension (PWA) |

This is the short summary of whatever I have faced yet + hours of bugs. I feel like it's near impossible to support most browsers/PWA to load `library.kiwix.org`. My suggestion is to support browsers that can load easily and leave the Chrome extension (library support). We should go back to the nested frame approach without sandbox and support chrome extension via `XMLHttpRequest` loading `download.kiwx.org`

Jaifroid:
@Rishabhg71 I'm sorry you're facing such obstacles. Well done for researching it so carefully. I wonder if cookies are a red herring, though -- if you download the HTML using `xmlHttpRequest` it certainly doesn't require any cookies. You can even insert that HTML into a div (or a Bootstrap panel) instead of an iframe if it helps.

However, it seems things are getting too complicated, so what I suggest is that you focus on the `library.kiwix.org` browser for those browsers that support it, and for the others, just open `download.kiwix.org` in a new browser tab, i.e., outside the app , and hence outside the sandbox.

Would that be workable?

Jaifroid:
Oh, sorry, I see the cookies are with library.kiwix.org, not download.kiwix.org. With nested iframes it should work, though, right?

Rishabhg71:
> I'm sorry you're facing such obstacles. Well done for researching it so carefully. 

Well, no worries I mean it's part of the job and learning new things. It's just bothersome to support so many browsers and platforms and I genuinely respect you for doing this for years 🫡

> I wonder if cookies are a red herring, though -- if you download the HTML using XMLHttpRequest it certainly doesn't require any cookies. You can even insert that HTML into a div (or a Bootstrap panel) instead of an iframe if it helps.

Yes, I feel the same thing doing this would be really complicated, and that too doesn't give a guarantee that it's going work 
I would rather make a small clone of `library.kiwix.org` inside the app (if you think it a good idea)

> However, it seems things are getting too complicated, so what I suggest is that you focus on the library.kiwix.org browser for those browsers that support it, and for the others, just open download.kiwix.org in a new browser tab, i.e., outside the app, and hence outside the sandbox.

Let's go with this, Please confirm and I will start this

> Oh, sorry, I see the cookies are with library.kiwix.org, not download.kiwix.org. With nested iframes it should work, though, right?

Well no, With my current knowledge when we enable a sandbox that iframe loses access to `document.cookies` (read/write) this can be overcome by adding `<iframe src="https://example.com" sandbox="allow-same-origin"></iframe>` But in case of chrome extension (manifest v3) doesn't have any `allow-same-origin` policy making it so any iframe will not have access to cookies. I mean wouldn't want an unsafe (using eval) extension to access cookies on any site

Jaifroid:
Happy to confirm going with the suggested path, and thanks for further explanation. So, I suggest that if it is Chrome in a local extension, which basically means Chrome in jQuery mode OR in localServiceWorker mode, we should open `library.kiwix.org` in a new browser tab. If it's Chrome using the browser-extension.kiwix.org code (which we strongly encourage users to do), then you should have no problems, because no code is blocked in that mode (`eval`, etc. all run fine).

Let me know if you don't follow that hasty explanation, or if I got anything wrong.

rollup.config.js
```diff
@@ -7,13 +7,28 @@ import copy from 'rollup-plugin-copy';
...
+    // input: ['www/js/app.js', 'www/js/lib/library.js'],
+    input: {
+        bundle: 'www/js/app.js',
+        library: 'www/js/lib/library.js'
     },
+    output: [
+        {
+            dir: 'dist/www/js',
+            // file: 'bundle.js',
+            format: 'umd',
```
Rishabhg71, line 22: To add use `library.js` we need to compile it using rollup too, so I have updated the config but it seems like the format `iife` and `umd` is not supported when compiling multiple files, I have limited knowledge when dealing with bundlers can you help a little here. Is this the right way to compile multiple files?

Jaifroid: @Rishabhg71  No, you don't want to compile this file. It's ONLY used (if I've understood correctly) by the html document you will load into the iframe. It's not used by the main application. Rollup config is only for the main application. You will just load this file directly, not as part of any bundle.

Rishabhg71: I am trying that but seems like some of the features of js are not available in IE11 like async functions. In that case i will just legacy JS code

Jaifroid: For a small file like that, it's much easier just to write conformant code than to try to configure Babel. Note that Promises are not natively supported in IE11 or Firefox OS (another old one we still support), and so you may just have to use old-style callbacks if you are using anything asynchronous. However, not sure you really need it?

Rishabhg71: Not only are promises missing, but also features like backticks. Nonetheless, I'll work with the writing lecgacy code.

Jaifroid:
I'm very happy for you to open in a new tab when library.kiwix.org is not supported. In fact it's what I though you'd decided to do in your response to https://github.com/kiwix/kiwix-js/pull/1127#issuecomment-1747511789. 

XMLHttpRequest does work in IE11 for most mirrors, but not all mirrors, because some do not send the right headers to force a download in the browser. Maybe you were stuck testing a mirror that doesn't work.

In any case, that comment is not implying you should try to fix it. As we discussed above, opening `download.kiwix.org` in a new tab when you detect that ´library.kiwix.org´ is not supported (that includes in a Chromium extension) seems a good compromise to me.

Rishabhg71:
@Jaifroid I feel this should be the final Review and one interesting thing I found is `library.kiwix.org` is supported in jQuery mode (new browsers)

Jaifroid:
> @Jaifroid I feel this should be the final Review and one interesting thing I found is `library.kiwix.org` is supported in jQuery mode (new browsers) Also, I think I should start working on #1032

Interesting! I'll take a look. And yes, please feel free to start work on #1032. I'll add some thoughts in that issue.

manifest.json
```diff
@@ -35,7 +35,6 @@
         "extension_pages": "script-src 'self' 'wasm-unsafe-eval'; object-src 'self';",
         "sandbox": "sandbox allow-scripts allow-downloads allow-forms allow-popups allow-modals; script-src 'self' 'unsafe-inline' 'unsafe-eval'; child-src 'self';"
     },
-
```
Jaifroid, line 38: Could you add this whitespace back in (carriage return) here? No need for this file to be part of this PR, and it's good to keep this file an almost identical version of the other browser extensions manifests (which have this space).

www/js/app.js
```diff
@@ -1296,6 +1306,23 @@ function handleFileDrop (packet) {
...
+    const libraryContent = document.getElementById('libraryContent');
+    const iframe = libraryContent.contentWindow.document.getElementById('libraryIframe');
+    try {
+        // eslint-disable-next-line no-new-func
+        Function('try{}catch{}')();
+        iframe.setAttribute('src', params.libraryUrl);
+    } catch (error) {
+        const fakeElement = document.createElement('a');
+        fakeElement.setAttribute('href', params.altLibraryUrl);
+        uiUtil.warnAndOpenExternalLinkInNewTab(e, fakeElement);
```
Jaifroid, line 1322: Two issues that are minor but could make a big difference to the UX:

1. I don't think it's necessary to ask the user if they wish to open a new tab: just do it. After all, the user has clicked on "Open library", and any user with half a brain will know that the library is an external URL...
2. Do a programmatic click on `btnConfigure` after you launch the new tab, so that the user isn't left staring at a disconcerting blank screen when they close the external tab... (see screenshot below).

image

www/js/init.js
```diff
@@ -77,6 +110,8 @@ params['contentInjectionMode'] = getSetting('contentInjectionMode') ||
 // A parameter to circumvent anti-fingerprinting technology in browsers that do not support WebP natively by substituting images
 // directly with the canvas elements produced by the WebP polyfill [kiwix-js #835]. NB This is only currently used in jQuery mode.
 params['useCanvasElementsForWebpTranscoding'] = null; // Value is determined in uiUtil.determineCanvasElementsWorkaround(), called when setting the content injection mode
+params['libraryUrl'] = 'https://library.kiwix.org/'; // Url for iframe that will be loaded to download new zim files
+params['altLibraryUrl'] = 'https://download.kiwix.org/zim/'; // Url for iframe that will be loaded to download new zim files
```
Jaifroid, line 114: Change description to "Alternative Url for iframe (for use with unsupported browsers) that will......"

Jaifroid (approved):
This is looking excellent to me! Well done, it makes a big difference to UX. You'll see I have some very minor suggestions that will probably take 5 or 10 minutes. While you're doing that, I'd like to check the code in an old Firefox OS IDE, which is also a good way of checking that there aren't any regressions with old browsers. So let me know when you've done the minor changes, so I can confirm I've completed that test.

Jaifroid:
Now I'm not sure what the best way to deal with this is. We could just forcibly unhide the buttons, but it makes for a cluttered UI (see below).

image

Jaifroid:
Or else, we could say that some reworking of the UI will be necessary for #656, so we could deal with it there.

Any thoughts?

Jaifroid:
When clicking on the button, I get the screen below. Any idea why it's untrusted? Are you using https? In any case, the user can click on "I understand the risks", and then the library opens, so I think that's acceptable. But we should just check that it's not redirecting to http: instead of https:...

image

Jaifroid:
OK, so I think it's just the age of the device, probably doesn't handle https very well. At least it gives the option to go ahead anyway.

So, when you're happy with your final fixes (let me know if you want me to check anything), go ahead and squash/merge.

PR #1131, opened 2023-10-11
[ADD] Native File System support to reload zim file

Rishabhg71:
Summary and Progress 👇
- [x] Using FS API get the file and try to load it
- [x] Save the file in indexDB and Load it again
- [x] UI for File selection under the config 
- [x] Adding logic and File selection to UI
- [x] Try to do the same steps but for the whole directory
- [x] Code cleanup
- [x] [LATER] add webkit as a fallback


Rishabhg71:
I have completed a small working prototype. You can Select File and it will reload it back. A few things I don't like are 
- If multiple files are selected browser is gonna request permission for each which will be annoying to the user (in case of split zim could be more than 10). So should I only let the user select a single file? That would mean that FS API will be unable to load split zims (will be only done via Folder select)
- As you saw I first asked for a confirmation to load the previous Zim file and then requested file access, Do you have any suggestions on how to request file

https://github.com/kiwix/kiwix-js/assets/47270636/e5181ab5-5bce-4a7a-a311-1a126d716227

Jaifroid:
@Rishabhg71 That's a great start!

Yes for the reason you state, I limit (in Kiwix JS PWA) the single file picker to a single file if the user is using the File System Access API. If they are using the normal file picker, then they can pick more and all the picked files will show in the dropdown list of archives.

And this is the reason I provided a "folder picker" as a separate button. It completely avoids the issue, because the user only has to give permission to access the (previously picked) folder, and automatically all files in the folder become accessible.

So, I advise you to finish the implementation for single file picking first -- don't try to make it into a multi-picking solution -- and then when you've done that (it seems it's nearly there), you can add folder picking.

Since every browser that supports the `showOpenFilePicker` method also supports the `showDirectoryPicker` method, it seems the logical way to handle multiple files and split files (through the directory picker only). It's by far the most convenient method, and I use it all the time in the PWA version.

Firefox is a special case: it doesn't support `showOpenFilePicker` at all, but it does support `webkitDirectory`, which is almost exactly the same as `showDirectoryPicker`.

(Don't forget to handle a single drag-dropped file: you can test if it has a filehandle, and if it has, then it's exactly the same as if you had picked it with `showOpenFilePicker`. You need to leverage the existing drag-drop handle in app.js -- `handleFileDrop()`).

Jaifroid:
> Firefox is a special case: it doesn't support `showOpenFilePicker` at all, but it does support `webkitDirectory`, which is almost exactly the same as `showDirectoryPicker`.

I should just qualify that: in Firefox, you can't store the permission in IndexedDB, the user has to re-pick the directory. But I made it so when a user clicks on the dropdown filename they want to open (stored in a cookie, and restored on app startup), the directory picker is opened, and the user only needs two clicks to re-authorize access.

Jaifroid:
@Rishabhg71 It seems to be working well! NB I didn't look at the code.

I think in the final version, instead of asking the user if they want to reopen the last ZIM file, it would be better to display it in the dropdown list of archives. Then the user simply clicks on it if they want to reopen it, or else they click on the file or directory picker if they want to pick a different archive. It's the same number of clicks to reopen (two - once on the file, once on Verify), but it's less intrusive.

Jaifroid:
NB In the future we will get persistent permissions for this API, at least if the app is installed. So we will want to try on startup to launch the last selected archive, and see if we have permission. For now, however, it's fine just to invoke the verify permission when the user makes a gesture. Unfortunately, launching the app isn't considered a sufficient user gesture to show the verify permission dialogue. You need an actual click in the UI.

Rishabhg71:
UPDATE: Directory pickup and selection are also done but I am not happy with the messy code so doing refactors now

Rishabhg71:
https://github.com/kiwix/kiwix-js/assets/47270636/690e0b16-f30d-472a-aa6f-edb6ba6dcf3b

Is the folder drop behavior a bug, or is it intended? @Jaifroid

Jaifroid:
> Is the folder drop behavior a bug, or is it intended?

@Rishabhg71 It was intended: if a user drags and drops, then they don't need file picker, so it can be tidied away. However, it might not be so ideal for your implementation, and can be changed. In Kiwix JS PWA (which has now changed from kiwix-js-windows to kiwix-js-pwa, and the development pwa is now at https://kiwix.github.io/kiwix-js-pwa/ ), dragging a file or a folder tucks away the file pickers behind a "Select storage" button, rather than hiding them completely, and then I decided that when a user has successfully picked a file, the file pickers should always be tucked away, so the UI is not so cluttered. Particularly important if we implement the OPFS in the future on Kiwix JS, because adding an OPFS quota panel clutters the interface even more...

But I'm happy for you to do what seems best for a smooth implementation at this stage!

Jaifroid:
@Rishabhg71 I'm presuming this isn't ready for review, as it's still marked as Draft, and tests are failing. But let me know if/when you need any input from me.

Rishabhg71:
PR is nearly done but I am stuck on a bug where the Dropping folder isn't saved into DB
It should be done by tomorrow or day after tomorrow

Jaifroid:
OK, no problem! I just wanted to check you weren't waiting for me.

Rishabhg71:
@Jaifroid could you quickly check out the functionality? I've got folder selection working for WebKit and FS. Still, there are a few test cases I need to add, like making sure it doesn't display non-ZIM files.

Also, can you guide me a little on how split-zim works? Let's say we have a zim split in 4 parts so I will get [`zima`, `zimb`, `zimc`, `zimd`] but what if it's more than 26 files?

Jaifroid:
> Also, can you guide me a little on how split-zim works? Let's say we have a zim split in 4 parts so I will get [`zima`, `zimb`, `zimc`, `zimd`] but what if it's more than 26 files?

You just need to get all the files in a directory that match `/\.zim\w\w$/i.test(fileName)` (and where the fileName without the split extension letters matches the selected file), then you present them to the file loader in any order, and it will take care to sort them into the right order. It doesn't matter if they are 2 splits or 58 splits.

There is code that handles looping through all files in a directory and matching against the selected archive in kiwix-js-pwa, look for the function `processDirectoryOfFiles` in `app.js`. The typical way to do this is something like:

```
return processNativeDirHandle(directoryHandle, function (fileHandles) {
     processDirectoryOfFiles(fileHandles, selectedArchive);
});
```

So, I take a native directory handle, iterate over the contents to get all the file handles it contains (at least those that match ZIM archives), and then take all those handles and pass them through `processDirectoryOfFiles`, with the string of the archive the user has selected (in the case of split files, this is just the `.zimaa` part). It then takes care of matching the correct file and getting hte ZIM parts for split archives.

Jaifroid:
@Rishabhg71 I just did a quick test (not the webkit part, just the File System Access API) of folder picking, and it's very impressive. Well done! It works smoothly, and works nicely after reload as well, with the files still listed in the dropdown.

There are some UI tweaks that could be made, but they can be done after core functionality is finished. Before I forget them, I'll list them here anyway, but focus on functionality first:

* Once a user has picked an archive or a folder of archives, it would be good to have a message above the dropdown displaying the number of archives it contains ("5 archives found in selected location:");
* Don't display "Select a file..." if there are no files in the dropdown, as it's confusing and distracts the user from noticing the folder selection button. It would be better for the default entry to be blank. However, once the dropdown is populated, it is worth displaying "Select a file..." (so the user only goes to the dropdown when they can see that there are files to select in it);
* I think the dropdown could be more like 60% of the div width, rather than 100%, as it looks a bit imposing!
* New UI elements will need to be translated to French and Spanish (but only at end).

Very encouraging work!

Rishabhg71:
@Jaifroid I feel this PR is now ready for a review

Jaifroid:
I just started some testing (haven't looked at code yet), and I noticed that drag-and-drop appears to be broken in IE11. Could you take a look?

I haven't tried to debug, but in case the issue turns out to be async / await not compiling well for older browsers, then that can usually be cured by converting any functions that will run on older browsers to Promises. All async code can easily be converted to Promises, as async / await are just sugar for Promises under the hood.

Jaifroid:
It's looking very good so far (I've only been testing the UI, haven't reviewed any code). Based on my testing, here are a few suggested UI / UX issues / improvements:

* When picking a folder of split ZIM files (e.g. our legacy Ray Charles test folder), the total number of archives reported above the dropdown list is given as the total number of files, whereas in fact there is only one archive that happens to be split (in the case of the Ray Charles ZIM). Since the dropdown list correctly shows only the `.zimaa` archive, you can simply get the correct total number of archives from the length of the dropdown list of archives once it has been populated;
* After a user has picked a folder of archives, it would be good to change the blank first line in the dropdown to "Select an archive...";
* If there is only one archive in the folder (including only one split archive), it seems to me that it would be friendly to launch that archive immediately (the same as happens if picking a single file), otherwise there are a lot of clicks to open an archive from a folder.

Rishabhg71:
> If there is only one archive in the folder (including only one split archive), it seems to me that it would be friendly to launch that archive immediately (the same as happens if picking a single file), otherwise, there are a lot of clicks to open an archive from a folder.

Everything is fixed except this, I tried to load the zim if there is only one file but it's making the code messy
(a lot of unnecessary returns)

Jaifroid:
OK, looking at it now!

Jaifroid (changes requested):
The functionality is very good. But I noticed these issues that need addressing:

* There is already an existing function in `app.js` called `populateDropDownListOfArchives()`, and there is an existing `archiveList` in the UI, which is a dropdown (search `index.html` for `archiveList`), but you don't use either of these and you've added your own custom ones. That leads to the problem in the screenshot below in Firefox OS, which uses the existing elements. Because you've added new ones instead of reusing, the display becomes disordered in apps that have access to the FS (but don't support the File System Access API).

image

* The exisitng `populateDropDownListOfArchives` function gets its list of archives from the `settingsStore`, rather than from `indexedDB`, and that seems more robust to me, given that some apps that will need to populate the list of archives (notably Firefox OS), don't support indexedDB, but do support localStorage (which is what `settingsStore` mostly uses). So I strongly suggest you populate the list using the existing method, and set and retrieve the list in `settingsStore`, unless you have a very good reason to maintain a separate list;
* The database in `indexedDB` is not named. It appears as `undefined`. See screenshot: image
 Did you forget to set a name for it in `cache.js`? In the PWA it is called `kiwix-assetsCache`, but it should be unique for this app, so I suggest `kiwixjs-assetsCache`. You probably forgot to set appropriate values for the constants at the top of `cache.js`, and you might need to add some appropriate params in `init.js`:
```
const CACHEAPI = params.cacheAPI; // Set the database or cache name here, and synchronize with Service Worker
const CACHEIDB = params.cacheIDB; // Slightly different name to disambiguate
var objStore = 'kiwix-assets'; // Name of the object store
const APPCACHE = 'kiwix-appCache-' + params.appVersion; // Ensure this is the same as in Service Worker
```
* There's one more issue I noticed that I'll post about in a separate comment.

Jaifroid:
This is the other issue mentioned above.

* After selecting a folder with several archives in for the first time, the dropdown is correctly populated, and the first time I clidk on an archive in the list, it opens immediately. But when I click on a subsequent archive, I am asked for permission to access it again. This shouldn't be necessary, as having access to the folder should give access to all its contents. See demo below.

FSA Issue

Jaifroid:
Apart from the functional issues above, there are these very minor UI issues, but they can be left to the end. I just note them here before I forget them:

* The button label "Select ZIM file(s)" should be "Select ZIM File(s)" (capital F) so that the capitalization is consistent with the other two buttons;
* However, it should only show "File(s)" when the app isn't using the File System Access API (i.e. when `window.showOpenFilePicker` isn't a function). Otherwise it should show "Select ZIM File", because if the FSA is in use, then only one file can (and should) be selected. I guess the UI for this could be set in `init.js`, or in `app.js` if there is an obvious place to do so;
* If there is only one archive found in selected location, then the text above the dropdown should read "1 archive found in selected location:" (not "1 archives found....");
* Can we make the font size of the text in the dropdown consistent with the rest of the app? It's too big and looks unprofessional because of that (like the UI was not properly designed!).

Rishabhg71:
I will quickly address these issues but I have some questions before I start 

> There is already an existing function in app.js called populateDropDownListOfArchives(), and there is an existing archiveList in the UI, which is a dropdown (search index.html for archiveList), but you don't use either of these and you've added your own custom ones. That leads to the problem in the screenshot below in Firefox OS, which uses the existing elements. Because you've added new ones instead of reusing, the display becomes disordered in apps that have access to the FS (but don't support the File System Access API).

I didn't know about these UI elements. Is there an emulator available to run Firefox OS? I'd like to test it on my own, but I'm not sure how to get started.


> The exisitng populateDropDownListOfArchives function gets its list of archives from the settingsStore, rather than from indexedDB, and that seems more robust to me, given that some apps that will need to populate the list of archives (notably Firefox OS), don't support indexedDB, but do support localStorage (which is what settingsStore mostly uses). So I strongly suggest you populate the list using the existing method, and set and retrieve the list in settings store, unless you have a very good reason to maintain a separate list;

I'm a bit confused here. Do I need to remove IndexDB and load the files (file handle) from local storage, or is it just about loading the filenames from local storage? 🤔😅


> After selecting a folder with several archives in for the first time, the dropdown is correctly populated, and the first time I clidk on an archive in the list, it opens immediately. But when I click on a subsequent archive, I am asked for permission to access it again. This shouldn't be necessary, as having access to the folder should give access to all its contents. See demo below.

I need to investigate this further since it's the first time I've encountered this behavior.

Jaifroid:
> I didn't know about these UI elements. Is there an emulator available to run Firefox OS? I'd like to test it on my own, but I'm not sure how to get started.

There is, but it's a bit complicated to set up, as it involves downloading an old version of Firefox (52) from PortableAppas, and setting up the FFOS emulator in it, adding archives to a fake SD card... What I suggest is that you simply refactor slightly your code to use the existing UI elements (they can be moved if necessary), then we'll see what the effect of that is. It could be all that's needed. If it proves to be tricky to solve the UI issues, I can help you set it up.

> I'm a bit confused here. Do I need to remove IndexDB and load the files (file handle) from local storage, or is it just about loading the filenames from local storage?

No, the file/folder handles can only be serialized in indexedDB. We're just talking about storing the filenames. Look at how it's currently done (the setting `listOfArchives` set in `populateDropDownListOfArchives`).

www/js/lib/fileSystem.js
```diff
@@ -0,0 +1,238 @@
+// refer to this article for easy explanation of File System API https://developer.chrome.com/articles/file-system-access/
```
Jaifroid, line 1: Add the same header, but include your GitHub name, as you find at the head of other Kiwix module files. You can use the header from `app.js`, but begin it with `/**` instead of `/*!` (the latter tells rollup.js to include the header in the built version, and we only want the header to appear once in the built version). It should start like this:

```
/**
 * fileSystem.js : BRIEF FILE DESCRIPTION GOES HERE
 * This file handles MORE DESCRIPTION, INCLUDE REFERENCE TO developer.chrome.com FROM ABOVE
 *
 * Copyright 2013-2023 Rishabhg71, Jaifroid and contributors
 * Licence GPL v3:
 *
 * This file is part of Kiwix.
 *
```
(etc.)

www/js/lib/fileSystem.js
```diff
@@ -0,0 +1,234 @@
...
+ */
+async function updateZimDropdownOptions (fileSystemHandler, selectedFile) {
+    const select = document.getElementById('zimSelectDropdown');
+    let options = '';
+    if (fileSystemHandler.files.length !== 0) options += '<option value="">Select an archive..</option>';
+
+    fileSystemHandler.files.forEach((fileName) => {
+        if (fileName.endsWith('.zim') || fileName.endsWith('.zimaa')) options += `<option value="${fileName}">${fileName}</option>`;
+    });
+    select.innerHTML = options;
```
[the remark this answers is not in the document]
Jaifroid: Maybe we can add escaping of meta-characters in order to avoid this warning?

Jaifroid (commented):
Just a few minor comments on the code, which looks to be very well written and organized. I'll do a proper code review when the other issues from earlier are addressed, but it's looking good. 😊

www/js/lib/fileSystem.js
Jaifroid, line 1: I've just remembered there is an existing module called `abstractFilesystemAccess.js`. It might be worth your moving your new functions into that module rather than using a new `fileSystem.js` module. See what you think.

Rishabhg71:
image
- Even after 5-10 mins of non-stop testing, I was unable to encounter this issue. If you encounter this issue again I have added a `console.log` of the file and its permission so next time you encounter this you can share logs as well


- I have refactored in such a way that now Options will be loaded in `archiveList` but I was unable to make use of `populateDropDownListOfArchives`. When I did try it was becoming messy with a lot of if-else statements 
 
- I am not sure if FFOS issues are gone or so you test it first if it still doesn't work I need to find a way to simulate it on my end
- I will be doing all the internationalization and small UI fixes at the end of the PR

Jaifroid:
> Even after 5-10 mins of non-stop testing, I was unable to encounter this issue. If you encounter this issue again I have added a console.log('[DEBUG]') of the file and its permission so next time you encounter this you can share logs as well

I encountered the issue on the built version of the app (i.e. in `dist`). Were you testing on that version, or on the raw unbundled version? Just to isolate whether the issue might be something "lost in translation"! 

> I have refactored in such a way that now Options will be loaded in archiveList but I was unable to make use of populateDropDownListOfArchives. When I did try it was becoming messy with a lot of if-else statements

OK, I understand.

> I am not sure if FFOS issues are gone or so you test it first if it still doesn't work I need to find a way to simulate it on my end

I'll test.

Rishabhg71:
> I found a little issue on Firefox 52. When selecting a folder of archives (webkitdirectory is supported in this browser), there is what looks like an unhandled Promise rejection and nothing is slected. Below is console log. In the pwa (pwa.kiwix.org) picking a folder works fine in this browser, so clearly it is supported, and you might be using a method that is unsupported by the browser. I can't quite tell what. Note that Firefox 52 is the version that runs Firefox OS Simulator (but this wasn't in the simulator).

seems like `element.replaceChildren()` is not available on Firefox < 78. I push the fix ASAP
https://developer.mozilla.org/en-US/docs/Web/API/Element/replaceChildren

Jaifroid:
Regarding FFOS simulator, I'm afraid the issue persists. But it's because this browser (and the OS is based on it) supports `webkitdirectory`, whereas on a Firefox OS device, the only way to pick an archive is (supposedly) by placing it in a folder in SD card and letting the device scan. There is code that handles this. That code is supposed to hide the file pickers completely, and only use the archiveList dropdown, but clearly that conflicts with the code you run that shows them again because it detects `webkitdirectory` is supported.

So I think all we need to do is find out where the files selectors are hidden by FFOS, and ensure it runs after your code that shows them (or sets a flag that prevents your code from doing it, or even from running, to avoid a race condition). It's completely acceptable on this old device to exempt showing file pickers at all due to webkitdirectory support being detected, as it's kind of useless on that device. On the other hand, this might conflict with ensuring webkitdirectory DOES work in standard Firefox 52.

I understand there still are a few users of FFOS devices, not me, but a previous main dev and a few others. I "owe it to him" to ensure continued support for a few iterations more!

Jaifroid:
I wonder if you think it will be worthwhile adding a UI test for folder picking? It would be a separate issue, but maybe worth making an issue for it, even if you don't do it yourself. It would only work on GitHub actions, not BrowserStack, because we have access to the FS there, and I don't think we do on BrowserStack. On the other hand, if it's not running on BrowserStack, then it's arguably of limited use, though it would test both FSA and webkitdirectory. I think I may just have given arguments why it wouldn't be worth the effort... But let me know what you think.

Jaifroid:
> seems like element.replaceChildren() is not available on Firefox < 78. I push the fix ASAP

Oh, weird that it wasn't picked up by corejs / Babel. Grrr. Maybe config isn't quite right.

Jaifroid:
Sorry for spam, but if you want to get an old Firefox to test against without going to BrowserStack, use this page: https://portableapps.com/apps/internet/firefox_portable/legacy . The one I was testing on was Firefox 52 (52.9 ESR). The main thing with these PortableApps is to avoid clicking yes if the browser itself asks you to update it to latest. It runs in its own little container isolated from the Registry (if using Windows version), so it doesn't interfere at all with any other Firefox you have installed.

www/index.html
```diff
@@ -487,10 +491,17 @@ <h2 data-i18n="configure-title">Configuration</h2>
                             <br /> Scanning for archives... Please wait <img src="img/spinner.gif" alt="Please wait..." />
                         </div>
                         <div id="chooseArchiveFromLocalStorage" style="display: none;">
-                            <br /> Please select the archive you want to use : <select id="archiveList"
-                                class="form-control"></select>
-                            <br /><button type="button" class="btn btn-light" id="btnRescanDeviceStorage">Rescan</button>
-                            Rescans your SD Cards and internal memory
+                            <div style="float: left;display: none;" id="fileCountDisplay">
```
Jaifroid, line 494: You shouldn't use these kinds of inline styles, and especially it shouldn't be necessary to float elements other than arguably some buttons. Use the correct Bootsrap classes for UI elements instead, if you're having difficulty getting alignment right. You have to assign classes to containing divs, e.g. `<div class="row">` to contain UI elements that should be aligned in the same row (rows might need to be wrapped in `<div class="container">`). I was looking at this because I wanted to ask you to add some vertical spacing in the screenshot below where the green line is. If you wrap textual elements in `<p>` rather than using `<br />`, they should get the appropriate vertical spacing (I realize there is a lot of legacy use of `<br />` in the original HTML, which I didn't write).

image

www/index.html
```diff
@@ -487,10 +491,17 @@ <h2 data-i18n="configure-title">Configuration</h2>
...
+                                <p id="numberOfFilesDisplay" style="float: left;">0</p>
+                                <span style="float: left;">&nbsp;archives found in selected location.&nbsp;</span>
```
Jaifroid, line 496: Two lines above don't make much sense. It's all one paragraph, so should all be wrapped in a single pair of `<p>...</p>` tags. The `numberOfFilesToDisplay` element should be a `<span>`, not a `<p>`, and no need for span tags around the "archives found in selected location" element, since you won't be making programmatic substitutions there at all (and it certainly shouldn't be floated left!).

www/index.html
```diff
@@ -487,10 +491,17 @@ <h2 data-i18n="configure-title">Configuration</h2>
...
+                            </div>
+                            <p style="float: left;">Please select the archive you want to use : </p>
+                            <select id="archiveList" class="form-control"></select>
+                            <br />
+                            <span id="rescanButtonAndText" style="display: none;">
```
Jaifroid, line 501: `<span>` is normally only used inside block elements. This seems like it should be a `<p>` not a `<span>`.

service-worker.js
```diff
@@ -125,6 +125,8 @@ const precacheFiles = [
     'www/js/lib/abstractFilesystemAccess.js',
     'www/js/lib/arrayFromPolyfill.js',
     'www/js/lib/filecache.js',
+    'www/js/lib/cache.js',
+    'www/js/lib/fileSystem.js',
```
Jaifroid, line 129: If you no longer have this file, you should remove it from here.

www/js/app.js
```diff
@@ -1360,6 +1439,7 @@ function archiveReadyCallback (archive) {
 function setLocalArchiveFromFileSelect () {
     setLocalArchiveFromFileList(document.getElementById('archiveFiles').files);
 }
+window.setLocalArchiveFromFileSelect = setLocalArchiveFromFileSelect;
```
Jaifroid, line 1442: Did you try simply defining the whole function as a variable: `window.setLocalArchiveFromFileSelect = function () { ... };`? It should still work everywhere as `setLocalArchiveFromFileSelect` (without `window`), so you don't have to redefine it here?

I'm a little concerned that this solution pollutes the global scope and breaks the modular architecture, but I see a previous senior Dev used a similar solution for loading files in Unit tests (and I adopted it when I rewrote the tests for more modern platforms), so I suppose it's OK, and I can't think of a better solution... (!)

www/js/app.js
```diff
@@ -1290,17 +1352,32 @@ function handleIframeDrop (e) {
...
+    document.getElementById('selectInstructions').style.display = 'none';
+    document.getElementById('fileSelectionButtonContainer').style.display = 'none';
     document.getElementById('downloadInstruction').style.display = 'none';
     document.getElementById('selectorsDisplay').style.display = 'inline';
-    setLocalArchiveFromFileList(files);
-    // This clears the display of any previously picked archive in the file selector
     document.getElementById('archiveFiles').value = null;
+
+    const isFSAPIsupported = typeof window.showOpenFilePicker === 'function'
+    const isWebkitSupported = 'webkitdirectory' in document.createElement('input')
```
Jaifroid, line 1367: This variable isn't descriptive enough. There are lots of `Webkit` APIs, and it sounds like you're testing whether the browser (Safari) is supported... I think you have to rename to `isWebkitDirAPISupported`.

However, you define these in several places. It would be better to define them and test them once. Since you need them across modules, you could add them as `params` (because `params` is a global object defined in `init.js`, and is available app-wide, including in modules like `abstractFileSystemAccess.js` (where you also need these variables).

The way to define them in `init.js` would be  add `@property {Boolean} isFSAPIsupported` and same for `isWebkitDirAPISupported` in the `AppParams` block in `init.js`. You don't actually need to list them and set them there, though you could do along with other params if you preferred. But it's probably best to set them in `app.js` ideally outside of any function block, and somewhere near the top along with other global defs, maybe after `appstate['pwaUpdateNeeded']`.

www/js/init.js
```diff
@@ -112,6 +112,8 @@ params['contentInjectionMode'] = getSetting('contentInjectionMode') ||
 params['useCanvasElementsForWebpTranscoding'] = null; // Value is determined in uiUtil.determineCanvasElementsWorkaround(), called when setting the content injection mode
 params['libraryUrl'] = 'https://library.kiwix.org/'; // Url for iframe that will be loaded to download new zim files
 params['altLibraryUrl'] = 'https://download.kiwix.org/zim/'; // Alternative Url for iframe (for use with unsupported browsers) that will be loaded to download new zim files
+params['cacheAPI'] = 'kiwix-js'; // Sets the database name for the IndexedDB cache
+params['cacheIDB'] = 'kiwix-zim'; // Not sure what this does
```
Jaifroid, line 116: You should add typedef properties for these two in the appropriate AppParams block towards top of `init.js`

www/css/app.css
```diff
@@ -94,6 +94,12 @@
     margin: 0 1rem;
 }
 
+#zimSelectDropdown {
+    width: 60%;
+    font-size: large;
+    height: 2em;
+}
+
```
Jaifroid, line 102: I think the above block is no longer used, so remove?

www/js/lib/abstractFilesystemAccess.js
```diff
@@ -88,6 +89,216 @@ StorageFirefoxOS.prototype.enumerate = function (path) {
...
+
+/**
+ * Gets the selected zim file from the IndexedDB
+ * @param {string} selectedFilename The name of the file to get back from DB
+ * @returns {Promise<Array<File>>} The selected File Object from cache
+ */
+function getSelectedZimFromCache (selectedFilename) {
+    return new Promise((resolve, _reject) => {
+        cache.idxDB('zimFiles', async function (fileOrDirHandle) {
+            console.log('[DEBUG] current file/dir permission status', await fileOrDirHandle.queryPermission(), fileOrDirHandle);
```
Jaifroid, line 158: I think this console.log should be removed before merge, as it adds some unnecessary async overhead (when compiled, it'll probably be a mass of code...).

www/js/lib/abstractFilesystemAccess.js
```diff
@@ -88,6 +89,216 @@ StorageFirefoxOS.prototype.enumerate = function (path) {
...
+                const file = await fileOrDirHandle.getFile();
+                resolve([file]);
+            }
+        });
+    });
+}
+
+/**
+ * @typedef {Object.<number, File>} WebkitFileList
+ */
```
Jaifroid, line 181: It looks like you really really want to write TypeScript! 😊

www/js/lib/abstractFilesystemAccess.js
```diff
@@ -88,6 +89,216 @@ StorageFirefoxOS.prototype.enumerate = function (path) {
...
+    for (const file of webKitFileList) {
+        if (regex.test(file.name) || file.name === filename) {
+            files.push(file);
+        }
+    }
+    return files;
+}
+
+/**
+ * Loads the Previously selected zim file via IndexedDB
```
Jaifroid, line 203: You're using localStorage now, so this needs changing (description).

www/js/app.js
Rishabhg71: I don't think there is any other way to do it at least I am not aware of.
I did try to set `setLocalArchiveFromFileSelect` as a global variable but that too didn't work even

Jaifroid: (just unresolving while you try it)

Rishabhg71: > Maybe just try window.setLocalArchiveFromFileSelect = function () { ... }; for the function def, and see if it works without the extra inserted line here? If it doesn't work, keep what you have. If it works, it might avoid having two functions hanging around that do the same thing (one in app.js and one attached to window).

There is a downside to this `setLocalArchiveFromFileSelect` is referenced in a lot of places and converting it to `window.setLocalArchiveFromFileSelect = function () { ... };` that means everywhere i need to update to 
`setLocalArchiveFromFileSelect` -> `window.setLocalArchiveFromFileSelect` (doesn't seem like a good idea to me)

Jaifroid: I completely agree, it's just that normally you can access any function in the `window` global scope merely by using the `functionName`, you don't have to use `window.functionName` usually. So it might just work without changing anything else. Try it! If it doesn't, keep what you have.

Rishabhg71:
> So I think all we need to do is find out where the files selectors are hidden by FFOS, and ensure it runs after your code that shows them (or sets a flag that prevents your code from doing it, or even from running, to avoid a race condition). It's completely acceptable on this old device to exempt showing file pickers at all due to webkitdirectory support being detected, as it's kind of useless on that device. On the other hand, this might conflict with ensuring webkitdirectory DOES work in standard Firefox 52.

Is there a flag or method to check if the FFOS native file supports it? That way I will not run any webkitdir code at all. I looked for any flag or condition but I cant find any

> I wonder if you think it will be worthwhile adding a UI test for folder picking?

Sure, In fact I was already planning to add some new tests

Jaifroid:
> Is there a flag or method to check if the FFOS native file support via SD card scan? That way I will not run any webkitdir code at all. I looked for any flag or condition but I cant find any

Sure, there's lots of code that calls abstractFileSystem, and it has methods for detecting FFOS storage. Just search for that and you should find something you can use.

Jaifroid:
On new tests, I would suggest making that a separate issue, so as not to delay this PR too much, though it's up to you of course!

Rishabhg71:
> On new tests, I would suggest making that a separate issue, so as not to delay this PR too much, though it's up to you of course!

Different PR seems good to me

> Sure, there's lots of code that calls abstractFileSystem, and it has methods for detecting FFOS storage. Just search for that and you should find something you can use.

I have added a condition to check `const isFireFoxOsNativeFileApiAvailable = typeof navigator.getDeviceStorages === 'function';` for Firefox OS. Could you check real quick if that does fix the issue?

Jaifroid:
Yes, it's working. The new UI elements no longer show on FFOS:

image

Jaifroid:
Oops, sorry, I was working on a different branch, and when I switched, I forgot to do ' npm run build-src`, so the dist folder was still serving a branch based on main. Unfortunately, the issue is still in your branch:

image

Jaifroid:
You probably have to have the new UI elements hidden by default and then unhidden for supported systems.

Rishabhg71:
In that case, maybe I need to install FFOS somehow I don't think I can fix it without running it locally. Any suggestions for me?

Jaifroid:
Although I'm happy to test, if you would rather get the simulator yourself, start by downloading Firefox 52.9 ESR as per previously posted link from PortableApps. When it opens, press Shift-F8 to open the IDE, then click on "Install Simulator", and choose Firefox OS 2.2 Simulator stable. Once installed, click on the simulator (should appear above "Install Simulator"), and it should start up.

Then you need to add a project and point it to `\kiwix-js\dist` (use absolute path, e.g. `C:\Repos\kiwix-js\dist`, to your dist folder for the Repo). If all that's working, let me know and I'll tell you how to add a small ZIM to a fake SD-card directory.

Jaifroid:
(our posts crossed)

Jaifroid:
In Firefox 52.9 refuse any prompts to update to Firefox 60.9 (click "ask later").

Jaifroid:
BTW, for UI translations to French and Spanish, I strongly recommend https://www.deepl.com/ . Alternatively, if you are using GitHub Copilot, you can usually get it to translate if you paste the English JSON strings into the French or Spanish in the appropriate position, then create a blank line above the string you want to translate and start typing the selector, it will usually realize what you're trying to do and complete it.

I know both languages, so can check.

Rishabhg71:
It took me some time to figure out but I got the simulator running. But I am getting this error (even on `main` branch too)
I have already attached a `fake-sdcard` by adding dir in the profile folder but I don't see it here. sorry to bother you by asking for a tutorial but seems like Firefox OS doesn't have a lot of info on the internet

image

Jaifroid:
Here's where I place a couple of small ZIMs, in these folders. You have to create the fake-sdcard folder, and obviously adapt this according to the location of your FirefoxPortableLegacy52 installation.

`C:\UTILS\FirefoxPortableLegacy52\Data\profile\extensions\fxos_2_2_simulator@mozilla.org\profile\storage\fake-sdcard`

It's enough to put a Ray Charles in there (I suggest a modern one, not the old legacy split one). You might have to restart simulator.

Jaifroid:
Seems like you found an unrelated bug. I think where `undefined` is showing we should in fact have a message about loading a ZIM archive... No idea why it's not showing properly. Anyway, it's an issue unrelated to your PR.

Rishabhg71:
@Jaifroid It seems like creating `fake-sdcard` didn't do anything FFOS is still not picking up on any zim/storage. I have also tried to follow the [official](https://devdoc.net/web/developer.mozilla.org/en-US/docs/Tools/Firefox_OS_Simulator/index.html) docs as well but that didn't work. I have pushed a fix do you mind checking if that fixed the issue

Jaifroid:
> @Jaifroid It seems like creating `fake-sdcard` didn't do anything FFOS is still not picking up on any zim/storage. I have also tried to follow the [official](https://devdoc.net/web/developer.mozilla.org/en-US/docs/Tools/Firefox_OS_Simulator/index.html) docs as well but that didn't work. I have pushed a fix do you mind checking if that fixed the issue

Oh, how strange! I'll take a look.

Jaifroid:
Looks like it's working! (I did remember to build the app this time.) I'll try building with build-src just to double-check,

image

Jaifroid:
Confirmed: I'm definitely running your branch.

Rishabhg71:
Great, Now I will clean the code, add translations, and then do a final review

www/js/app.js
```diff
@@ -1290,17 +1353,29 @@ function handleIframeDrop (e) {
...
     document.getElementById('selectorsDisplay').style.display = 'inline';
-    setLocalArchiveFromFileList(files);
-    // This clears the display of any previously picked archive in the file selector
     document.getElementById('archiveFiles').value = null;
+
+    // value will be set to true if a folder is dropped then there will be no need to
+    // call the `setLocalArchiveFromFileList`
+    let loadZim = true;
+
+    if (params.isFileSystemApiSupported) loadZim = await abstractFilesystemAccess.handleFolderDropViaFileSystemAPI(packet)
```
Rishabhg71, line 1371: Is file drop supported by Firefox I don't think there is any file drop. If Filedrop is supported then I need to add the same conditions here as well

Jaifroid: Do you mean drag-and-drop? Firefox supports this, so long as you haven't turned it off in Expert Settings! However, I'm not sure if it supports dragging a folder in. It supports dragging one or more files in the same way as picking one or more files with the (legacy) File picker. Files dragged act in exactly the same way as if they had been picked.

It's possible that a folder dragged in has the webkitdirectory property, in which case you could use it just like picking a folder in Firefox. I didn't implement that in the PWA, because it seemed unlikely to me that anyone would drag a folder as opposed to dragging some files from the folder.

www/js/lib/abstractFilesystemAccess.js
```diff
@@ -88,6 +90,223 @@ StorageFirefoxOS.prototype.enumerate = function (path) {
...
+    }
+    return files;
+}
+
 export default {
-    StorageFirefoxOS: StorageFirefoxOS
-};
\ No newline at end of file
+    StorageFirefoxOS: StorageFirefoxOS,
+    updateZimDropdownOptions,
```
Jaifroid, line 304: Throughout the app we use the format `updateZimDropdownOptions: updateZimDropdownOptions,` rather than this shorthand format. For consistency we should probably do the same here. When I was developing the bundling with rollup, I ran into some problems when I tried to use the shorthand format. Maybe those problems have gone away, but I'd feel a bit more reassured if we use the same format throughout. If it can be changed, then we should have a separate issue to change them all.

Jaifroid (commented):
This is looking great! I've been through the code and have some minor suggestions. While you're attending to these, I'll do some testing on various platforms.

www/js/app.js
Jaifroid: File dragging is indeed broken in Firefox in your implementation, so that will need to be fixed, as it works in main.

Just treat the file as a File object and use the existing function for that (can't' remember if it's selectArchiveFromFileList or something). Or check how it's done on main.

Dragging a folder works fine in Firefox in your implementation.

Jaifroid:
By the way, the dropdown list of archives has become 100% width again, which looks incongruous to me, as we don't have any archives with such long filenames. You should probably set it to a percentage but with a min-width, so on narrow screens it uses the full width, but only a percentage when full screen.

Jaifroid:
After I've selected a folder of archives (testing in Firefox here), and I press Ctrl-R to reload the app, I get this appearing:

image

The archives themselves are listed in the dropdown, but the dropdown also is missing "Select an archive". There might be a conflict with the translation software here. It's possible the translation happens after you've made your substitutions, and since the translation has the raw template string `{{numberOfFiles]]`, it gets substituted back in (this also happens with the English strings, even though they're already in the HTML). I don't know for sure if it's a race condition, but if you coded the substitution to happen on app launch, then it's most likely what is happening. In that case, it would be best to check that the language files are loaded, or else chain your function to happen after the Promise resolves for i18n substitution -- see `loadTranslationStrings()` in `translateUI.js`.

Jaifroid:
I published your implementation to `kiwix.github.io/kiwix-js/dist/` to test on BrowserStack (by the way, if you see this tomorrow morning, we might need to publish it again, as I think it gets reset back to main overnight, not 100% sure). It seems the implementation is broken in Edge 18, whereas both file and folder picking (using `webkitdirectory` property) are fully working in the PWA. See screenshot of yours on left, and PWA on right, both after picking a folder with two archives in it. But file picking is also broken on Edge 18.

image

Console is showing these errors. It looks like it could be relatively easy to fix.

image

Rishabhg71:
> Minor, but niggly: it would be good to fix the singular/plural when there is only one archive in the archive list (remove s in all three languages if there is only 1 archive):

I don't feel it is a good idea to have one more line just for this case

> The archives themselves are listed in the dropdown, but the dropdown also is missing "Select an archive". There might be a conflict with the translation software here. It's possible the translation happens after you've made your substitutions, and since the translation has the raw template string {{numberOfFiles]], it gets substituted back in (this also happens with the English strings, even though they're already in the HTML). I don't know for sure if it's a race condition, but if you coded the substitution to happen on app launch, then it's most likely what is happening. In that case, it would be best to check that the language files are loaded, or else chain your function to happen after the Promise resolves for i18n substitution -- see loadTranslationStrings() in translateUI.js.

Hmm a little weird as it should be hidden when the app is reloaded. Nonetheless I have added a edge case where {{numberOfFiles}} is replaced with "0"

> I published your implementation to kiwix.github.io/kiwix-js/dist/ to test on BrowserStack (by the way, if you see this tomorrow morning, we might need to publish it again, as I think it gets reset back to main overnight, not 100% sure). It seems the implementation is broken in Edge 18, whereas both file and folder picking (using webkitdirectory property) are fully working in the PWA. See screenshot of yours on left, and PWA on right, both after picking a folder with two archives in it. But file picking is also broken on Edge 18.

Checking this issue right now

Rishabhg71:
@Jaifroid I have gone through all the issues and fixed them. do you mind checking again?

i18n/es.jsonp.js
```diff
@@ -19,8 +19,14 @@ document.localeJson = {
...
       "configure-about-usage-link": "Información (Uso)",
       "configure-btn-library": "Biblioteca ZIM",
+      "configure-btn-folderselect": "Seleccione carpeta",
+      "configure-btn-rescan": "Volver a escanear",
+      "configure-about-rescan-tip": "Vuelve a escanear las tarjetas SD y la memoria interna",
+      "configure-select-file-numbers": "archivo(s) encontrados en la ubicación seleccionada. ",
       "configure-download-instructions": "Esta aplicación necesita un archivo ZIM para funcionar.<br />Para instrucciones completas, vea la sección",
       "configure-select-instructions": "Seleccione o arrastre y suelte un archivo .zim (o todos los .zimaa, .zimab etc en caso de un archivo dividido):",
+      "configure-select-file-instructions": "Seleccione el archivo que desea utilizar:",
+      "configure-select-file-first-option": "Selecciona un archivo..",
```
Jaifroid, line 29: I just realized this should be "Seleccione" (like the line above) instead of "Selecciona". (The latter is informal, which is not usually used in software or operating systems.) Also, you're missing one extra full stop to make ellipsis.

i18n/fr.jsonp.js
```diff
@@ -19,8 +19,14 @@ document.localeJson = {
       "configure-title": "Configuration",
       "configure-about-usage-link": "Informations (Utilisation)",
       "configure-btn-library": "Bibliothèque ZIM",
+      "configure-btn-folderselect": "Sélectionner un dossier",
+      "configure-btn-rescan": "Rechercher",
```
Jaifroid, line 23: Change to "Réanalyser" like line below. Using "Rechercher" is ambiguous, because it means "search", and is the term used elsewhere in the app for article/title search.

i18n/fr.jsonp.js
```diff
@@ -19,8 +19,14 @@ document.localeJson = {
...
+      "configure-about-rescan-tip": "Réanalyser la carte SD et la mérmoire interne",
+      "configure-select-file-numbers": "archive(s) trouvées dans le lieu sélectionné. ",
```
Jaifroid, line 25: You also need to put parentheses round the plural form of the adjective, i.e. "trouvée(s)"

i18n/es.jsonp.js
```diff
@@ -19,8 +19,14 @@ document.localeJson = {
       "configure-title": "Configuración",
       "configure-about-usage-link": "Información (Uso)",
       "configure-btn-library": "Biblioteca ZIM",
+      "configure-btn-folderselect": "Seleccione carpeta",
+      "configure-btn-rescan": "Volver a escanear",
+      "configure-about-rescan-tip": "Vuelve a escanear las tarjetas SD y la memoria interna",
+      "configure-select-file-numbers": "archivo(s) encontrados en la ubicación seleccionada. ",
```
Jaifroid, line 25: You also need to add parentheses round the plural form of the adjective, i.e. "encontrado(s)"

www/css/app.css
```diff
@@ -98,6 +98,18 @@
     margin: 0 1rem;
 }
 
+#filesSelectionInfoAndCount p{
+    display: inline;
+}
+
+#archiveList {
+    width: 60%;
+}
```
Jaifroid, line 107: See codefactor warning below. You have duplicate selectors for archiveList in the same ruleset. Best to eliminate this one as the next one is more complete. (I know you have a third below in a different media type ruleset, which is fine.)

www/index.html
```diff
@@ -487,10 +491,19 @@ <h2 data-i18n="configure-title">Configuration</h2>
...
                         </div>
                         <div id="chooseArchiveFromLocalStorage" style="display: none;">
-                            <br /> Please select the archive you want to use : <select id="archiveList"
-                                class="form-control"></select>
-                            <br /><button type="button" class="btn btn-light" id="btnRescanDeviceStorage">Rescan</button>
-                            Rescans your SD Cards and internal memory
+                            <div id="filesSelectionInfoAndCount">
+                                <p id="numberOfFilesCount" style="display: none;">0</p>
+                                <p id="fileCountDisplay" style="display: none;" data-i18n="configure-select-file-numbers">
+                                    archives found in selected location.&nbsp;
```
Jaifroid, line 497: You need to add the parentheses, `archive(s)` here too, as this can act as fallback if the browser can't load the translation strings.

www/index.html
```diff
@@ -487,10 +491,19 @@ <h2 data-i18n="configure-title">Configuration</h2>
...
+                                </p>
+                                <p data-i18n="configure-select-file-instructions">Please select the archive you want to use : </p>
```
Jaifroid, line 499: Eliminate space between "use" and colon.

i18n/en.jsonp.js
```diff
@@ -19,8 +19,14 @@ document.localeJson = {
       "configure-title": "Configuration",
       "configure-about-usage-link": "About (Usage)",
       "configure-btn-library": "Browse ZIM Library",
+      "configure-btn-folderselect": "Select Folder",
+      "configure-btn-rescan": "Rescan",
+      "configure-about-rescan-tip": "Rescans your SD Cards and internal memory",
+      "configure-select-file-numbers": "archive(s) found in selected location. ",
       "configure-download-instructions": "This application needs a ZIM archive to work.<br />For full instructions, please see the section",
       "configure-select-instructions": "Please select or drag and drop a .zim file (or all the .zimaa, .zimab etc in case of a split ZIM file):",
+      "configure-select-file-instructions": "Please select the archive you want to use :",
```
Jaifroid, line 28: Eliminate space between "use" and colon. That's French style, but not used in English.

i18n/en.jsonp.js
```diff
@@ -19,8 +19,14 @@ document.localeJson = {
...
+      "configure-select-file-first-option": "Select an archive..",
```
Jaifroid, line 29: You're missing one extra full stop after archive to make an ellipsis.

i18n/fr.jsonp.js
```diff
@@ -19,8 +19,14 @@ document.localeJson = {
...
       "configure-download-instructions": "Cette application a besoin d'un fichier ZIM pour fonctionner.<br />Pour des instructions complètes, veuillez consulter la section",
       "configure-select-instructions": "Veuillez sélectionner ou glisser-déposer un fichier .zim (ou tous les .zimaa, .zimab etc. dans le cas d'un fichier ZIM découpé)&nbsp;:",
+      "configure-select-file-instructions": "Veuillez sélectionner l'archive que vous souhaitez utiliser :",
```
Jaifroid, line 28: Space before colon **is** correct here.

i18n/fr.jsonp.js
```diff
@@ -19,8 +19,14 @@ document.localeJson = {
...
+      "configure-select-file-first-option": "Sélectionner une archive..",
```
Jaifroid, line 29: Missing extra full stop to make an ellipsis.

www/js/lib/abstractFilesystemAccess.js
```diff
@@ -88,6 +89,216 @@
...
+    let count = 0;
+    if (files.length !== 0) options += '<option value="">Select an archive..</option>';
+
+    files.forEach((fileName) => {
+        if (fileName.endsWith('.zim') || fileName.endsWith('.zimaa')) {
+            options += `<option value="${fileName}">${fileName}</option>`;
+            count++;
+        }
+    });
+    select.innerHTML = options;
```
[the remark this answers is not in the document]
Jaifroid: I think you can dismiss this alert. I can't see any risk, as we're not using PHP or other technologies vulnerable to injection AFAIK.

Jaifroid (approved):
Just a few minor issues I've just noticed in the translations, and one or two other small things. The code looks very good now.

Just make these small changes, and I'd like to do a few more tests of edge cases before you merge. I'll let you know when I've done them.

Jaifroid:
This is cosmetic, but I wonder if it's easy to fix the spacing in the screenshot below. There is too little space before the "2 archiv(s) found" line, and too much space after the list of archives. It should be easy to redistribute so it doesn't look so crowded.

image

Jaifroid:
By the way, Folder picking works on Safari, though the list of archives is not re-populated after I refresh the page. Not sure if that's intended.

Jaifroid:
Testing on iOS and Android, I saw a couple of confusing ergonomic issues:

* The folder picker is shown, but on Android, you are essentially offered a file picker, as there is no way to pick a folder. Nevertheless, you can pick a file in this picker, but it doesn't actually open at all. Something similar, but a bit different (and inconsistent) happens on iOS. In any case, on either platform, there seems to be no way to pick a folder. So I think this folder picker needs to be disabled on mobile.
* The file picker (on the left) offers the possibility of picking multiple archives (and it has to, because of split ZIM archives, at least when `window.showOpenFilePicker` is not available), but only one of those picked is actually shown and launched (I'm talking about picking unsplit archives).

Now, I think the second issue is also present on master. It's not something you've introduced in your PR, but it is now exposed because we have a dropdown list of archives that could now be populated with all the picked files. Therefore, I think bullet point two could be assigned to a separate issue. However, bullet point one (disabling the folder picker on mobile) should be attended to in this PR unless you can find a consistent way to use these. Otherwise it's confusing and a bit misleading to offer the possibility.

I have a reliable method for detecting iOS and Android in the PWA. I'll find it for you.

Jaifroid:
Here's how I test appType:

https://github.com/kiwix/kiwix-js-pwa/blob/main/www/js/init.js#L285

Jaifroid:
Also just noticed that when I've picked a single archive (in browser supporting `window.showOpenFilePicker`) and then I refresh the page, the `Select an archive...` message is still not being shown (on latest code). The same happens with the folder picker.

image

Jaifroid:
Final thing I noticed: on modern Firefox, like on Safari, after I've picked a folder and then I refresh the app (or re-launch it), the dropdown list of archives isn't re-populated. Since all it takes is storing it in settings (see populateDropDownListofArchives function, which does this), it shoudl be very easy to set and then restore the folder contents. Clicking on one of those files should simply evoke the folder picker again, so the user can re-pick the folder quickly and easily. (That's how it works in the PWA.)

Jaifroid (changes requested):
As I've now identified a number of inconsistencies through my UI tests, it would be good to see if some of those can be ironed out. Don't get me wrong, you've done a great job on all these different APIs, and it's a huge improvement. These are just the last niggly details to ensure a good and consistent UX. `ㄟ( ▔, ▔ )ㄏ`

i18n/fr.jsonp.js
Rishabhg71: I am not sure how to do it. In the code we are using `.innerText` property to change text `&nbsp` will just show as text

Rishabhg71:
image
> By the way, Folder picking works on Safari, though the list of archives is not re-populated after I refresh the page. Not sure if that's intended.

I tested on safari it seems like it doesn't support FileSystemAPI (safari 17)

Jaifroid:
> I tested on safari it seems like it doesn't support FileSystemAPI (safari 17)

Safari supports `webkitdirectory` just like Firefox. So it's exactly the same case as I mentioned in https://github.com/kiwix/kiwix-js/pull/1131#issuecomment-1788166074. A fix in Firefox would (should) also fix Safari and all other browsers that support `webkitdirectory` (like Edge 18).

Rishabhg71:
- [ ] Remove JS injection (XSS)
- [ ] Android/IOS file picker inconsistencies
- [ ] Open dir pickup on file select (refresh)

Rishabhg71:
@Jaifroid Can you review again? I have added a feature that the last file will be loaded automatically on folder select but it doesn't work on folder drop. I tried my best to make it work on folder drop but it's making the code messy

Jaifroid:
> @Jaifroid Can you review again? I have added a feature that the last file will be loaded automatically on folder select but it doesn't work on folder drop. I tried my best to make it work on folder drop but it's making the code messy

No problem, I'm taking a look now. I don't think I supported folder drop either in the PWA.

Jaifroid:
It's working really well! I tested in Edge and Firefox, and actually I got good results with the folder drop in Firefox. I don't think it's possible to do it better: it seemed to remember the files in the folder between app launches, and prompted me to re-upload the folder when I selected an archive from the dropdown. It's not possible to get it only to ask for a permission prompt because with webkitdirectory property, there is not possibility of storing the directory handle between sessions. So you've done as well as is possible with the API, congratulations!

I'll test on IE11 and a few other things after our conversation, but it's looking great!

www/js/app.js
```diff
@@ -1300,8 +1317,75 @@ function displayFileSelect () {
         });
         globalDropZone.addEventListener('drop', handleFileDrop);
     }
-    // This handles use of the file picker
-    document.getElementById('archiveFiles').addEventListener('change', setLocalArchiveFromFileSelect);
+
+    if (!isFireFoxOsNativeFileApiAvailable) {
```
Jaifroid, line 1321: This seems really counterintuitive, that you have to keep saying "only run this if a defunct OS is not supported! I realize it does the job, but wouldn't it be easier just to put the final legacy picker in its own function `useLegacyFilePicker()` and do something like:

```
if (isFireFoxOsNativeFileApiAvailable) {
    useLegacyFilePicker();
    return;
}

```
Then  you can get rid of this test and all the other FireFoxOS tests below. What do you think?

www/js/app.js
```diff
@@ -1300,8 +1317,75 @@ function displayFileSelect () {
...
+        })
+    }
+    if (params.isFileSystemApiSupported && !isFireFoxOsNativeFileApiAvailable) {
+        // Handles File selection when showOpenFilePicker is supported and uses the filesystem api
+        document.getElementById('archiveFiles').addEventListener('click', async function (e) {
+            e.preventDefault();
+            const files = await abstractFilesystemAccess.selectFileFromPickerViaFileSystemApi(e);
+            setLocalArchiveFromFileList(files);
+        });
+    } else {
```
Jaifroid, line 1377: So this would become:
```
} else {
    useLegacyFilePicker();
}
```
And you then put the code below in a function of its own (outside of this function scope), above `handleGlobalDragover` below.

Jaifroid (approved):
This looks very good. Well done! I've just got some final suggestions that hopefully you'll find helpful. Don't feel you have to use them, but it might just make things a bit more intuitive and not make some of the code appear to depend on the non-existence of a defunct Firefox OS!

The only possible "bug" I noticed is that on Safari (tested on macOS Sonoma on BrowserStack), the folder picker fails to open when clicking on one of the archives in a previously picked folder after refreshing the page. It works fine on Firefox and on Edge 18 (both of which support webkitdirectory, like Safari), so it's puzzling that it doesn't work there. It works with the PWA on the same Safari version, so I suspect you've accidentally ruled out Safari on macOS somewhere. I'm sure it's a quick fix. I've set kiwix.github.io/kiwix-js/ and kiwix.github.io/kiwix-js/dist/ to the to the code on this branch, so it can be tested and debugged on BrowserStack easily.

After you've done those, I don't need to see the code again, so when you feel it's ready, please feel free to go ahead and Squash/Merge. And thank you for another great PR!

Jaifroid:
PS It would be a good idea to create an issue for the File System UI and/or Unit tests you mentioned. Definitely a separate issue, as I'd really like to get this merged asap, so we can do #1143.

Issue #886, opened 2022-09-10
Add support for favicon extracted from ZIM metadata


Jaifroid:
As explained in https://github.com/openzim/mwoffliner/issues/1658, some scrapers now produce ZIMs with new `illustration` metadata entries in the M namespace. See example below from `3dprinting.stackexchange.com_en_all_2022-08.zim`, showing the metadata entries and one of the extracted icons (the larger one).

This is now documented in the OpenZIM specification, and is currently only used by Kiwix Serve. It would not be difficult to extract these images, though it's not 100% clear to me *how* they should be used -- e.g. as primary favicons or as fallbacks in case a page doesn't already point to a favicon in the ZIM.

It would be neat to set the brower tab icon appropriately, according to the loaded ZIM. MWOffliner ZIMs currently don't have these metadata entries. After quick testing, I found them in new Stackexchange and Zimit ZIMs so far.

<img width="562" alt="image" src="https://user-images.githubusercontent.com/4304337/189482462-5ff4dbf6-f453-4c17-8261-728d6308db81.png">

<img width="123" alt="image" src="https://user-images.githubusercontent.com/4304337/189482487-7472d3c8-f505-4325-ad29-327260133a78.png">

depuc8:
"Hi @Jaifroid, I'd like to tackle this.

I see that PR #1133 likely added the backend infrastructure to extract this metadata. My plan is to:

    Use the existing metadata extraction logic to retrieve the M/Illustration_48x48@1 blob.

    Convert it to an ObjectURL.

    Inject it into the <link rel='icon'> tag.

If this approach aligns with the current codebase state, I can start on a PR."

PR #1133, opened 2023-10-16
Add facility to extract and set ZIM metadata

Jaifroid:
This backports a facility in Kiwix JS PWA.

### DIFF
--- a/www/js/lib/zimArchive.js
+++ b/www/js/lib/zimArchive.js
@@ -33,10 +33,12 @@ import utf8 from './utf8.js';
 /**
  * ZIM Archive
  *
- *
  * @typedef ZIMArchive
- * @property {ZIMFile} _file The ZIM file (instance of ZIMFile, that might physically be split into several actual files)
- * @property {String} _language Language of the content
+ * @property {ZIMFile} file The ZIM file (instance of ZIMFile, that might physically be split into several actual files)
+ * @property {String} creator Creator of the content
+ * @property {String} language Language of the content
+ * @property {String} name Name of the archive
+ * @property {String} zimType Extended property: currently either 'open' for OpenZIM file type, or 'zimit' for the warc2zim file type used by Zimit
  */
 
 /**
@@ -65,17 +67,16 @@ var LZ;
  */
 function ZIMArchive (storage, path, callbackReady, callbackError) {
     var that = this;
-    that._file = null;
-    that._language = ''; // @TODO
+    that.file = null;
     var createZimfile = function (fileArray) {
         zimfile.fromFileArray(fileArray).then(function (file) {
-            that._file = file;
+            that.file = file;
             // Clear the previous libzimWoker
             LZ = null;
             // Set a global parameter to report the search provider type
             params.searchProvider = 'title';
             // File has been created, but we need to add any Listings which extend the archive metadata
-            that._file.setListings([
+            that.file.setListings([
                 // Provide here any Listings for which we need to extract metadata as key:value obects to be added to the file
                 // 'ptrName' and 'countName' contain the key names to be set in the archive file object
                 {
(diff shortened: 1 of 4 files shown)


Jaifroid:
Tested and working on Firefox OS, the only implementation currently using the dropdown list of archives.

Jaifroid:
Properly fixes #395. Also provides dev info for #886.

Jaifroid:
@Rishabhg71 I'd like you to check over this PR if you have a chance. It adds the most important (and actually used) metadata that are contained in the M/ namespace (see https://openzim.org/wiki/Metadata) to the ZIMArchive object.

It extracts the metadata in two phases. This is because it's a slightly costly operation at least on old, slow browsers, as it's actually extracting a series of BLOBs, albeit a very small string in each case. First are the time-critical metadata that are/may be needed on ZIM archive load. Then, in the background, I add the principal remaining items (actually used ones, not all those listed in the above spec). If a Dev needs others, they can use the extraction function manually when needed.

The reason for this PR is that there are some items that I need early access to in kiwix-js-pwa, in particular, I need to be able to detect the ZIM file Creator at load time. I also add the language here because in the future the language could well determine automatically setting the RTL or LTR of the UI, or even auto-selecting UI language (as an option).

None of the rest are likely to be time-critical, but if any are, they can be added in the appropriate background loading section.

Once the principal metadata are loaded, the object will be displayed in the console.debug log, which is useful for developers, and also for the curious.

We could think of adding a display of archive metadata in the UI, as an expandable list hidden under a small down arrow next to the loaded archive name (in Configuration) in the future, but for now, displaying it in console.debug seems sufficient.

www/js/lib/zimArchive.js
```diff
@@ -33,10 +33,17 @@ import utf8 from './utf8.js';
 /**
  * ZIM Archive
  *
- *
  * @typedef ZIMArchive
- * @property {ZIMFile} _file The ZIM file (instance of ZIMFile, that might physically be split into several actual files)
- * @property {String} _language Language of the content
```
Jaifroid, line 39: I felt it was unnecessary to mark all of these metadata, including the abstraction of the ZIMFile, as private variables. So, I've changed them to normal (removed the starting _).

However, I agree that ZIMArchive.file._files (the _files part) should remain as a private variable, but that is defined in ZIMFile.

www/js/lib/zimfile.js
```diff
@@ -105,7 +105,6 @@ var readInt = function (data, offset, size) {
  * @property {Integer} mimeListPos Position of the MIME type list (also header size)
  * @property {Integer} mainPage Main page or 0xffffffff if no main page
  * @property {Integer} layoutPage Layout page or 0xffffffffff if no layout page
- * @property {String} zimType Extended property: currently either 'open' for OpenZIM file type, or 'zimit' for the warc2zim file type used by Zimit (set in zimArchive.js)
```
Jaifroid, line 108: This has been moved to the higher level (property of ZIMArchive instead of ZIMFile), as I feel it more appropriately belongs with the metadata. It describes the Archive type, not the File type.

Rishabhg71:
@Jaifroid, I've read through the issues and what you've done, but I'm not quite sure what's expected from me. Can you clarify? 😅👍

Jaifroid:
It's just good to get a second opinion on PRs. I prefer a PR to be reviewed if possible. A second pair of eyes can sometimes spot issues or else see a better way of organizing code. If you don't see any issues, then I'd appreciate if you could approve the PR.

Rishabhg71:
I had a look at the code it looks good to me (honestly I may not be the best person to comment on that code )
Nonetheless looking at the code got me thinking why aren't we using typescript and OOPS (+ design patterns)

If we are using rollup isn't that gonna take care of supporting all the browsers? I mean all of our code is more like a "script" than an app (kinda ironic for me as it is called `javascript` 😄)

After working for a month on the project I would say the developer experience is not the best. Would it be something you want? I would love to do a refactor after my tasks

Rishabhg71 (approved):

Jaifroid:
Thanks @Rishabhg71! I always think of review on GitHub as peer review, even if the reviewer is not an expert. It's useful to have someone else sanity check some things.

Regarding typescript, it could certainly be done, and all the work I put into introducing a modern build system was heading in that direction. So it's certainly a possibility. The only thing to consider is not making the entry level too hard for new contributors. We can talk more about this! JavaScript is a subset of TrypeScript in any case.

PR #1135, opened 2023-10-20
Update babel-core and fix vulnerabilities

Jaifroid:
Seeing if tests work when pushed by maintainer.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4,6 +4,7 @@
   "requires": true,
   "packages": {
     "": {
+      "name": "kiwix-js",
       "dependencies": {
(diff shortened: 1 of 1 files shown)


Jaifroid:
@Rishabhg71 FYI, it looks like the issue with the dependabot updates not passing tests is that dependabot doesn't have access to the site secrets, so can't run tests on BrowserStack. It seems like it has enough access to be considered an internal contributor, but not enough to run the BrowserStack tests, hence the failures.

In any case, by updating babel-core and running npm install, and then npm audit fix, I got the same updates, and have pushed them to this PR, which is passing tests. So I'll merge.

Issue #732, opened 2021-06-13
Use generic event listener for link processing in jQuery mode


Jaifroid:
This is point 1 discussed in #725. Rather than a new feature for jQuery, I see it as code cleanup, and it will prevent seeing so many "violations" for adding non-passive event listeners to hundresds of document links that show up in the Chromium console. Instead, we add a single `onclick` event listener on the iframe's contentWindow. Then we use almost exactly the same code as we currently have to analyse the clicked link. So changes would be minor.

It's the same solution as proposed for Option 2 of #404, except that of course for SW mode the function would only intervene if the user has clicked a link that navigates away from the document. The code for this could be re-used between both modes, or not used at all for SW mode if it is determined not to be safe.

I did bash together a prototype yesterday (it only took 15 minutes) as a quick test for how Option 2 of #404 would work, but if not used there, the same principle can be used for this PR. It is in branch [Check-for-external-links-and-warn](https://github.com/kiwix/kiwix-js/tree/Check-for-external-links-and-warn) and only works in SW mode to check for external links and asks the user if they want to access the site. It can be used for testing the viability of #404, but also for testing the principle of this PR before writing any more code.

Jaifroid:
As discussed at the Hackathon, we may be forced to apply some link processing in SW mode to enable opening external links in a new tab or window. The method outlined above (minimalized for the single requirement of opening external links) could be used for this purpose.

Issue #725, opened 2021-05-22
Ideas for future simplification of document processing and API


Jaifroid:
In working on #680, I've stumbled across certain possibilities that could reduce the amount of pre-processing of elements that we currently do in jQuery mode. I'm recording these here before I forget them, as possible ideas for future simplification.

* No need to attach event listeners to every ZIM link in the document. Instead, a single click event listener can be added to the document, and when it fires we check the `event.target`. The target will contain the innermost element that was clicked, as opposed to the `event.currentTarget` which would be element to which the event listener was attached. This seems more efficient to me, and would be a relatively small change at code level.
* The second idea is to generalize the use of [Window.postmessage()](https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage), which we currently use with the Service Worker, to communicate with the iframe and/or any opened windows or tabs. It might be possible to end up with a unified API for jQuery and Service Worker code, as regards data requests from the article window and dispatching a response from the back-end to it. It would be necessary to add a controlling script to `article.html` which would act as a layer to request data from the back-end and to inject received data into the document. That script would do some of the processing of the document we currently do with parsing anchors and images. Of course this would only emulate fetch requests, it could never actually trap all the fetch requests in the way that the Service Worker does. A side effect of this is that we would eliminate the CORS issues we have with the iframe if running from the file:// protocol (though this is not particularly important given the existence of extensions).

mossroy:
A generic event listener might be a good idea, and might speed up some page loading.

About the second idea, I'm not sure I understand.
I suppose the controlling script of `article.html` would parse the DOM (like `displayArticleContentInIframe` currently does) but, instead of reading directly the content through `selectedArchive.*` function calls, it would do a `postmessage()` instead, and the message could be handled by the same functions used by SW mode?
That might work, but I don't really see the pros of this technique, except to workaround the CORS issues.
Except maybe that it might allow to move most of the jQuery-only code in a separate file? (that might be interesting to clean up the code)

Jaifroid:
Yes, I think the main advantage of this would be universalizing and simplifying the API for requesting data from the app and the backend. A bit like abstracting a pseudo-Fetch interface for the window containing the document and how it requests data. You're right that the practical outcome for us would in effect be code cleanup: the code is currently a bit schizophrenic, with two different and quite complicated code pathways to get to the same end result (from the user's perspective). I don't know how easy it would be to achieve, but it would be neat if it's possible.

PR #1140, opened 2023-10-27
Slide away UI elements on scroll

### LINKED ISSUE #1139: Slide away the search bar when scrolling down, slide it back in when scrolling up
This would be an intermediate step on the way to #523. It would already be a significant UI improvement in itself, and should be relatively easy to implement.

Jaifroid:
Closes #1139.

It's still a bit rough, but working in essence.

### DIFF
--- a/www/js/app.js
+++ b/www/js/app.js
@@ -87,6 +87,62 @@ appstate['search'] = {
 // A Boolean to store the update status of the PWA version (currently only used with Firefox Extension)
 appstate['pwaUpdateNeeded'] = false; // This will be set to true if the Service Worker has an update waiting
 
+// Placeholders for the article container and the article window
+const articleContainer = document.getElementById('articleContent');
+const articleWindow = articleContainer.contentWindow;
+const articleDocument = articleContainer.contentDocument;
+const region = document.getElementById('search-article');
+let throttle = false;
+let oldScrollY = 0;
+let newScrollY = 0;
+const header = document.getElementById('top');
+const footer = document.getElementById('footer');
+const navbarDim = header.getBoundingClientRect();
+header.style.transition = 'transform 500ms';
+articleContainer.style.transition = 'transform 500ms';
+articleContainer.style.zIndex = 0;
+footer.style.transition = 'transform 500ms';
+
+// Add an event listener to the article window to slide away UI elements when the user scrolls up or down
+var slideAway = function () {
+    newScrollY = articleWindow.pageYOffset;
+    const headerStyles = getComputedStyle(header);
+    const headerHeight = parseFloat(headerStyles.height) + parseFloat(headerStyles.marginBottom);
+    const footerStyles = getComputedStyle(footer);
+    const footerDim = parseFloat(footerStyles.height) + parseFloat(footerStyles.marginTop);
+    // Hide the toolbars if user has scrolled and search elements are not selected
+    if (newScrollY - oldScrollY > 0 && document.activeElement !== document.getElementById('prefix')) {
+        // If the header and/or footer have not already been hidden
+        if (/\(0p?x?\)/.test(header.style.transform)) {
+            if (newScrollY > navbarDim.height) {
+                header.style.transform = 'translateY(-' + headerHeight + 'px)';
+                articleContainer.style.transform = 'translateY(-' + headerHeight + 'px)';
+                var iframeHeight = parseFloat(articleContainer.style.height.replace('px', ''));
+                articleContainer.style.height = iframeHeight + headerHeight + 'px';
+                footer.style.transform = 'translateY(' + footerDim + 'px)';
+                region.style.height = window.innerHeight + headerHeight + 10 + 'px';
+            }
+        }
+    } else if (newScrollY - oldScrollY < 0) {
+        header.style.zIndex = 1;
+        header.style.transform = 'translateY(0)';
+        // Needed for Windows Mobile to prevent header disappearing beneath iframe
+        articleContainer.style.transform = 'translateY(-1px)';
+        footer.style.transform = 'translateY(0)';
+        articleContainer.style.height = window.innerHeight - headerHeight + 'px';
+        region.style.height = window.innerHeight + 10 + 'px';
+    }
+    oldScrollY = newScrollY;
+    throttle = false;
+};
+
+var scroller = function () {
+    if (throttle) return;
+    throttle = true;
+    clearTimeout(slideAway);
+    setTimeout(slideAway, 200);
+};
+
 switchHomeKeyToFocusSearchBar();
 
 // We check here if we have to warn the user that we switched to ServiceWorkerMode
@@ -137,10 +193,8 @@ darkPreference.onchange = function () {
  */
 function resizeIFrame () {
     const headerStyles = getComputedStyle(document.getElementById('top'));
-    const articleContent = document.getElementById('articleContent');
     const libraryContent = document.getElementById('libraryContent');
-    const frames = [articleContent, libraryContent];
-    const region = document.getElementById('search-article');
+    const frames = [articleContainer, libraryContent];
     const nestedFrame = libraryContent.contentWindow.document.getElementById('libraryIframe');
 
     for (let i = 0; i < frames.length; i++) {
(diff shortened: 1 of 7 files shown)


Jaifroid:
Seems ready. Checked on IE11 and Firefox OS. Need to check on Edge Legacy.

Jaifroid:
@Rishabhg71 Would you be able to test this and see if you think UX is good?

Any comments on the code would be appreciated, but optional!

If you don't have any comments and are happy with it, please approve. Thanks!

www/js/app.js
```diff
@@ -87,6 +87,66 @@ appstate['search'] = {
...
+}
+
+let throttle = false;
+
+// Throttles the slide-away function
+function scroller () {
+    if (throttle) return;
+    // If the user has scrolled to the bottom of the article, stop processing
+    // const articleDocument = articleWindow.document.documentElement;
+    // if (articleDocument.scrollTop === (articleDocument.scrollHeight - articleDocument.clientHeight)) return;
```
Jaifroid, line 141: No longer necessary. Remove commented-out code.

Jaifroid:
PS, just so you're aware, sliding away of header/footer is disabled if the search bar (prefix) has focus. Since prefix is automatically focused on loading the landing page of a ZIM, it means that the landing page won't show the scroll effect unless you click somewhere in the article. Subsequent articles will work without the extra click.

Jaifroid:
On Edge 15 (on Windows Mobile), the header remains hidden under the iframe until the user scrolls back to the very top of the document. I probably need to set the zindex to 1.

Rishabhg71:
@Jaifroid 
image
I have seen this behavior a lot of times usually refreshing the page the load it up properly. This only happens when running the vite server. Any idea why + Am I the only one getting this behavior?
try visiting this URL (GH codespaces) https://musical-yodel-xvpjvv4pwpvcprpp-5173.app.github.dev/www/index.html

Jaifroid:
Yes, that's a known issue with the vite server, and is usually fixed by ensuring Bypass AppCache setting is turned ON. When it's on, and when using vite server, it's not necessary to refresh the page when you make a change, because the server forces the browser to refresh. However, if you do get the disordered display, then a refresh will fix it, or sometimes Ctrl-Shift-R with devtools in focus.

This is nothing to do with this PR/branch, it's because Vite uses [esbuild](https://esbuild.github.io/) as a fast bundler when using Hot Module Replacement, whereas it uses rollup for production builds (but I don't actually use Vite's version of rollup, because it's configured way too aggressively -- we just use pure rollup for our bundling).

Jaifroid:
You can use http-server instead if it's bothersome (`npm run web-server`).

Rishabhg71:
> This is nothing to do with this PR/branch, it's because Vite uses [esbuild](https://esbuild.github.io/) as a fast bundler when using Hot Module Replacement, whereas it uses rollup for production builds (but I don't actually use Vite's version of rollup, because it's configured way too aggressively -- we just use pure rollup for our bundling).

I had an idea it was a problem with vite because the normal HTTP server works now coming to the main I feel UI/UX is good but IMHO I found it annoying that every time I scroll up and down. A simple solution is to only show the header/footer if a minimum distance has been crossed by the scroll wheel.
On a side note, There is a weird bug on Brave where the header hides but leaves that red area It might be an issue on Brave or all Chromium-based browsers. I know we don't officially support Brave so maybe we can leave it

https://github.com/kiwix/kiwix-js/assets/47270636/22a2f7db-502f-47de-bc02-581c0197a625

Jaifroid:
> On a side note, There is a weird bug on Brave where the header hides but leaves that red area It might be an issue on Brave or all Chromium-based browsers. I know we don't officially support Brave so maybe we can leave it

Oh, well it's definitely not the case with all Chromium browsers, as it doesn't show on Edge. I'll check on Chrome. However, I should try to fix this, unless it really is a bug with Brave. I'll check.

> A simple solution is to only show the header/footer if a minimum distance has been crossed by the scroll wheel.

Yes, I could try that. However, I think that should apply to scrolling down (in fact I did make it that the user needs to scroll at least 50 pixels from the top, but not on subsequent scrolls). On scrolling up, it's more likely the user wants to see the header so they can go somewhere else or search for something, so they might be annoyed if it doesn't appear straight away.

Anyway, thanks for checking, and I'll see if I can make improvements.

(By the way, if a user is really annoyed, they can always turn the feature off in Configuration.)

Jaifroid:
Quick question, do you think this feature should be off by default, and users have to turn it on if they want it?

I was kind of taking my cue from the Android app where it's always on (and I don't think it can be turned off).

Rishabhg71:
> Quick question, do you think this feature should be off by default, and users have to turn it on if they want it?

Turning this feature on by default makes more sense. Most of the users don't care for these minor inconveniences and also get used to them after some time. One more suggestion you can try playing around with the animation time I feel it's a little on the slow side

Jaifroid:
So I've done these changes:

* Animation reduced to 0.3s instead of 0.5s. You're right, this makes it feel more responsive.
* User has to scroll down at least 50 pixels (from any starting position) to activate hiding. This means that really slow scrolling won't activate it (due to my intentional throttling of the onscroll event), but I think that's quite good because the result is that you need a slightly more decisive scroll in order to activate it. Keyboard down-arrow mostly activates it, but occasionally only after second down-arrow. Scrolling up doesn't have the 50 pixel margin, because users will usually be doing that after they've finished reading and want to show the header/footer so they can navigate.
* I now only apply the z-index and position: relative workaround in IE11 and Edge Legacy (only Edge Legacy actually needs it).
* I think the last change (z-index) has fixed the Brave bug. At least, I can't reproduce it now using portable Brave (https://portapps.io/app/brave-portable/) installed in Windows Sandbox. I believe having changed the position of the header from static to relative is what caused the Brave bug. It's back to static for all browsers except IE11 and Edge Legacy.

www/js/init.js
```diff
@@ -82,6 +83,8 @@ params['storeType'] = getBestAvailableStorageAPI();
 // The key prefix used by the settingsStore.js (see comment there for explanation), but we also need it below
 params['keyPrefix'] = 'kiwixjs-';
 params['hideActiveContentWarning'] = getSetting('hideActiveContentWarning') === true;
+// A parameter to determine whether to slide away the header and footer when scrolling (feaults to true)
```
Jaifroid, line 86: Typo -> defaults

Jaifroid:
@Rishabhg71 If you're happy that the changes address the issues you identified, I'd be grateful if you could approve the PR! Thanks for helping to improve this PR -- it's always good to have a second opinion, especially for UI changes which can be quite contentious for users.

Rishabhg71 (approved):

Jaifroid:
Last tests:

- [ ] Firefox 64 JQuery and SW modes
- [ ] FFOS double-check
- [ ] IE11 double-check
- [ ] As a Chromium extension
- [ ] As a Firefox extension

Jaifroid:
Encountering a bug on Firefox OS whereby links clicked with a mouse stop working after a few clicks. Touch still works. Very odd. Need to check if this is also happening on main, to see if this PR caused it.

Jaifroid:
FFOS bug appears to be triggered when the transition is set on the iframe. It appears to cause some href elements in the document to lose their click event listeners, though touch still works.. It doesn't happen if the feature is turned off.

It might be time to implement the first bullet point of #725, which is an overdue simplification of link handling, and probably really quick to do. I'll test.

Jaifroid:
OK, so I've determined that #732 doesn't fix this. It's caused by a weird bug whereby the `event.target` of a clicked link no longer points to the right bit of HTML due to the iframe having been shifted up by several pixels (by the transform translateY style having been added and shifting the iframe up). This of course should have nothing to do with the document inside the iframe, so it's definitely a framework bug.

I tried restoring the position of the iframe immediatley on click, but this is too late, and doesn't change what the event.target points to.

Now there are two mitigating factors:

1. Real FFOS devices are touchscreen devices, and at least in the simulator, touch seems to provide the correct information, and is unaffected;
2. It can be turned off.

So, I propose to add a check for FFOS, and NOT turn this feature on by default for that platform. When user turns it on anyway, a warning can be added if it's FFOS, to use at own risk.

Issue #435, opened 2018-11-11
In PhET ZIMs the load button is hidden under the bottom bar


Jaifroid:
The title is self-explanatory, and I include a screenshot to show the problem. Note that this occurs in desktop-wide views. The button can be accessed by narrowing the window. This may be related to #386,  but as it occurs in SW mode, it may need a different solution. Or maybe we can find a generic one that solves both issues.

![image](https://user-images.githubusercontent.com/4304337/48319577-9c863200-e607-11e8-802d-8477187d4af3.png)

This may be better addressed through #120 or #337.

kelson42:
For me it does not work at all with a recent version, even with Chrome & Service Worker.

kelson42:
@mossroy @Jaifroid Should we really keep that ticket considering #339?

Jaifroid:
> For me it does not work at all with a recent version, even with Chrome & Service Worker.

@kelson42, PhET ZIMs work in Service Worker mode, for which you need to run the extension version of the app (Firefox or Chrome) and select "Service Worker" in Configuration. PhET experiments work in jQuery mode only in my JavaScript Support branch, but then you have to search for the experiment title in the search bar, because the ReactJS interface is not working in that mode yet. It works fine in SW mode.

Jaifroid:
> @mossroy @Jaifroid Should we really keep that ticket considering #339?

I think we do need to keep this ticket, because we do have PhET working in SW mode (in master). However, this ticket may be superseded by #523. Until we've made more progress on that, I think we should keep this issue.

mossroy:
I agree with you @Jaifroid.
@kelson42 : I confirm PhET is fully working in SW mode with Firefox or Chrome, since version 2.5. Including as a Chrome extension.

kelson42:
FYI The phets scraper has been rewritten an a version 2.0 has been published a few days ago. New ZIM files have been released. Phets related bugs should probably be re-confirmed.

Jaifroid:
Indeed, and actually we've rewritten the algorithm for calculating the visible area, so it *may* be fixed our end. Will check.

kelson42:
@Jaifroid Might be time to refresh that one

Jaifroid:
[...]

PR #1142, opened 2023-10-30
Slide away UI elements with swipe even on non-scrollable pages

### LINKED ISSUE #1141: Allow slide away of header and footer even if there is no scroll possible in iframe
Once I have merged #1140 (#1139), it will finally be possible to address the issues in #337 and #435, as we'll have tested functions for hiding and showing the UI elements.

We'll need to handle mousedown and touchstart events on the window (possible also on the iframe), and test if there has been a "swipe down" or "swipe up" without a scroll. Boilerplate code would look something like this (untested):

```
let start = null;

function startHandler(e) {
  if (e.type === 'touchstart') {
    start = e.changedTouches[0];
  } else if (e.type === 'mousedown') {
    start = e;
  }
}

function endHandler(e) {
  let end;
  if (e.type === 'touchend') {
    end = e.changedTouches[0];
  } else if (e.type === 'mouseup') {
    end = e;
  }

  if (end.screenY - start.screenY > 0) {
    console.log('scrolling up');
    document.getElementById('status').innerHTML = 'scrolling up';
  } else if (end.screenY - start.screenY < 0) {
    console.log('scrolling down');
    document.getElementById('status').innerHTML = 'scrolling down';
  }
}

window.addEventListener('touchstart', startHandler);
window.addEventListener('mousedown', startHandler);
window.addEventListener('touchend', endHandler);
window.addEventListener('mouseup', endHandler);

```
This code listens for both touchstart/mousedown and touchend/mouseup events on the window. When a touch or mouse button press starts, it records the starting position. When the touch ends or mouse button is released, it checks if the end position is greater than or less than the start position to determine if the swipe was up or down.

Jaifroid:
Fixes #1141, #435 and #337.

Code working, but needs tidying.

On fixed (non-scrollable) pages, a swipe up or down, a mousewheel (or trackpad) swipe up or down, or Ctrl/Commad UpArrow or DownArrow on the article window wil toggle the visibility of header and footer.


Jaifroid:
So in fact IE11 doesn't support whell or touch events on the iframe window, but it does support both the scroll and the keydown events, so we provide support for IE11 (and also for others) with the Ctrl/Cmd + DownArrow/UpArrow key combinations.

PR #1158, opened 2023-11-13
Make a generic browserAction for backgroundscript

Jaifroid:
This attempts to make the browserAction generic, so that developers don't need to alter backgroundscript when testing manifest v2 and manifest v3 extensions.


Jaifroid:
Tested and working in all extensions.

PR #1160, opened 2023-11-13
Use libzim for reading ZIM contents

Rishabhg71:


Rishabhg71:
@Jaifroid Basic UI is done. What I wanted to know is whether there is any implementation for libzim or do I start from scratch (except for other PR) ?

Jaifroid:
For now we can only implement the functions that are currently exposed in the libzim bindings over at javascript/libzim, and for all intents and purposes, I believe those are full-text search (already implemented in our production code, so there is nothing to do to that at this stage), retrieving an article by path, and retrieving assets by path or URL. All this is enough to be able to read and browse the contents of a ZIM.

It would be a good idea to allow developers to load (from the interface), any one of the four implementations of libzim that we have, as each will be helpful in analysing and decoding issues:

1. The default would be off (use existing implementation, or disable libzim for article loading)
2. Use the production WASM (i.e., minified and production-ready version) for article loading;
3. Use the dev version of the WASM implementation (this is built with error assertions, so we get meaningful feedback from errors occurring inside the WASM);
4. Use the production ASM version for older browsers;
5. Use the dev ASM version (useful because it is pure JavaScript, so can be debugged with DEV tools).

You can see an example of the kind of choice I'm suggesting over at the PWA (https://pwa.kiwix.org), where I included such a dropdown under Troubleshooting and Development in Configuration. However, that dropdown **only** chooses the implementation for full-text searching, it doesn't allow the WASM to be loaded for article retrieval. Your implementation should not be about search, but should instead be about using the WASM for reading articles.

Needless to say, you mustn't load the WASM twice, in other words, your dropdown should control the EXISTING loading of the WASM in `zimArchive.js`. It is that version that you will then call like in the lbizm experiments PR #766. If the user disables WASM use in your dropdown, full-text searching should still work (in other words, Disable would only turn off reading articles with libzim WASM, not search).

I'm not sure I've expressed myself clearly. Let me know if I can clarify anything.

Jaifroid:
By the way, I think the title of this PR isn't very clear. We already use WASM (and ASM) for XZ and ZSTD. This PR is about libzim, not whether it is ASM or WASM. I suggest changing the title to something like "Use libzim for reading ZIM contents" (or something similar)!

Rishabhg71:
@Jaifroid I was working on `getEntryByPath` from libzim_binding but I don't what function does what. There are no docs here as well is there any place I could directly look for comments on these functions https://github.com/openzim/javascript-libzim/blob/main/libzim_bindings.cpp#L84C2-L84C2

Jaifroid:
@Rishabhg71 You have to look at the libzim source itself here: https://github.com/openzim/libzim/tree/main/src . Basically, the bindings try to expose in JS the functions that we need from libzim.

As well as the source code, there is this:

https://libzim.readthedocs.io/en/stable/

Hopefully some of that helps!

However, remember we've already got the most useful bindings exposed, so you don't need to replicate those. It's great if you can get your head round them, though, as at least one of them doesn't seem to have proper error catching.

Rishabhg71:
@Jaifroid I am still confused as I was before. I will try explaining with an example like there is a function called `getEntryByPath` now I can call this easily but I don't know which args to pass. Looking into the docs I get the vague idea that I need to pass some sort of path but which path?? (I tried `/`| `index.html` | `this.file.mainPage`). I can't go on and create a proper mapping if I don't know which args to pass. Do you have any idea how do I proceed?

Jaifroid:
The PWA is an easy way to see all the URL paths in a ZIM. Open a ZIM, and in the search bar, type  space + / (space followed by forward slash). That will evoke the URL index, and then you can clearly see what kind of ZIM you're dealing with, a Type 0 (articles in A/ namespace), a Type 1 (all user content in C/ namespace) or a hybrid Zimit type (with namespaces that look like C/A(... for user content, and C/H/... for response headers).

If you are interested in a particular article, you can type part of its path into the search box, like `A/Far-`. So long as you begin with a valid namespace, it will search the URL index (see image below).

As far as I know, the "path" is in fact the ZIM URL as shown in the image below. Try it and see. Note that in Wikimedia ZIMs, it's still A/..., whereas in most other newer ZIMs it should be C/...

image

Jaifroid:
PS In both the Kiwix JS code (current code on main) and the Kiwix PWA code, you can find in Dev Tools a printout of the ZIM File properties, as well of the ZIM Archive properties after you open a ZIM. There it also tells you what ZIM type you're dealing with.

image

image

Jaifroid:
And did you try out the libzim [test implementation](https://openzim.github.io/javascript-libzim/tests/prototype/) linked from here: https://github.com/openzim/javascript-libzim? That's a really basic implementation that allows you to see clearly how the calls are done to the WASM version (by examining in DEV tools). It's kind of tuned for using a Ray Charles (modern) ZIM, but you can use any so long as you alter the paths. Beware that some ZIMs currently don't work for article loading due to the lack of error-handling I've mentioned before (one thing we need to fix, see https://github.com/openzim/javascript-libzim/issues/64).

Rishabhg71:
@Jaifroid It seems like after the latest update WASM files are not exporting the functions at all. Reverting back to this commit is working wasm is working `53bcf308389dd5f3827b2007c46c1a318c39874a`
image

Jaifroid:
OK, I'll check. I updated this branch from main, but didn't test properly, as it's an experimental branch.

Jaifroid:
Oh, that's simply because this branch needs to be fixed as I outlined in #1169. As I wrote to you there, you need to call `whenZimReady` only after the Promise for the libzim has completed. Currently it's being called too early. That's all. You should keep the current call to `whenZimReady`, but make it conditional on whether the user has selected to use libzim for file reading. As I suggested, it would be good to set a `params.useLibzim` when the user selects that option, so you can check for it here and call `whenZimReady` at the appropriate point.

Rishabhg71:
Thanks to your advice I have a working prototype that I can now build upon. The first issue I have encountered here is only certain Zim files are being loaded properly as you can see in the video below. It seems like something related to namespace working Gutenberg has a namespace of `A/Home.html` and another one has `W/mainPage`. Its a guess but could you point me in right direction?

https://github.com/kiwix/kiwix-js/assets/47270636/beb1e984-4482-47f8-80a8-5a7e8f43459d

Jaifroid:
Good news that it's (partially) working!

I don't think that's an issue with the ZIMs. All ZIMs have a W namespace entry, standing for "Well-known entry", aka, pointer to the landing page. That pointer is read when the ZIM is first opened, to determine the location of the landing page directory entry. That should redirect (automatically) to something like `index.html` (but it won't always be that).

You shouldn't have to handle any of this. It should all be transparent.  It's puzzling indeed that it works for one case, but not for another. Could it be that in one case, our backend is resolving `W/mainPage`, whereas in the other, you're passing that request directly to libzim? Our backend will resolve `W/mainPage` to `C/Home.html` in the case of the ro ZIM, but resolving that is the job of a function that may not be available in our bindings, something like `getMainPageDirEntry()` (that's what we call it).

You could pause on that function in zimArchive.js to see if it's being called in some cases, but not in others, and then work back from there.

Rishabhg71:
image

I used the `ZIMArchive.resolveRedirect` method to obtain the redirected URL, but it seems that the backend (libzim) is unable to retrieve the content from the redirected URL. This issue might be related to the presence of Unicode characters in the URL. Additionally, I tested a simple file without redirects, `wikipedia_en_100_maxi_2023-11.zim`, and found that this ZIM file loaded perfectly, with all pages functioning properly.

Is there a ZIM file that contains redirected content, but all the data is in English? I'd like to check if Unicode characters are causing the problem.

Jaifroid:
If you look at the OpenZIM specification, you'll see that UTF8 is the standard and mandatory encoding for all ZIM content. Libzim support UTF8. 

Without debugging your code in detail, I would hazard these guesses:

* You're running into https://github.com/openzim/javascript-libzim/issues/64, and need to compile new versions of the libzim ASM and WASM. You can tell if this is the case by testing in a Chromium browser, where the error circled in the screenshot above is clearly shown. If you make a PR over there with just the exception catching mentioned (you only need to insert it in two places in the makefile), then the PR will automatically build new versions for you and attach them to the workflow output. You can then test with those.
* Alternatively, you're not doing decodeURI, or decodeURIComponent (better) on the URL before sending it to libzim.

I strongly suggest you take a look at, and test, your ZIMs in the libzim experiments branch, where the only errors I noticed were the ones mentioned above (Uncaught ....), and we know what the easy solution to that is. If it's working there, and not working in your branch, then the problem is with your branch...

www/js/app.js
```diff
@@ -886,6 +904,60 @@ function initServiceWorkerMessaging () {
...
+    const messagePort = event.ports[0];
+    try {
+        const ret = await selectedArchive.getEntryDirByWasm(title);
+        if (ret === null) {
+            console.error('Title ' + title + ' not found in archive.');
+            messagePort.postMessage({ action: 'giveContent', title: title, content: '' });
+            return;
+        }
+
+        if (ret.mimetype === 'unknown') {
```
Rishabhg71, line 926: I couldn't get the `redirect` variable as true for `W/mainPage`

Rishabhg71:
I finally have a basic working version where most of the zims are worker Gutenberg, StackExchange, and some of the Wikipedia zims. After checking why those zims are not working it seems like they are not hitting the `handleMessageChannelMessage` function. Could you maybe give me some advice

Zims which are not working 
- wikipedia_en_ray_charles_2015-06.zimaa
- wikiwel_en_all_maxi_2018-10.zim

When I am done with these zims I can finally add the UI controls

Jaifroid:
That's good news!

Regarding the issues you are seeing, the first one is easy to explain -- see https://github.com/openzim/javascript-libzim/issues/16. Feel free to look into why we're not able to load split ZIM files: I confess I haven't taken the time to look into it, because it's not critical and we have another way to load them in our legacy backend. It could be a small issue or something more fundamental).

Regarding the older wikimed, I guess that's using XZ compression rather than zstandard, but in principle it shouldn't matter as libzim should support both.

I would say neither of these are critical. Feel free to look into them if you wish, or else to advance your PR without solving those issues, and for now just leave a note saying that only recent unsplit ZIMs are currently supported, pending further investigation.

Rishabhg71:
> see https://github.com/openzim/javascript-libzim/issues/16..., we're not able to load split ZIM files...
@Jaifroid I will check these once this PR gets merged, I think it would be a good time for a first review

Jaifroid:
@Rishabhg71 I note tests are failing. I'd normally wait for tests to be passing before reviewing. I can take a look, but I think it would be good for you to investigate and fix the tests if you can.

www/js/app.js
```diff
@@ -841,7 +848,8 @@ function initServiceWorkerMessaging () {
                     // Until we find a way to tell where it is coming from, we allow the request through on all controlled clients and try to load the content
                     console.warn('>>> Allowing passthrough of SW request to process Zimit video <<<');
                 }
-                handleMessageChannelMessage(event);
+                if (params.useLibzim) handleMessageChannelByWasm(event);
```
Jaifroid, line 851: I don't think this name is accurate. The other message channel also uses WASM, and there are also ASM versions of libzim (which we need to support also for browsers that can't handle WASM). So I'd suggest you rename this to `handleMessageChannelForLibzim`. Or another name if you prefer, but it would be good to reference libzim explicitly, to be clear about what is being used in that message channel.

www/js/app.js
Jaifroid: Do you mean the dirEntry property `isredirect`? Does libzim return that property or something similar? It might be a function rather than a property (`isredirect()`). Worth checking by pausing when the dirEntry is returned and inspecting its properties. Some are different from the ones we're used to. E.g. mimetype is a string rather than a function. Come to think of it, I'm not sure we can return functions over postMessage, but I may be wrong.

www/js/init.js
```diff
@@ -55,6 +55,8 @@
  * @property {string} cacheIDB - Name of the Indexed DB database
  * @property {boolean} isFileSystemApiSupported - A boolean indicating whether the FileSystem API is supported.
  * @property {boolean} isWebkitDirApiSupported - A boolean indicating whether the Webkit Directory API is supported.
+ * @property {boolean} useLibzim - A boolean indicating weather to use the libzim to load zim files.
+ * @property {"wasm-dev" | "wasm" | "asm" | "asm-dev" | "default"} libzimMode - A value indicating which libzim mode is selected.
```
Jaifroid, line 59: I don't think these are types. You should probably just use `{String}` and probably use capital initial letter for types (I may be wrong, but I think that's what we normally use). You can list the acceptable strings in the description instead.

www/js/init.js
```diff
@@ -123,6 +125,8 @@ params['cacheAPI'] = 'kiwix-js'; // Sets name of the prefix used to identify the
 params['cacheIDB'] = 'kiwix-zim'; // Sets name of the Indexed DB database
 params['isFileSystemApiSupported'] = typeof window.showOpenFilePicker === 'function'; // Sets a boolean indicating whether the FileSystem API is supported
 params['isWebkitDirApiSupported'] = 'webkitdirectory' in document.createElement('input'); // Sets a boolean indicating whether the Webkit Directory API is supported
+params['libzimMode'] = getSetting('libzimMode') ?? 'default'; // Sets a value indicating which libzim mode is selected
```
Jaifroid, line 128: What's wrong with `getSetting('libzimMode') || 'default';` ? I think it's more comprehensible. Basically, if it's `null`, then you set it to default, otherwise it'll be whatever value is set in the setting.

www/js/init.js
```diff
@@ -123,6 +125,8 @@ params['cacheAPI'] = 'kiwix-js'; // Sets name of the prefix used to identify the
...
+params['useLibzim'] = params.libzimMode !== 'default'; // Sets a value indicating which libzim mode is selected
```
Jaifroid, line 129: So you don't have a separate checkbox setting for `params.useLibzim` in the UI? Wouldn't that be cleaner than depending on a potentially confusing side-effect here?

www/js/lib/zimArchive.js
```diff
@@ -501,6 +506,16 @@ ZIMArchive.prototype.findDirEntriesFromFullTextSearch = function (search, dirEnt
     });
 };
 
+ZIMArchive.prototype.getEntryDirByWasm = async function (path) {
```
Jaifroid, line 509: Elsewhere we unifromly use `getDirEntry`, not `getEntryDir`. It's confusing to have one function named differently. Also, as mentioned above, I don't think `ByWasm` is accurate, as we have WASMs of the other decompressors used as standard.

Jaifroid (changes requested):
I haven't tested yet, but here's some initial code review. This branch should have libzim off by default (maybe that's the case already) so that tests will pass on the branch. (You can't make tests on libzim work yet because of the split ZIM issue, and that's fine, as we're talking about an experimental feature.) So, to be sure your code is working, you need to make sure tests are passing, or else find out why they're failing. By the way, you should see whether your code works with Zimit-based archives. I expect it won't without some work, and that's OK (I haven't enabled libzim with Zimit on the PWA, because it would need a lot of integration work in the message channel. That can be a separate issue.)

www/js/init.js
Rishabhg71: These are the literal types a string could be anything but this property will only expect certain strings
image

www/js/init.js
Rishabhg71: My preference just leaked in this code.

www/js/init.js
Rishabhg71: I don't think we need an extra button to enable libzim since
- It will be enabled nonetheless (in case of default for full-text search)
- I added `useLibzim` only so that anyone knows the file read will be also done by libzim

I don't have a strong opinion here if you think I should directly use `params.libzimMode !== "default"`. I will make those changes

www/js/lib/zimArchive.js
Rishabhg71: My bad I left this here. I added this when I was prototyping

www/js/app.js
Rishabhg71: After calling `await this.callLibzimWorker({ action: 'getEntryByPath', path: "W/mainPage" })` We get the result back with mime-type as `unknown` and redirect as `false`. Basically, libzim is unable to tell if the requested title is a redirect or not so if libzim does return false. I am checking if it's a redirect or not by `selectedArchive.resolveRedirect`

Rishabhg71: `await this.callLibzimWorker({ action: 'getEntryByPath', path: "W/mainPage" })` returns this 
`{
    "content": {},
    "mimetype": "unknown",
    "isRedirect": false
}`

www/js/init.js
Jaifroid: OK, I'll check when I test the UI.

www/js/app.js
Jaifroid: OK, that's strange. It should show as a redirect when examining the dirEntry, because W/mainPage is supposed to be a redirect to an actual HTML (usually) entry. I'll see if I can troubleshoot this.

Jaifroid (commented):
Congratulations, it's working well, by my tests, on non-Zimit-based archives, which is great!

I have to say I do find the way of enabling this reading mode to be confusing in the UI. Because with the dropdown, it's not clear what the difference is between "Default" and, say, "WASM". In fact, both of those settings use the production WASM version of libzim, but with default it is only used for search, and with WASM it is used for both search and reading the ZIM contents. But it's really not clear to the user what's going on here.

Additionally, the label seems to imply it only enables WASM, whereas in fact it has five options, only two of which will force use of WASM. It also implies that the "WASM API" is unstable (WASM is a very stable technology, what is unstable is the libzim port, whether used in WASM or ASM mode).

I think this can be cleared up by having the dropdown govern which version of libzim is loaded and used for full-text search, and additionally a user can enable using the selected version of libzim also for reading the ZIM contents (with a checkbox underneath the dropdown), something like:

  [ ] Also use the selected libzim for reading ZIM content (**unstable**)

This should be a simple addition. The checkbox would enable and remember the option `params.useLibzim`.

Jaifroid:
In the meantime, I'll try to debug what's going on with `W/mainPage`.

Jaifroid:
@Rishabhg71 Thanks for changing that -- the UI is much clearer now, and you've also managed to declutter by hiding the dropdown till the user selects to use libzim. Could you change the tooltip that appears when hovering over the dropdown from "Libzim wasm mode" to "Uses the selected version of libzim to access the ZIM contents (ServiceWorker mode only)"? Also, change the label from "Libzim version: (Select which version to use)" to "Select libzim version to use:" (in bold). And you can remove the tooltip for the Use libzim checkbox (we don't need two similar tooltips).

I still haven't had a chance to debug as I said above, but in the meantime, when you're satisfied with the UI, perhaps you could work on the translations?

Rishabhg71:
@Jaifroid The Requested changes are done. Could you check the translations if they are correct? It was auto-generated by Copilot

i18n/fr.jsonp.js
```diff
@@ -93,6 +93,11 @@ document.localeJson = {
       "configure-expert-disabledragdrop-tip": "Cette option désactive la fonctionnalité de glisser-déposer des fichiers ZIM pour les charger dans l'application. Il n'est pas recommandé de désactiver cette fonctionnalité.",
       "configure-expert-bypassappcache": "<strong>Désactiver AppCache</strong> (<i>désactive l'utilisation hors ligne de cette application !</i>)",
       "configure-expert-bypassappcache-tip": "AVERTISSEMENT&nbsp;: Désactiver cette option désactivera l'utilisation hors ligne de la PWA. Le paramètre effacera tous les caches de l'API de Cache, mais le cache des actifs sera utilisé à moins qu'il ne soit également désactivé ci-dessus. Il facilite le développement de nouveaux codes dans la PWA.",
+      "configure-expert-useLibzim-tip": "Utiliser la bibliothèque ZIM pour lire les fichiers ZIM (nécessite un redémarrage).",
+      "configure-expert-useLibzim": "Utiliser la bibliothèque ZIM ",
+      "configure-expert-useLibzim-warning": "Utilise l'API libzim instable pour lire zim",
+      "configure-expert-libzimMode-tip": "Mode de la bibliothèque ZIM (défaut&nbsp;: 0)",
```
Jaifroid, line 99: It says "Mode of the ZIM library (default: 0)" !!!

i18n/fr.jsonp.js
```diff
@@ -93,6 +93,11 @@ document.localeJson = {
...
+      "configure-expert-libzimMode": "Mode de la bibliothèque ZIM",
```
Jaifroid, line 100: Again, wrong.

www/index.html
```diff
@@ -694,6 +694,23 @@ <h3 data-i18n="configure-expert-settings-title">Expert settings</h3>
...
                                     </label>
                                 </div>
+                                <div class="checkbox" id="">
+                                    <label data-i18n-tip="configure-expert-useLibzim-tip" title="Uses the selected version of libzim to access the ZIM contents (ServiceWorker mode only)">
+                                        <input type="checkbox" name="useLibzim" id="useLibzim">
+                                        <span><strong data-i18n="configure-expert-useLibzim">Use Libzim for loading zim</strong>(<i data-i18n="configure-expert-useLibzim-warning">Uses unstable libzim api for reading zim</i>)</span>
+                                    </label>
+                                </div>
+                                <div class="checkbox" id="libzimMode" style="display: none;">
+                                    <label data-i18n-tip="configure-expert-libzimMode-tip" title="Libzim version: (Select which version to use)">
```
Jaifroid, line 704: You shouldn't use a tooltip title that merely says the same thing as the visible text (in essence). So, either remove the tooltip here, or else put something more informative in (e.g. that the setting will refresh the page).

www/js/app.js
```diff
@@ -886,6 +899,49 @@ function initServiceWorkerMessaging () {
...
+        const ret = await selectedArchive.callLibzimWorker({ action: 'getEntryByPath', path: title })
+        if (ret === null) {
+            console.error('Title ' + title + ' not found in archive.');
+            messagePort.postMessage({ action: 'giveContent', title: title, content: '' });
+            return;
+        }
+
+        if (ret.mimetype === 'unknown') {
+            // We have a redirect to follow
+            // this is still a bit flawed, as we do not check if the redirect or the file doesnt exist
```
Jaifroid, line 923: I would say ""if the redirect or the file actually exist". In any case, `doesn't` needs and apostrophe if you keep it.

www/js/init.js
Jaifroid: You are listing permitted properties, but the actual type is `{String}`. A value's type and its permitted values are not the same thing in my books. But I may be wrong, and this could be as JSDoc convention. Can you point me to any JSDoc documentation that this is the correct way to list accepted values? I think a parallel might be listing a type a `{Boolean}` as opposed to listing it as `{True | False | Null | 1 | 0 | "" | Undefined}`, because those are all permitted Boolean values.

Jaifroid (changes requested):
Thank you very much for this PR, we're nearly there, but unfortunately you've fallen victim to the AI hallucination effect, as the translations are pretty much all hallucinated. I think only one was correct! Not your fault, but I'd always advice to check a translation by doing reverse translation back into English. A few other things noted.

www/js/init.js
Rishabhg71: https://www.geeksforgeeks.org/what-are-string-literal-types-in-tpescript/
I could not find any proper article explaining string literal in JS but here is an article explaining using typescript.


If we use the type Boolean, we ought to stick to either true or false. However, because JS isn't type-safe, it doesn't enforce this, so we're treating truthy/falsy values as booleans here. The reason for preferring string literals over strings is their resemblance to enums. Rather than permitting any string value, we restrict it to specific strings as the type. When a new user edits this value, the editor will indicate the exact type, aiding the developer in understanding the precise values a particular variable might hold.

Jaifroid: OK, you've clearly researched this, thanks for the info. I learnt something.😉

i18n/en.jsonp.js
```diff
@@ -93,6 +93,11 @@ document.localeJson = {
       "configure-expert-disabledragdrop-tip": "This option disables drag-and-drop of ZIM files onto the app. It is not recommended to disable this feature.",
       "configure-expert-bypassappcache": "<b>Bypass AppCache</b> (<i>disables offline use of this app!</i>)",
       "configure-expert-bypassappcache-tip": "WARNING: Leaving this checked will prevent offline usage of the PWA. Setting will clear all existing Cache API caches, but assetsCache will be used unless also disabled above. For testing new code with the PWA.",
+      "configure-expert-useLibzim-tip": "Uses the selected version of libzim to access the ZIM contents (ServiceWorker mode only)",
+      "configure-expert-useLibzim": "Use libzim for loading ZIM",
+      "configure-expert-useLibzim-warning": "uses unstable libzim API for reading ZIM",
+      "configure-expert-libzimMode-tip": "Libzim version: (Select which version to use)",
```
Jaifroid, line 99: This tool tip still seems redundant to me. I'd recommend removing it here, in the other languages, and in the HTML. Tooltips should add info, not paraphrase what's already obvious from the UI.

i18n/es.jsonp.js
```diff
@@ -93,6 +93,11 @@ document.localeJson = {
...
       "configure-expert-bypassappcache-tip": "ADVERTENCIA: Dejar esto marcado deshabilitará el uso sin conexión de la PWA. El ajuste borrará todas las cachés de la API de Caché, pero se utilizará la caché de activos a menos que también se desactive arriba. Facilita el desarrollo de nuevo código en la PWA.",
+      "configure-expert-useLibzim-tip": "Si está usando el modo ServiceWorker, puede elegir entre usar la biblioteca Libzim o la biblioteca ZIMjs para cargar los archivos ZIM. La biblioteca Libzim es más rápida, pero la biblioteca ZIMjs es más robusta y puede manejar archivos ZIM más grandes. Si tiene problemas con la biblioteca Libzim, intente cambiar a la biblioteca ZIMjs.",
+      "configure-expert-useLibzim": "Usar la biblioteca Libzim ",
+      "configure-expert-useLibzim-warning": "¡ADVERTENCIA! Esta opción sólo se aplica al modo ServiceWorker.",
+      "configure-expert-libzimMode-tip": "Seleccione la versión de la biblioteca Libzim que desea usar. La versión 5 es más rápida, pero la versión 6 es más robusta y puede manejar archivos ZIM más grandes. Si tiene problemas con la versión 5, intente cambiar a la versión 6.",
```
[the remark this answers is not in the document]
Jaifroid: Sorry, but you haven't changed this laughable version! It still says "Version 5 is faster but Version 7 is more robust and can deal with larger ZIMs...." Etc. Pure hallucination. However, as per recommendation above, I'd just remove this.

i18n/es.jsonp.js
```diff
@@ -93,6 +93,11 @@ document.localeJson = {
...
+      "configure-expert-libzimMode": "Versión de la biblioteca Libzim",
```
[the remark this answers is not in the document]
Jaifroid: Translation is still wrong. You need "Seleccionar versión libzim para usar:" (be sure to add colon at end)

i18n/fr.jsonp.js
Jaifroid: Umm, you didn't address this point! Again, pure hallucination.

i18n/fr.jsonp.js
Jaifroid: Same, you marked this as resolved, but didn't change it.

Jaifroid (changes requested):
Unfortunately, you marked some of my review comments from last time as resolved when they haven't changed at all, and still have hallucinated translations in them. Maybe you didn't expand the conversations in the conversations tab (GitHub tends to hide some if there are too many). In future, it's best to look at review comments directly in the Files changed tab, where they will always show.

www/js/init.js
Jaifroid: I would just add that as our convention throughout is to use single quotes, please change the use of double quotes to single here, otherwise it looks incongruous.

Rishabhg71:
Could you take another look it weird I did the translations via [deepl](https://www.deepl.com/) maybe I messed up something

Jaifroid:
> Could you take another look it weird I did the translations via [deepl](https://www.deepl.com/) maybe I messed up something

The problem was the old translations that got missed probably due to the GitHub comment-hiding "feature", not so much the new ones. I'm looking now.

Jaifroid (approved):
Thanks, that's a great improvement! There are some last minor but niggly little things -- please carefully go through the comments below, and when you're happy you've addressed these, feel free to squash and merge, I don't need to check again. One you missed yesterday/today, was converting the double-quotes to single-quotes in your type declaration to match the house style, so don't forget to include it in your last revisions. If you see some comments are hidden by GitHub in the list below, please expand them and/or go to the Files Changed tab and view the Diffs on all files.

Jaifroid:
Overall, many thanks for this PR -- it's a nice final one for your internship. I'm very pleased with your overall achievements in this repo: your PRs make a real and tangible difference to the Kiwix JS app! Well done!

Issue #1146, opened 2023-11-05
Try to do without the `initOrKeepAliveServiceWorker` code by leveraging `clients.matchAll` in the Service Worker


Jaifroid:
We currently have some clunky code that forcibly keeps the Service Worker alive every 30 seconds. This works around a key CPU-sparing feature of Service Workers, which is that they are killed when not being used, and restart automatically when one of their events is fired. Unfortunately, we depend on a message channel sent from app.js for two-way communication. It is perfectly possible that a user will keep the app open doing other stuff for several minutes, and then will click a link. This link will be trapped by the newly awoken Service Worker, but it will no longer have access to the message channel that was sent from `app.js` before it was put to sleep. So, it has nowhere to send its request to.

However, if we use `clients.matchAll` with the `activate` event of the SW, we can communicate from the SW to all clients and get the client's context. We then only send the message to the client (or clients) that is (are) handling that particular context. In our case, a context would be a specific ZIM.

This ought to allow having multiple clients handling different ZIM archives. In other words, this would enable a multi-ZIM, multi-window configuration with a single Service Worker.

Jaifroid:
See https://web.dev/articles/two-way-communication-guide for more information.

Jaifroid:
This technique is working very well in the PWA, and opens up a number of possible future avenues.

PR #1163, opened 2023-11-15
Add rudimentary static Zimit support

Jaifroid:
This adds some rudimentary, support for entirely static Zimit archives. It doesn't attempt to run JavaScript, and leverages jQuery mode under the hood to transform static links only. I figure it is better than just dropping users into a blank page. We still tell users with a dialogue box that the ZIM is unsupported, so they don't have any illusions...

image


Jaifroid:
I think we've reached the point where we need #1146 before we can proceed any further.

Jaifroid:
@Rishabhg71 I'd like your views on whether to merge this PR. As you can see, it expands the range of ZIMs that the app can provide some support for, but it's not by any means a complete Zimit support. It ignores JavaScript completely. But if a site can be browsed without JS, then the site will most likely work in this version. That's the case at least for the Internet Encyclopedia of Philosophy, the Low Tech Magazine, and a few others. But other sites will look broken. I show the dialogue box below when a user first opens a ZIM of this type (or any time they go to the landing page), so no-one can say they were misled.

The question is, is it worth putting in an intermediate support like this? The full support of Zimit is technically very challenging, and will take some time, especially as the format is subject to change. But the fact is that we currently show a completely blank page, which is not helpful at all to end users (with a dialogue box). I feel that merging this and making it available to users at next release is an improvement over what we currently do. On the other hand, users might say "Why are you tempting us with a half-way house?"... I'm in two minds.

image

Rishabhg71:
I think it's okay to add it as an experimental feature. As a user, I would like to have a choice rather than none. It's up to me to decide how I want to use the product

PR #1164, opened 2023-11-15
Remove unnecessary initOrKeepAliveServiceWorker

### LINKED ISSUE #1146: Try to do without the `initOrKeepAliveServiceWorker` code by leveraging `clients.matchAll` in the Service Worker

Jaifroid:
Fixes #1146.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -94,7 +94,7 @@ jobs:
           local-testing: start
           local-identifier: random
 
-      - name: Run App locally in background
+      - name: Start server locally in background
         if: github.event_name == 'push' || github.event.pull_request.head.repo.full_name == github.repository
         env:
           BROWSERSTACK_USERNAME: ${{ secrets.BROWSERSTACK_USERNAME }}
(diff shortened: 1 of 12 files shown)


Jaifroid:
This PR enables independent browsing of different ZIMs. We can see this in Firefox 70 SW mode below, but I'm having issues with tests on this branch. The new way of communicating the message channel port seems not to sit well with the timings in some tests.

image

PR #1169, opened 2023-11-21
Avoid side effects when loading archive

Jaifroid:
This provides better async support for loading archives.

It also provides a whenZimReady function in the loading chain that makes it easier to wait for the ZIM: this will be important for libzim loading (much slower to initialize).

Additionally adds a flag to the ZimArchive object `libzimReady`, which can be 'loading', 'ready' or 'error'.

### DIFF
--- a/tests/unit/index.html
+++ b/tests/unit/index.html
@@ -1,6 +1,7 @@
 <!DOCTYPE html>
 <html>
     <head>
+        <meta name="viewport" content="width=device-width, initial-scale=1">
         <meta charset="utf-8">
         <base href="../../">
         <title>Kiwix-js Unit tests</title>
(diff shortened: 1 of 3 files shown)


Jaifroid:
@Rishabhg71 Just so you're aware, this PR (now merged) adds more robust process for loading ZIM archives, and in particular of interest to you will be the `whenZimReady` funciton added in zimArchive.js. At the moment, this is called after the legacy backend has loaded the ZIM archive, but before libzim is loaded (given that libzim is only used for full-text search currently, so we don't have to wait for its initialization in order to display the landing page of a ZIM).

I suggest you add a param when the user chooses to uses libzim for reading the archive, `params.useLibzim` (Boolean). You will then be able to add conditionality to the loading in `zimArchive.js`. If `params.useLibzim` is true, you should only call `whenZimReady()` after the `.libzimready` property is set to 'ready' (that should be obvious in the code).

Issue #1174, opened 2023-11-27
Fancy library doesn't work in browsers that do not support replaceAll


Jaifroid:
`replaceAll()` is used to construct the magnet link, and this throws an error in browser < Chrome/Edge 85 and Firefox 77. To be clear, this error only surfaces when you try to download a file. But as this is the whole point of the library, it really needs to be fixed...

See https://caniuse.com/?search=replaceAll.

We should probably polyfill that anyway, as the polyfill is available in core-js.

Jaifroid:
Actually, it's not yet in core-js 3. It's scheduled for 4 (a proposal).

We can get a polyfill like this:

`npm install --save es6-string.prototype.replaceall`

And where needed:

`import 'es6-string.prototype.replaceall';`

This would have to be added to the dependency grid. See https://github.com/kiwix/kiwix-js/blob/main/ADDING_DEPENDENCIES_NODE_MODULES.md for the method to do this.

Jaifroid:
@Rishabhg71 This is one that could interest you, as it could effectively fix your library feature for a wider range of browsers.

Greeshmanth1909:
Will it be possible to "artificially " define a string.prototype.replaceAll method in the global scope of the program as a temporary fix to this issue until the release?

Jaifroid:
> Will it be possible to "artificially" define a string.prototype.replaceAll method in the global scope of the program as a temporary fix for this issue until the release?

@Greeshmanth1909 I'll entertain any appropriate fix! We're probably a couple of weeks away from releasing v4.0, and it would certainly be good to get this fixed before that release. I don't think replaceAll should be too hard to emulate, and a full polyfill may be overkill.

Jaifroid:
@Greeshmanth1909 has discovered that this cannot be polyfilled in Kiwix JS due to CORS. However, it could be fixed upstream in kiwix-tools.

Jaifroid:
Since this isn't polyfilled, and we offer an alternative for such browsers, we can close this as won't fix.

PR #1190, opened 2024-01-08
Added String.prototype.replaceAll method

Greeshmanth1909:
resolves https://github.com/kiwix/kiwix-js/issues/1174
Added String.prototype.replaceAll method in case a browser doesn't support it by default.
Here are the tests I performed:
- Ran `npm run serve`
- Ran `npm run preview`
- Tested with `npm test`
In all the above cases I was able to browse and open zim files in both serviceWorker and JQuery modes in Chrome, Firefox (only JQuery) and Brave. I could't test the magnet link generation for older browsers.
- Tested the extension for Firefox.
- Tested the extension for Chrome in both JQuery and ServiceWorker modes.

### DIFF
--- a/www/js/app.js
+++ b/www/js/app.js
@@ -1388,7 +1388,7 @@ function displayFileSelect () {
 
     document.getElementById('archiveList').addEventListener('change', function (e) {
         // handle zim selection from dropdown if multiple files are loaded via webkitdirectory or filesystem api
-        localStorage.setItem('previousZimFileName', e.target.value);
+        settingsStore.setItem('previousZimFileName', e.target.value, Infinity);
         if (params.isFileSystemApiSupported) {
             return abstractFilesystemAccess.getSelectedZimFromCache(e.target.value).then(function (files) {
                 setLocalArchiveFromFileList(files);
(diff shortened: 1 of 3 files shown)


Jaifroid:
@Greeshmanth1909 Thank you! I'll test this on a Chrome < 85 and Firefox < 77 with the download library.

Jaifroid:
I should get to this this weekend.

Greeshmanth1909:
Pushed the code for a completely different issue to this pr by accident. Will fix it asap.

Jaifroid (changes requested):
Thanks very much for this PR. I tested it in Firefox Portable ESR 70.1, and unfortunately it doesn't appear to be working as intended. I get the error `magnetLink.replaceAll is not a function` when clicking on the "Download" link in the library. See screenshot below. Perhaps you haven't injected the function into the correct context? The library code runs in an iframe,

image

Jaifroid:
You can get different versions of Firefox Portable here: https://sourceforge.net/projects/portableapps/files/Mozilla%20Firefox%2C%20Portable%20Ed./

Greeshmanth1909:
Ill try fixing it. Thanks for the resource.

Jaifroid:
I was thinking you might run into CORS issues attempting to inject this function into the library iframe. In fact you almost definitely will, unless there is a known way to polyfill browser functions in a way that works also in iframes with third-party content. It seems unlikely, as it would be considered a vector for attack.

Given this, the only solution -- I think -- would be to make a PR (or suggest a PR) over at kiwix-tools (https://github.com/kiwix/kiwix-tools). I'm sorry if the issue misled you here, I only realized this might be an issue after polyfilling the function in `init.js` appears to have failed.

Greeshmanth1909:
I've also tried defining the polyfill in an inline script of `library.html` it didn't work.

Greeshmanth1909:
Shall I close this pr? @Jaifroid

PR #1192, opened 2024-01-13
Added a feature that prompts a user when a new Zim file is opened.

### LINKED ISSUE #974: Prompt user to consider whether they trust the source of a ZIM before allowing the user to open it
This is a follow-on from #753. Since we cannot fully mitigate against malicious ZIM content, we could instead ask a user whether they trust the source of the ZIM before loading it, and warn them of bad consequences if it contains malicious scripts.

Greeshmanth1909:
Might resolve #974
An alert will appear when a user opens a file for the first time asking weather they trust the source of the file or not. If they do, the file will be read in the usual way. If the user does not trust the source, the file will be read in JQuery mode.
The option to disable this feature was not added.
Once the user says they trust a particular file, the same prompt will not appear for that file in the future.


Greeshmanth1909:
@Jaifroid I was able to view and browse files once I confirmed the source with the 'trust' button in the pop up. 
The alert box that warns the user before opening a new file might be causing the errors we are seeing here imo. What do you think?

Jaifroid:
I'm pretty sure the tests are failing due to the alert. I think you'll have to add an option to bypass this (trust all local ZIM archives), and enable the bypass for tests. You can add a simple `params.trustAllLocalZIMArchives` Boolean for now, and set it to true for tests.

Greeshmanth1909:
@Jaifroid  Only one of the e2e tests failed. Can you re-run?

Jaifroid:
A couple of quick observations before you go much further with this PR:

* You seem to be asking the user if they trust the ZIM in the `readArticle` function. However, that seems too late to me, especially as the query to the user is async, so in fact the readArticle will go ahead while the user's consent is being requested (I haven't checked your code in detail, so not sure if you accounted for this already). However, it seems to me that this question should be asked in `archiveReadyCallback`, i.e. once the archive is ready, but before `goToMainAritcle` is called, and hence before any assets (JS in particular) can be loaded.
* I think it will be necessary to allow the user to turn off the warnings completely in the UI (probably an Expert Setting, with appropriate warning about security threat).

Greeshmanth1909:
The user can enable this feature in the expert settings via the "Enable source verification" checkbox which will be unchecked by default for testing purposes. This can however be changed to have the feature checked by default if necessary.

Jaifroid:
> The user can enable this feature in the expert settings via the "Enable source verification" checkbox which will be unchecked by default for testing purposes. This can however be changed to have the feature checked by default if necessary.

I would say that the feature should be on by default, since it is a security feature: users should be protected by default. In a testing context, of course, it will be turned off in code.

Jaifroid:
Also, you will see that code scanning is producing a high-risk alert. I'm not at all sure why it's being flagged as high risk -- I can only assume it is based on the naming of the variables (with the word `trusted`) that is flagging this as not a trustworthy way of storing trusted data.

What is your assessment of the actual risk here? I can't see a way of encrypting the data without including the keys for encrypting and decrypting them in the app. Since this is an offline solution, we can't use server-based encryption/decryption.

Greeshmanth1909:
This feature can be enabled by default by simply changing a Boolean to true. I disabled it by default so that it can be tested, thinking it can be changed in production. 

Either we automatically disable this during testing or tweak the tests to accommodate this feature.
 
Do you see another way @Jaifroid ?

Greeshmanth1909:
We store the names of the zim files that the user flags as "trusted" in local storage as a continuous string of file names separated by "|".
I named the key of this value as 'trustedZimFiles' for simplicity. 
As we are only reading the file names, I don't see any security risk here.

Greeshmanth1909:
Should I try changing the key to something else?

Jaifroid:
On security, I agree. I think this is a false positive based on matching the word "trusted". However, I don't think we should change the variable name because of that.

On the default, I suggest turning the feature on by default. In the testing script(s) you then need to set the appropriate param so that tests will be able to load the ZIMs they test against without interruption. Note that Unit tests contain their own `init.js`, so you could set the param there. CI tests will also need adjusting in the scripts that run those, or by testing for a CI environment. Note that we test on GitHub, but we also test on Browser Stack, but the Browser Stack tests don't run for third-party contributors like yourself. I will need to add a mirror PR for them to run.

Greeshmanth1909:
So I just have to manually change the params object in the tests/unit/js/init.js
I don't completely understand what to do for the CI test though.

Jaifroid:
Yes, for Unit tests, just set the param in that `init.js`. For the end-to-end tests, there are test files `legacy-ray_charles.e2e.spec.js` and  `gutenberg_ro.e2e.spec.js` under `tests/e2e/spec/`. These tests run against compiled versions of the code, so you can't just edit an init file. There are two approaches I can think of:

1. Write code in these test files that tests for and dismisses the alert box when the archives are opened (you can copy the style of code for dismissing other alerts already in these files - search for "dismiss");
2. Alternatively, set the `params.sourceVerifiaction` Boolean by passing it to the app by using `await driver.executeScript(...)`. This would need to be done after the app has loaded, but before the archives are loaded into the app in these scripts.

The first would be the more useful in a way because it would also be testing your code's UI. The same test runners run in GitHub actions as in Browser Stack, so you should be able to see if your code is working in your PR, and when it's working with the GitHub e2e tests, I can make a mirror PR (which I won't merge) that will allow your code to run on Browser Stack as well, to be sure all is working well.

Greeshmanth1909:
Thanks a lot for the info! 
I'll fix this asap.

Greeshmanth1909:
I've made the necessary changes, please go ahead with the Browser Stack tests @Jaifroid

Jaifroid:
@Greeshmanth1909 Thanks! I ran the Linux tests twice. The first one appeared to be a random failure, but the second one looks like an issue with the PR. You can see the error messages here: https://github.com/kiwix/kiwix-js/actions/runs/7708040000/job/21006685376?pr=1192, and you can examine the session on Browser Stack (and view the video) here: https://automate.browserstack.com/builds/02e2058e5efa9ad2a556c82e00ff2ebaa6322d33/sessions/39e1472142df23ca14a8c9ea71fc746bd2081e29?auth_token=e6d8704fd572b3fdee6dfa06f7d84dbc7367cdf8f437f38e5db0ed06439297d5

One thing I noticed is that the failed session was already running in jQuery mode, but your dialogue box asking if the user trusts the source still popped up. If you add code to prevent the popup showing if the user is already in jQuery mode (just check `params.contentInjectionMode`) it will allow us to progress these tests to the next stage and see if they are also failing in SW mode.

The tests will need to be adjusted also so that the detection of the dialogue box doesn't occur in the jquery test. You can easily tell in the test code if it is running in jquery or sw mode with `mode === 'jquery'` (will be true or false).

Jaifroid:
I ran tests twice with the same result -- looks like Gutenberg SW-mode tests are systematically failing. Possibly also Ray Charles SW tests, though that's not so uniform.

Greeshmanth1909:
Yeah got similar results on my machine. 
It looks like the button that must be clicked is either not clickable or not scrolled into view. I've tried increasing the wait time and scrolling the element into view, nothing seems to be happening.

Greeshmanth1909:
Quick question, as you can see in `app.js` i'm also testing if serviceworker API is available with `isServiceWorkerAvailable()`. If the Fireforx version that was mentioned earlier only supports jquery, shouldn't the condition fail? What am I missing here?

Jaifroid:
> Quick question, as you can see in `app.js` I'm also testing if serviceworker API is available with `isServiceWorkerAvailable()`. If the Fireforx version that was mentioned earlier only supports jquery, shouldn't the condition fail? What am I missing here?

I'm not sure if it affects your code, but you don't need to test if the Service Worker is available, because if `params.contentInjectionMode === 'serviceworker'`, then the SW must be available. What I can say is that on some versions of Firefox (old ones, possible this one), the Service Worker is available, but we can't actually use it due to the sandbox. This is mentioned in the README I think. In any case it should be immaterial in a way, because all you need to test is whether we're running in jQuery mode, and disable your dialogue box if so. The tests on the affected browsers are already set only to run in jQuery mode.

Jaifroid:
By the way, you can run tests locally on your own machine with `npm run test-e2e-firefox` (or chrome, or edge), and it can help you observe what is going on. Maybe you're already doing this? The problem is that these tests run on new browsers, so they don't always reflect 100% what will happen on old browsers.

Greeshmanth1909:
Here are the changes I've made:
- The dialog box will only appears if the user is in SW mode.
- If the user switches to jquery mode, the check box that enables/disables source verification will disappear and will reappear once the app is switched back to SW mode.
- The e2e tests have also been modified to disable source verification by clicking the corresponding checkbox. This was done as clicking the `Trust Source` in the automated environment lead to inconsistency in the test results.
- I've throughly tested this feature manually in both Firefox and Chrome for both modes and everything behaves as expected.

Jaifroid:
> Here are the changes I've made:
> 
> * The dialog box will only appear if the user is in SW mode.
> * If the user switches to jquery mode, the check box that enables/disables source verification will disappear and will reappear once the app is switched back to SW mode.
> * The e2e tests have also been modified to disable source verification by clicking the corresponding checkbox. This was done as clicking the `Trust Source` in the automated environment lead to inconsistency in the test results.
> * I've throughly tested this feature manually in both Firefox and Chrome for both modes and everything behaves as expected.

That sounds like excellent work! I'll update the mirror branch. In the meantime, take a look at the minor formatting issues identified by codefactor. It's cosmetic stuff, but should be fixed.

Jaifroid:
Congrats, it's passing all Browser Stack tests! Only the CodeFactor cosmetic stuff to deal with. I'll review the code.

tests/e2e/spec/gutenberg_ro.e2e.spec.js
```diff
@@ -182,6 +182,12 @@ function runTests (driver, modes) {
                     console.log('\x1b[33m%s\x1b[0m', '      Skipping SW mode tests because browser does not support API');
                     await driver.quit();
                 }
+                if(mode === 'serviceworker') {
```
Jaifroid, line 185: Please add a comment so that future developers know what this code does.

tests/e2e/spec/legacy-ray_charles.e2e.spec.js
```diff
@@ -194,8 +194,14 @@ function runTests (driver, modes) {
                     console.log('\x1b[33m%s\x1b[0m', '      Skipping SW mode tests because browser does not support API');
                     await driver.quit();
                 }
+                if(mode === 'serviceworker') {
```
Jaifroid, line 197: Ditto about comment.

tests/unit/js/init.js
```diff
@@ -28,7 +28,8 @@
 // Define global params needed for tests to run on existing app code
 // eslint-disable-next-line no-unused-vars
 var params = {};
-
+// setting source verification to false will enable proper testing.
```
Jaifroid, line 31:
```suggestion
// We need to turn off source verification so that the test files can be loaded normally without interruption
```

www/index.html
```diff
@@ -694,6 +694,12 @@ <h3 data-i18n="configure-expert-settings-title">Expert settings</h3>
                                         <span data-i18n="configure-expert-bypassappcache"><b>Bypass AppCache</b> (<i>disables offline use of PWA!</i>)</span>
                                     </label>
                                 </div>
+                                <div class="checkbox" id="enableSourceVerificationCheckBox">
+                                    <label data-i18n-tip="configure-expert-enaable-source-verification" title="Warning: A few zim files might run malicious programs in your browser. This can however be prevented by using jquery mode, kiwix will automatically switch to jquery mode if you  mention you donot trust the source. Disabling this option will result in using serviceworker mode by default, if available." >
```
Jaifroid, line 698: Just some suggested corrections for clarity:

```suggestion
                                    <label data-i18n-tip="configure-expert-enaable-source-verification" title="Warning: Some ZIM archives from untrusted sources could run malicious code in your browser. This can be prevented by using JQuery mode, which cannot run active content from the ZIM. Highly dynamic ZIMs will probably fail in JQuery mode, but ZIMs with largely static content should work. If you trust the source of all of your ZIMs, then disabling this option will use ServiceWorker mode by default, if available.">
```

www/index.html
```diff
@@ -694,6 +694,12 @@ <h3 data-i18n="configure-expert-settings-title">Expert settings</h3>
...
+                                        <input type="checkbox" name="disableFileVerification" id="enableSourceVerification" >
+                                        <span data-i18n="configure-expert-disable-source-verification"><strong>Enable source verification of new files</strong> (once checked, kiwix will ask to verify file source before opening.)</span>
```
Jaifroid, line 700: Another suggestion for clarity:

```suggestion
                                        <span data-i18n="configure-expert-disable-source-verification"><strong>Enable source verification of new files</strong> (<i>recommended</i>: you will only be prompted the first time you open a ZIM)</span>
```

www/js/app.js
```diff
@@ -496,6 +496,25 @@ document.getElementById('disableDragAndDropCheck').addEventListener('change', fu
         }
     });
 });
+// Handle switching from jQuery to serviceWorker modes.
+document.getElementById('serviceworkerModeRadio').addEventListener('click', function() {
+    if (this.checked) {
+        document.getElementById('enableSourceVerificationCheckBox').style.display = 'block'
```
Jaifroid, line 502: Better to use `style.display = '';` rather than guessing if it's `block` or `inline`.

Also note missing semicolon at end of line.

www/js/app.js
```diff
@@ -496,6 +496,25 @@ document.getElementById('disableDragAndDropCheck').addEventListener('change', fu
         }
     });
 });
+// Handle switching from jQuery to serviceWorker modes.
+document.getElementById('serviceworkerModeRadio').addEventListener('click', function() {
+    if (this.checked) {
+        document.getElementById('enableSourceVerificationCheckBox').style.display = 'block' 
+        params.sourceVerification = getSetting('sourceVerification') === null ? true : getSetting('sourceVerification')
```
Jaifroid, line 503: Missing semicolon.

www/js/app.js
```diff
@@ -496,6 +496,25 @@ document.getElementById('disableDragAndDropCheck').addEventListener('change', fu
...
+    }
+}) 
+document.getElementById('jqueryModeRadio').addEventListener('click', function() {
+    if (this.checked) {
+        document.getElementById('enableSourceVerificationCheckBox').style.display = 'none'
```
Jaifroid, line 508: Ditto

www/js/app.js
```diff
@@ -496,6 +496,25 @@ document.getElementById('disableDragAndDropCheck').addEventListener('change', fu
...
+    }
+})
+// Source verification is only makes sense in SW mode as doing the same in jQuery mode is redundant.
+document.getElementById('enableSourceVerificationCheckBox').style.display = params.contentInjectionMode === 'serviceworker' ? 'block' : 'none';
+
+document.getElementById('enableSourceVerification').addEventListener('change', function() {
+    params.sourceVerification = this.checked;
+    settingsStore.setItem('sourceVerification', this.checked, Infinity)
```
Jaifroid, line 516: Ditto

www/js/app.js
```diff
@@ -496,6 +496,25 @@ document.getElementById('disableDragAndDropCheck').addEventListener('change', fu
...
+})
```
Jaifroid, line 517: Ditto

www/js/app.js
```diff
@@ -496,6 +496,25 @@ document.getElementById('disableDragAndDropCheck').addEventListener('change', fu
...
+    if (this.checked) {
+        document.getElementById('enableSourceVerificationCheckBox').style.display = 'block' 
+        params.sourceVerification = getSetting('sourceVerification') === null ? true : getSetting('sourceVerification')
+    }
+}) 
+document.getElementById('jqueryModeRadio').addEventListener('click', function() {
+    if (this.checked) {
+        document.getElementById('enableSourceVerificationCheckBox').style.display = 'none'
+    }
+})
```
Jaifroid, line 510: Ditto

www/js/app.js
```diff
@@ -1650,6 +1669,27 @@ function archiveReadyCallback (archive) {
...
         }
     }
+    if (settingsStore.getItem('trustedZimFiles') === null) {
+        settingsStore.setItem('trustedZimFiles', '', Infinity);
+    }
+    if (params.sourceVerification && params.contentInjectionMode === 'serviceworker'){
+        // Check if source of the zim file can be trusted.
+        if (!(settingsStore.getItem('trustedZimFiles').includes(archive.file.name))) {
+            // Alert user about unknown source.
+            console.log("check22")
```
Jaifroid, line 1679: Be sure to remove this before merge.

www/js/app.js
```diff
@@ -1650,6 +1669,27 @@ function archiveReadyCallback (archive) {
...
+            const response = await uiUtil.systemAlert('Do you trust this source?\n You can still read the zim file in Safe Mode incase you cannot verify the source. Closing this window also opens the file in Safe Mode. This option can be disabled in Expert Settings', 'Alert!', true, 'Open in Safe Mode', 'Trust Source')
```
Jaifroid, line 1680: As this will be inserted as HTML, it's no good using newline codes, they'll be ignored. You should instead wrap each separate paragraph in `<p>...</p>` inside this string.

Some stylistic suggestions:

```suggestion
            const response = await uiUtil.systemAlert('Is this ZIM archive from a trusted source?\n If not, you can still read the ZIM file in Safe Mode (aka JQuery mode). Closing this window also opens the file in Safe Mode. This option can be disabled in Expert Settings', 'Security alert!', true, 'Open in Safe Mode', 'Trust Source');
```

www/js/init.js
```diff
@@ -125,6 +125,7 @@ params['cacheAPI'] = 'kiwix-js'; // Sets name of the prefix used to identify the
 params['cacheIDB'] = 'kiwix-zim'; // Sets name of the Indexed DB database
 params['isFileSystemApiSupported'] = typeof window.showOpenFilePicker === 'function'; // Sets a boolean indicating whether the FileSystem API is supported
 params['isWebkitDirApiSupported'] = 'webkitdirectory' in document.createElement('input'); // Sets a boolean indicating whether the Webkit Directory API is supported
+params['sourceVerification'] = params.contentInjectionMode === 'serviceworker' ? (getSetting('sourceVerification') === null ? true : getSetting('sourceVerification')) : false;// Sets a boolean indicating weather a user trusts the source of zim files.
```
Jaifroid, line 128: "Boolean" should be capitalized (initial capital B). Also, not full stop at end of line (as per surrounding style -- we only use full stops if comment is more than one sentence).

Jaifroid (changes requested):
This is looking good, mostly minor suggestions in my review. NB I haven't yet tested the code. Note that once you have finalized (and I've checked) the text of the dialogue box and tool tips, etc., you'll need to translate them into French and Spanish and add the strings to the English, French and Spanish translation `jsonp` files under `i18n`. You can use [deepL](https://www.deepl.com/translator) to get good translations, and I know those languages so can check for machine translation glitches.

Greeshmanth1909:
I totally overlooked that!

Greeshmanth1909:
@Jaifroid  I've made all the changes you've requested. Please note that no code was added to let the users know that they are in Safe Mode/jQuery Mode yet.
Also codefactor points to a trailing white space error that doesn't make sense.

Greeshmanth1909:
@Jaifroid I've made the requested changes. Please note that no code has been added to let the user know that they're in Safe Mode/Jquery mode yet.
Also I think codefactor is pointing to a false positive error in this case. I've checked for trailing white spaces and couldn't find any.

Jaifroid:
@Greeshmanth1909 I think the trailing space is what is shown in my copy of VS Code in screenshot below. It's labelled as the wrong line in Codefactor. Just remove the space characters that are underlined in screenshot. You don't have to close up vertical space. However, I'm just running extended tests, so wait till those have finished before doing this and updating the branch to latest.

image

Jaifroid:
Tests completed successfully.

Jaifroid:
I think at the same time that you put the app into temporary jQuery mode (safe mode), you need to issue this instruction:

`document.getElementById('jqueryModeRadio').checked = true;`

This should set the radio switch to jQuery mode without triggering the full setContentInjectionMode code (and hence a jump to local code in the extension). It also means the user can set it back to SW mode by clicking the button.

Likewise, when the user opens a new ZIM, you will need to reset the UI to SW mode (if that was the original mode) with:

`document.getElementById('serviceworkerModeRadio').checked = true`

I think it's important that the switch doesn't get desynchronized from the actual state of the app.

Jaifroid:
There's one complication: there is a third setting, `serviceWorkerLocal` that happens when a Chromium extension is using local code. So you may need to check if `params.serviceWorkerLocal` is true before switching the UI back to "ServiceWorker mode" in the way described abvoe. If it's true, then you'll need to use this instruction instead:

`document.getElementById('serviceworkerLocalModeRadio').checked = true`

www/js/app.js
```diff
@@ -1650,6 +1680,31 @@ function archiveReadyCallback (archive) {
             params.originalContentInjectionMode = null;
         }
     }
+    // Set contentInjectionMode to serviceWorker when opening a new archive in case the user switched to Safe Mode/jquery Mode when opening the previous archive
+    if (params.contentInjectionMode === 'jquery') {
+        params.contentInjectionMode = settingsStore.getItem('contentInjectionMode');
```
Jaifroid, line 1685: Be careful, the value in settingsStore can be `serviceworkerlocal`. You have to test for that, and if it is `serviceworkerlocal`, you need to set `params.serviceWorkerLocal = true;` and params.contentInjectionMode = `serviceworker`.

Jaifroid:
@Greeshmanth1909 Release of v4.0 of Kiwix JS is getting close, and it would be great if we could include your PR in that release. I think your PR is very close to being ready. I can help with translation of the UI to French and Spanish once you've finalized the code. Do you think it's feasible to finalize soon?

Greeshmanth1909:
Yes, I will finish this today!

Greeshmanth1909:
@Jaifroid I think the code is ready for the final review. I've added the changes you requested and tested the chrome extension, it worked as expected.

www/js/app.js
```diff
@@ -496,6 +496,60 @@ document.getElementById('disableDragAndDropCheck').addEventListener('change', fu
...
+});
+document.getElementById('jqueryModeRadio').addEventListener('click', function() {
+    if (this.checked) {
+        document.getElementById('enableSourceVerificationCheckBox').style.display = 'none';
+    }
+});
+// Handle switching to serviceWorkerLocal mode for chrome-extension
+document.getElementById('serviceworkerLocalModeRadio').addEventListener('click', async function() {
+    document.getElementById('enableSourceVerificationCheckBox').style.display = ''; 
+    params.sourceVerification = getSetting('sourceVerification') === null ? true : getSetting('sourceVerification');
```
Jaifroid, line 525: My IDE says getSetting is undefined.

www/js/app.js
```diff
@@ -496,6 +496,60 @@ document.getElementById('disableDragAndDropCheck').addEventListener('change', fu
...
+            }
+    }
+});
+document.getElementById('jqueryModeRadio').addEventListener('click', function() {
+    if (this.checked) {
+        document.getElementById('enableSourceVerificationCheckBox').style.display = 'none';
+    }
+});
+// Handle switching to serviceWorkerLocal mode for chrome-extension
+document.getElementById('serviceworkerLocalModeRadio').addEventListener('click', async function() {
```
Jaifroid, line 523: Please add a space between `function` and `()` as per style.

www/js/app.js
```diff
@@ -496,6 +496,60 @@ document.getElementById('disableDragAndDropCheck').addEventListener('change', fu
...
+    document.getElementById('enableSourceVerificationCheckBox').style.display = ''; 
+    params.sourceVerification = getSetting('sourceVerification') === null ? true : getSetting('sourceVerification');
+    if (selectedArchive.isReady() && !(settingsStore.getItem("trustedZimFiles").includes(selectedArchive.file.name))) {
```
Jaifroid, line 526: Use single quote marks, not double, please.

www/js/app.js
```diff
@@ -496,6 +496,60 @@ document.getElementById('disableDragAndDropCheck').addEventListener('change', fu
...
+            if (response) {
+                params.contentInjectionMode = 'serviceworker';
+                var trustedZimFiles = settingsStore.getItem('trustedZimFiles');
+                var updatedTrustedZimFiles = trustedZimFiles + archive.file.name + '|';
+                settingsStore.setItem('trustedZimFiles', updatedTrustedZimFiles, Infinity);
+                
+            } else {
+                params.contentInjectionMode = 'jquery';
+                document.getElementById('jqueryModeRadio').checked = true;
+            }
```
Jaifroid, line 538: Reduce indentation of these lines so that they are at same level as line 527.

www/js/app.js
```diff
@@ -496,6 +496,60 @@ document.getElementById('disableDragAndDropCheck').addEventListener('change', fu
...
+                params.contentInjectionMode = 'serviceworker';
+                var trustedZimFiles = settingsStore.getItem('trustedZimFiles');
+                var updatedTrustedZimFiles = trustedZimFiles + archive.file.name + '|';
+                settingsStore.setItem('trustedZimFiles', updatedTrustedZimFiles, Infinity);
+            } else {
+                params.contentInjectionMode = 'jquery';
+                document.getElementById('jqueryModeRadio').checked = true;
+            }
+    }
+});
```
Jaifroid, line 516: Apply same fixes to code block above as identified in code block for `serviceworkerLocalModeRadio` below. Same undefined variables and formatting issues.

Jaifroid (changes requested):
Thanks, we're nearly there, but some errors crept in! Should be easy to clean up, and consolidate repeated code.

Greeshmanth1909:
Sorry for the oversight. Fixing it.

www/js/app.js
Greeshmanth1909: the line wasn't doing anything, I added it by accident.

Greeshmanth1909:
Done!

Greeshmanth1909:
Please do not review the code yet, I might have found another bug.

Greeshmanth1909:
In the chrome extension, the checkbox that controls source verification behaves as expected (appearing/disappearing when switching between `serviceWorkerLocal` and `jQuery` modes)

swl

But, when i switched to `ServiceWorker` mode, the html of this checkbox disappeared (as shown by the console output) which is not the case in jQuery mode.

Screenshot 2024-02-16 at 4 23 47 PM

Here in `Jquery` mode the checkbox is hidden with css, but the html is still present.

Screenshot 2024-02-16 at 4 12 58 PM

Greeshmanth1909:
I don't understand what is causing the HTML to disappear like that. @Jaifroid I need your help figuring this out.

Jaifroid:
I'm not 100% sure, but one issue might be that you need to double-check when you are using `enableSourceVerification`, and when you are using `enableSourceVerificationCheckBox`. It looks to me like you misnamed these in `index.html`, because you call the whole div the `CheckBox`, and the checkbox itself is just called `enableSourceVerification`. I don't know if there might be some mixup in the code?

Another possibility:
When you jump from the extension to ServiceWorker mode, you're jumping out of your PR to the PWA hosted at https://browser-extension.kiwix.org, in which this verificationCheckBox doesn't exist. It looks to me like your code is executing when the remote extension sends a message to the local extension to say "I loaded correctly". But the local extension doesn't load the whole app, it only loads a small section of code, enough for the code to store a success key.

Now, I'm not convinced by this second explanation either, because you don't appear to be setting the state of the checkbox in any way that is different from what other code does in `init.js`. Yours shouldn't be the only one causing an exception.

In any case, one solution might be to wrap the offending code in an `if` statement, so it only tries to apply the style if it can find the element (in this case, the source verification check box I presume).

Greeshmanth1909:
It seems like your second explanation might be correct. I made a copy of one of the other check boxes that was visible in `serviceWorker` mode previously and the duplicate didn't show in `serviceWorker` mode while the original did!

Greeshmanth1909:
@Jaifroid I think its ready for another review.

www/js/app.js
```diff
@@ -601,6 +624,21 @@ function focusPrefixOnHomeKey (event) {
         }, 0);
     }
 }
+// Verify the source for a loaded archive
+async function verifyLoadedArchive () {
+    if (selectedArchive.isReady() && !(settingsStore.getItem("trustedZimFiles").includes(selectedArchive.file.name)) && params.sourceVerification) {
+        const response = await uiUtil.systemAlert('Is the loaded ZIM archive from a trusted source?\n If not, you can still read the ZIM file in Safe Mode (aka JQuery mode). Closing this window also opens the file in Safe Mode. This option can be disabled in Expert Settings', 'Security alert!', true, 'Open in Safe Mode', 'Trust Source');
```
Jaifroid, line 630: Strings need to be put into a translation-ready format. Look at some other examples of `uiUtil.systemAlert()` to see how to do this. You need to do it for each of the strings here that will need to be translated, i.e. all four strings.

www/js/app.js
```diff
@@ -601,6 +624,21 @@ function focusPrefixOnHomeKey (event) {
         }, 0);
     }
 }
+// Verify the source for a loaded archive
+async function verifyLoadedArchive () {
+    if (selectedArchive.isReady() && !(settingsStore.getItem("trustedZimFiles").includes(selectedArchive.file.name)) && params.sourceVerification) {
```
Jaifroid, line 629: Change `"trustedZimFiles"` to single quotes as per house style.

www/js/app.js
```diff
@@ -1650,6 +1688,46 @@ function archiveReadyCallback (archive) {
...
+        }
+    }
+    if (settingsStore.getItem('trustedZimFiles') === null) {
+        settingsStore.setItem('trustedZimFiles', '', Infinity);
+    }
+    if (params.sourceVerification && (params.contentInjectionMode === 'serviceworker' || params.contentInjectionMode === 'serviceworkerlocal')){
+        // Check if source of the zim file can be trusted.
+        if (!(settingsStore.getItem('trustedZimFiles').includes(archive.file.name))) {
+            // Alert user about unknown source.
+            const response = await uiUtil.systemAlert('Is this ZIM archive from a trusted source?\n If not, you can still read the ZIM file in Safe Mode (aka JQuery mode). Closing this window also opens the file in Safe Mode. This option can be disabled in Expert Settings', 'Security alert!', true, 'Open in Safe Mode', 'Trust Source');
```
Jaifroid, line 1711: I'm not too keen on repeated code like this. This is the same message as above. Can't you just use your `verifyLoadedArchive()` function here again and eliminate repeated code It looks like it's the same? I don't just mean the message, but all the stuff here that is about verifying the file. Looks like you could get rid of most of this part?

www/js/app.js
```diff
@@ -1650,6 +1688,46 @@ function archiveReadyCallback (archive) {
...
+            console.log(response);
+            if (response) {
+                params.contentInjectionMode = 'serviceworker';
+                var trustedZimFiles = settingsStore.getItem('trustedZimFiles');
+                var updatedTrustedZimFiles = trustedZimFiles + archive.file.name + '|';
+                settingsStore.setItem('trustedZimFiles', updatedTrustedZimFiles, Infinity);
+                // Change radio buttons accourdingly
```
Jaifroid, line 1718: Typo: `accordingly`. But as it's spelled correctly above in the `verify` function, I think this is part of the code that could be replaced by a call to that function.

www/js/app.js
```diff
@@ -1650,6 +1688,46 @@ function archiveReadyCallback (archive) {
...
+    }
+    if (settingsStore.getItem('trustedZimFiles') === null) {
+        settingsStore.setItem('trustedZimFiles', '', Infinity);
+    }
+    if (params.sourceVerification && (params.contentInjectionMode === 'serviceworker' || params.contentInjectionMode === 'serviceworkerlocal')){
+        // Check if source of the zim file can be trusted.
+        if (!(settingsStore.getItem('trustedZimFiles').includes(archive.file.name))) {
+            // Alert user about unknown source.
+            const response = await uiUtil.systemAlert('Is this ZIM archive from a trusted source?\n If not, you can still read the ZIM file in Safe Mode (aka JQuery mode). Closing this window also opens the file in Safe Mode. This option can be disabled in Expert Settings', 'Security alert!', true, 'Open in Safe Mode', 'Trust Source');
+            console.log(response);
```
Jaifroid, line 1712: Remove console.log before merge.

www/js/init.js
```diff
@@ -124,7 +124,8 @@ params['altLibraryUrl'] = 'https://download.kiwix.org/zim/'; // Alternative Url
 params['cacheAPI'] = 'kiwix-js'; // Sets name of the prefix used to identify the cache in Cache API
 params['cacheIDB'] = 'kiwix-zim'; // Sets name of the Indexed DB database
 params['isFileSystemApiSupported'] = typeof window.showOpenFilePicker === 'function'; // Sets a boolean indicating whether the FileSystem API is supported
-params['isWebkitDirApiSupported'] = 'webkitdirectory' in document.createElement('input'); // Sets a boolean indicating whether the Webkit Directory API is supported
+params['isWebkitDirApiSupported'] = 'webkitdirectory' in document.createElement('input'); // Sets a Boolean indicating whether the Webkit Directory API is supported
+params['sourceVerification'] = params.contentInjectionMode === 'serviceworker' ? (getSetting('sourceVerification') === null ? true : getSetting('sourceVerification')) : false;// Sets a boolean indicating weather a user trusts the source of zim files
```
Jaifroid, line 128: Very minor, but add a space before the semicolon and the start of the comment.

www/js/app.js
```diff
@@ -496,6 +496,29 @@ document.getElementById('disableDragAndDropCheck').addEventListener('change', fu
...
+    verifyLoadedArchive();
+});
+document.getElementById('jqueryModeRadio').addEventListener('click', function () {
+    if (this.checked) {
+        document.getElementById('enableSourceVerificationCheckBox').style.display = 'none';
+    }
+});
+// Handle switching to serviceWorkerLocal mode for chrome-extension
+document.getElementById('serviceworkerLocalModeRadio').addEventListener('click', async function () {
+    document.getElementById('enableSourceVerificationCheckBox').style.display = '';
```
Jaifroid, line 511: My IDE tells me there's a trailing space at end of this line.

www/js/app.js
```diff
@@ -1650,6 +1688,46 @@ function archiveReadyCallback (archive) {
...
+                break;
+            case 'serviceworkerlocal':
+                document.getElementById('serviceworkerLocalModeRadio').checked = true;
+                break;
+        }
+    }
+    if (settingsStore.getItem('trustedZimFiles') === null) {
+        settingsStore.setItem('trustedZimFiles', '', Infinity);
+    }
+    if (params.sourceVerification && (params.contentInjectionMode === 'serviceworker' || params.contentInjectionMode === 'serviceworkerlocal')){
```
Jaifroid, line 1707: Add a space before opening bracket at end of this line, as per house style.

Jaifroid (commented):
Thank you very much for this PR, my tests show it's working really well now! There are some last minor cleanup things, some of them cosmetic. They should be very quick to do. I still need to test properly in extension mode, which I'll do tomorrow. And you need to add the translation strings to the file `en.jsonp.js` in the `i18n` folder. Then we'll need to do same for Spanish and French.

www/js/app.js
Greeshmanth1909: Apart from that, its pretty much the same. A few modifications have to be made to the function in order for it to work in both the cases. Will do it soon.

Greeshmanth1909:
I've translated the strings into French and Spanish using [deepl](https://www.deepl.com/translator).

Jaifroid (changes requested):
Many thanks for making a stab at the translations! I've added my suggested changes, and there are a few strings you forgot. In this review, I only focused on the translation strings, and didn't test the code. I'll do that for the final time when you've completed the UI.

UI is always pesky: lots of fiddly details that coders don't really see the point of, but users always obsess over... So, it's like the icing on the cake that makes the whole thing desirable form a UX perspective!

i18n/es.jsonp.js
```diff
@@ -184,6 +185,8 @@ document.localeJson = {
       "dialog-serviceworker-defaultmodechange-title": "Cambio del modo de inyección de contenido predeterminado",
       "dialog-serviceworker-defaultmodechange-message": "<p>Le hemos cambiado al modo ServiceWorker (ahora es el modo predeterminado). Soporta más tipos de archivo ZIM y es mucho más robusto.</p><p>Si tiene problemas con este modo, puede volver al modo JQuery (ahora obsoleto). En tal caso, por favor, informe de los problemas que ha experimentado (véase la página Información).</p>",
       "dialog-slideawaycheck-message": "Esta configuración puede no funcionar correctamente en Firefox OS. Si encuentra que algunos enlaces ZIM no responden, se recomienda desactivar esta configuración.",
+      "dialog-sourceverification-alert": "¿Es este archivo ZIM de una fuente de confianza?\n Si no es así, aún puede leer el archivo ZIM en Modo Seguro (también conocido como modo JQuery). Al cerrar esta ventana también se abre el archivo en Modo Seguro. Esta opción puede desactivarse en Configuración Experto",
```
Jaifroid, line 188: Just saw this: change "Configuración Experto" to "Opciones avanzadas" (it's what we use in the UI).

Greeshmanth1909:
I should get this done by tonight!

www/js/app.js
```diff
@@ -601,6 +628,29 @@ function focusPrefixOnHomeKey (event) {
         }, 0);
     }
 }
+/**
+ * Verifies the given archive and switches contentInjectionMode accourdingly
+ * @param {archive} the archive that needs verification
+ * */
+async function verifyLoadedArchive (archive) {
+    const response = await uiUtil.systemAlert(translateUI.t('dialog-sourceverification-alert'), translateUI.t('dialog-sourceverification-title'), true, 'Open in Safe Mode', 'Trust Source');
```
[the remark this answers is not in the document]
Jaifroid: Yes, same way, you just want the code to convert the string to the appropriate language and then send the converted string `||` the fallback string which should appear in the code (so devs know what it is!).

Greeshmanth1909:
There's a 110% chance I missed something here, lol.

i18n/es.jsonp.js
```diff
@@ -24,7 +24,8 @@ document.localeJson = {
       "configure-changed-content-note": "(Vuelva a seleccionar carpeta si hay contenido actualizado en su dispositivo)",
       "configure-btn-rescan": "Volver a escanear",
       "configure-about-rescan-tip": "Volver a escanear las tarjetas SD y la memoria interna",
-      "configure-expert-enable-source-verification-tip": "Advertencia: Algunos archivos ZIM de fuentes no fiables podrían ejecutar código malicioso en tu navegador. Esto puede evitarse utilizando el modo JQuery, que no puede ejecutar contenido activo desde el ZIM. Los ZIM muy dinámicos probablemente fallarán en el modo JQuery, pero los ZIM con contenido mayoritariamente estático deberían funcionar. Si confías en la fuente de todos tus ZIMs, desactivando esta opción utilizarás el modo ServiceWorker por defecto, si está disponible.",
+      "configure-expert-enable-source-verification-tip": "Si confía en la fuente de todos sus archivos ZIM, puede desactivar esta opción y se utlizará el modo ServiceWorker por defecto (si está disponible).",
```
Jaifroid, line 27: You deleted the first part of this tool tip, and only left the revised last sentence!

i18n/es.jsonp.js
```diff
@@ -24,7 +24,8 @@ document.localeJson = {
...
+      "configure-expert-enable-source-verification-check-box": "<strong>Activar la verificación del origen de los archivos nuevos</strong> (<i>recomendado</i>: Sólo se le pedirá la primera vez que abra una ZIM)",
```
Jaifroid, line 28: Change "una ZIM" to "un archivo ZIM)"

i18n/fr.jsonp.js
```diff
@@ -24,7 +24,8 @@ document.localeJson = {
...
+      "configure-expert-enable-source-verification-tip": "Avertissement : Certaines archives ZIM provenant de sources non fiables peuvent exécuter du code malveillant dans votre navigateur. Vous pouvez éviter cela en utilisant le mode JQuery, qui ne peut pas exécuter de contenu actif à partir du ZIM. Les ZIM très dynamiques échoueront probablement en mode JQuery, mais les ZIM dont le contenu est essentiellement statique devraient fonctionner. Si vous faites confiance à la source de tous vos ZIM, la désactivation de cette option permet d'utiliser le mode ServiceWorker par défaut, s'il est disponible.",
+      "configure-expert-enable-source-verification-check-box": "<strong>Activer la vérification de la source</strong> (<i>recommandé</i>: vous ne serez invité à le faire que la première fois que vous ouvrirez un ZIM)",
```
Jaifroid, line 28: Place a `&nbsp;` before the colon after `<i>recommandé</i>`.

i18n/fr.jsonp.js
```diff
@@ -185,8 +186,10 @@ document.localeJson = {
...
       "dialog-slideawaycheck-message": "Ce paramètre peut ne pas fonctionner correctement chez Firefox OS. Si vous constatez que certains liens ZIM deviennent inactifs, veuillez désactiver ce paramètre.",
-      "dialog-sourceverification-alert": "Cette archive ZIM provient-elle d'une source fiable ? Si ce n'est pas le cas, vous pouvez toujours lire le fichier ZIM en mode sans échec (également appelé mode JQuery). La fermeture de cette fenêtre ouvre également le fichier en mode sans échec. Cette option peut être désactivée dans Expert Settings",
+      "dialog-sourceverification-alert": "Cette archive ZIM provient-elle d'une source fiable&nbsp;? Si ce n'est pas le cas, vous pouvez toujours lire le fichier ZIM en mode sécurisé (également appelé mode JQuery). La fermeture de cette fenêtre ouvre également le fichier en mode sans échec. Cette option peut être désactivée dans Options avancées.",
```
Jaifroid, line 189: You forgot to change the second "mode sans échec" to "mode sécurisé".

i18n/fr.jsonp.js
```diff
@@ -185,8 +186,10 @@ document.localeJson = {
...
       "dialog-sourceverification-title": "Alerte sécurité!",
+      "dialog-sourceverification-trust-button": "Source de confiance",
+      "dialog-sourceverification-safe-mode-button": "Ouvrir en mode sans échec",
```
Jaifroid, line 192: Change `mode sans échec` to `mode sécurisé`.

i18n/fr.jsonp.js
```diff
@@ -184,6 +186,10 @@ document.localeJson = {
       "dialog-serviceworker-defaultmodechange-title": "Changement du mode d'injection de contenu par défaut",
       "dialog-serviceworker-defaultmodechange-message": "<p>Nous vous avons basculé en mode ServiceWorker (c'est maintenant le mode par défaut). Il prend en charge plus de types de fichiers ZIM et est beaucoup plus robuste.</p><p>Si vous rencontrez des problèmes avec ce mode, vous pouvez revenir au mode JQuery (maintenant obsolète). Dans ce cas, veuillez signaler les problèmes que vous avez rencontrés (voir la page Informations).</p>",
       "dialog-slideawaycheck-message": "Ce paramètre peut ne pas fonctionner correctement chez Firefox OS. Si vous constatez que certains liens ZIM deviennent inactifs, veuillez désactiver ce paramètre.",
+      "dialog-sourceverification-alert": "Cette archive ZIM provient-elle d'une source fiable&nbsp;? Si ce n'est pas le cas, vous pouvez toujours lire le fichier ZIM en mode sécurisé (également appelé mode JQuery). La fermeture de cette fenêtre ouvre également le fichier en mode sans échec. Cette option peut être désactivée dans Options avancées.",
+      "dialog-sourceverification-title": "Alerte sécurité!",
```
Jaifroid, line 190: Add `de` (separate word) between `Alerte` and `sécurité`.

Jaifroid:
Last few now! Hopefully we'll be done pending final testing.

Greeshmanth1909:
The same test failed last time too, is this something that we should be worried about?

Jaifroid:
> The same test failed last time too, is this something that we should be worried about?

This is just a flaky test, unfortunately. There's some race condition because sometimes it works and sometimes it fails, and it seems to depend on the overall load on the system...

i18n/es.jsonp.js
```diff
@@ -184,6 +186,10 @@ document.localeJson = {
       "dialog-serviceworker-defaultmodechange-title": "Cambio del modo de inyección de contenido predeterminado",
       "dialog-serviceworker-defaultmodechange-message": "<p>Le hemos cambiado al modo ServiceWorker (ahora es el modo predeterminado). Soporta más tipos de archivo ZIM y es mucho más robusto.</p><p>Si tiene problemas con este modo, puede volver al modo JQuery (ahora obsoleto). En tal caso, por favor, informe de los problemas que ha experimentado (véase la página Información).</p>",
       "dialog-slideawaycheck-message": "Esta configuración puede no funcionar correctamente en Firefox OS. Si encuentra que algunos enlaces ZIM no responden, se recomienda desactivar esta configuración.",
+      "dialog-sourceverification-alert": "¿Proviene este archivo ZIM de una fuente de confianza?\nDe no ser así, aún puede leer el archivo en Modo Seguro (también conocido como modo JQuery). Al cerrar esta ventana también se abre el archivo en Modo Seguro. Esta verificación puede desactivarse.",
```
Jaifroid, line 189: Very last change: you missed "en Opciones avanzadas" before the full stop right at the end (after `desactivarse`).

Jaifroid (approved):
Congratulations on this PR! It all looks good to me now, apart from one tiny little correction below. I'll do a bit more testing of the switch from a local extension to the remote code and vice versa, but in the meantime, after you've done the one little correction, let me know if you're happy for the PR to be merged.

Jaifroid:
Just to confirm that the flow seems to work well from JQuery mode in an extension through to SW mode. The user will likely be asked twice if they trust the source because the list of secure archives in the local code is not the same list as is maintained in remote code -- but this is inevitable, and will only affect a user who is switching to SW mode for the first time and has an archive loaded (a rare situation, since most users capable of running the app in SW mode will have switched when they first loaded the app). So, all OK! Thanks once again for this PR which will increase people's awareness of security. I'll merge when the Linux test is complete.

PR #1193, opened 2024-01-17
Update CONTRIBUTING.md

skushagra9:
Already the Bypass App Cache in Expert Settings is turned on


Jaifroid (changes requested):
I don't think you can simply remove any mention of the Bypass AppCache option. It's important to make developers aware of its importance, even though I added code that would turn it on automatically in some contexts. However, it isn't turned on automatically if developers are using, e.g., http-server instead of vite. So I think you need to modify the text rather than remove it.

skushagra9:
Ok, thanks for mentioning

Jaifroid:
Just to say that an edit would be welcomed, it's not necessary for you to abandon this PR just because I suggested editing the text!

Issue #1199, opened 2024-01-22
Add support for Zimit v2.0 archive types


Jaifroid:
There are a few parts to this:

- [x] #1200
- [x] #1183
- [x] #1206 

Further info on the second item: clicking internal links currently works for the single archive tested, but the existing implementation is faulty with regard to Zimit classic due to the many edge cases, and we need to add the same support as for Kiwix PWA, which correctly detects external links and PDFs so they may be opened outside the sandbox. I fully expect this to be broken for Zimit 2 archives and will need handling like in libkiwix: https://github.com/kiwix/libkiwix/pull/1036, or through a similar method.

For the first item, we just need to backport https://github.com/kiwix/kiwix-js-pwa/pull/537.

Jaifroid:
While it may not be 100% complete, we only have five or so test ZIMs, and these are all working. Further issues can be added as they occur.

PR #1201, opened 2024-01-23
Open external links and PDFs outsde zimit sandbox

### LINKED ISSUE #1183: In Zimit archives when detecting PDFs for load in new tab, do a Header lookup before deciding a resource is external
Currently, we have to intervene manually to prevent PDFs from launching in the sandboxed iframe, which does not work in most browsers due to CSP restrictions. However, in some archives, e.g. `harmony-project.org_en_all_2023-09.zim`, the PDFs are stored in the ZIM, but in a different domain, and are redirected via a header. So, we need to check the header.

The reason we can't currently let the Replay Worker handle this process is that we have to prevent the link from launching before it reaches the Service Worker, or else the CSP kicks in.

There may be a better way to do this, but for now, there is code in Kiwix PWA that already does this, and it's a simple process. It needs to be added to `handleClickOnReplayLink()`.

Jaifroid:
Fixes #1183 and second bullet point of #1199.

### DIFF
--- a/www/js/app.js
+++ b/www/js/app.js
@@ -2092,6 +2092,7 @@ function handleClickOnReplayLink (ev, anchor) {
         ev.preventDefault();
         ev.stopPropagation();
         // Note that true in the fourth argument instructs getDirEntryByPath to follow redirects by looking up the Header
+        // DEV: CURRENTLY NON-FUNCTION IN KIWIX-JS -- NEEDS FIXING
         return selectedArchive.getDirEntryByPath(zimUrl, null, null, true).then(function (dirEntry) {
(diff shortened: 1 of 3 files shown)


Jaifroid:
External links and PDFs are confirmed to be opening outside the sandbox (tested with a zimit2 version of military medicine). Also tested general support with solar.lowtechmagazine and mesquartierschinois.

Jaifroid:
However, loading PDFs from classic Zimit files is not fixed, indeed is broken by this PR. Grrr.

Jaifroid:
All seems to be working now.

Issue #1214, opened 2024-02-22
Remove unnecessary space below search bar when screen is narrow


Jaifroid:
Although this is not noticeable on desktop, on narrow form factors, the excessive amount of space below the search bar is extremely noticeable. While the ultimate solution to this is #523, in the meantime, it would be a quick fix to remove the space now. See screenshot.

![image](https://github.com/kiwix/kiwix-js/assets/4304337/35c112b1-e84b-42f6-935d-32b078fbc8cf)

Sky-De:
@Jaifroid i would like to fix this issue, please assign this to me

Jaifroid:
@Sky-De Please take a look at https://github.com/kiwix/kiwix-js/blob/main/CONTRIBUTING.md (read it all the way through), set up your development environment, and make sure you're happy with the process here. If all is well, please come back here outlining your suggestion of how to complete this issue so that I can assign you.

Sky-De:
@Jaifroid i willdo that tomorrow and will infotm you, thanks for quick response

Aditya0049:
Before
<img width="709" alt="Screenshot 2024-02-26 004445" src="https://github.com/kiwix/kiwix-js/assets/88885329/236fbc93-cb7e-4f04-9d8c-e73df5123f19">

After
<img width="612" alt="Screenshot 2024-02-26 004502" src="https://github.com/kiwix/kiwix-js/assets/88885329/fb0c3af8-a560-4770-8fac-46d78a47cfff">

chec here #1216

amr-salahuddin:
Have this issue been solved already?

Jaifroid:
> Have this issue been solved already?

@amr-salahuddin Yes, @Aditya0049 already has a PR:

CREDO23:
I want to work on this @Jaifroid !

Jaifroid:
OK. Please check out the previous PR and what was wrong with it. Be sure to follow instructions in CONTRIBUTING.md so you know how to set up your development environment and test properly.

z1shivam:
This issue has been sitting for a while, so I thought maybe I could help finally fix this. I read the convo at #1216 and the follow up there makes simple sense. Extending on that, I noticed a couple of other issues with the `formArticleSearch` container.

**For one**, the "random article" button has its font-size reduced at smaller widths (768px) which in-turn reduces its height, making it look awkward next to the other elements.
[...]

PR #1216, opened 2024-02-25
reduced margin between search button and result text

Aditya0049:
solved this issue #1214
Remove unnecessary space below search bar when screen is narrow

### DIFF
--- a/www/css/app.css
+++ b/www/css/app.css
@@ -346,6 +346,12 @@ button {
         padding-top: 0.5rem;
         padding-bottom: 0.5rem;
     }
+
+    /* Reduce margin between search button and result text */
+    #formArticleSearch > .container{
+        margin-bottom: 0rem;
+    }
+
     /* Reduce size of the top buttons */
     .btn-outline-secondary {
         font-size: 10px !important;


Jaifroid:
@Aditya0049 Thanks for the PR. I'm just running automated tests, then I'll test it myself and get back to you.

Jaifroid:
@Aditya0049 It's looking good, but I think it needs two things:

1. Make the space above the bar equal to the space below it (see screenshot below, where I drew squiggly yellow line);
2. Reduce the size of the search bar and the search button when the screen is narrow (they should be at least the size of the Random button).

You'll see in the CSS some code that reduces the size of the bottom toolbar buttons. We need something similar for the search bar area. You can probably just copy (roughly) the CSS rules. You should ensure that Random button, Search button and search text field are the same height so they match.

I realize there are a few things here that weren't in the original issue, but seeing it after your changes makes the need for them more obvious!

image

ehtishama:
Hi @Aditya0049, I noticed there hasn't been any activity in a while. Are you still working on this issue? If not, I'd like to take on this issue.

Aditya0049:
> Hi @Aditya0049, I noticed there hasn't been any activity in a while. Are you still working on this issue? If not, I'd like to take on this issue.

I'm working ....it'll be done till next 2 days

Jaifroid (changes requested):
See changes mentioned here: https://github.com/kiwix/kiwix-js/pull/1216#issuecomment-1964274103

Jaifroid:
Please feel free to re-open if you want to complete this.

PR #1219, opened 2024-03-01
Fix errors enumerating ZIM archives in directory

### LINKED ISSUE #1220: Use of FSA API and webkitdirectory stores unnecessary keys in localStorage and webkitdirectory fails to clear when new directory is loaded
> Additionally, when a user picks a new directory using webkitdirectory API, the old contents are not cleared from the list, and the new contents are, confusingly, added to the dropdown list of archives.

_Originally posted by @Jaifroid in https://github.com/kiwix/kiwix-js/issues/1156#issuecomment-1837762841_

Furthermore, subdirectory names and all their contents, regardless of whether they are ZIM archives or not, get stored in the kiwixjs-zimFilenames key in localStorage.

Jaifroid:
There are a few issues with the code that enumerates archives in a directory:

* The first is that the code stores ALL files in the directory in the zimFilenames key. This PR adds a filter so only files matching the pattern `.zim\w?\w?$` (case insensitive) will be stored;
* The second is that the code was unnecessarily calling await functions several times to get values that could simply be stored in a variable;
* The third is that to test for a previously selected archive, a regex was being made from filenames without filtering for regex meta-charecters, which could easily lead to errors. In fact that's not necessary, as we can simply use `indexOf` without needing to do a regex test on the whole filename;
* The fourth is a longstanding bug where permissions appeared not be being retrieved correctly on the first go. The cleanup in this PR **may** have fixed that, but it needs monitoring and testing;
* Fifth is that when loading a new `webkitdirectory`, the old contents are not cleared

Fixes #1220.

### DIFF
--- a/www/js/app.js
+++ b/www/js/app.js
@@ -1528,7 +1528,7 @@ function displayFileSelect () {
     }
     if (params.isWebkitDirApiSupported) {
         // Handles Folder selection when webkitdirectory is supported but showDirectoryPicker is not
-        folderSelect.addEventListener('change', async function (e) {
+        folderSelect.addEventListener('change', function (e) {
             e.preventDefault();
             const filenames = [];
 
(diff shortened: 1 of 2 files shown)


Jaifroid:
Tested on Chrome 49, and doesn't seem to be affected by webkitRelativePath bug in Chrome < 72, at least if there are ZIM archives in the directory selected.

Jaifroid:
Tested and working on Firefox and Edge Legacy.

Issue #1217, opened 2024-02-29
Update info about file and directory picking to include File System Access persistent permissions


Jaifroid:
Persistent permissions have landed in Chromium desktop browsers > 122 for sites with a secure origin. See https://developer.chrome.com/blog/persistent-permissions-for-the-file-system-access-api. This means that the Chromium browser extension (Chrome, Edge) in ServiceWorker mode can now give persistent access to files and directories if the user answers the one-time prompt that appears the second time a user attempts to access a previously picked directory or file with "Allow on every visit" (see screenshot). This will probably only work with full ServiceWorker mode, i.e., when the extension is acting as a PWA from the secure browser-extension.kiwix.org origin.

So, we need to update info in README and in-app (in all languages), to signal this enhancement. We currently say that it's not technically possible to avoid permission prompts in browser context, but this is not universally true any more.

![Persistent Pwrmissions](https://github.com/kiwix/kiwix-js/assets/4304337/e2484bd2-6d0e-4f42-a57f-452b1bb8d8aa)

PR #1221, opened 2024-03-02
Reopen last selected archive with File System Access API

### LINKED ISSUE #1218: Re-open previous archive if app has permission via persistent permissions, like with Firefox OS
Up until now the only Kiwix JS app (other than the Kiwix PWA variant) able to re-open a previously picked archive on re-launch of the app has been the Firefox OS version.

This autolaunch behaviour can be leveraged now for Chromium desktop browsers as well. See #1217. Such code already exists in the PWA. It should be sufficient to attempt access in most contexts, as the access will only be granted by the browser if the user has answered the persistent permissions prompt to allow access every time.

Jaifroid:
In browsers that support the File System Access API (only Chromium browsers currently), it is now possible to re-open the last selected archive automatically if the user has given permanent access permissions.

This is a great step towards giving the app native-level UX.

However, there are situations where this could be annoying, so we need to provide a UI option to disable auto-loading.

Nota bene: this does not re-open the last read article like in Kiwix PWA, as that would require more code and more UI to avoid privacy concerns. It might be added in the future.

Fixes #1218.

### DIFF
--- a/README.md
+++ b/README.md
@@ -124,12 +124,18 @@ best to display static content, but much functionality is likely to be broken. H
 You can switch between these content injection modes in Configuration, but if your browser supports ServiceWorker mode as an offline-first PWA, you are
 strongly advised to remain in this mode.
 
-### Limitations
+### File access and other limitations
 
-It is not yet technically possible automatically re-open a selected ZIM file between sessions. However, browsers that support the File System Access API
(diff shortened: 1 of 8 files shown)


Jaifroid:
Also fixes #1217.

PR #1225, opened 2024-03-06
Remove remaining JQuery

Jaifroid:
It may be time to remove the last remnants of JQuery if possible.


Jaifroid:
Unfortunately, we cannot yet remove jQuery completely, because the bootstrap version used in this Repo still relies on it. We would need to upgrade, but the upgrade may not be compatible with IE11. It would need testing whether that can be overcome with Babel or not.

Jaifroid:
Fixes #367.

PR #1227, opened 2024-03-08
update regex to match PhET

### LINKED ISSUE #1157: Active content warning not detecting PhET ZIMs
At least in Firefox, the user is just presented with a blank page if attempting to access a PhET ZIM in jQuery mode. This won't affect most users, as those with the ability to run a Service Worker should already be in SW mode, but for completeness, this regression should be fixed.

Greeshmanth1909:
resolve #1157
Updated the regex to also match `ractive.js`. I was able to view the active content warning when PhET zims were opened in Safe mode.


Jaifroid (approved):
Excellent, thank you very much! I'll merge after doing a few small tests myself!

PR #1228, opened 2024-03-13
Fix duplicate credits in about section

dan-niles:
The credits in the About section are duplicated as shown below:
image

I fixed it by seperating the bullet points into seperate entries in the i18n translation files.


Jaifroid:
Thank you! I'll take a look.

Jaifroid:
I confirm this is all fine! I'll merge now.

Jaifroid:
Many thanks for the PR!

www/index.html
```diff
@@ -385,9 +385,9 @@ <h3 id="feedback" data-i18n="about-feedback">Feedback / helping / contributing</
                     <h3 id="credits" data-i18n="about-credits">Credits</h3>
                     <p data-i18n="about-credits-gratitude">We owe a big debt of gratitude to:
```
Jaifroid, line 386: Although your code works, I think it only works because the browser auto-adds a close `</p>` at the end of this line. So in effect, I believe it works due to a side effect! We probably shouldn't rely on that.

All you have to do is add a `</p>` at the end of this line, and remove the `z/p>` at the bottom of the block (see below).

www/index.html
```diff
@@ -385,9 +385,9 @@ <h3 id="feedback" data-i18n="about-feedback">Feedback / helping / contributing</
...
                         <ul>
-                            <li>Kelson and all the Kiwix community for welcoming us to this project;</li>
-                            <li>Wikipedia teams and contributors for helping to make knowledge available to everybody;</li>
-                            <li>All the developers of libraries and tools we used, for releasing them as Free software (see libraries and tools paragraph).</li>
+                            <li data-i18n="about-credits-gratitude-part1">Kelson and all the Kiwix community for welcoming us to this project;</li>
+                            <li data-i18n="about-credits-gratitude-part2">Wikipedia teams and contributors for helping to make knowledge available to everybody;</li>
+                            <li data-i18n="about-credits-gratitude-part3">All the developers of libraries and tools we used, for releasing them as Free software (see libraries and tools paragraph).</li>
                         </ul>
                     </p>
```
Jaifroid, line 392: Remove this `</p>`. You'll also need to move the indentation of the whole `<ul>...`</ul>` block back by four spaces.

Jaifroid (changes requested):
Hi, I spoke a bit too soon about merging, and have noticed that you're relying on a side-effect. But the good news is it will take about 1 minute for you to fix!

www/index.html
dan-niles: Ahhh I see. I'll fix it.

Jaifroid (approved):
Great, all fixed now! Will merge.

PR #1230, opened 2024-03-14
Workaround for over-encoded URLs in zimit2

### LINKED ISSUE #1229: Work around the workaround: deal with querystrings in zimit2 ZIMs where the URL separators have been encoded
Pending the outcome of https://github.com/openzim/libzim/issues/865, we may need to introduce a workaround in Kiwix JS readers.

See original issue for full detail, but the summary is that, on HTML pages that have been encoded by warc2zim2, querystrings are incorrectly presented with encoded separators. For example, `index.html?thematique=internet` might be presented as `index.html%3Fthematique%3Dinternet`. This currently causes KJS to fail, as it (correctly) does `decodeURI()` when converting the HTML representation to a ZIM URL, not `decodeURIComponent()`.

If we decode the URI separators as well, then we could be left with anomalous situations like `what_is_a_cookie?.html?title=What is a cookie?&theme=dark`. So we will have to take care in such situations with how we handle the querystring.

Jaifroid:
The first commit is a workaround for over-encoded URLs in zimit2.

However, we could also usefully check to see that we are handling clicks on URLs in Wikipedia articles where an encoded question mark that is not a querystring needs to be decoded. For example, an HTML URL like `What_About_Me%3F_(Kenny_Rogers_album)` needs to be decoded to `A/What_About_Me?_(Kenny_Rogers_album)` in order to find it in the ZIM and the material after the `?` must not be removed.

Should fix #1229. Should only be merged pending the outcome of https://github.com/openzim/libzim/issues/865.


Jaifroid:
Since further test ZIMs are not pending, we should merge and deal with any subsequent issues in a separate PR.

PR #1237, opened 2024-03-29
Fix bug for empty search box on initial ZIM load

### LINKED ISSUE #1083: Clicking the search button when the search box is empty sometimes fails to show all titles
I believe this is due to a protection feature in the search, which prevents lookup if the same string is entered twice in a row (instead, it should just display the previous search). However, there is no previous search if the field is blank and the user has just loaded a ZIM.

I believe the solution is to exempt comparison of current search and prefix if the search box is empty / blank / null, and/or checking that we have search results to display before invoking the protection.

dan-niles:
resolves #1083

I updated the protection check to exempt comparison of current search and prefix if the search box is empty.


dan-niles:
I made a mistake by removing the `!` from the protection condtion that checks the status of the search. I was doing some testing  and accidently committed it without changing it. Added it back in [5d74859](https://github.com/kiwix/kiwix-js/pull/1237/commits/5d7485923407b2e32f00c9082bf57f6af2cdc988).

Jaifroid:
> I made a mistake by removing the `!` from the protection condtion that checks the status of the search. I was doing some testing and accidently committed it without changing it. Added it back in [5d74859](https://github.com/kiwix/kiwix-js/pull/1237/commits/5d7485923407b2e32f00c9082bf57f6af2cdc988).

No problem. I'm just re-running tests.

Jaifroid (approved):
As far as I can tell, this should fix the stated issue. Thank you! Let me know when you're Ok for me to squash-merge.

dan-niles:
Sure, you can go ahead and merge. 👍

PR #1238, opened 2024-03-29
Change method for patching gitignore

Jaifroid:
This switches from the unwieldly git patch method to a simple Bash script which can be updated much more easily when new dependencies are added.


Jaifroid:
All tested and working.

PR #1241, opened 2024-04-05
add ToC as in kiwix pwa

### LINKED ISSUE #1212: Need Sidebar with sub heading menu
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

Kiwix extension chrome rendering zim article without subtopic menu and hyperlink for the same

Greeshmanth1909:
resolves #1212
Most of the code was taken from the actual pwa implementation. Variations from the original implementation include:
- The `setupTableOfContents` function would be called once the `ToC` drop down is clicked. The drop down only appears if an article is loaded in the iframe hence, the table of contents would be constructed once the user actually clicks the drop down.
- `params.relativeFontSize` was set to 100
 
This is how it looks:
Screenshot 2024-04-05 at 7 11 58 PM


Jaifroid:
Thank you very much. I was away, so didn't have a chance to look at it till now. I'll take a look asap.

Jaifroid:
I've had a quick look (I haven't tested the code so far). It's functional, which is great! I noticed a few cosmetic issues. The first is that the ToC button isn't vertically aligned in the same way as the other buttons on the bottom navbar. The second is that the ToC list opens attached to the bottom right of the screen instead of being attached to the button itself like in the PWA. And finally, there is no margin / padding around the text in the ToC list, which makes it difficult to distinguish from the article text. See screenshot. Could you kindly take a look at these issues?

image

Greeshmanth1909:
Yes, will get this done by the end of this day.

Jaifroid:
Hmm, I'm not sure. Maybe we should take the cue from Wikipedia here, where the ToC appears on the left (at least for left-to-right languages) as in screenshot. But then the question is whether to replace the Home button, or move it. In some ways it's redundant, as we have a Home link in the top bar. But maybe we can try out putting ToC far-left, then Home, then the rest.

I also wonder if it would be better to use the `rectangle-list` font-awesome icon instead of "ToC", so that we don't have to deal with translations. You can see it here: https://fontawesome.com/icons/rectangle-list?s=solid

image

Greeshmanth1909:
I've added the icon and moved the list to the left, the coloring is a little off though.

Screenshot 2024-04-09 at 4 28 31 PM

Greeshmanth1909:
All my attempts at manually matching the color of the icon to the rest were futile, obviously. Is there a workaround?

www/index.html
```diff
@@ -789,7 +789,14 @@ <h3 data-i18n="configure-expert-settings-title">Expert settings</h3>
                         <span id="alertMessage"></span>
                     </div>
                 </div>
-                <div id="navigationButtons" class="btn-group btn-block">
+                <div id="navigationButtons" class="btn-group btn-block" style="padding-left: 5%;">
+                    <div class="dropup" >
+                        <a href="#" class="btn btn-default dropdown-toggle col-xs-4" role="button" id="dropup" data-toggle="dropdown" aria-haspopup="true" aria-expanded="True" style="font-size: 14px; padding-top: 12px;">
```
Jaifroid, line 794: I think it's because you have copied code in the HTML from Kiwix PWA, whereas this Repo is using s different bootstrap. Look carefully at how the other buttons are wrapped, and copy it. For a start, the `<a ... >` container for the button here should have a class of `btn btn-lg`, rather than btn-default. Also, try to use the class-based shorthand for the button, don't copy and past the SVG code inline.

www/index.html
```diff
@@ -789,7 +789,14 @@ <h3 data-i18n="configure-expert-settings-title">Expert settings</h3>
...
+                            <svg xmlns="http://www.w3.org/2000/svg" height="20" width="22.5" viewBox="0 0 576 512"><path fill="#74C0FC" d="M0 96C0 60.7 28.7 32 64 32H512c35.3 0 64 28.7 64 64V416c0 35.3-28.7 64-64 64H64c-35.3 0-64-28.7-64-64V96zM128 288a32 32 0 1 0 0-64 32 32 0 1 0 0 64zm32-128a32 32 0 1 0 -64 0 32 32 0 1 0 64 0zM128 384a32 32 0 1 0 0-64 32 32 0 1 0 0 64zm96-248c-13.3 0-24 10.7-24 24s10.7 24 24 24H448c13.3 0 24-10.7 24-24s-10.7-24-24-24H224zm0 96c-13.3 0-24 10.7-24 24s10.7 24 24 24H448c13.3 0 24-10.7 24-24s-10.7-24-24-24H224zm0 96c-13.3 0-24 10.7-24 24s10.7 24 24 24H448c13.3 0 24-10.7 24-24s-10.7-24-24-24H224z"/></svg>
+                            <span class="caret"></span>
+                        </a>
+                        <ul id="ToCList" class="dropdown-menu flex-container" style="overflow-y:auto; padding-top: 10px; padding-bottom: 10px; padding-right: 10px; padding-left: 10px; position: absolute; left: 10px;"></ul>
+                    </div>
                     <a href="#" data-i18n-tip="home" class="btn btn-lg" id="btnHomeBottom" title="Home"><i class="fas fa-home"></i></a>
```
Jaifroid, line 800: Look at how it's done here. Use the `<i class="fas fa-rectangle-list">` (or similar) code.

Jaifroid: And don't include an `<svg ...>` block. That gets expanded automatically AFAIK.

Greeshmanth1909:
The provided link led to the latest version (6.x.x) it seems like we're using version 5.15.4. As I was initially trying to use the class based representation from the latest version, it didn't work. This new icon is from one of the older versions.

Jaifroid:
That looks better, thanks! Sorry about having misled you. We are indeed currently using an older Bootstrap, but there is a PR in progress to upgrade it. I think your PR is likely to be ready first. Nevertheless, at last for now, this icon works well. I'll make some further comments in the code.

Greeshmanth1909:
The clean up messed up the functionality. I'll fix it now.

www/js/app.js
```diff
@@ -2956,6 +2956,56 @@ function pushBrowserHistoryState (title, titleSearch) {
     }
     window.history.pushState(stateObj, stateLabel, urlParameters);
 }
+var dropup = document.getElementById('dropup');
+dropup.addEventListener('click', function () {
+    var ToCList = document.getElementById('ToCList');
+    ToCList.style.display = ToCList.style.display === 'block' ? 'none' : 'block';
```
Jaifroid, line 2962: I think we're unnecessarily building the ToC every time the user presses the button, even if the click is supposed to hide the ToC. So, I suggest something like this:

```
    if (ToCList.style.display !== 'none') {
        ToCList.style.display = 'none';
    } else {
        ToCList.style.display = 'block';
        ... rest of code ...
    }
```

Greeshmanth1909: Yeah, makes sense. I'll get it done tomorrow and add some comments.

Jaifroid:
I haven't done a proper code review yet, as I realized we need to add some code to dismiss/hide the ToC if the user clicks outside of it, or it otherwise loses focus. I suggest you experiment with adding a 'blur' event listener to the dropup, and simply hide the ToC in that event:
```
dropup.addEventListener('blur', function () {
    ...
});
```
NB, this needs testing, because it's possible that focus is lost when the document is scrolled to the selected heading, though I don't think that should be the case.

I think this should take care of the other issue, which is that currently the old ToC remains in place if a user navigates to a new article, and it doesn't get rebuilt when the new article is loaded. However, if the blur function works, then navigating to a new article will hide the ToC anyway, and it will get rebuilt when the user clicks on the button.

Jaifroid:
I also noticed this issue when testing the display in a mobile format in dev tools (the ToC gets squashed up against the left margin). You may need to add a margin instead of padding, or else use !important in the style rule to ensure the padding is respected even when the display goes into responsive mode.

Ultimately, the style rules will need to be moved to `app.css`, but for now you can keep them inline while you experiment and iron out bugs. But you should be aware that there is a section in `app.css` which alters some styles when the width is less than a certain number of pixels. However, I think it's Bootstrap that's displaying things differently on narrow screens rather than our CSS, so an !important might do the job to force the padding to display.

Greeshmanth1909:
I won't be able to get to this until Sunday as I have exams.

Jaifroid:
Of course, no rush. Good luck!

Greeshmanth1909:
@Jaifroid What function gets called when a user clicks on a link that takes them to another article?

Jaifroid:
> @Jaifroid What function gets called when a user clicks on a link that takes them to another article?

In Safe Mode (called jQuery mode in the code), listeners are attached to each link in an article. In SW mode, there is no attached function: the Service Worker catches the browser's Fetch request and diverts it to fetch an article from the ZIM.

Are you looking to tear down the old ToC on new article load? I think that would be a bit complicated, and unnecessary. Simply rebuild it when a new article is loaded. All you have to do is ensure the current ToC gets hidden if it loses focus, for which I think the blur event is easiest.

Greeshmanth1909:
The list disappears as expected but the auto scroll stopped working. Is this because of conflicting eventListeners for `blur` and `click` for the list elements?

Jaifroid:
> The list disappears as expected but the auto scroll stopped working. Is this because of conflicting eventListeners for `blur` and `click` for the list elements?

Without testing it, which I can't do just now, it's hard to speculate, but I do notice that you seem to have added the blur event to the button itself, which may not be ideal. Try adding it to the ToCList, as that's what needs to blur, not the button.

Greeshmanth1909:
I added a timeout within the exiting event listener and everything is working as expected.

Jaifroid:
Hi @Greeshmanth1909 With your latest code, I'm seeing a couple of problems: 1. I have to press the ToC button twice before it responds; 2. Clicking on one of the links fails to scroll the article. Maybe the timeout has introduced a race condition? It would probably be better to solve it without a timeout if at all possible, because timeouts tend to work differently on different devices or under different conditions.

Jaifroid:
Yes, that is my experience too. I forgot to mention that.

Jaifroid:
@Greeshmanth1909 No rush, but I just wanted to check that you're not awaiting anything from me on this PR.

Greeshmanth1909:
@Jaifroid It looks like the list element is never focused. The focus shifts between the button, the main article and the iframe. Keeping this in mind, there are three instances I can think of in which the list element has to be hidden:
1. User clicks on the heading they want to checkout, this line executes after the heading is scrolled into view and the element disappears as expected.
`ToCList.style.display = 'none';`
2. When the user clicks away from the displayed list.
3. When the user clicks on another link that takes them to a different page.

In both the above mentioned scenarios, the button looses focus. As you've mentioned earlier, the blur event listener can be used to hide the list once the button looses focus. This caused a problem with the automatic scrolling as the button also looses focus when one of the toc contents is clicked.

The timeout approach solved both the problems. I think the present inconsistency regarding the automatic scroll can be addressed by increasing the timeout. 
The other ways I found online for this involved the use of JQuery. The timeout approach seemed simpler.

Jaifroid:
@Greeshmanth1909 You're right that a timeout is much preferable to reintroducing jQuery (which it took a lot of work to eliminate!). Please do try increasing the timeout, or else check what the solution used in Kiwix PWA is (I'm sorry, I can't remember without looking at it in detail).

Jaifroid (changes requested):
Changes are discussed in Conversation thread - some functionality still a bit erratic.

Greeshmanth1909:
Will get this done by this weekend.

Greeshmanth1909:
I've increased the wait time to hide the list element when it looses focus and decreased the wait time for scrolling. It seems to be working fine.

Jaifroid:
@Greeshmanth1909 Thanks, but I'm still seeing the issues in screen capture below. To summarize: 1. I still have to click twice on the ToC icon; 2. Selecting a heading doesn't actually jump to the heading.

https://github.com/kiwix/kiwix-js/assets/4304337/78383479-7972-42de-acc4-f9cf0ab1c23e

To be clear, I double-checked that I pulled all your changes. Possibly when you merged main into your branch, it might have reintroduced issues?

If it works for you, and not for me, it's likely your timeouts introduced a race condition. If I remember rightly, in the PWA, I set up the ToC every time a page is loaded, so that it is already available when the user clicks the ToC button and doesn't have to be composed. Of course it should be possible to compose it on click time instead, but your current code appears to have a race condition, which is likely what is causing the need to click twice. Adding timeouts is not a good solution to such race conditions, because you can never be sure to allow enough time for very slow devices, and you don't want to slow it down for people with fast devices. An alternative would be to use a Promise or Async code to get the headings. It would be much safer than a timeout.

So, I think you have two options: 1. Compose the ToC at page load time (i.e., once the HTML has been injected into the iframe DOM); 2. Compose the ToC with async code when the user hits the button, if your attempts to use synchronous code are not working.

Of course there could be some other issue why the ToC doesn't display on the first click. Please debug and test carefully.

Jaifroid:
I'm closing this as there hasn't been any progress for several months. However, please feel free to reopen if you want to complete it.

Issue #1037, opened 2023-07-31
Drag of images causes to disconnect from the loaded zim file


FractalU:
In the kiwix browser extension 3.9.0 of both firefox and chrome, whenever I drag an image by mouse the loaded zim file gets disconnected and I get sent back to the configuration page. This also happens with the progressive web app. I tested the progressive web app on https://pwa.kiwix.org/www/index.html. What I expect is that the dragging of images does nothing as it's the default behavior of any web page. The kiwix browser extension and the progressive web app are both based on kiwix-js. I guess there is a bug in kiwix-js. That's why I'm writing here.

FractalU:
@Jaifroid Thank you for this option. It fixed the issue on the kiwix browser extension on both firefox and chrome as well as on the progressive web app on those browsers. I guess the drag and drop mechanism on the configuration page is still active after the zim file has been loaded and the pages from that zim file are being displayed. When I accidentally drag an image while the pages from the zim file are being displayed, the drag mechanism comes in and thinks there might be another zim file to load. It sends me back to the configuration page and shows the dotted red lines to drop the dragged item. My suggestion is to have the drag and drop mechanism active only on the configuration page. That would solve the issue. There would also be no need to have the disable drag-and-drop option anymore.

Jaifroid:
@FractalU I'm keen to make opening ZIM files as frictionless as possible, hence the idea that the whole app is a target for drag-and-drop, because it means that a user doesn't need to click any buttons to open a new ZIM file. They can simply drag a zim anywhere into the app, even if another ZIM is open.

Having said that, it would be perfectly possible to make it an option so that only the Config page receives drag-and-drop if a ZIM is already open. The control could be a dropdown with the options:

* Default (drag-and-drop always active)
* Config only (only the Config page can receive drag-and-drop)
* Disable

I think that would probably address the issue fully.

FractalU:
[...]

PR #1245, opened 2024-04-29
Non-file Dragging Bug Fix

D3V-D:
For issue #1037 

Tests done:
`npm run serve`
`npm run preview`
`npm run build`, followed by` npm run serve` and using the localhost:xxxx/dist/? page. (should be same as running the extension)

Tested on both firefox and chromium.
Tested with both mdn web docs (which uses workers) and a smaller page that doesn't use workers. Tested in both jQuery and workers mode. Passed `npm test`. 

Ran `e2e-test` for firefox, just one error as seen below:
```
1) Gutenberg_ro test [ZSTD compression] [SW mode]
       Sorting books by name:

      AssertionError [ERR_ASSERTION]: 'Poezii' == 'Creierul, O Enigma Descifrata'
      + expected - actual

      -Poezii
      +Creierul, O Enigma Descifrata
```
On e2e-test for chromium, also just one error, as seen below:
```
1) Gutenberg_ro test [ZSTD compression] [SW mode]
       Change Language:
     JavascriptError: javascript error: {"status":11,"value":"Element is not currently visible and may not be manipulated"}
  (Session info: chrome=124.0.6367.91)
```

I suspect these are unrelated to my changes, but please let me know if I broke something.

Other than that, I do have one question; is there a way to know if the ZIM page has already been loaded? With my current solution, it takes them unnecessarily to home if they dragleave the page, even if no ZIM file has been loaded. Also, if there's a way to keep track of their last visited page, that could greatly improve the experience when having dragged a file over but not having dropped it on the page (right now, it will take you back to the home page of the ZIM). @Jaifroid

### DIFF
--- a/www/css/app.css
+++ b/www/css/app.css
@@ -94,10 +94,6 @@
     padding-top: 1rem;
 }
 
-/* #search-article * {
-    pointer-events: none;
-} */
-
 #configuration {
     margin: 0 1rem;
 }
--- a/www/css/app.css
+++ b/www/css/app.css
@@ -65,7 +65,7 @@
 }
 
 .dragging-over * {
-    pointer-events: none !important;
+    pointer-events: none;
 }
 
 .form-control {
--- a/www/css/app.css
+++ b/www/css/app.css
@@ -64,6 +64,10 @@
     padding-left: 1px !important;
 }
 
+.dragging-over * {
+    pointer-events: none !important;
+}
+
 .form-control {
     margin-top: 2px;
 }
--- a/www/css/app.css
+++ b/www/css/app.css
@@ -94,6 +94,10 @@
     padding-top: 1rem;
 }
 
+/* #search-article * {
+    pointer-events: none;
+} */
+
 #configuration {
     margin: 0 1rem;
 }
--- a/www/js/app.js
+++ b/www/js/app.js
@@ -1640,7 +1640,6 @@ function handleGlobalDragenter (e) {
 }
 
 function handleGlobalDragover (e) {
-    console.log('glb over')
     e.preventDefault();
 
     if (e.dataTransfer.types.includes('Files') && !hasInvalidType(e.dataTransfer.types)) {
@@ -1654,7 +1653,6 @@ function handleGlobalDragover (e) {
 function handleGlobalDragleave (e) {
     e.preventDefault();
     globalDropZone.style.border = '';
-    console.log(e.target)
     if (e.dataTransfer.types.includes('Files') && !hasInvalidType(e.dataTransfer.types)) {
         /** can we somehow check if a page has been loaded? no need to go home if no page loaded yet. just keep on config */
         if (lastenter === e.target) {
@@ -1666,7 +1664,6 @@ function handleGlobalDragleave (e) {
 
 function handleIframeDragover (e) {
     e.preventDefault();
-    // add type check for chromium browsers
     if (e.dataTransfer.types.includes('Files') && !hasInvalidType(e.dataTransfer.types)) {
         globalDropZone.classList.add('dragging-over')
         e.dataTransfer.dropEffect = 'link';
@@ -1674,6 +1671,7 @@ function handleIframeDragover (e) {
     }
 }
 
+// add type check for chromium browsers, since they count images on the same page as files
 function hasInvalidType (typesList) {
     for (var i = 0; i < typesList.length; i++) {
         if (typesList[i].startsWith('image') || typesList[i].startsWith('text') || typesList[i].startsWith('video')) {
--- a/www/js/app.js
+++ b/www/js/app.js
@@ -1524,6 +1524,7 @@ function displayFileSelect () {
         globalDropZone.addEventListener('dragover', handleGlobalDragover);
         globalDropZone.addEventListener('dragleave', handleGlobalDragleave);
         globalDropZone.addEventListener('drop', handleFileDrop);
+        globalDropZone.addEventListener('dragenter', handleGlobalDragenter);
     }
 
     if (isFireFoxOsNativeFileApiAvailable) {
@@ -1633,14 +1634,17 @@ let noPointerEventsStyle = '#search-article * { pointer-events: none; };'
 let noPointerEventsStyleSheet = document.createElement('style');
 noPointerEventsStyleSheet.innerText = noPointerEventsStyle;
 
+function handleGlobalDragenter (e) {
+    e.preventDefault();
+    document.head.append(noPointerEventsStyleSheet);
+}
+
 function handleGlobalDragover (e) {
     e.preventDefault();
-   
-    document.head.append(noPointerEventsStyleSheet);
 
     if (e.dataTransfer.types.includes('Files')) {
         e.dataTransfer.dropEffect = 'link';
-        globalDropZone.style.border = '3px dotted red';
+        globalDropZone.style.border = '3px dashed red';
         document.getElementById('btnConfigure').click();
     }
 }
@@ -1659,6 +1663,7 @@ function handleIframeDragover (e) {
     // add type check for chromium browsers
     if (e.dataTransfer.types.includes('Files') && !hasInvalidType(e.dataTransfer.types)) {
         e.dataTransfer.dropEffect = 'link';
+        document.head.append(noPointerEventsStyleSheet);
         document.getElementById('btnConfigure').click();
     }
 }
--- a/www/js/app.js
+++ b/www/js/app.js
@@ -1655,9 +1655,8 @@ function handleGlobalDragover (e) {
 function handleGlobalDragleave (e) {
     e.preventDefault();
     globalDropZone.style.border = '';
-    if (hasType(e.dataTransfer.types, 'Files') && !hasInvalidType(e.dataTransfer.types) && enteredElement === e.target) {
+    if (enteredElement === e.target) {
         globalDropZone.classList.remove('dragging-over');
-            
         // Only return to page if a ZIM is actually loaded
         if (selectedArchive.isReady()) {
             returnToCurrentPage();
--- a/www/js/app.js
+++ b/www/js/app.js
@@ -1628,22 +1628,24 @@ document.getElementById('archiveFilesLbl').addEventListener('keydown', function 
     }
 });
 
-// while being dragged over, disable pointer events
-// on children so dragleave fires appropriately
-let noPointerEventsStyle = '#search-article * { pointer-events: none; };'
-let noPointerEventsStyleSheet = document.createElement('style');
-noPointerEventsStyleSheet.innerText = noPointerEventsStyle;
+// keep track of entrance event so we only fire the correct leave event 
+var lastenter;
 
 function handleGlobalDragenter (e) {
     e.preventDefault();
-    document.head.append(noPointerEventsStyleSheet);
+    // disable pointer-events on children
+    // so they don't interfere with dragleave events
+    globalDropZone.classList.add('dragging-over')
+    lastenter = e.target;
 }
 
 function handleGlobalDragover (e) {
+    console.log('glb over')
     e.preventDefault();
 
-    if (e.dataTransfer.types.includes('Files')) {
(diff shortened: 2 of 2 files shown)


Jaifroid:
@D3V-D Thanks very much for this PR! It's passed testing, and I've also tested manually -- it seems to be working perfectly as intended. Well done. I'll do a code review shortly.

D3V-D:
Sounds good! Do let me know if you find any answers to my questions above.

Jaifroid:
@D3V-D Regarding your test failures, I'm not seeing them in the CI (you can probably look at the CI test runs below), so they are probably a result of your environment / speed of CPU, etc. E2E tests are quite sensitive to timings and have been optimized to run on GitHub, where they are passing. We also have BrowserStack texts, but they can't run from a PR without access to the secrets, but I don't anticipate any issues given what your PR touches (we'll know for sure after merge).

Regarding knowing whether a ZIM is loaded: yes, simply call `selectedArchive.isReady()` in `app.js` (not currently available in other modules, but that could be changed if necessary). If you want to know whether a specific ZIM is loaded, then look at the `selectedArchive.name` field (if it exists, then the archive is loaded).

www/css/app.css
```diff
@@ -64,6 +64,10 @@
     padding-left: 1px !important;
 }
 
+.dragging-over * {
+    pointer-events: none !important;
```
Jaifroid, line 68: Is `!important` necessary? Not a problem if it's needed, but did you test without?

www/js/app.js
```diff
@@ -1521,17 +1521,11 @@ function displayFileSelect () {
...
-        configDropZone.addEventListener('dragleave', function () {
-            configDropZone.style.border = '';
-        });
-        // Also set a global drop zone (allows us to ensure Config is always displayed for the file drop)
-        globalDropZone.addEventListener('dragover', function (e) {
-            e.preventDefault();
-            if (configDropZone.style.display === 'none') document.getElementById('btnConfigure').click();
-            e.dataTransfer.dropEffect = 'link';
-        });
+        // do globally, since the whole page functions as a drop zone, so indicate properly
```
Jaifroid, line 1524: This comment implies the dev reading it will have knowledge of the deleted comments/code above. I suggest you start `// Set a global drop zone, so that the whole page is enabled for drag-and-drop`. I'm not sure what "so indicate properly" refers to, so I think you could delete.

www/js/app.js
```diff
@@ -1635,27 +1629,71 @@ document.getElementById('archiveFilesLbl').addEventListener('keydown', function
     }
 });
 
+/** Drag and Drop handling for ZIM files */
+
+// keep track of entrance event so we only fire the correct leave event
```
Jaifroid, line 1634: Very minor, but our house style tries to start comments with a capital letter, so it looks more like a well written sentence (however, only use full stop at end if you need more than one sentence).

www/js/app.js
```diff
@@ -1635,27 +1629,71 @@ document.getElementById('archiveFilesLbl').addEventListener('keydown', function
...
 
+/** Drag and Drop handling for ZIM files */
+
+// keep track of entrance event so we only fire the correct leave event 
+var enteredelement;
+
+function handleGlobalDragenter (e) {
+    e.preventDefault();
+    // disable pointer-events on children
+    // so they don't interfere with dragleave events
```
Jaifroid, line 1640: Ditto with above comments

www/js/app.js
```diff
@@ -1635,27 +1629,71 @@ document.getElementById('archiveFilesLbl').addEventListener('keydown', function
     }
 });
 
+/** Drag and Drop handling for ZIM files */
+
+// keep track of entrance event so we only fire the correct leave event 
+var enteredelement;
```
Jaifroid, line 1635: For variables we use camelCase to make more readable. Change to `enteredElement`.

www/js/app.js
```diff
@@ -1635,27 +1629,71 @@ document.getElementById('archiveFilesLbl').addEventListener('keydown', function
...
+        globalDropZone.style.border = '3px dashed red';
+        document.getElementById('btnConfigure').click();
+    }
+}
+
+function handleGlobalDragleave (e) {
+    e.preventDefault();
+    globalDropZone.style.border = '';
+    if (e.dataTransfer.types.includes('Files') && !hasInvalidType(e.dataTransfer.types)) {
+        /** can we somehow check if a page has been loaded? no need to go home if no page loaded yet. just keep on config */
```
Jaifroid, line 1660: Yes, see earlier comment in conversation.

www/js/app.js
```diff
@@ -1635,27 +1629,71 @@ document.getElementById('archiveFilesLbl').addEventListener('keydown', function
...
+    }
 }
 
 function handleIframeDrop (e) {
-    e.stopPropagation();
     e.preventDefault();
+    e.stopPropagation();
+}
+
+// add type check for chromium browsers, since they count images on the same page as files
```
Jaifroid, line 1682: Ditto about comment.

Jaifroid (changes requested):
It's looking good! Just some very minor suggestions mostly about house style.

www/css/app.css
D3V-D: I think I did test, but the issue at the time was something else (not 100% sure on that) and I never changed it back. It may well be unnecessary

www/js/app.js
D3V-D: Okay I'll go back and try and improve the comprehensiveness

www/js/app.js
D3V-D: Gotcha I'll fix with other comments

D3V-D:
Ok! I'll work on the commenting and variable names. Thanks for the answer about seeing if a page has been loaded and what page it is; I'll try to add those to the implementation while I'm at it.

D3V-D:
I'm not at my computer (not home rn) so I don't know if this would be trivial or more difficult, but if I store the current page at time of drag over, I should be able to just load back into that page after going to home right? (When they drag leave)

Jaifroid:
@D3V-D I tested the compiled version in IE11, and unfortunately drag-and-drop now only works if Config is open. It no longer works dragging the ZIM onto the app with a ZIM already open and displayed. Additionally (and it's probably related), the red "drop-zone" outline no longer shows in IE11, whereas it does in modern browsers. On main, both of these things work.

We use IE11 as a proxy for older browsers (if it works in IE11 it'll work everywhere!). I think this is why I originally coded two drop zones, because the global drop zone isn't working in all browsers (I realize I should have left a comment explaining this if it's the case). Alternatively, the explanation may be because some of the newer events you are using are not supported in older browsers, and in that case it might be enough to test if the API exists before attempting to use it, and fall back to the default API if not.

In Windows, you can easily test on IE11 by using IE11 Mode in Microsoft Edge. However, debugging is a bit trickier in that context, and it may be easier to use a simple trick to open IE11 itself (I can give you details via Slack if you have access to Windows 10/11).

www/css/app.css
Jaifroid: OK, if it turns out not to be necessary, best to remove it.

D3V-D:
Hey, yeah I was a little worried about IE11 since I'm on a Linux machine and I don't have it installed. If needed I can run a windows VM and run it for testing; let me know if that would work

Jaifroid:
> I'm not at my computer (not home rn) so I don't know if this would be trivial or more difficult, but if I store the current page at time of drag over, I should be able to just load back into that page after going to home right? (When they drag leave)

No need to store anything. In fact there is an easy way to restore the currently viewed article without reloading the landing page. Look for function `showReturnLink` in `uiUtil.js`, and copy what happens in the `viewArticle` click event listener in that function. That hides config and shows the currently loaded article, without reloading anything.

D3V-D:
That sounds very convenient; thanks for letting me know about it.

Jaifroid:
> Hey, yeah I was a little worried about IE11 since I'm on a Linux machine and I don't have it installed. If needed I can run a windows VM and run it for testing; let me know if that would work

Yes, that would work. Or you can try using https://caniuse.com/ to check which APIs you're using aren't available on IE11 and add a fallback, and I'd be happy to test. If that doesn't work, you may have to restore the Config zone in addition to the global drop zone.

D3V-D:
> @D3V-D I tested the compiled version in IE11, and unfortunately drag-and-drop now only works if Config is open. It no longer works dragging the ZIM onto the app with a ZIM already open and displayed. Additionally (and it's probably related), the red "drop-zone" outline no longer shows in IE11, whereas it does in modern browsers. On main, both of these things work.
> 

image
Won't know till I can do some testing, but this may be part of the problem, seems to be an IE11 quirk. Do contact me on slack if you want to discuss it more in-depth

D3V-D:
I figured out the issues with IE11; both the .startsWith() and the .includes() functions are not supported, so I had to use some more basic functions that effectively achieve the same goal. 

Additionally, I implemented the other features, so now dragging over on config page when no ZIM is loaded no longer unnecessarily takes you back to the home page, and dragging over and exiting doesn't unnecessarily load the home page of the current ZIM but rather just restores the last page that was open.

I also fixed up the commenting and variable names. 

Please test for IE11 @Jaifroid and let me know if its working as intended now!

Jaifroid:
OK, thanks -- I'll check, although note that we use core-js so startsWith, etc., should be polyfilled when the app is built (which is necessary for IE11 to work in any case). However, there's no problem with using more universal JS in any case -- often a lot of modern methods are just sugar.

Jaifroid:
Congratulations @D3V-D, I confirm that IE11 is now working fine! Tests are passing, so I'll do another code review asap.

www/js/app.js
```diff
@@ -1635,39 +1629,105 @@ document.getElementById('archiveFilesLbl').addEventListener('keydown', function
...
+        globalDropZone.classList.add('dragging-over');
+        globalDropZone.style.border = '3px dashed red';
+        document.getElementById('btnConfigure').click();
+    }
+}
+
+function handleGlobalDragleave (e) {
+    e.preventDefault();
+    globalDropZone.style.border = '';
+    if (hasType(e.dataTransfer.types, 'Files') && !hasInvalidType(e.dataTransfer.types) && enteredElement === e.target) {
```
Jaifroid, line 1658: Why do you need to do these checks when the user leaves dragover? Isn't it enough simply to remove the red border and return to the article (if one is loaded)? I may be misunderstanding something, I didn't test that.

Jaifroid (commented):
This is looking very good. I've got some minor suggestions and a query. But this is functioning very well!

www/js/app.js
```diff
@@ -1635,39 +1629,105 @@ document.getElementById('archiveFilesLbl').addEventListener('keydown', function
...
+}
+
+// Function to switch back to currently loaded page from config page after dragleave event
+function returnToCurrentPage () {
+    document.getElementById('liConfigureNav').classList.remove('active');
+    document.getElementById('liHomeNav').classList.add('active');
+    uiUtil.tabTransitionToSection('home', params.showUIAnimations);
+    const welcomeText = document.getElementById('welcomeText');
+    welcomeText.style.display = 'none';
+    viewArticle.style.display = 'none';
```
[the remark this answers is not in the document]
D3V-D: Sure I don't mind doing that, should be pretty simple

D3V-D:
Didn't mean to resolve that one, but please do read the comment under it. 

As for the others, I'll make some last edits.

D3V-D:
Alright, I've removed the blank line and shortened the checks on the `if` to just use the `enteredElement`.

Please let me know if there's anything else @Jaifroid! If not, I can make that PR for moving the function to uiUtil once this has been merged.

Jaifroid (approved):
That's great! My last manual tests -- including dropping all the parts of a split ZIM -- all passed, so this is ready to merge. Let me know when you're happy for me to go ahead. It would be great if you could make an issue for moving the `returnToCurrentPage` function to `uiUtil.js` (before making a PR for it), so I can formally assign you to it. It should be a very quick PR.

D3V-D:
Feel free to merge right now. I'll write up the issue

Issue #1247, opened 2024-05-03
Add ZIM info to the security dialogue box when opening a ZIM for the first time


Jaifroid:
See Kiwix PWA for existing implementation. I propose simply providing the same info as in screenshot:

![image](https://github.com/kiwix/kiwix-js/assets/4304337/f009a48d-7196-470c-896d-7610003da5b7)

D3V-D:
I'd also be interested in helping in this issue if help is needed.

Jaifroid:
> I'd also be interested in helping in this issue if help is needed. @Jaifroid

Happy for you to do it -- it should be pretty easy to use the PWA template/code for this. The only complication is adding the translation strings. Also, instead of "can easily be spoofed" I think it would be more accurate to say "can be spoofed" (and I'll probably change that in the PWA when I've finished another PR I'm working on).

Jaifroid:
Oh, and you'll need to make sure all the required metadata are available when the ZIM first opens, which may require a small change to `zimArchive.js`.

D3V-D:
Yeah, once I take a look at it ill let you know what I think. I have yet to work with "translation strings" - is that the alternate language versions of the text i assume?

Jaifroid:
@D3V-D Yes, I've implemented French and Spanish (as well as English) in Kiwix JS (but not in the PWA). It's sort of obvious how it's done when you look at it -- there are lots of examples of invoking a dialogue box like that with the language string keys, and then the translation backend takes care of displaying the appropriate language. You can use https://www.deepl.com/translator to get the French and Spanish versions (but that comes at the end, after code is working). I can check French and Spanish, as I know these languages.

D3V-D:
Sorry I haven't had time to take a good look at the issue. I have exams for the next week or two, but after that I'll be able to do some work on this issue.

Just a heads-up.

PR #1250, opened 2024-05-13
Added ZIM metadata to security dialogue box when opening ZIM for the first time

D3V-D:
Fix for #1247 
Ran `npm run serve` and `npm run test` and `npm run preview`
Also did `npm run build` followed by `npm run serve` and tested on the /dist page.

Tested firefox + chromium. (not tested on ie11 yet)

Pretty sure jQuery mode testing is not applicable as the security dialogue is only for serviceWorker mode (correct me if I'm wrong).

Added french and spanish translations (and English) with deepl, feel free to double check.

Please review, @Jaifroid . Thanks!


D3V-D:
Just wanted to add, there's a lot of other ZIM metadata I think could be used well here; so I don't know if we'd want to pursue that? Like including date of creation, description, etc.

Jaifroid:
> Just wanted to add, there's a lot of other ZIM metadata I think could be used well here; so I don't know if we'd want to pursue that? Like including date of creation, description, etc.

There are indeed, but each one has a "cost", in that we have to load them up-front from the ZIM for them to be available at the time these metadata are shown. Currently, most metadata aren't needed at ZIM load time, and they load in the background in order to prioritize opening the ZIM as fast as possible for the user. So I judged that these are the most important ones (when I coded it for the PWA).

Jaifroid:
I've tested on Edge -- all functionality appears to work as intended. In IE11 of course the security dialogue doesn't show because it's already running in jQuery, aka "Safe", mode, but the code doesn't cause any issues for the browser. So, all working well!

I'll do a code review asap.

D3V-D:
Yep, please let me know if I did something strange when retrieving the metadata for the alert, the solution felt somewhat obtuse.

Jaifroid:
> Also, did you link this PR to the corresponding issue?

Done. Normally it links automatically if you write "Fixes .....", maybe because you wrote "Fix for ..." (and it's not a very intelligent algorithm)!

Jaifroid:
Actually the algo is more intelligent than I give it credit for:

image

www/js/lib/uiUtil.js
```diff
@@ -183,15 +183,27 @@ function slideAway (e) {
  * @param {String} closeMessageLabel  The text to display on the close alert message button (optional, Default = "Okay")
  * @param {String} hideOptionLabel  The text to display on the hide option button (optional, Default = "Don't ask again")
  * @param {Boolean} displayHideOption If true, option to permanently hide the modal will be shown (currently only implemented for hideExternalLinkWarning)
+ * @param {Object} archiveMetadata If warning message for unknown ZIM file, contains the information to show user to help them confirm validity of ZIM file
+ * @param {String} archiveMetadata.name The name of the ZIM file
+ * @param {String} archiveMetadata.creator The creator of the ZIM file
+ * @param {String} archiveMetadata.publisher The publisher of the ZIM file
+ * @param {String} archiveMetadata.scraper The scraper used to create the ZIM file
```
Jaifroid, line 190: Hmm. It's clunky to create these custom properties extending the systemAlert function for a single purpose in the app. The existing function works absolutely fine if you simply pass it the HTML with the text already translated. As this is a basic and generic function, I don't think you should alter it for such a specific purpose.

www/js/lib/uiUtil.js
```diff
@@ -183,15 +183,27 @@ function slideAway (e) {
...
  * @returns {Promise<Boolean>} A promise which resolves to true if the user clicked Confirm, false if the user clicked Cancel/Okay, backdrop or the cross(x) button
  */
-function systemAlert (message, label, isConfirm, declineConfirmLabel, approveConfirmLabel, closeMessageLabel, hideOptionLabel, displayHideOption) {
+function systemAlert (message, label, isConfirm, declineConfirmLabel, approveConfirmLabel, closeMessageLabel, hideOptionLabel, displayHideOption, archiveMetadata, metadataWarningLabel) {
```
Jaifroid, line 193: So, I suggest you revert this function to the generic state and send the HTML to the alert function with the necessary text already translated. You can easily do the pre-translatio in the `verifyLoadedArchive` function.

www/js/lib/uiUtil.js
```diff
@@ -183,15 +183,27 @@ function slideAway (e) {
...
     hideOptionLabel = hideOptionLabel || (translateUI.t('dialog-hide') || "Don't ask again");
     displayHideOption = displayHideOption || false;
     label = label || (isConfirm ? 'Confirmation' : 'Message');
+    metadataWarningLabel = metadataWarningLabel || translateUI.t('dialog-metadata-warning') || 'Warning: above data can be spoofed!';
+    const metadataLabels = {
+        name: translateUI.t('dialog-metadata-name') || 'Name: ',
+        creator: translateUI.t('dialog-metadata-creator') || 'Creator: ',
+        publisher: translateUI.t('dialog-metadata-publisher') || 'Publisher: ',
+        scraper: translateUI.t('dialog-metadata-scraper') || 'Scraper: '
+    }
```
Jaifroid, line 206: Similarly, this code really doesn't belong here. In general, avoid mixing the code from a new feature into existing generic functions, otherwise things get confusing! It's a good principle of functional programming (although most of KJS was written with an object-oriented programming style...).

www/js/lib/uiUtil.js
```diff
@@ -204,6 +216,16 @@ function systemAlert (message, label, isConfirm, declineConfirmLabel, approveCon
...
+            // Set archive metadata if provided. Avoid using innerHTML since data can be spoofed.
+            document.getElementById('modal-archive-metadata-container').style.display = archiveMetadata ? 'block' : 'none';
+            if (archiveMetadata) {
+                document.getElementById('confirm-archive-name').textContent = metadataLabels.name + (archiveMetadata.name || '-');
+                document.getElementById('confirm-archive-creator').textContent = metadataLabels.creator + (archiveMetadata.creator || '-');
+                document.getElementById('confirm-archive-publisher').textContent = metadataLabels.publisher + (archiveMetadata.publisher || '-');
+                document.getElementById('confirm-archive-scraper').textContent = metadataLabels.scraper + (archiveMetadata.scraper || '-');
+            }
+            document.getElementById('modal-archive-metadata-warning').textContent = metadataWarningLabel;
+            document.getElementById('modal-archive-metadata-warning').style.display = archiveMetadata ? 'block' : 'none';
```
Jaifroid, line 228: Ditto, move all this into your own pre-formatting function. You can either gather the translated variables you need and then simply insert them into HTML, or you can create the element with `var securityModal = document.createElement(.....)`, populate it with DOM methods like here, and then simply grab the HTML `var securityModalHTML = sucurityModal.outerHTML;` and send the HTML to `uiUtil.systemAlert`.

www/js/lib/zimArchive.js
```diff
@@ -85,7 +85,7 @@ function ZIMArchive (storage, path, callbackReady, callbackError) {
             console.debug('ZIMArchive ready, metadata will be added in the background');
             // Add non-time-critical metadata to archive in background so as not to delay opening of the archive
             // DEV: Note that it does not make sense to extract illustration (icon) metadata here. Instead, if you implement use of the illustration
-            // metadata as icons for the loaded ZIM [kiwix-js #886], you should simply use the ZIMArdhive.getMetadata() function when needed
+            // metadata as icons for the loaded ZIM [kiwix-js #886], you should simply use the ZIMArchive.getMetadata() function when needed
```
Jaifroid, line 88: Thanks!

www/js/lib/zimArchive.js
```diff
@@ -668,6 +668,25 @@ ZIMArchive.prototype.getMetadata = function (key, callback) {
...
+                that.getMetadata('Publisher', function (publisher) {
+                    that.getMetadata('Scraper', function (scraper) {
+                        resolve({ name: name, creator: creator, publisher: publisher, scraper: scraper });
+                    });
+                });
+            });
+        });
+    })
+}
+
```
Jaifroid, line 689: It's good that you worked out how to get the metadata! However, it's reinventing the wheel, because the current code has two places to get metadata: the first is while the ZIM is loading, and the second is asynchronously after the ZIM has loaded (for non-critical data). All you need to do is add the time-critical metadata (i.e. these keys) to the list here: https://github.com/kiwix/kiwix-js/blob/main/www/js/lib/zimArchive.js#L82, and remove any of these keys that appear in the list of non-critical data here: https://github.com/kiwix/kiwix-js/blob/main/www/js/lib/zimArchive.js#L91. Also, JFYI, the more elegant way to gather a series of Promises is to use `Promise.all()` as is used in those functions, rather than a chain of Promises as you do here. Once you've modified the code around line 82, you can remove this function. The metadata you need will then be available in `archive.file.xxxx` in `archiveReadyCallback` (or `selectedArchive.file.xxxx` in the rest of `app.js`).

Jaifroid (changes requested):
@D3V-D Many thanks for this PR. It works, as I mentioned, and you came up with some creative solutions. There are, however, a few things that could do with refactoring to avoid putting your code into generic functions, and also to ensure metadata are fetched at the appropriate time in a way that can be used by other app functions. All explained in comments.

Regarding translations, thanks for doing these. I'll check these properly once the code is ready.

Don't hesitate to ask if you have any queries!

D3V-D:
@Jaifroid many thanks for the feedback! Yeah I did think of passing in HTML directly, but I wasn't sure if that was a good idea or not, but I guess it would be better for the reasons you stated.

I'll also look into the different ways of getting the metadata, and get back to you with an update

Jaifroid:
> @Jaifroid many thanks for the feedback! Yeah I did think of passing in HTML directly, but I wasn't sure if that was a good idea or not, but I guess it would be better for the reasons you stated.
> 
> I'll also look into the different ways of getting the metadata, and get back to you with an update

Thanks! I don't think it will take very long, as it's mostly moving a few things around.

D3V-D:
Alright, everything should be in order now. Please review again (and any testing necessary to ensure it's still functional, but it all seems fine on my end).

i18n/es.jsonp.js
```diff
@@ -178,6 +178,11 @@ document.localeJson = {
       "dialog-readzim-error-message": "Error al leer archivo",
       "dialog-reset-warning-title": "¡AVISO!",
       "dialog-reset-warning-message": "¡Se restablecerá la aplicación al estado original, borrando todas las cachés y configuraciones!",
+      "dialog-metadata-warning": "Atención: ¡los datos anteriores pueden ser falsificados!",
```
Jaifroid, line 181: I would change to: "¡Atención! Estos datos pueden ser falsificados."

i18n/es.jsonp.js
```diff
@@ -178,6 +178,11 @@ document.localeJson = {
...
+      "dialog-metadata-name": "Nombre: ",
+      "dialog-metadata-creator": "Creador: ",
+      "dialog-metadata-publisher": "Editor: ",
+      "dialog-metadata-scraper": "Rascador: ",
```
Jaifroid, line 185: The correct word is usually "Raspador" for a Web scraper. "Rascador" means something like "paint scraper".

www/js/app.js
```diff
@@ -640,15 +640,63 @@ function focusPrefixOnHomeKey (event) {
         }, 0);
     }
 }
+
 /**
  * Verifies the given archive and switches contentInjectionMode accourdingly
- * @param {archive} the archive that needs verification
+ * @param {archive} archive the archive that needs verification
```
Jaifroid, line 646: I would change `{archive}` to `{Object}` (with capital letter. Then, instead of `the archive ...` I would change to `The archive ...` (i.e. capitalized `The`).

www/js/app.js
```diff
@@ -640,15 +640,63 @@ function focusPrefixOnHomeKey (event) {
...
  * */
 async function verifyLoadedArchive (archive) {
-    const response = await uiUtil.systemAlert(translateUI.t('dialog-sourceverification-alert') ||
-    'Is this ZIM archive from a trusted source?\n If not, you can still read the ZIM file in Safe Mode. Closing this window also opens the file in Safe Mode. This option can be disabled in Expert Settings',
-    translateUI.t('dialog-sourceverification-title') || 'Security alert!', true, translateUI.t('dialog-sourceverification-safe-mode-button') || 'Open in Safe Mode',
-    translateUI.t('dialog-sourceverification-trust-button') || 'Trust Source');
+    // We construct a HTML element to show the user in the alert with the metadata contained in it
```
Jaifroid, line 649: `an HTML element` (rather than `a HTML element`)
`show the user the alert` (not `in the alert`)

i18n/fr.jsonp.js
```diff
@@ -178,6 +178,11 @@ document.localeJson = {
       "dialog-readzim-error-message": "Erreur lors de la lecture du fichier ZIM",
       "dialog-reset-warning-title": "ATTENTION&nbsp;!",
       "dialog-reset-warning-message": "Cela réinitialisera l'application à l'état d'installation originale. Tous les caches et paramètres seront effacés&nbsp;!",
+      "dialog-metadata-warning": "Attention: les données ci-dessus peuvent être falsifiées!",
+      "dialog-metadata-name": "Nom: ",
+      "dialog-metadata-creator": "Créateur: ",
+      "dialog-metadata-publisher": "Éditeur: ",
+      "dialog-metadata-scraper": "Grattoir: ",
```
Jaifroid, line 185: In all the above translations, please put `&nbsp;` between the word and the colon (but the space after the colon can remain as a normal space). Also, in the first case, add `&nbsp;` before the exclamation mark. This is because French uses spacing around colons, question marks, exclamation marks and guillemets. It needs to be a non-breaking space so that the punctuation doesn't get orphaned on narrow screens.

i18n/en.jsonp.js
```diff
@@ -178,6 +178,11 @@ document.localeJson = {
       "dialog-readzim-error-message": "Error reading ZIM file",
       "dialog-reset-warning-title": "WARNING!",
       "dialog-reset-warning-message": "This will reset the app to a freshly installed state, deleting all app caches and settings!",
+      "dialog-metadata-warning": "Warning: above data can be spoofed!",
+      "dialog-metadata-name": "Name: ",
+      "dialog-metadata-creator": "Creator: ",
+      "dialog-metadata-publisher": "Publisher: ",
+      "dialog-metadata-scraper": "Scraper: ",
```
Jaifroid, line 185: Please move this block above so it is immediately below the last `dialog-messagechannel...` translation (so it's in roughly alphabetical order).

i18n/es.jsonp.js
Jaifroid, line 185: Ditto about moving the block so it's in alphabetical order.

i18n/fr.jsonp.js
Jaifroid, line 185: Ditto about moving block so it's in alphabetical order.

www/js/app.js
```diff
@@ -640,15 +640,63 @@ function focusPrefixOnHomeKey (event) {
...
+        name: translateUI.t('dialog-metadata-name') || 'Name: ',
+        creator: translateUI.t('dialog-metadata-creator') || 'Creator: ',
+        publisher: translateUI.t('dialog-metadata-publisher') || 'Publisher: ',
+        scraper: translateUI.t('dialog-metadata-scraper') || 'Scraper: '
+    }
+
+    const verificationBody = document.createElement('div');
+
+    // text & metadata box
+    const verificationText = translateUI.t('dialog-sourceverification-alert') || 'Is this ZIM archive from a trusted source?\n If not, you can still read the ZIM file in Safe Mode. Closing this window also opens the file in Safe Mode. This option can be disabled in Expert Settings';
```
Jaifroid, line 660: Add missing full stop at the end of `Settings`.

www/js/app.js
```diff
@@ -640,15 +640,63 @@ function focusPrefixOnHomeKey (event) {
...
+    const metadataLabels = {
+        name: translateUI.t('dialog-metadata-name') || 'Name: ',
+        creator: translateUI.t('dialog-metadata-creator') || 'Creator: ',
+        publisher: translateUI.t('dialog-metadata-publisher') || 'Publisher: ',
+        scraper: translateUI.t('dialog-metadata-scraper') || 'Scraper: '
+    }
+
+    const verificationBody = document.createElement('div');
+
+    // text & metadata box
```
Jaifroid, line 659: Use capital `T` for `Text`.

www/js/app.js
```diff
@@ -640,15 +640,63 @@ function focusPrefixOnHomeKey (event) {
...
+    verifyScraper.id = 'confirm-archive-scraper';
+    verifyScraper.classList.add('archive-metadata');
+    verifyScraper.innerText = metadataLabels.scraper + (archive.scraper || '-');
+
+    const verifyWarning = document.createElement('p');
+    verifyWarning.id = 'modal-archive-metadata-warning';
+    verifyWarning.innerText = translateUI.t('dialog-metadata-warning') || 'Warning: above data can be spoofed!';
+
+    metadataBox.append(verifyName, verifyCreator, verifyPublisher, verifyScraper);
+    verificationBody.append(verificationText, metadataBox, verifyWarning);
```
Jaifroid, line 690: I think it should be OK to use `append()` here because the code will not be executed by IE11 or by any browser that does not support Service Workers.

Jaifroid (commented):
That's much better now, many thanks! It's looking very good. Just some really minor last things (punctuation, translations, ...).

Jaifroid:
I also just noticed this, which doesn't show on main. You are probably using `textContent` somewhere and it might literally interpret `&nbsp;` as text rather than as HTML? You can probably just substitute `innerHTML` at the appropriate point.

image

D3V-D:
Yeah, I'll take a look at all those things.

D3V-D:
Okay, I've fixed pretty much all the things you said, just one issue.

When putting the text for the verification metadata, I avoided using innerHTML and instead used innerText to avoid (since the metadata can be "spoofed") any kind of security issues. However, this makes it hard to use the &nbsp; for French. If necessary to use that instead of a normal blank space, I can try using innerHTML for the translated text then followed by innerText to insert the actual metadata.

Jaifroid:
OK, great. I understand the security concern with innerHTML if we are inserting a string that we do not control and may be a vector for attack. We do, of course, control the metadata labels, so it should be safe to insert at least those with innerHTML. The alternative would be to sanitize the metadata themselves, potentially in the getMetadata function (or whatever it's called).

www/js/app.js
```diff
@@ -640,15 +640,64 @@ function focusPrefixOnHomeKey (event) {
         }, 0);
     }
 }
+
 /**
  * Verifies the given archive and switches contentInjectionMode accourdingly
- * @param {archive} the archive that needs verification
+ * @param {Object} archive The archive that needs verification
```
Jaifroid, line 646: Apologies, I just remembered that we have defined a type definition `{ZIMArchive}`, which my IDE recognizes, so perhaps best to use that here instead of `{Object}`.

Jaifroid:
Actually, no need to complicate the code, as you're right that we don't actually need `&nbsp;` where a space will do and there is no risk of the colons being orphaned (no screen will ever be that narrow). So leave it as you have it now.

Jaifroid (approved):
All looking good now! Perhaps just change that type from Object to ZIMArchive, and it's ready.

www/js/app.js
D3V-D: I actually did see that before, but I wasn't sure if it was the exact same class or not. I'll fix that

D3V-D:
All done! Feel free to merge whenever @Jaifroid

Jaifroid:
Great, thanks for the PR!

Issue #908, opened 2022-11-06
Add DarkReader plugin as an optional dark theme


Jaifroid:
I have been experimenting (successfully) in KJSW with the open-source DarkReader plugin: https://github.com/darkreader/darkreader. It is a simple JavaScript dependency (single file). It would work only in ServiceWorker Mode. All that is needed is to inject the script into the HTML document, and it analyses the document and generates specific CSS for a dark theme. The advantage of this is that maintenance of it is crowd-sourced, and that there are various options that can be tweaked.

There is one potential disadvantage: to work without the screen flashing to white between each page load, it may be necessary to inject the script into the HTML before it is rendered by the browser, which means adding it at the point at which the HTML is received fromt he back end. It may be possible to avoid this by hiding the document until the script is added through DOM methods, and then allowing it to render. Either way, I think it can be injected in a way that would not interfere with #870 (i.e. using new libzim WASM).

I would propose including this alongside our exisitng Dark Theme options.

mossroy:
It might indeed be better to use an existing tool.
However, I don't think injecting code in the HTML stream is a good idea in SW mode. It takes the risk to inject it in files that should not have it, or to break them.
I understand the flashing issue, and it might be complicated to avoid it.

However, I see there are different generation modes (maybe the less smart ones can be applied without having to wait for the whole page to be loaded), and some APIs to interact programmatically with it

Jaifroid:
Thank you @mossroy, I agree, and from my experiments I think it should be possible to add the JS using DOM methods only before the page renders, and that this may be sufficient. Yes, there are various modes that require less (or more) processing power.

PR #1252, opened 2024-05-19
Backport popover code from Kiwix PWA

### LINKED ISSUE #719: Support "page preview" feature of Wikipedia and Wikivoyage
This is a feature that is default for signed-out users of Wikipedia, though signed-in users may need to switch it on. It consists in showing a small preview of the lede, together with an image (if the article has one) when a user hovers over (or long-presses on mobile) an intenal link. It looks like in the screenshot below (the Île-de-France pop-up). It seems to me that this would be quite easy to support in Kiwix JS, especially now that block caching gives us quick lookup capabilities. The equivalent of hovering over a link in mobile is long-press.

Would this be of interest? An issue might be how to get the target of a hover in SW mode without interfering with the DOM. I think it can be done with a mouseover event listener on the iframe **window** (i.e. above the DOM), from which the target of the event can be retrieved and tested. It would need to run with a suitable timeout delay, so that it is not triggered excessively with each mouse movement, but this is precisely what is required for a hover event. This method would avoid having to attach listeners to each link in the DOM.

image

Jaifroid:
Fixes #719. This backports the new feature from the PWA.

So far, I have ported the mouseover and the focus events, together with the popover code. Focus event provides keyboard support (tabbing into a link).

To do:

- [x] Add touch support (long press)
- [x] Add pointerdown fallback for browsers that don't support touchstart (e.g. Safari on macOS, and Edge legacy)
- [x] Fix hover over popover in Safari for macOS
- [x] Disable the contextmenu event if popovers are enabled
- [x] Enable for Safe mode (without image support, as in the PWA)
- [x] Add the option to disable popover preview in the UI, with translations
- [x] Fix support for dark modes
- [x] Move as many inline styles as possible to stylesheet
- [x] Suppress popover loading if user has clicked the link
- [x] Add e2e test for presence of popover on tab into link (Ray Charles ZIM)
- [x] Make the breakout icon open a new window
- [x] Fix clicking on a link inside the popover
- [x] ~~Add popover support to new windows~~ Open a new issue to add popover support to new tabs/windows: #1253
- [x] Clean up code hangovers from the PWA that are not needed here
- [x] Final testing


www/js/lib/uiUtil.js
```diff
@@ -1003,6 +1007,382 @@ function warnAndOpenExternalLinkInNewTab (event, clickedAnchor, archive) {
...
+                            return !/^\s*$/.test(text) && text.length >= 50;
+                        });
+                        if (nonEmptyParagraphs.length > 0) {
+                            var cumulativeCharCount = 0;
+                            // Add enough paras to complete the word count
+                            for (let i = 0; i < nonEmptyParagraphs.length; i++) {
+                                // Get the character count: to fill the larger box we need ~850 characters (815 plus leeway)
+                                var plainText = nonEmptyParagraphs[i].innerText;
+                                cumulativeCharCount += plainText.length;
+                                // In Restricted mode, we risk breaking the UI if user clicks on an embedded link, so only use innerText
```
Jaifroid, line 1053: Refer to it as "Safe mode" in this codebase

www/js/lib/uiUtil.js
```diff
@@ -1003,6 +1007,382 @@ function warnAndOpenExternalLinkInNewTab (event, clickedAnchor, archive) {
...
+                            style.parentNode.removeChild(style);
+                        });
+                        const paragraphs = Array.from(articleBody.querySelectorAll('p'));
+                        // Filter out empty paragraphs or those with less than 50 characters
+                        const nonEmptyParagraphs = paragraphs.filter(para => {
+                            const text = para.innerText.trim();
+                            return !/^\s*$/.test(text) && text.length >= 50;
+                        });
+                        if (nonEmptyParagraphs.length > 0) {
+                            var cumulativeCharCount = 0;
```
Jaifroid, line 1047: Use `let` or `const` in new code instead of legacy `var`.

www/js/lib/uiUtil.js
```diff
@@ -1003,6 +1007,382 @@ function warnAndOpenExternalLinkInNewTab (event, clickedAnchor, archive) {
...
+                archive.resolveRedirect(dirEntry, function (reDirEntry) {
+                    resolve(readArticle(reDirEntry));
+                });
+            }).catch(error => {
+                return Promise.reject(error);
+            });
+        } else {
+            return Promise.resolve(readArticle(dirEntry));
+        }
+    });
```
Jaifroid, line 1107: Handle catch?

www/js/lib/uiUtil.js
```diff
@@ -1003,6 +1007,382 @@ function warnAndOpenExternalLinkInNewTab (event, clickedAnchor, archive) {
...
+
+/**
+ * A function to attach the tooltip CSS for popovers (NB this does not attach the box itself, only the CSS)
+ * @param {Document} doc The document to which to attach the blloon.css styelesheet
+ * @param {Boolean} dark An optional parameter to adjust the background colour for dark themes
+ */
+function attachKiwixPopoverCss (doc, dark) {
+    const colour = dark ? '#darkgray' : '#black';
+    const backgroundColour = dark ? '#111' : '#ebf4fb';
+    var cssLink = document.createElement('link');
```
Jaifroid, line 1118: Be consistent with let or const.

www/js/lib/uiUtil.js
```diff
@@ -1003,6 +1007,382 @@ function warnAndOpenExternalLinkInNewTab (event, clickedAnchor, archive) {
...
+      appstate.expectedArticleURLToBeDisplayed === 'A/Wikivoyage:Offline_reader_Expedition/Home_page')) {
+        return;
+    }
+    link.popoverisloading = true;
+    // Do not disply a popover if one is already showing for the current link
+    var kiwixPopover = ev.target.ownerDocument.querySelector('.kiwixtooltip');
+    if (kiwixPopover && kiwixPopover.dataset.href === linkHref) return;
+    // console.debug('Attaching popover...');
+    var currentDocument = ev.target.ownerDocument;
+    var articleWindow = currentDocument.defaultView;
```
Jaifroid, line 1197: Rename all var declarations

www/js/lib/uiUtil.js
```diff
@@ -1003,6 +1007,382 @@ function warnAndOpenExternalLinkInNewTab (event, clickedAnchor, archive) {
...
+    setTimeout(function () {
+        // Do not show popover if the user has initiated an article load
+        if (link.articleloading) return;
+        // Check if the link is still being hovered over, and abort display of popover if not
+        if (!link.matches(':hover') && currentDocument.activeElement !== link) {
+            link.popoverisloading = false;
+            return;
+        }
+        var div = document.createElement('div');
+        div.popoverisloading = true; // console.debug('div.popoverisloading', div.popoverisloading);
```
Jaifroid, line 1208: Remove comment

www/js/lib/uiUtil.js
```diff
@@ -1003,6 +1007,382 @@ function warnAndOpenExternalLinkInNewTab (event, clickedAnchor, archive) {
...
+                    position: absolute;
+                    top: ${triangleY}px;
+                    left: ${triangleX}px;
+                `;
+                div.appendChild(span);
+            }
+            // Programme the icons
+            addEventListenersToPopoverIcons(link, div, currentDocument);
+            setTimeout(function () {
+                div.popoverisloading = false; // console.debug('div.popoverisloading', div.popoverisloading);
```
Jaifroid, line 1306: Remove comment

www/js/lib/uiUtil.js
```diff
@@ -1003,6 +1007,382 @@ function warnAndOpenExternalLinkInNewTab (event, clickedAnchor, archive) {
...
+ *
+ * @param {Document} doc The document from which to remove any popovers
+ */
+function removeKiwixPopoverDivs (doc) {
+    var divs = doc.getElementsByClassName('kiwixtooltip');
+    setTimeout(function () {
+        Array.prototype.slice.call(divs).forEach(function (div) {
+            if (div.popoverisloading) return;
+            var timeoutID;
+            var fadeOutDiv = function () {
```
Jaifroid, line 1363: Explain what this does (in comment)

tests/e2e/spec/legacy-ray_charles.e2e.spec.js
```diff
@@ -205,7 +205,7 @@ function runTests (driver, modes) {
             it('Load legacy Ray Charles and check index contains specified article', async function () {
                 if (!serviceWorkerAPI) {
                     console.log('\x1b[33m%s\x1b[0m', '    - Following test skipped:');
-                    return;
+                    this.skipe();
```
Jaifroid, line 208: Correct typo

tests/e2e/spec/legacy-ray_charles.e2e.spec.js
```diff
@@ -274,12 +274,37 @@ function runTests (driver, modes) {
...
+                await driver.switchTo().frame('articleContent');
+                // Focus on the link "Hallelujah" with id="mwVw"
+                let link = await driver.findElement(By.id('mwVw'));
+                await driver.executeScript('arguments[0].focus();', link);
+                await driver.sleep(2000);
+                // Focus on the next link "A Fool for You" with id="mwWw"
+                await driver.executeScript('document.getElementById("mwWw").focus();');
+                // Wait for the popover to appear
+                await driver.sleep(2500); // DEV: Adjust this delay if failing on older, slower browsers
+                // Use standard JavaScript methods to find the popover element
```
Jaifroid, line 297: Explain that Safari 14 fails with webdriver methods

tests/e2e/spec/legacy-ray_charles.e2e.spec.js
```diff
@@ -274,12 +274,37 @@ function runTests (driver, modes) {
...
+                let popover = await driver.executeScript('return document.querySelector(".kiwixtooltip").outerHTML;');
+                let popoverContainsText = /bluesy/.test(popover);
```
Jaifroid, line 299: Add comment to explain

www/js/app.js
```diff
@@ -2227,11 +2226,11 @@ function filterClickEvent (event) {
             console.debug('filterClickEvent opening external link in new tab');
             clickedAnchor.newcontainer = true;
             uiUtil.warnAndOpenExternalLinkInNewTab(event, clickedAnchor);
-        } else if (/\.pdf([?#]|$)/i.test(href) && selectedArchive.zimType !== 'zimit') {
+        } else if (clickedAnchor.newcontainer || /\.pdf([?#]|$)/i.test(href) && selectedArchive.zimType !== 'zimit') {
             // Due to the iframe sandbox, we have to prevent the PDF viewer from opening in the iframe and instead open it in a new tab
```
Jaifroid, line 2230: Expand comment to add the necontainer functionality

www/js/app.js
```diff
@@ -2490,6 +2545,10 @@ function handleMessageChannelMessage (event) {
                     window.timeout = setTimeout(function () {
                         uiUtil.spinnerDisplay(false);
                     }, 1000);
+                    if (/\bhtml/i.test(mimetype)) {
```
Jaifroid, line 2548: Add xhtml and explain problem with MIME type format

www/js/app.js
```diff
@@ -2820,30 +2879,42 @@ function displayArticleContentInIframe (dirEntry, htmlArticle) {
...
                 } else {
                     // It's a relative URL, so we need to calculate the full ZIM URL
-                    zimUrl = uiUtil.deriveZimUrlFromRelativeUrl(uriComponent, baseUrl);
+                    zimUrl = uiUtil.deriveZimUrlFromRelativeUrl(uriComponent, appstate.baseUrl);
                 }
                 goToArticle(zimUrl, downloadAttrValue, contentType);
             });
         });
+        // The popover feature requires as a minimum that the browser supports the css matches function
+        // (having this condition prevents very erratic popover placement in IE11, for example, so the feature is disabled)
```
Jaifroid, line 2895: Explain why code is repeated here

www/js/lib/uiUtil.js
```diff
@@ -1003,6 +1007,389 @@ function warnAndOpenExternalLinkInNewTab (event, clickedAnchor, archive) {
...
+ * A function to attach the tooltip CSS for popovers (NB this does not attach the box itself, only the CSS)
+ * @param {Document} doc The document to which to attach the blloon.css styelesheet
+ * @param {Boolean} dark An optional parameter to adjust the background colour for dark themes (generally not needed for inversion-based themes)
+ */
+function attachKiwixPopoverCss (doc, dark) {
+    const colour = dark ? '#darkgray' : '#black';
+    const backgroundColour = dark ? '#111' : '#ebf4fb';
+    const cssLink = document.createElement('link');
+    doc.head.appendChild(cssLink);
+    replaceCSSLinkWithInlineCSS(cssLink, `
```
Jaifroid, line 1122: Explain that FFOS blocks loading stylesheets into DOM content even if it is same origin

www/js/lib/uiUtil.js
```diff
@@ -1003,6 +1007,389 @@ function warnAndOpenExternalLinkInNewTab (event, clickedAnchor, archive) {
...
+ *
+ * @param {Element} anchor The anchor which launched the popover
+ * @param {Element} popover The containing element of the popover (div)
+ * @param {Document} doc The doucment on which to operate
+ */
+function addEventListenersToPopoverIcons (anchor, popover, doc) {
+    const breakout = function (e) {
+        e.preventDefault();
+        e.stopPropagation();
+        anchor.newcontainer = true;
```
Jaifroid, line 1335: Add comment to explain how this works

audiodude:
image

Jaifroid:
> image

NIce one @audiodude! For now, I'm still working through things and fixing stuff, but I'll be very please to request a review when it's ready! Still some rough stuff...

www/js/lib/uiUtil.js
```diff
@@ -1003,6 +1007,405 @@ function warnAndOpenExternalLinkInNewTab (event, clickedAnchor, archive) {
     });
 }
 
+/**
+ * Parses a linked article in a loaded document in order to extract the first main paragraph (the 'lede') and first
+ * main image (if any). This function currently only parses Wikimedia articles. It returns an HTML string, formatted
+ * for display in a popover
+ *
```
Jaifroid, line 1014: Remove blank line

www/js/lib/uiUtil.js
```diff
@@ -1003,6 +1007,405 @@ function warnAndOpenExternalLinkInNewTab (event, clickedAnchor, archive) {
...
+            link.articleloading = false;
+            link.dataset.touchevoked = false;
+            link.popoverisloading = false;
+        });
+    }, 600);
+}
+
+/**
+ * Adds event listeners to the popover's control icons
+ *
```
Jaifroid, line 1341: Remove blank line

www/js/lib/uiUtil.js
```diff
@@ -1003,6 +1007,405 @@ function warnAndOpenExternalLinkInNewTab (event, clickedAnchor, archive) {
...
+        e.preventDefault();
+        e.stopPropagation();
+        closePopover(popover);
+    }, true);
+    breakoutIcon.addEventListener(eventName, breakout, true);
+}
+
+/**
+ * Remove any preview popover DIVs
+ *
```
Jaifroid, line 1374: Remove blank line

www/js/lib/uiUtil.js
```diff
@@ -1003,6 +1007,405 @@ function warnAndOpenExternalLinkInNewTab (event, clickedAnchor, archive) {
...
+                } else {
+                    timeoutID = setTimeout(fadeOutDiv, 250);
+                }
+            };
+            timeoutID = setTimeout(fadeOutDiv, 0);
+        });
+    }, 400);
+}
+
+// Directly close a popover div
```
Jaifroid, line 1399: Use proper JSDoc

Jaifroid:
Pointerdown fallback is working on Edge Legacy at least, but user has to dismiss the popover with the X icon if evoked by touch rather than by hover. That seems like an acceptable compromise for such an old browser. However, it will be important to prevent popup from loading if a click event is registered, as currently with pointerdown fallback, the browser opens the popover and simultaneously navigates to the requested page if user completes a click (rather than long press for example).

Jaifroid:
Now working in Safari on iOS, but user has to lift finger after a long-press for the popover to show. Apparently iPad users are used to this weird behaviour, so it's OK... ㄟ( ▔, ▔ )ㄏ

Jaifroid:
On Safari for macOS, the popover loads fine when hovering a ZIM link, but it does not remain open when the user moves the mouse over the popover (to select a link in it, for example). However, this works in the PWA on the same browser, so I need to debug the difference between the two implementations. On other browsers, the popover does remain open while hovering the popover div... Sigh.

www/js/app.js
```diff
@@ -2303,6 +2318,111 @@ function articleLoadedSW (iframeArticleContent) {
...
+                // Check if a popover div is currently being hovered
+                const divs = iframeDoc.getElementsByClassName('kiwixtooltip');
+                let divIsHovered = false;
+                Array.prototype.slice.call(divs).forEach(function (div) {
+                    if (div.matches(':hover')) divIsHovered = true;
+                });
+                // Only add a popover to the link if a current popover is not being hovered (prevents popovers showing for links in a popover)
+                if (!divIsHovered) {
+                    // Prevent text selection while popover is open
+                    a.style.userSelect = 'none';
```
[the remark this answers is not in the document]
Jaifroid: Actually, turns out they are.

Jaifroid:
@audiodude I'm not sure if this PR will make much sense outside of the convoluted context of Kiwix JS, but you did kindly offer to take a look! Some considerations / background:

* The main things to focus on are the functions `attachPopoverTriggerEvents` and `handlePopoverEvents` in www/js/app.js, and `getArticleLede`, `populateKiwixPopoverDiv` and `createNewkiwixPopoverContainer` from www/js/lib/uiUtil.js. Most other changes are either incidental variable changes/renaming and a few other utility functions for closing stuff and cleaning up.
* Be aware that this Repo supports pretty much every browser since IE11, given target users often have very old devices/OSs, and this means lots of little quirks and tweaks and fallbacks, worked out empirically with manual testing... i.e., a royal pain in the proverbial. For example, Firefox OS (the app's original target!) doesn't support adding stylesheets to the DOM in an iframe even if same origin, so instead I have to add a style element with the style content. Loads of other stuff like that even for some modern browsers.
* Historic coding target was ES5 with JQuery and Common-JS modules. I've ripped out almost all JQuery, converted the codebase to ES6 modules, and we now use rollup and core-js as bundlers / polyfillers, but they don't cover everything.
* Because of historic coding target, there are no classes or other more complex structures, and the original coding style was object-oriented as opposed to functional. Constructors are functions rather than classes, etc. We can use async functions thanks to rollup, but a lot is originally coded with Promises, even callbacks. I tend to use Promises liberally because I got used to them...
* The app runs in two different content injection modes, one for browsers that support Service Workers, and a mode originally called "JQuery mode" now renamed "Safe Mode" in the UI, but you'll see references to jquery mode in code. The popovers are simplified in this mode: only textContent and no image.
* We have e2e tests of most old platforms via BrowserStack. I added a UI test for these popovers in tests/e2e/spec/legacy-ray_charles.e2e.spec.js. It tests tabbing into a link (hovering is way too complicated to test in WebDriver) and then tests that a popover was displayed.
* Unit tests are........... very basic and use a really old QUnit. They run in browser context rather than in Node context....... (I know)..... And basically need a major upgrade. So I haven't (shock horror) written Unit tests for this PR. Testing relies entirely on the e2e UI test. That's not satisfactory, but until I've done the work to upgrade Unit tests to Mocha, it would be pointless to try to write such tests now.

Anyway, I hope this context is useful! Please don't feel you have to spend much time on this. It's a large new feature arising from the Hackathon...

www/js/app.js
```diff
@@ -2303,6 +2318,112 @@ function articleLoadedSW (iframeArticleContent) {
...
+                if (!divIsHovered) {
+                    // Prevent text selection while popover is open in modern browsers
+                    a.style.userSelect = 'none';
+                    // Resolve the true app theme
+                    const isDarkTheme = uiUtil.isDarkTheme(params.appTheme);
+                    // Get and populate the popover corresponding to the hovered or focused link
+                    uiUtil.populateKiwixPopoverDiv(event, a, appstate.baseUrl, isDarkTheme, selectedArchive);
+                }
+                const outHandler = function (e) {
+                    console.debug('outHandler', e.type);
```
Jaifroid, line 2395: Comment out or remove before merge.

www/js/lib/uiUtil.js
```diff
@@ -1003,6 +1011,428 @@ function warnAndOpenExternalLinkInNewTab (event, clickedAnchor, archive) {
...
+    const uriComponent = removeUrlParameters(href);
+    const zimURL = deriveZimUrlFromRelativeUrl(uriComponent, baseUrl);
+    console.debug('Previewing ' + zimURL);
+    return archive.getDirEntryByPath(zimURL).then(function (dirEntry) {
+        const readArticle = function (dirEntry) {
+            return new Promise((resolve, reject) => {
+                archive.readUtf8File(dirEntry, function (fileDirEntry, htmlArticle) {
+                    const parser = new DOMParser();
+                    const doc = parser.parseFromString(htmlArticle, 'text/html');
+                    // const articleBody = doc.getElementById('mw-content-text');
```
Jaifroid, line 1034: Remove comment.

www/js/lib/uiUtil.js
```diff
@@ -1003,6 +1011,428 @@ function warnAndOpenExternalLinkInNewTab (event, clickedAnchor, archive) {
...
+ * @param {String} href The href of the article link from which to extract the lede
+ * @param {String} baseUrl The base URL of the currently loaded article
+ * @param {Document} articleDocument The DOM of the currently loaded article
+ * @param {ZIMArchive} archive The archive from which to extract the lede
+ * @returns {Promise<String>} A Promise for the linked article's lede HTML including first main image URL if any
+ */
+function getArticleLede (href, baseUrl, articleDocument, archive) {
+    const uriComponent = removeUrlParameters(href);
+    const zimURL = deriveZimUrlFromRelativeUrl(uriComponent, baseUrl);
+    console.debug('Previewing ' + zimURL);
```
Jaifroid, line 1027: Add comment to explain what is happening here.

www/js/lib/uiUtil.js
```diff
@@ -1003,6 +1011,428 @@ function warnAndOpenExternalLinkInNewTab (event, clickedAnchor, archive) {
...
+                                }
+                                // Get the transformed HTML. Note that in Safe mode, we risk breaking the UI if user clicks on an
+                                // embedded link, so only use innerText in that case
+                                const content = params.contentInjectionMode === 'jquery' ? plainText
+                                    : nonEmptyParagraphs[i].innerHTML;
+                                balloonString += '<p>' + content + '</p>';
+                                // If we have enough characters to fill the box, break
+                                if (cumulativeCharCount >= 850) break;
+                            }
+                        }
```
Jaifroid, line 1079: Add comment to explain what we are doing now.

www/js/lib/uiUtil.js
```diff
@@ -1003,6 +1011,428 @@ function warnAndOpenExternalLinkInNewTab (event, clickedAnchor, archive) {
...
+                                }
+                            }
+                        }
+                        if (firstImage) {
+                            // Calculate root relative URL of image
+                            const imageZimURL = encodeURI(deriveZimUrlFromRelativeUrl(firstImage.getAttribute('src'), balloonBaseURL));
+                            firstImage.src = rootRelativePathPrefix + imageZimURL;
+                            balloonString = firstImage.outerHTML + balloonString;
+                        }
+                        // console.debug(balloonString);
```
Jaifroid, line 1099: Remove

www/js/lib/uiUtil.js
```diff
@@ -1003,6 +1011,428 @@ function warnAndOpenExternalLinkInNewTab (event, clickedAnchor, archive) {
...
+                        }
+                    } else {
+                        reject(new Error('No article body found'));
+                    }
+                });
+            });
+        }
+        if (!dirEntry) {
+            return Promise.reject(new Error('No directory entry found'));
+        } else if (dirEntry.redirect) {
```
Jaifroid, line 1113: Add comment

www/js/lib/uiUtil.js
```diff
@@ -1003,6 +1011,428 @@ function warnAndOpenExternalLinkInNewTab (event, clickedAnchor, archive) {
...
+      appstate.expectedArticleURLToBeDisplayed === 'A/Wikivoyage:Offline_reader_Expedition/Home_page')) {
+        return Promise.resolve();
+    }
+    link.popoverisloading = true;
+    // Do not display a popover if one is already showing for the current link
+    const kiwixPopover = ev.target.ownerDocument.querySelector('.kiwixtooltip');
+    if (kiwixPopover && kiwixPopover.dataset.href === linkHref) return Promise.resolve();
+    // console.debug('Attaching popover...');
+    const currentDocument = ev.target.ownerDocument;
+    const articleWindow = currentDocument.defaultView;
```
Jaifroid, line 1224: Add comment to explain why we remove any popovers first

www/js/lib/uiUtil.js
```diff
@@ -1003,6 +1011,428 @@ function warnAndOpenExternalLinkInNewTab (event, clickedAnchor, archive) {
...
+    removeKiwixPopoverDivs(currentDocument);
+    setTimeout(function () {
+        // Check if the user has moved away from the link or has clicked it, and abort display of popover if so
+        if (link.articleisloading || !link.matches(':hover') && !link.touched && currentDocument.activeElement !== link) {
+            console.debug('Aborting popover display for ' + linkHref + ' because user has moved away from link or clicked it');
```
Jaifroid, line 1229: Comment out or remove debug statement.

www/js/lib/uiUtil.js
```diff
@@ -1003,6 +1011,428 @@ function warnAndOpenExternalLinkInNewTab (event, clickedAnchor, archive) {
...
+            link.popoverisloading = false;
+            return Promise.resolve();
+        }
+        // Create a new Kiwix popover container
+        const divWithArrow = createNewKiwixPopoverCointainer(articleWindow, link, ev);
+        const div = divWithArrow.div;
+        const span = divWithArrow.span;
```
Jaifroid, line 1236: Add comment to explain what happens now

www/js/lib/uiUtil.js
```diff
@@ -1003,6 +1011,428 @@ function warnAndOpenExternalLinkInNewTab (event, clickedAnchor, archive) {
...
+    // If we're less than margin to the left, shift it to margin px from left
+    if (divRectX < margin) {
+        triangleX += divRectX;
+        divRectX = margin;
+        triangleX -= divRectX;
+    }
+    // Adjust triangleX if necessary
+    if (triangleX < 10) triangleX = 10;
+    if (triangleX > divWidth - 10) triangleX = divWidth - 10;
+    // Now set the calculated x and y positions, taking into account the zoom factor
```
Jaifroid, line 1345: Remove reference to zoom factor (only relevant in PWA)

www/js/lib/uiUtil.js
```diff
@@ -1003,6 +1011,428 @@ function warnAndOpenExternalLinkInNewTab (event, clickedAnchor, archive) {
...
+    // Check if we have restricted screen height
+    const divHeight = screenHeight < 512 ? 160 : 256;
+    div.style.width = divWidth + 'px';
+    div.style.height = divHeight + 'px';
+    div.style.display = 'flex';
+    div.style.justifyContent = 'center';
+    div.style.alignItems = 'center';
+    div.className = 'kiwixtooltip';
+    div.innerHTML = '<p>Loading ...</p>';
+    div.dataset.href = linkHref;
```
Jaifroid, line 1303: Explain why we have to attach the div to the article's document in order to calculate accurate positioning.

Jaifroid:
Implementations of this branch for in-browser testing:

* Unbundled (only works in browsers supporting ES6 modules natively); https://kiwix.github.io/kiwix-js/www/index.html
* Bundled: https://kiwix.github.io/kiwix-js/dist/www/index.html

audiodude:
Okay! Let me dive and and take a look at this tomorrow (Thursday)!

www/js/init.js
```diff
@@ -74,6 +75,14 @@
  */
 var params = {};
 
+/**
+ * A global object for storing app state
+ *
+ * @type Object
+ */
+// eslint-disable-next-line no-unused-vars
+var appstate = {};
```
audiodude, line 84: Looks like this isn't used, but you've acknowledged that with the eslint comment. Should it be removed?

audiodude: edit: Okay now I see that it was removed from app.js. Why?

www/js/app.js
```diff
@@ -2303,6 +2318,112 @@ function articleLoadedSW (iframeArticleContent) {
...
+    // (having this condition prevents very erratic popover placement in IE11, for example, so the feature is disabled for such browsers)
+    if (appstate.wikimediaZimLoaded && params.showPopoverPreviews && 'matches' in Element.prototype) {
+        const iframeDoc = win.document;
+        if (iframeDoc) {
+            // Attach the popover CSS to the current article document
+            uiUtil.attachKiwixPopoverCss(iframeDoc);
+            // Add event listeners to iframe window to check for links
+            win.addEventListener('mouseover', handlePopoverEvents, true);
+            win.addEventListener('focus', handlePopoverEvents, true);
+            // Conditionally add event listeners to support touch events with fallback to pointer events
```
audiodude, line 2336: Aren't these proxies for "click" events? Are we showing the popover when someone clicks or only when they hover/focus?

www/js/app.js
```diff
@@ -2303,6 +2318,112 @@ function articleLoadedSW (iframeArticleContent) {
...
+/**
+ * Attaches popover trigger events to the given window
+ * @param {Window} win The window to which to attach popover trigger events
+ */
+function attachPopoverTriggerEvents (win) {
+    // The popover feature requires as a minimum that the browser supports the css matches function
+    // (having this condition prevents very erratic popover placement in IE11, for example, so the feature is disabled for such browsers)
+    if (appstate.wikimediaZimLoaded && params.showPopoverPreviews && 'matches' in Element.prototype) {
+        const iframeDoc = win.document;
+        if (iframeDoc) {
```
audiodude, line 2330: nit: prefer `if (!iframedoc) return;`

In fact you have an if nested directly under another if. Consider:

```
const iframeDoc = win.document;
if (!iframeDoc || !appstate.wikimediaZimLoaded || !params.showPopoverPreviews  || !'matches' in Element.prototype) {
  return;
}
```

www/js/app.js
```diff
@@ -2303,6 +2318,112 @@ function articleLoadedSW (iframeArticleContent) {
...
+
+/**
+ * Event handler for attaching preview popovers
+ * @param {Event} event The event produced by the mouseover or focus action
+ */
+function handlePopoverEvents (event) {
+    // Check if the hovered or focused element or its parent is a link (but throttle its activation)
+    if (window.popoverThrottle) return;
+    window.popoverThrottle = true;
+    setTimeout(function () {
```
audiodude, line 2354: nit: Somewhat hard to read a giant anonymous function inside of a setTimeout. Probably better to:

1. Refactor the whole function into a named `onPopover` function oustdie of `handlePopoverEvents`, or
2. At the least, do `function onPopover() {` and then `setTimeout(onPopover, 10)`, within this function

www/js/app.js
```diff
@@ -2303,6 +2318,112 @@ function articleLoadedSW (iframeArticleContent) {
...
+}
+
+/**
+ * Event handler for attaching preview popovers
+ * @param {Event} event The event produced by the mouseover or focus action
+ */
+function handlePopoverEvents (event) {
+    // Check if the hovered or focused element or its parent is a link (but throttle its activation)
+    if (window.popoverThrottle) return;
+    window.popoverThrottle = true;
```
audiodude, line 2353: Global variables, ew. How are you tracking these "scattered" globals versus things like `appState` which are more explicitly declared?

www/js/app.js
```diff
@@ -2303,6 +2318,112 @@ function articleLoadedSW (iframeArticleContent) {
...
+    setTimeout(function () {
+        let a = event.target;
+        const iframeDoc = a.ownerDocument;
+        if (iframeDoc) {
+            const iframeWindow = iframeDoc.defaultView;
+            while (a && a !== iframeWindow && a.nodeName !== 'A') {
+                a = a.parentNode;
+            }
+            // If a link was hovered or focused, process it
+            if (a && a.nodeName === 'A') {
```
audiodude, line 2363: nit: same thing, reverse the condition so you can return early and outdent.

if (!a || a.nodeName !== 'A') return;

audiodude: You should be sure that `a.nodeName === 'A'` at this point right? it was a condition for breaking out of the `while` directly above.

www/js/app.js
```diff
@@ -2303,6 +2318,112 @@ function articleLoadedSW (iframeArticleContent) {
...
+/**
+ * Event handler for attaching preview popovers
+ * @param {Event} event The event produced by the mouseover or focus action
+ */
+function handlePopoverEvents (event) {
+    // Check if the hovered or focused element or its parent is a link (but throttle its activation)
+    if (window.popoverThrottle) return;
+    window.popoverThrottle = true;
+    setTimeout(function () {
+        let a = event.target;
```
audiodude, line 2355: nit: took me a second, but I realize that this is a as i "anchor". Might be better to spell out `anchor` for readability?

www/js/app.js
```diff
@@ -2303,6 +2318,112 @@ function articleLoadedSW (iframeArticleContent) {
...
+                }
+                if (a.style.userSelect === undefined) {
+                    // This prevents selection of the text in a touched link in Safari for iOS and Edge Legacy / UWP
+                    a.style.webkitUserSelect = 'none';
+                    a.style.msUserSelect = 'none';
+                }
+                // Check if a popover div is currently being hovered
+                const divs = iframeDoc.getElementsByClassName('kiwixtooltip');
+                let divIsHovered = false;
+                Array.prototype.slice.call(divs).forEach(function (div) {
```
audiodude, line 2382: Why do you have to invoke `Array.prototype.slice` statically? Is this for pollyfills? You can't just do `divs.forEach()`?

www/js/app.js
```diff
@@ -2303,6 +2318,112 @@ function articleLoadedSW (iframeArticleContent) {
...
+                    if (div.matches(':hover')) divIsHovered = true;
```
audiodude, line 2383: Not sure of your polyfill situation, can you use Array.some?

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some

So it would be `const divIsHoevered = divs.some(() => div.matches(':hover'))` in modern, say, Typescript

www/js/app.js
```diff
@@ -2303,6 +2318,112 @@ function articleLoadedSW (iframeArticleContent) {
...
+                if (!divIsHovered) {
+                    // Prevent text selection while popover is open in modern browsers
+                    a.style.userSelect = 'none';
+                    // Resolve the true app theme
+                    const isDarkTheme = uiUtil.isDarkTheme(params.appTheme);
+                    // Get and populate the popover corresponding to the hovered or focused link
+                    uiUtil.populateKiwixPopoverDiv(event, a, appstate.baseUrl, isDarkTheme, selectedArchive);
+                }
+                const outHandler = function (e) {
+                    // console.debug('outHandler', e.type);
```
audiodude, line 2395: nit: remove commented console log

www/js/app.js
```diff
@@ -2303,6 +2318,112 @@ function articleLoadedSW (iframeArticleContent) {
...
+                        a.style.msUserSelect = 'auto';
+                        a.style.userSelect = 'auto';
+                        a.removeEventListener(e.type, outHandler);
+                        a.removeEventListener('contextmenu', suppressContextMenuHandler, true);
+                    }, 250);
+                };
+                if (/mouseover/.test(event.type)) {
+                    a.addEventListener('mouseleave', outHandler);
+                }
+                if (/pointerdown/.test(event.type)) {
```
audiodude, line 2412: Are these meant to be non-mutually exclusive? That is, not using `else if`? Just to be clear if both of them are true we want to do both actions and attach multiple event listeners?

audiodude (commented):
Okay I've reviewed app.js. I don't have time right now to review uiutil.js, but I'll get to it It'll probably just be another day (sorry!).

I think this looks reasonable considering the limitations you're working with. Just a few comments.

www/js/app.js
Jaifroid: Not really. There is some tricky event handling in this feature, due to the need to support: 1. Mouse-based devices; 2. Touchscreen laptops and mobile; 3. Keyboard. To be clear:

1. These event listeners are added to the article's window (not to each link individually) - the code looks for the hovered / touched element and only executes if it is an Anchor;
2. `mouseover` (hovering over a link without clicking it), `touchstart` and `pointerdown` will all show the popover if the user hovers / touches / holds mouse down for longer than 600ms (this is a timeout defined in `uiUtil.js`). We only listen to `pointerdown` if `touchstart` is not available in the browser (e.g. the UWP framework, which uses Edge Legacy, supports touch, but not Touch events, only Pointer events);
3. `focus` supports keyboard users, who may tab into a link, but have not "clicked" it (usually achieved by using spacebar after a link is focused). Note that hovering does not focus an element, and `touchstart` / `pointerdown` occur before focus in all browsers. In some only the touch events will fire and not focus. In others, both events will fire, but because touch/pointer occurs first, the second `focus` activation gets cancelled due to flags set.

www/js/app.js
Jaifroid: Thanks, will check.

www/js/app.js
Jaifroid: This was just me trying to be concise. I'll get rid of them here.

www/js/app.js
Jaifroid: Old habit. I can just use `Array.from()` because we have that polyfilled. Before the polyfill, we always used this formula in order to support older browsers that cannot do `forEach` over a Node list. So it became second nature... Will simplify.

www/js/app.js
Jaifroid: As mentioned above, we have `Array.from` polyfilled. It would be interesting to see if core-js picks up Array.some -- will try, but I suspect I'll be limited to Array.from.

www/js/app.js
Jaifroid: Just want to evoke the cleanup routine, so if some or any of them have been defined, their "opposite" will be used to clean up.

Jaifroid:
Thanks for those useful comments 😊. I'll deal with those before you get to `uiUtil.js`.

www/js/app.js

www/js/app.js
Jaifroid: Turns out Array.some() has wide support without polyfills, so I've incorporated it, thanks!

www/js/app.js
Jaifroid: It's `pointerdown` and `touchstart` that are mutually exclusive, but I don't think adding and `else` makes any difference, as if one of these doesn't exist its opposite (`pointerup' ,  'touchend' simply never gets attached). I think of this code as a broom sweeping away anything it finds.... 🧹

www/js/app.js
```diff
@@ -2303,6 +2325,116 @@ function articleLoadedSW (iframeArticleContent) {
...
+function attachPopoverTriggerEvents (win) {
+    const iframeDoc = win.document;
+    // The popover feature requires as a minimum that the browser supports the css matches function
+    // (having this condition prevents very erratic popover placement in IE11, for example, so the feature is disabled for such browsers)
+    if (!iframeDoc || !appstate.wikimediaZimLoaded || !params.showPopoverPreviews || !('matches' in Element.prototype)) {
+        return;
+    }
+    // Attach the popover CSS to the current article document
+    uiUtil.attachKiwixPopoverCss(iframeDoc);
+    // Add event listeners to iframe window to check for links
```
Jaifroid, line 2341:
```suggestion
    // Add event listeners to the iframe window to check when anchors are hovered, focused or touched
```

www/js/app.js
```diff
@@ -2276,13 +2298,13 @@ function articleLoadedSW (iframeArticleContent) {
...
+    var iframeWindow = iframeArticleContent.contentWindow;
+    if (iframeWindow) {
         // Configure home key press to focus #prefix only if the feature is in active state
-        if (params.useHomeKeyToFocusSearchBar) { iframeArticleContent.contentWindow.onkeydown = focusPrefixOnHomeKey; }
-        if (params.openExternalLinksInNewTabs) {
-            // Add event listener to iframe window to check for links to external resources
-            iframeArticleContent.contentWindow.onclick = filterClickEvent;
-        }
+        if (params.useHomeKeyToFocusSearchBar) { iframeWindow.onkeydown = focusPrefixOnHomeKey; }
+        // Add event listener to iframe window to check for links to external resources
```
Jaifroid, line 2305:
```suggestion
        // Add event listeners to iframe window to check for links to external resources and for actions that trigger popovers
```

www/js/app.js
```diff
@@ -2062,15 +2067,15 @@ function findDirEntryFromDirEntryIdAndLaunchArticleRead (dirEntryId) {
 }
 
 /**
- * Check whether the given URL from given dirEntry equals the expectedArticleURLToBeDisplayed
+ * Check whether the given URL from given dirEntry equals the appstate.expectedArticleURLToBeDisplayed
```
Jaifroid, line 2070:
```suggestion
 * Check whether the given URL from given dirEntry matches the expected article
```

www/js/app.js
```diff
@@ -2303,6 +2325,116 @@ function articleLoadedSW (iframeArticleContent) {
...
+    win.addEventListener('focus', evokePopoverEvents, true);
+    // Conditionally add event listeners to support touch events with fallback to pointer events
+    if (window.navigator.maxTouchPoints > 0) {
+        win.addEventListener('touchstart', evokePopoverEvents, true);
+    } else {
+        win.addEventListener('pointerdown', evokePopoverEvents, true);
+    }
+}
+
+// Throttle the popover event handler to prevent multiple activations with mouse movement
```
Jaifroid, line 2352:
```suggestion
// Throttle for the popover event handler to prevent multiple activations with mouse movement
```

www/js/app.js
```diff
@@ -2303,6 +2325,116 @@ function articleLoadedSW (iframeArticleContent) {
...
+    let anchor = ev.target;
+    const iframeDoc = anchor.ownerDocument;
+    if (!iframeDoc) return;
+    const iframeWindow = iframeDoc.defaultView;
+    while (anchor && anchor !== iframeWindow && anchor.nodeName !== 'A') {
+        anchor = anchor.parentNode;
+    }
+    // If we're not hovering a link, then we can exit
+    if (!anchor || anchor.nodeName !== 'A') return;
+    // console.debug(event.type, event.target, a);
```
Jaifroid, line 2383: I'd like to keep this in the code (commented out) for ease of future debugging.

www/js/app.js
```diff
@@ -2303,6 +2325,116 @@ function articleLoadedSW (iframeArticleContent) {
...
+                uiUtil.removeKiwixPopoverDivs(iframeDoc);
+                anchor.touched = false;
+            }
+            anchor.style.webkitUserSelect = 'auto';
+            anchor.style.msUserSelect = 'auto';
+            anchor.style.userSelect = 'auto';
+            anchor.removeEventListener(e.type, outHandler);
+            anchor.removeEventListener('contextmenu', suppressContextMenuHandler, true);
+        }, 250);
+    };
```
Jaifroid, line 2423:
```suggestion
    };
    // Clean up when user stops hovering, lifts pointer, stops touching, or unfocuses (blurs) the link
```

Jaifroid (commented):
Cleanup self-review

Jaifroid:
@audiodude All done, should be ready now for you to move on to `uiUtil.js` when/if you have time. I've left a few comments open above for you to review, and I think everything else you suggested is resolved.

www/js/lib/uiUtil.js
```diff
@@ -1003,6 +1011,436 @@ function warnAndOpenExternalLinkInNewTab (event, clickedAnchor, archive) {
...
+    });
+}
+
+/**
+ * A function to attach the tooltip CSS for popovers (NB this does not attach the box itself, only the CSS)
+ * @param {Document} doc The document to which to attach the popover stylesheet
+ * @param {Boolean} dark An optional parameter to adjust the background colour for dark themes (generally not needed for inversion-based themes)
+ */
+function attachKiwixPopoverCss (doc, dark) {
+    const colour = dark ? '#darkgray' : '#black';
```
Jaifroid, line 1139:
```suggestion
    const colour = dark ? 'darkgray' : 'black';
```
Oops... Code wasn't tested because not used with current dark style, but will be used if/when we add `darkReader` to this Repo. See #908.

www/js/lib/uiUtil.js
```diff
@@ -1003,6 +1011,436 @@ function warnAndOpenExternalLinkInNewTab (event, clickedAnchor, archive) {
...
+    }
+    link.popoverisloading = true;
+    // Do not display a popover if one is already showing for the current link
+    const kiwixPopover = ev.target.ownerDocument.querySelector('.kiwixtooltip');
+    if (kiwixPopover && kiwixPopover.dataset.href === linkHref) return Promise.resolve();
+    // console.debug('Attaching popover...');
+    const currentDocument = ev.target.ownerDocument;
+    const articleWindow = currentDocument.defaultView;
+    // Remove any existing popover(s) that the user may not have closed before creating a new one
+    removeKiwixPopoverDivs(currentDocument);
```
Jaifroid, line 1230:
```suggestion
    removeKiwixPopoverDivs(currentDocument);
    // Timeout below ensures that popovers are not loaded if a user is simply moving their mouse around on a page
    // without hovering. It provides a 600ms pause before app begins the process of binary search and decompression
```

www/js/lib/uiUtil.js
```diff
@@ -1003,6 +1011,436 @@ function warnAndOpenExternalLinkInNewTab (event, clickedAnchor, archive) {
...
+    }, true);
+    breakoutIcon.addEventListener('mousedown', breakout, true);
+    // Additionally register either pointerdown or touchstart if supported for faster response
+    const eventName = window.PointerEvent ? 'pointerdown' : 'touchstart';
+    closeIcon.addEventListener(eventName, function (e) {
+        e.preventDefault();
+        e.stopPropagation();
+        closePopover(popover);
+    }, true);
+    breakoutIcon.addEventListener(eventName, breakout, true);
```
Jaifroid, line 1402: Code may not be necessary if `mousedown` is universally supported. Can't remember why I added it -- I think I was having trouble getting it to work at some point, but logically shouldn't be needed. Will test.

www/js/app.js
audiodude: Sure np

www/js/app.js
audiodude: Ah okay I missed that they were being added to the window. Sounds good.

www/js/lib/uiUtil.js
audiodude, line 1: Okay so this might be a "spicy" take. I've generally come to the opinion that a "util.js" or "util/" directory is an antipattern. It's a symptom of "I don't know what to name this" and "I don't know where to put this", but it hurts readability and code discoverability.

Here we have a collection of 4 or 5 functiosn that, collectively, allow you to create, destroy, add listeners, and remove listeners for popover divs. Why not `popovers.js`?

In the same vein, why does getting the article lede (`getArticleLede`) have anything to do with the "UI", much less something as generic as `uitUtil.js`? Wouldn't it make sense, at the least, to move it to `popovers.js`, perhaps with the logic that this is the only place it's used, or even better, `zimfile.js` because it involves extracting data from the ZIM?

I'll let you decide if you want to make any of these changes before reviewing further.

audiodude (commented):
Just one comment, about file structure/code location. I don't want to review more until you resolve because it would change where my comments go and I would get the stupid github "outdate" nonsense.

www/js/lib/uiUtil.js
Jaifroid: Actually that does make a lot of sense. I put it all in `uiUtil.js` because I thought of it all as being related to the UI (and I thought it was only going to be one function at first, t.b.h., until I went down a rabbit hole). But you're right that a dedicated `popovers.js` would be much more discoverable / readable / in the spirit of modularity. And it will be easy to do. I'll resolve and let you know.

www/js/lib/popovers.js
```diff
@@ -0,0 +1,455 @@
...
+ */
+function getArticleLede (href, baseUrl, articleDocument, archive) {
+    const uriComponent = uiUtil.removeUrlParameters(href);
+    const zimURL = uiUtil.deriveZimUrlFromRelativeUrl(uriComponent, baseUrl);
+    console.debug('Previewing ' + zimURL);
+    // Do a binary search in the URL index to get the directory entry for the requested article
+    return archive.getDirEntryByPath(zimURL).then(function (dirEntry) {
+        const readArticle = function (dirEntry) {
+            // Wrap legacy callback-based code in a Promise
+            return new Promise((resolve, reject) => {
```
audiodude, line 47: You could probably do the same thing here:

```
function readArticleInner(resolve, reject) {
  ...
}

return new Promise(readArticleInner)
```

I feel like strategies like this help a lot when dealing with interwoven promise and callback code.

www/js/lib/popovers.js
```diff
@@ -0,0 +1,455 @@
...
+                    const doc = parser.parseFromString(htmlArticle, 'text/html');
+                    const articleBody = doc.body;
+                    if (articleBody) {
+                        // Establish the popup balloon's base URL and the absolute path for calculating the ZIM URL of links and images
+                        const balloonBaseURL = encodeURI(fileDirEntry.namespace + '/' + fileDirEntry.url.replace(/[^/]+$/, ''));
+                        const docUrl = new URL(articleDocument.location.href);
+                        const rootRelativePathPrefix = docUrl.pathname.replace(/([^.]\.zim\w?\w?\/).+$/i, '$1');
+                        let balloonString = '';
+                        // Remove all standalone style elements, because their content is shown by both innerText and textContent
+                        const styleElements = Array.from(articleBody.querySelectorAll('style'));
```
audiodude, line 60: Perhaps you could extract this into a `removeAllStyleElementsFromBody` function? Then getArticleLede would be a lot shorter and easier to read.

www/js/lib/uiUtil.js
Jaifroid: Done.

www/js/lib/uiUtil.js
Jaifroid: Done.

www/js/lib/popovers.js
```diff
@@ -0,0 +1,455 @@
...
+                            return !/^\s*$/.test(text) && text.length >= 50;
+                        });
+                        if (nonEmptyParagraphs.length > 0) {
+                            let cumulativeCharCount = 0;
+                            // Add enough paras to complete the word count
+                            for (let i = 0; i < nonEmptyParagraphs.length; i++) {
+                                // Get the character count: to fill the larger box we need ~850 characters (815 plus leeway)
+                                const plainText = nonEmptyParagraphs[i].innerText;
+                                cumulativeCharCount += plainText.length;
+                                // In ServiceWorker mode, we need to transform the URLs of any links in the paragraph
```
audiodude, line 77: Also a candidate for a helper function, `transformServiceWorkerLinks`.

I've found that nice readability 'hack' is to even define these helpers inline inside the main function, especially if you're worried that they won't be useful at the top level. Something like:

```
function bigComplicatedFunction(data) {
  function preProcessFile(file) {
    // Lots of code
  }

  function rewriteLinks(body, links) {
    // more code ...
   }

   // Main function code
   data.forEach(() => {
      var foo = Bar.thing()
      preProcessFile(foo);
    
     if (foo == 'baz') {
        rewriteLinks(foo, data);
     }
  }
}
```

Jaifroid:
New test deployment as v4.0.9 here:

* (browsers with ES module support) https://kiwix.github.io/kiwix-js/www/index.html
* (bundled but not minified) https://kiwix.github.io/kiwix-js/dist/www/index.html

Ensure it has updated to v4.0.9 (message in Configuration should appear after 15s) and restart or reset if necessary. In Firefox a Reset (Expert Settings) is often needed due to few-hours limit on Service Worker checking.

www/js/lib/popovers.js
```diff
@@ -0,0 +1,455 @@
...
+                                // embedded link, so only use innerText in that case
+                                const content = params.contentInjectionMode === 'jquery' ? plainText
+                                    : nonEmptyParagraphs[i].innerHTML;
+                                balloonString += '<p>' + content + '</p>';
+                                // If we have enough characters to fill the box, break
+                                if (cumulativeCharCount >= 850) break;
+                            }
+                        }
+                        // If we have a lede, we can now add an image to the balloon
+                        const images = articleBody.querySelectorAll('img');
```
audiodude, line 98: Again, an `addImage` helper function would be sweet! 😀

www/js/lib/uiUtil.js
Jaifroid: OK, refactored now with `popovers.js`, just waiting on a re-run test, and I want to do a quick manual test of something on iPad via BrowserStack.

www/js/lib/popovers.js
```diff
@@ -0,0 +1,455 @@
...
+        }
+        if (!dirEntry) {
+            return Promise.reject(new Error('No directory entry found'));
+        } else if (dirEntry.redirect) {
+            // If the dirEntry is a redirect, we need to resolve it before reading the article
+            return new Promise((resolve, reject) => {
+                archive.resolveRedirect(dirEntry, function (reDirEntry) {
+                    resolve(readArticle(reDirEntry));
+                });
+            }).catch(error => {
```
audiodude, line 136: Based on the code above, this promise can't reject right?

www/js/lib/popovers.js
```diff
@@ -0,0 +1,455 @@
...
+                return Promise.reject(error);
```
audiodude, line 137: I might be rusty on Promises, but isn't it redundant to `catch` then just  return a `Promise.reject`? But see comment above anyway.

www/js/lib/popovers.js
```diff
@@ -0,0 +1,455 @@
...
+            });
+        } else {
+            // Directory entry was found, so now read the article data
+            return Promise.resolve(readArticle(dirEntry));
```
audiodude, line 141: Again, might be rusty, but you're already in a function that is called from a `then()` right? So you shouldn't need `Promise.resolve()`, you should just be able to `return readArticle(dirEntry)`?

www/js/lib/popovers.js
```diff
@@ -0,0 +1,455 @@
...
+    });
+}
+
+/**
+ * A function to attach the tooltip CSS for popovers (NB this does not attach the box itself, only the CSS)
+ * @param {Document} doc The document to which to attach the popover stylesheet
+ * @param {Boolean} dark An optional parameter to adjust the background colour for dark themes (generally not needed for inversion-based themes)
+ */
+function attachKiwixPopoverCss (doc, dark) {
+    const colour = dark ? 'darkgray' : 'black';
```
audiodude, line 154: colour....  😂

www/js/lib/popovers.js
```diff
@@ -0,0 +1,455 @@
...
+    // Do not show popover if there is no href or with certain landing pages
+    if (!linkHref || /^wikivoyage/i.test(archive.file.name) &&
+      (appstate.expectedArticleURLToBeDisplayed === archive.landingPageUrl ||
+      appstate.expectedArticleURLToBeDisplayed === 'A/Wikivoyage:Offline_reader_Expedition/Home_page')) {
+        return Promise.resolve();
+    }
+    link.popoverisloading = true;
+    // Do not display a popover if one is already showing for the current link
+    const kiwixPopover = ev.target.ownerDocument.querySelector('.kiwixtooltip');
+    if (kiwixPopover && kiwixPopover.dataset.href === linkHref) return Promise.resolve();
```
audiodude, line 239: Do you have to unset `popoverisloading` here before you return?

www/js/lib/uiUtil.js
Jaifroid: Manual test on Safari for iPad, iPhone and on Edge Legacy (the most problematic browsers/platforms!) are fine. Also tested on Android for good measure. Seems to fix #1255.

Jaifroid: So that's resolved, fingers crossed!

www/js/lib/popovers.js
```diff
@@ -0,0 +1,455 @@
...
+    const currentDocument = ev.target.ownerDocument;
+    const articleWindow = currentDocument.defaultView;
+    // Remove any existing popover(s) that the user may not have closed before creating a new one
+    removeKiwixPopoverDivs(currentDocument);
+    setTimeout(function () {
+        // Check if the user has moved away from the link or has clicked it, and abort display of popover if so
+        if (link.articleisloading || !link.matches(':hover') && !link.touched && currentDocument.activeElement !== link) {
+            // console.debug('Aborting popover display for ' + linkHref + ' because user has moved away from link or clicked it');
+            link.popoverisloading = false;
+            return Promise.resolve();
```
audiodude, line 252: I don't think this does what you think it does. You're return a promise from the anonymous function inside of the setTimeout?

www/js/lib/popovers.js
```diff
@@ -0,0 +1,455 @@
...
+                // Do not close any div which is being hovered
+                if (!div.matches(':hover')) {
+                    closePopover(div);
+                } else {
+                    timeoutID = setTimeout(fadeOutDiv, 250);
+                }
+            };
+            timeoutID = setTimeout(fadeOutDiv, 0);
+        });
+    }, 400);
```
audiodude, line 434: Maybe add comment on why you have to wait 400ms before removing?

www/js/lib/popovers.js
```diff
@@ -0,0 +1,455 @@
...
+ * Closes the specified popover div, with fadeout effect, and removes it from the DOM
+ * @param {Element} div The div to close
+ */
+function closePopover (div) {
+    div.style.opacity = '0';
+    setTimeout(function () {
+        if (div && div.parentElement) {
+            div.parentElement.removeChild(div);
+        }
+    }, 200);
```
audiodude, line 447: Same here, why can't this be done immediately?

audiodude (commented):
Looks good in general. I can see what you mean about the pain of having to deal with Promise and callbacks in the same code base, ick. I've made lots of suggestions about refactoring and extracting local helper functions, I think it will really help readability especially in the aforementioned context.

www/js/lib/popovers.js
```diff
@@ -0,0 +1,455 @@
...
+ * @param {Document} articleDocument The DOM of the currently loaded article
+ * @param {ZIMArchive} archive The archive from which to extract the lede
+ * @returns {Promise<String>} A Promise for the linked article's lede HTML including first main image URL if any
+ */
+function getArticleLede (href, baseUrl, articleDocument, archive) {
+    const uriComponent = uiUtil.removeUrlParameters(href);
+    const zimURL = uiUtil.deriveZimUrlFromRelativeUrl(uriComponent, baseUrl);
+    console.debug('Previewing ' + zimURL);
+    // Do a binary search in the URL index to get the directory entry for the requested article
+    return archive.getDirEntryByPath(zimURL).then(function (dirEntry) {
```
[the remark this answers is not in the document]
Jaifroid: Yes, that's the nightmare of Kiwix JS code! I've kind of got used to the patterns, and now replicate them as I immediately recognize what they're doing and these patterns are all over the codebase. But I do agree, and it provides a big entry hurdle for potential contributors, so I'll look to simplify as you suggest.

www/js/lib/popovers.js
```diff
@@ -0,0 +1,455 @@
...
+            // console.debug('Aborting popover display for ' + linkHref + ' because user has moved away from link or clicked it');
+            link.popoverisloading = false;
+            return Promise.resolve();
+        }
+        // Create a new Kiwix popover container
+        const divWithArrow = createNewKiwixPopoverCointainer(articleWindow, link, ev);
+        const div = divWithArrow.div;
+        const span = divWithArrow.span;
+        // Get the article's 'lede' (first main paragraph or two) and the first main image (if any)
+        return getArticleLede(linkHref, appstate.baseUrl, currentDocument, archive).then(function (html) {
```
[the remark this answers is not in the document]
Jaifroid: Yep, again this was harmless, but misleading. The caller is not expecting a Promise, so we don't need to return one. Removed the return here and the final return.

Jaifroid (commented):
So in the next push to the PR, I'm refactoring the code to address the identified issues. Some explanatory comments below. I'll close conversations before pushing, or else it gets hard to close them even using the GitHub UI. Many thanks, I think the code is more accessible now, and unneeded (if harmless) Promise resolving should have been fixed.

Jaifroid:
New test deployment as v4.0.91 here:

* (browsers with ES module support) https://kiwix.github.io/kiwix-js/www/index.html
* (bundled but not minified) https://kiwix.github.io/kiwix-js/dist/www/index.html

Ensure it has updated to v4.0.91 (message in Configuration should appear after 15s) and restart or reset if necessary. In Firefox a Reset (Expert Settings) is often needed due to few-hours limit on Service Worker checking.

Jaifroid:
Screenshots in light and dark mode:

image

image

Jaifroid:
@audiodude Are you happy with the resolved issues? OK to merge after final testing?

NB future PRs will certainly not be as massive as this! This was the outcome of work at the Hackathon, hence a major feature. It's rare to do major features! At some point there will be a PR (see #1253) to add these popovers to new tabs / windows controlled by the app's Service Worker, but not for this PR which is feature-complete as far as I can manage for now.  Adding to new tabs or windows needs careful thought, as we have to keep track of opened windows, which we currently don't do explicitly (just let the SW do that job, but it knows nothing about the DOM).

audiodude (approved):

audiodude:
Sorry for the delay. LGTM.

Jaifroid:
Last manual tests:

* Works fine in Firefox 65 (tested on Big Sur). Below Firefox 65, WebP images are not supported by Firefox, so the images will not show. I think it would complicate the code tremendously to try to run the WebP Hero polyfill on popover images for such barely used and obsolete browsers. If user wants to see image, they can open the page and the polyfill will then be run. To be clear, there are only a tiny number of Firefox versions affected: those that don't support WebP and DO support Service Workers, i.e. Firefox 56-64 (our minimum supported Firefox version for ServiceWorker mode is Firefox 56, though in practice SW mode only works reliably on all platforms on non-ESR versions of Firefox 60+).
* Works fine in Safari 14 on macOS
* Works fine in Safari 14 on iPad
* Works fine on Edge 18 on W10
* Works fine (Safe Mode only) in Edge 16 on W10,

Good to merge.

PR #1264, opened 2024-07-08
Add generic URI schema support when clicking ZIM links

### LINKED ISSUE #1260: Add filtering for custom URI schemata (e.g. `mailto:`, `skype:`) to the filterClickEvent to force these to open outside the sandbox
It turns out that attempting to open a `mailto:` link in a Chromium extension (or in a PWA in Chromium browsers) is seen by Chrome as an insecure protocol, which triggers sandbox blocking. See https://github.com/kiwix/kiwix-tools/issues/680. In Zimit1, the ReplayWorker handled this kind of stuff, I think, but we may have to handle it ourselves. There's currently already such a filter in JQuery/Safe mode.

However, we should first try a simple fix by adding `mailto:` as an allowed protocol in the sandbox Content Security Policy.

Jaifroid:
Fixes #1260. This should add generic support for different "protocols" (URI schemata), so that they are forced to open outside the sandbox. This avoids triggering a CSP security exception (CORS), which then leaves the app in a completely unusable state.

The PR generically catches any URI scheme that is different from the app's location.protocol. It is this difference that can trigger CORS protection. Doing it generically means we don't have to list a whole bunch of schemata like `skype:`, `zoomus:`, `mailto:`, etc.., especially since custom protocols can be defined nowadays. While some of these don't cause exceptions in Firefox (e.g. `mailto:`), they all cause CORS exceptions in Chromium browsers.


Jaifroid:
See also https://github.com/kiwix/kiwix-tools/issues/680 for analogue issue in Kiwix Serve.

Jaifroid:
Hmm, although this code works in the PWA, it's not working here. I think it is tested too late, after some other conditions have been triggered.

Jaifroid:
OK, stupidly forgot to copy some detection code in `handleClickOnReplayLink`. Now working with `mailto:` in Military Medicine ZIM.

image

Jaifroid:
@audiodude If the Linux (Browser Stack) tests fail again it's probably because the tests are a bit flaky (timing issues), and just need to be re-run, which I'll do if necessary in the morning here. Note that there isn't an e2e test covering this PR because we don't currently have a (small) ZIM in the test suite that has a `mailto:` link, and also because it would be hard to detect whether such a link had performed a custom action in Selenium. So I'm limited to manual testing.

What this PR does is add two rules, one for Replay-based ZIMs (zimit1, but also zimit2 in most circumstances), and one for openZIM ZIMs, which effectively test whether the protocol of a clicked anchor is the same as the protocol of the document loaded in the iframe. If they aren't the same, then (with the exception of some internal protocols such as `javascript:` or `about:`) we force the browser to open the link in a new tab or window, i.e. outside of the sandbox. Then it's no longer our problem...

audiodude:
I assume from your prior comments @Jaifroid that this fixes the case where our CSP says "Didn't allow such and such links" but then we serve documents that contain them? Probably because any old things could end up in a ZIM?

Jaifroid:
> I assume from your prior comments @Jaifroid that this fixes the case where our CSP says "Don't allow such and such links" but then we serve documents that contain them? Probably because any old things could end up in a ZIM?

Yes, it's a combination of the strict CSP, sandbox of iframe and CORS protection: designed to ensure that the app can't "call home" or run active content from remote scripts, or download external images / iframes etc.. The main issue we're working around is that CORS protection is triggered when a user tries to load any URI as the source of the iframe that the browser doesn't perceive as same-origin for the app itself. So, since the app must be served from a secure URL (https:...) in order to run Service Workers, trying to load a `skype:` URI as the source of the iframe (which is what happens if we click on a `skype:` URL) triggers CORS protection, and puts the app into an unusable and unrecoverable state. If we redirect such URIs to a different tab, then CORS isn't triggered.

audiodude:
Just a random thought. Would it be better/cheaper to detect those links on DOM load and add a `target="blank"` attribute to them? Might make debugging easier in the future, and prevent having to remember to add this logic in new places as the architecture changes.

audiodude:
Let me look closer in the morning

Jaifroid:
> Just a random thought. Would it be better/cheaper to detect those links on DOM load and add a `target="blank"` attribute to them? Might make debugging easier in the future, and prevent having to remember to add this logic in new places as the architecture changes.

There's something of a taboo on altering DOM content from the ZIM unless it's the only solution, and I've been strongly discouraged from doing so at least in this app (which follows a purist philosophy that is certainly very sound if sometimes a bit frustrating!). One issue with adding `target="blank"` to all ZIM links is that it will alter the behaviour of a page if it is already loaded at top level, i.e. outside the iframe in a tab or window without any app chrome (in this mode, the SW acts as a transparent proxy). In that case, none of this link filtering is needed: link filtering is only active if there is an iframe to which the click event listener is attached. And it's only necessary because of the CSP protection on the iframe. Otherwise, we just let the browser do its thing and the app acts just like Kiwix Serve does when run in frameless mode, only in our case it's a Service Worker rather than a Web Server that is interfacing with the decompressor.

Apart from those reasons, it wouldn't be cheaper to process every single ZIM link in an article in advance. It is much cheaper only to process the single anchor on which a user has clicked at click time. FYI, we do have DOM-trawling code in the legacy JQuery mode, but the idea of adding a Service Worker was to prevent us having to simulate browser functionality for every click, every image loaded, every script called.... (it was a nightmare!)... Instead, the SW intercepts Fetch requests, decides if they have to come from the ZIM or from the FS, and responds accordingly. Unfortunately, the SW can't intercept out-of-scope URLs...

You can test browsing chromeless by going to https://browser-extension.kiwix.org/current/www/index.html, opening a ZIM and then command-clicking a ZIM link. This should open the clicked article in a new tab which is fully browsable, with the Service Worker serving content that the browser requests from the ZIM.

audiodude:
Thanks for the extensive explanation! Maybe we can VC at some point and you can explain the architecture to me (again?) because I am definitely a bit lost.

www/js/app.js
```diff
@@ -2236,9 +2236,10 @@ function filterClickEvent (event) {
         }
         // DEV: The href returned below is the href as written in the HTML, which may be relative
         var href = clickedAnchor.getAttribute('href');
-        // We assume that, if an absolute http(s) link is hardcoded inside an HTML string, it means it's a link to an external website.
-        // We also do it for ftp even if it's not supported any more by recent browsers...
-        if (/^(?:http|ftp)/i.test(href)) {
+        // We assume that, if an absolute http(s) link is hardcoded inside an HTML string, it means it's a link to an external website
+        // (this assumption is only safe for non-Replay archives, but we deal with those separately above: they are routed to handleClickOnReplayLink).
+        // Additionally, by comparing the protocols, we can filter out protocols such as `mailto:`, `tel:`, `skype:`, etc. (these should open in a new window).
+        if (/^(?:http|ftp)/i.test(href) || clickedAnchor.protocol && clickedAnchor.protocol !== ':' && articleWindow.location.protocol !== clickedAnchor.protocol) {
```
audiodude, line 2242: Technically `articleWindow.location.protocol` will never be `ftp:` right? So you can probably remove that from the regex to make the condition more self explanatory.

Also what links have a protocol of `:`?

www/js/app.js
```diff
@@ -2445,10 +2446,20 @@ function handleClickOnReplayLink (ev, anchor) {
...
-    // If it's for a different protocol (e.g. javascript:) we should let Replay handle that, or if the paths are identical, then we are dealing
-    // with a link to an anchor in the same document, or if the user has pressed the ctrl or command key, the document will open in a new window
-    // anyway, so we can return. Note that some PDFs are served with a protocol of http: instead of https:, so we need to account for that.
-    if (anchor.protocol.replace(/s:/, ':') !== document.location.protocol.replace(/s:/, ':') || pseudoDomainPath === containingDocDomainPath) return;
+    // If the paths are identical, then we are dealing with a link to an anchor in the same document
+    if (pseudoDomainPath === containingDocDomainPath) return;
+    // If it's for a different protocol (e.g. javascript:) we may need to handle that, or if the user has pressed the ctrl or command key, the document
+    // will open in a new window anyway, so we can return. Note that some PDFs are served with a protocol of http: instead of https:, so we need to account for that.
+    if (anchor.protocol && anchor.protocol.replace(/s:/, ':') !== document.location.protocol.replace(/s:/, ':')) {
```
audiodude, line 2453: Maybe extract `anchor.protocol.replace(/s:/, ':')` and `document.location.protocol.replace(/s:/, ':')` into local variables, so you can make a comment like "normalize https to http"?

audiodude (approved):
LGTM

www/js/app.js
Jaifroid: You're right, that's old code that's only really there to show a dev how to test for different protocols, but it's a bit redundant with the following `||` test. We do, however, need to keep the `^http` test here because we shouldn't normally find absolute `^http`-prefixed protocols for a ZIM link except in the case of Replay-based ZIMs (where wombat.js is in charge of overriding these and transforming them to ZIM links, but only on-the-fly). So, if we do find an `http`-prefixed protocol it is an external link and has to be opened outside the sandbox.

Regarding `:`, I was surprised myself to find these, but I was forced to put in this condition. It seems to be an edge case in Chromium browsers, probably a bug t.b.h., where some internal code has stripped the protocol because the URL resolves to a local URL that is same-origin, but the colon has been left in whatever search-and-replace operation happens internally. Browser quirks are the bane of my life....

I'll amend.

audiodude: As a random guess, I thought maybe relative links would have an "empty" protocol (the MDN docs say that the protocol always ends with `:`). But that's not the case in Chrome:

Jaifroid: I'll see if I can find the condition and browser that produced this... But I was definitely forced to add it at some point... I should have documented it I guess.

Jaifroid: So with some quick testing I haven't yet identified anything, but I'm reluctant to remove it because I know I was forced to put it in the code, and it could be some quirk in Edge Legacy or another browser I can't test right now. Or it could be a side-effect of wombat.js processing of links from when this function originally tried to cover that case and before I added in specific processing of Replay links...

What I'll do is add a DEV note to say that the usefulness of the condition should be evaluated and if no longer necessary after careful testing it can be removed.

Jaifroid:
Manual tests to be performed:

- [ ] IE11 in Restricted mode
- [ ] Chrome/Firefox in Restricted mode
- [ ] Safari and Edge Legacy in SW mode (via BrowserStack)
- [ ] Chromium Extension in ServiceWorkerLocal mode

Jaifroid:
All manual and automatic tests are passing, so I think it's safe to merge.

PR #1265, opened 2024-07-09
Rename Safe mode to Restricted mode

### LINKED ISSUE #1259: Rename "Safe Mode" to "Restricted Mode"
With regard to #1257, I think the use of the term "Safe Mode" may be steering some users towards not allowing the app to run in full ServiceWorker mode. I now think it would be better to use the term "Restricted Mode" (the term used in Visual Studio Code for when code is opened in a strong sandbox). This better describes the fact that the ZIM will not have all capabilities if read in this mode. The mode is now primarily intended as a "no-JavaScript" mode where untrusted ZIMs can be viewed without loading potentially unsafe scripts so that the user can judge the safety of the ZIM. Of course it also doubles as the fallback mode for devices or frameworks where Service Workers are not supported or banned (e.g. Chrome or Firefox on iOS, due to Apple restrictions, IE11, EdgeMobile, Firefox OS).

Jaifroid:
Fixes #1259. "Restricted" is what is used in Visual Studio Code when a file outside of a trusted Repo is opened. I think we risk implying that the mode is fully "Safe" by calling it "Safe". In fact it isn't, because inline JS can still run (at least outside of a Chromium Browser Extension), even though all attached scripts are disabled. It is "safer" than ServiceWorker mode or ServiceWorkerLocal mode, but it isn't 100% "safe".

Additionally, calling it "Safe" makes it sound more attractive than it is: it's a highly restrictive mode in which many things will be broken in all but the most static of ZIMs. It's a good way to view the contents of a ZIM in a relatively static environment before deciding if the ZIM is trusted, but it's not viable for everyday use except for Wikipedia and Stack Exchange ZIMs.

Please note, we never actually released a public version of the app with JQuery mode renamed as Safe mode, so this won't be a confusing change for users. They will simply see that "JQuery mode" (not really intelligible to the average user) is now renamed as "Restricted mode", which better describes what it does.


Jaifroid:
@audiodude There are no actual code changes here, just a mass search-and-rename (with care taken over capitalization and i18n). There may be a few missed cases I'll search for during testing.

audiodude (approved):
LGTM

Issue #1257, opened 2024-05-31
Users seem confused when the Chrome extension won't do some stuff in ServiceWorkerLocal mode


Jaifroid:
Not really sure what we can do about this, as we clearly post a warning message at the top of the landing page "Unable to display active content..." and tell them what to do about it (see screenshot). But two experienced users have recently ignored this and have reported video isn't working. It can't work, because Chrome browser extensions prohibit any inline JS as well as eval and other stuff. Also, when first installing the extension, it asks for permission to fetch the browser-extension.kiwix.org PWA, which solves the issue (and works offline thereafter). Same issue in Firefox, only worse, because Service Workers are prohibited in local Firefox extensions, so to do anything reasonable at all, users must agree to fetch the browser-extension PWA.

Is there anything more we could do? Maybe make that warning into a modal that users must actively acknowledge (could be annoying...)? @kelson42, any UX advice here?

![image](https://github.com/kiwix/kiwix-js/assets/4304337/b47064df-b482-4eae-90e3-774748e4da5e)

audiodude:
In particularly it's not clear what "active content" is, or what the difference between the "ServiceWorkerLocal" and "ServiceWorker" modes are and why I would want to switch. Even "content may be available by clicking search button above (or type a letter of the alphabet)" seems both vague and oddly specific. It should probably read "Some content may still be available via searching, above" or something.

Jaifroid:
@audiodude Thanks for the feedback. I suppose it's one of those cases of "developer-speak". The wording "active content" and what is called in the code "activeContentWarning" has been used ever since Service Workers were introduced in the app and we had to warn users that things would be broken if they didn't start using it. There are a few versions of this warning.

I agree it needs simplification. I propose something like:

> **Warning: you must [switch to ServiceWorker mode](#something) to view all content in this ZIM**.  Some content may still be available by searching above (tap the search icon, or type a letter of the alphabet in the search field).

The instruction on how to search is a workaround for those people who don't have a browser that supports Service Workers, but still want to access content in ZIMs like Gutenberg. The link [switch to ServiceWorker mode](#something) takes them straight to the setting they need to change.

FYI The problem here is a technical one: as mentioned, Firefox doesn't support Service Workers in browser extensions (not considered a secure context like `https:`), and while Chrome does support Service Workers in extensions, it blocks a lot of JS, so many things break in modern ZIMs.

[...]

PR #1268, opened 2024-07-11
Do not jump to first file when using FSA API

### LINKED ISSUE #1254: When opening a folder of archives with FSA API, do not jump straight to the first file
I'm not sure when this bug crept in, but opening a folder of archives in a browser where the File System Access API is available immediately opens the first archive in the folder. This is a bit disorientating, and not very user-friendly given the potential security concerns that we flag to the user. Opening a specific archive should be a conscious choice by the user, even if there is only one archive in the folder.

An exception would be re-opening the last-opened archive, IF it is in the list of trusted archives. And only if it was the last archive opened by the user.

Jaifroid:
Fixes #1254.

This should be the last of the issue-fixing PRs prior to preparing release of v4.1.

It ensures that a ZIM file is not actually opened if the user selected a directory of archives using the File System Access `window.showDirectoryPicker()` method, and if that folder contains more than one archive. Without this fix, the first archive in the directory is inadvertently opened, which is unexpected and potentially insecure behaviour.

I need to test whether the same applies to the `webkitdirectory` method used by Firefox and Edge Legacy.

### DIFF
--- a/www/js/app.js
+++ b/www/js/app.js
@@ -1617,7 +1617,7 @@ function displayFileSelect () {
         folderSelect.addEventListener('click', async function (e) {
             e.preventDefault();
             const previousZimFiles = await abstractFilesystemAccess.selectDirectoryFromPickerViaFileSystemApi()
-            if (previousZimFiles.length !== 0) setLocalArchiveFromFileList(previousZimFiles);
+            if (previousZimFiles.length === 1) setLocalArchiveFromFileList(previousZimFiles);
         });
     }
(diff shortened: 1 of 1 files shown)


Jaifroid:
Also works as intended with `webkitdirectory` method.

Jaifroid:
Ah, I forgot, I also need to do #1257 as we had agreed. So there will be one more issue-fixing PR after this.

audiodude (approved):
LGTM

PR #1269, opened 2024-07-11
Simplify active content warning

### LINKED ISSUE #1257: Users seem confused when the Chrome extension won't do some stuff in ServiceWorkerLocal mode

Jaifroid:
Fixes #1257, with the message we agreed there.


audiodude (approved):
LGTM

PR #1271, opened 2024-07-12
Add i189 for Chromium store entries

### LINKED ISSUE #1118: Translate the Store information (Mozilla, Chrome, Edge, Ubuntu) into at least French and Spanish
While I believe a lot of Stores provide automatic translation, at least the output should be checked for the languages of the app (i.e. French and Spanish, as well as English, for now). If the translations are poor, we should take control of them and produce our own translations.

Jaifroid:
Fixes #1118. According to the messages I receive when uploading an extension to the Edge or Chrome Stores, in order to internationalize the Store entry it is necessary to internationalize the manifests, adding those languages supported by the app. I've followed https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Internationalization, but I don't know for sure that this will do the trick until I attempt to upload the generated extension ZIPs.

We don't use the Chrome i18n module for i18n within the app, as we use a much simpler string-substitution system, and I'm not going to change that any time soon, especially as i18n for this app needs to work in multiple contexts, not just extensions.

### DIFF
--- a/_locales/en/messages.json
+++ b/_locales/en/messages.json
@@ -1,0 +1,11 @@
+{
+  "extensionName": {
+    "message": "Kiwix JS Browser Extension",
+    "description": "Name of the extension."
+  },
+
+  "extensionDescription": {
+    "message": "Kiwix is an offline reader for content from Wikipedia, TED Talks, Stackexchange, and many other sources.",
+    "description": "Description of the extension."
+  }
+}
--- a/_locales/es/messages.json
+++ b/_locales/es/messages.json
@@ -1,0 +1,11 @@
+{
+  "extensionName": {
+    "message": "Extensión de navegador Kiwix JS",
+    "description": "Nombre de la extensión."
+  },
+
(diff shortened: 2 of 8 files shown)


audiodude (approved):
LGTM

Jaifroid:
Thanks, I'm going to see if I can do a test draft upload from this branch to the Edge Store to see if this actually does the job I want it to do...

Jaifroid:
Hmm, something is wrong uploading and zipping the _locales directory. I'll debug tomorrow.

Jaifroid:
LOL, I forgot to add the `_locales` directory int he rollup config, doh...

Jaifroid:
Yay, it works, finally!

image

Now I just have to work out why the "Description" field is empty when I thought included it....

Jaifroid:
OK, so description in Edge Store at least isn't taken from the manifest, so I think this PR is complete now.

PR #1272, opened 2024-07-12
Extend time between reloads for failed SW registration

### LINKED ISSUE #1255: Wait longer for the SW to cache assets before rebooting if the SW fails to register
The app contains code to cache its own assets, according to the SW specification, but it occasionally has to reboot itself when the SW fails to register. In KJS, this reboot is being issued too quickly, and sometimes the app gets into a loop whereby it reboots before assets have finished caching, causing a boot loop in rapid succession. I saw this some time ago in the PWA, and slowed the process down by giving it more time. Something similar should be added here -- or that whole code needs some debugging. I've only noticed this when testing on iOS on an iPhone (not on an iPad). It's not a target for this app, but it does expose an issue that should be fixed.

Jaifroid:
Fixes #1255. Sorry, should be last one. Again, its very simple, but I forgot I said I'd observed this again while testing popover code and it does need fixing.

### DIFF
--- a/www/js/app.js
+++ b/www/js/app.js
@@ -991,10 +991,10 @@ function initServiceWorkerMessaging () {
             action: 'init'
         });
     } else if (serviceWorkerRegistration) {
-        // If this is the first time we are initiating the SW, allow Promises to complete by delaying potential reload till next tick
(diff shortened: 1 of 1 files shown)


audiodude (approved):
LGTM

PR #1278, opened 2024-10-01
Fixed Brief Description Popover extraction Issue

### LINKED ISSUE #1276: Brief article description is showing inappropriately in popovers from Wikivoyage (at least) ZIMs scraped from new mobile HTML endpoint
See screenshot to understand the issue. I believe these new brief descriptions are specific to Wikivoyage, and are only showing in ZIMs scraped from the new mobile HTML endpoint. There is nothing wrong with them, but they are not well displayed and are probably redundant in popovers. It should be easy to filter them out in `popovers.js` given that they have an id. Test ZIM for this is the 08-2024 Wikivoyage English maxi ZIM https://download.kiwix.org/zim/wikivoyage/wikivoyage_en_all_maxi_2024-08.zim.

image

THEBOSS0369:
This PR Fixes #1276 

Hello Everyone!

## Description
In this PR i have fixed the popover inappropriate description extraction by adding regex logic so that only Required information will be shown. 
**NOTE**: This PR fixes almost every popover however there are still 3 - 5 popover's that i haven't able to fix for example calsacaus in Europe , Greenland and other 3 from USA. 

## Test
1. Checked in both "**Restricted**" and "**ServiceWorker**" Everything is working fine check the ss below.
3. End-to-end (e2e) `tests-e2e-iemode`-> There was some e2e.runner.js error which i have commented on this PR's Issue thread. I tried alternative http-server there wasn't any issue.
5. extension versions with production code tested
6. Browser Test -> Microsoft Edge, Chrome, Firefox and Brave

### Before
rail
mount
safari

### After (Test - eServiceworker / Non restricted)
rail
mount
safari

### After (Test - Restricted)
high speed rail
asia
greece

Thanks

### DIFF
--- a/www/js/lib/popovers.js
+++ b/www/js/lib/popovers.js
@@ -102,7 +102,7 @@ function cleanUpLedeContent (node) {
     // Define an array of exclusion filters
     // (note `.exclude-this-class` is a dummy class used as an example for any future exclusion filters)
     const exclusionFilters = ['#pcs-edit-section-title-description', '.exclude-this-class'];
-    // Constrct the :not() selector thing
+    // Construct the `:not()` CSS exclusion selector list
     const notSelector = exclusionFilters.map(filter => `:not(${filter})`).join('');
 
     // Remove all standalone style elements from the given DOM node, because their content is shown by innerText and textContent
--- a/www/js/lib/popovers.js
+++ b/www/js/lib/popovers.js
@@ -261,7 +261,7 @@ function populateKiwixPopoverDiv (ev, link, state, dark, archive) {
     const linkHref = link.getAttribute('href');
     // Do not show popover if there is no href or with certain landing pages
     if (!linkHref || /^wikivoyage/i.test(archive.file.name) &&
-       (state.expectedArticleURLToBeDisplayed === archive.landingPageUrl ||
+      (state.expectedArticleURLToBeDisplayed === archive.landingPageUrl ||
       state.expectedArticleURLToBeDisplayed === 'A/Wikivoyage:Offline_reader_Expedition/Home_page')) {
         return Promise.resolve();
     }
--- a/www/js/lib/popovers.js
+++ b/www/js/lib/popovers.js
@@ -109,7 +109,8 @@ function cleanUpLedeContent (node) {
         // DEV: Note that innerText is not supported in Firefox OS, so we need to use textContent as a fallback
         // The reason we prefer innerText is that it strips out hidden text and unnecessary whitespace, which is not the case with textContent
         const innerText = para.innerText ? para.innerText : para.textContent;
-        return innerText.length >= 50;
+        const text = innerText.trim();
+        return !/^\s*$/.test(text) && text.length >= 50;
     });
     return parasWithContent;
 }
--- a/www/js/lib/popovers.js
+++ b/www/js/lib/popovers.js
@@ -36,7 +36,7 @@ import uiUtil from './uiUtil.js';
  * @param {ZIMArchive} archive The archive from which to extract the lede
  * @returns {Promise<String>} A Promise for the linked article's lede HTML including first main image URL if any
  */
-function getArticleLede (href, baseUrl, articleDocument, archive) {
+function getArticleLede(href, baseUrl, articleDocument, archive) {
     const uriComponent = uiUtil.removeUrlParameters(href);
     const zimURL = uiUtil.deriveZimUrlFromRelativeUrl(uriComponent, baseUrl);
     console.debug('Previewing ' + zimURL);
@@ -98,7 +98,7 @@ function getArticleLede (href, baseUrl, articleDocument, archive) {
 };
 
 // Helper function to clean up the lede content
-function cleanUpLedeContent (node) {
+function cleanUpLedeContent(node) {
     // Remove all standalone style elements from the given DOM node, because their content is shown by innerText and textContent
     const styleElements = Array.from(node.querySelectorAll('style'));
     styleElements.forEach(style => {
@@ -111,13 +111,18 @@ function cleanUpLedeContent (node) {
         // The reason we prefer innerText is that it strips out hidden text and unnecessary whitespace, which is not the case with textContent
         const innerText = para.innerText ? para.innerText : para.textContent;
         const text = innerText.trim();
-        return !/^\s*$/.test(text) && text.length >= 50;
+
+        // removing the para with less than 50 characters
+        // regex to check the paragraph if its too short or a brief description
+        const briefDescriptionRegex = /^.{1,100}$/;
+
+        return !briefDescriptionRegex.test(text) && text.length >= 50;
     });
     return parasWithContent;
 }
 
 // Helper function to concatenate paragraphs to fill the balloon
-function fillBalloonString (paras, baseURL, pathPrefix) {
+function fillBalloonString(paras, baseURL, pathPrefix) {
     let cumulativeCharCount = 0;
     let concatenatedText = '';
     // Add enough paras to complete the word count
@@ -149,7 +154,7 @@ function fillBalloonString (paras, baseURL, pathPrefix) {
 }
 
 // Helper function to get the first main image from the given node
-function getImageHTMLFromNode (node, baseURL, pathPrefix) {
+function getImageHTMLFromNode(node, baseURL, pathPrefix) {
     const images = node.querySelectorAll('img');
     let firstImage = null;
     if (images) {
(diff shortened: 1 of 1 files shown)


Jaifroid:
Your issue is probably due to the lack of a standard Firefox installation locally. I'm running tests on GH Actions, so let's see.

Jaifroid:
Hi, all tests passed, and I've also now had a chance to test this on the Wikivoyage that was showing it. It's working fine, well done! I just need to check for regressions with any other Wikimedia archives, which I'll do now.

www/js/lib/popovers.js
```diff
@@ -111,7 +111,12 @@ function cleanUpLedeContent (node) {
         // The reason we prefer innerText is that it strips out hidden text and unnecessary whitespace, which is not the case with textContent
         const innerText = para.innerText ? para.innerText : para.textContent;
         const text = innerText.trim();
-        return !/^\s*$/.test(text) && text.length >= 50;
+
+        // removing the para with less than 50 characters
+        // regex to check the paragraph if its too short or a brief description
+        const briefDescriptionRegex = /^.{1,100}$/;
```
Jaifroid, line 117: I don't think this is a safe way to filter out these short descriptions. We don't know in advance the amount of text, so setting a limit of 100 characters seems arbitrary and will likely have unintended consequences elsewhere, for example filtering out legitimate paragraphs with only 100 characters in them.

Jaifroid (changes requested):
Please see comment on why we can't use length as a selection criterion. However, I saw while testing that these title descriptions do in fact have an ID which can be used to filter them out. See screenshots:

image

image

All you need to do, then, is refine the `querySelectorAll` function on line 107 so that it selects all paragraphs EXCEPT those with the id `#pcs-edit-section-title-description`.

Jaifroid:
Good idea to update the branch too, see button "Update branch" below. Once you've done that, be sure to pull the change in your local copy of the Repo.

www/js/lib/popovers.js
Jaifroid: The reason we do have a filter requiring at least 50 characters is that there are lots of small empty paragraphs or empty paragraphs with things like `<span>&nbsp;</span>` which can't be filtered out any other way. Through testing, we determined that 50 characters caught nearly all these while never removing any nodes with meaningful text.

THEBOSS0369:
Hey @Jaifroid !

Apologies for late reply, i wasn't able to open github as i had exams which ended today. 

So for this PR i was also not very confident because i was directly changing the hex code, that's why i created it thanks for identifying this.
Also can you tell me how did you find the Description's id in inspect mode like if i have to find any id i used to search the whole components ID (i know its childish) but i never thought that we can do this with console as well😅😅.

Thanks!

www/js/lib/popovers.js
THEBOSS0369: Got it Sir! i will change the logic.

Jaifroid:
@THEBOSS0369 I simply put a break on line 109 of popover.js using the browser's dev tools, and then when the execution paused, I inspected the value of the `node` variable. The first paragraph of the collection clearly contains the ID `#pcs-edit-section-title-description`. Hope that helps!

THEBOSS0369:
Hey @Jaifroid ! 

I have fixed the requested changes. Please review it. This time every popover is working fine. 

This the ss for the inspect mode to verify the `#pcs-edit-section-title-description` is not coming in the website
image

and these are some ss of popover which weren't working before but now working. 
image
image
image

Thanks!

www/js/lib/popovers.js
```diff
@@ -104,14 +104,12 @@ function cleanUpLedeContent (node) {
     styleElements.forEach(style => {
         style.parentNode.removeChild(style);
     });
-    const paragraphs = Array.from(node.querySelectorAll('p'));
-    // Filter out empty paragraphs or those with less than 50 characters
```
Jaifroid, line 108: This comment shouldn't be deleted, but it should be moved above line 108 (above line beginning `const parasWithContent = ...`.

www/js/lib/popovers.js
```diff
@@ -104,14 +104,12 @@ function cleanUpLedeContent (node) {
...
+    const paragraphs = Array.from(node.querySelectorAll('p:not(#pcs-edit-section-title-description)'));
```
Jaifroid, line 107: This is good, but I think we should generalize this function, as in the future we may want to exclude other elements by css selector. See overall comment.

www/js/lib/popovers.js
```diff
@@ -104,14 +104,12 @@ function cleanUpLedeContent (node) {
...
     const parasWithContent = paragraphs.filter(para => {
         // DEV: Note that innerText is not supported in Firefox OS, so we need to use textContent as a fallback
         // The reason we prefer innerText is that it strips out hidden text and unnecessary whitespace, which is not the case with textContent
         const innerText = para.innerText ? para.innerText : para.textContent;
-        const text = innerText.trim();
-        return !/^\s*$/.test(text) && text.length >= 50;
```
Jaifroid, line 114: I don't understand why you've deleted these lines and replaced it with the one below. This reduces cleanup functionality. Please restore.

Jaifroid (changes requested):
This is a great improvement, many thanks!

Apart from the specific comments, here's how you could generalize the `:not` selector:

At the beginning of the `cleanupLedeContent` function, add the variables defining exclusion styles, e.g.:

```
// Define an array of exclusion filters (note .exclude-this-class is a dummy class as an example)
const exclusionFilters = ['#pcs-edit-section-title-description', '.exclude-this-class'];
// Construct the :not() selector string
const notSelector = exclusionFilters.map(filter => `:not(${filter})`).join('');
```
Then change your `querySelectorAll` definition to use this, e.g. ``p${notSelector}`. Also add a comment before this, to explain what we're doing. Something like "Apply the style-based exclusion filters to remove unwanted paragraphs".

This uses the same solution you came up with, but would make it more flexible for future developers who need to add more exclusion filters.

www/js/lib/popovers.js
THEBOSS0369: Ohhhhh sorry sorry!!!!! I was playing with the code to see what will happen to website and pushed it without fixing it. fixing this right away.

www/js/lib/popovers.js
THEBOSS0369: Got it , Fixing this.

THEBOSS0369:
@Jaifroid ! I have fixed all the changes you asked me. Everything is working fine after these changes. Here is another ss for the proof which was in the issue.

image

Thanks !

www/js/lib/popovers.js
```diff
@@ -99,13 +99,21 @@ function getArticleLede (href, baseUrl, articleDocument, archive) {
 
 // Helper function to clean up the lede content
 function cleanUpLedeContent (node) {
+    // Define an array of exclusion filters
+    // (note .exclude-this-class is a dummy class used as an exmple which will used to exclude classes in future)
```
Jaifroid, line 103:
```suggestion
    // (note `.exclude-this-class` is a dummy class used as an example for any future exclusion filters)
```

www/js/lib/popovers.js
```diff
@@ -99,13 +99,21 @@ function getArticleLede (href, baseUrl, articleDocument, archive) {
...
+    const exclusionFilters = ['#pcs-edit-section-title-description', '.exclude-this-class'];
+    // Constrct the :not() selector thing
```
Jaifroid, line 105:
```suggestion
    // Construct the `:not()` CSS exclusion selector list
```

www/js/lib/popovers.js
```diff
@@ -99,13 +99,21 @@ function getArticleLede (href, baseUrl, articleDocument, archive) {
...
     const styleElements = Array.from(node.querySelectorAll('style'));
     styleElements.forEach(style => {
         style.parentNode.removeChild(style);
     });
-    const paragraphs = Array.from(node.querySelectorAll('p'));
-    // Filter out empty paragraphs or those with less than 50 characters
+    // Apply this style-based exclusion filter to remove unwanted paragraphs in the popover
+    const paragraphs = Array.from(node.querySelectorAll(`p${notSelector}`));
+
+    // Filter out empty paragraphs or those with less than 50 character
```
Jaifroid, line 116:
```suggestion
    // Filter out empty paragraphs or those with less than 50 characters
```

Jaifroid (commented):
This is looking good. Just a few typos to correct! I've done these as suggestions, so you can simply accept them if you wish.

THEBOSS0369:
> This is looking good. Just a few typos to correct! I've done these as suggestions, so you can simply accept them if you wish.

@Jaifroid These suggestions are definitely great i have commited those changes! Also Is there anything else needed to be done for this issue or The issue is fixed Now.

Jaifroid (approved):
@THEBOSS0369 Thank you very much for this PR. It looks great now! I'll do some final manual testing before merging it. Are you OK for me to merge it after testing if all is good? The PR will be in your name and credited to you.

THEBOSS0369:
> @THEBOSS0369 Thank you very much for this PR. It looks great now! I'll do some final manual testing before merging it. Are you OK for me to merge it after testing if all is good? The PR will be in your name and credited to you.

@Jaifroid Yes Sir! No Problem at all!

PR #1284, opened 2024-11-04
 Reopening last article via double clicking on menu items/refreshing when double clicking on home

D3V-D:
Works in `npm run serve` and passes `npm run test`.

Works fine in Firefox & Chromium for me, I do not have edge to test that so some help there would be much appreciated!

Also, if you could check `npm run preview` for me as the changes are not reflected for me, that'd be great.

Should be completely keyboard compatible

### DIFF
--- a/www/css/app.css
+++ b/www/css/app.css
@@ -313,15 +313,20 @@ iframe._invert, iframe._mwInvert {
 /* Change focus styles on top nav-bar buttons (keep border-color for keyboard users) */
 .btn-nav-top:focus {
     background: none;
-    border-color: grey;
-    box-shadow: none;
+    border-color: #80bdff;
+    box-shadow: 0 0 0 .2rem rgba(0,123,255,.25);
 }
 
-.active-btn {
+/* Note: This one is !important to override the .active > button box-shadow */
+.btn-nav-top:hover {
+    box-shadow: 0 0 0 -.1rem rgba(216,217,219,.8) !important;
+}
+
+.active > button {
     color: #212529 !important;
     background-color: #e2e6ea !important;
     border-color: #dae0e5 !important;
-    box-shadow: 0 0 0 -.1rem rgba(216,217,219,.5) !important;
+    box-shadow: 0 0 0 .2rem rgba(216,217,219,.5) !important;
 }
 
 /* Add missing margin to buttons */
--- a/www/css/app.css
+++ b/www/css/app.css
@@ -310,6 +310,20 @@ iframe._invert, iframe._mwInvert {
     color: #337ab7;
 }
 
+/* Change focus styles on top nav-bar buttons (keep border-color for keyboard users) */
+.btn-nav-top:focus {
+    background: none;
+    border-color: grey;
+    box-shadow: none;
+}
+
+.active-btn {
+    color: #212529 !important;
+    background-color: #e2e6ea !important;
+    border-color: #dae0e5 !important;
+    box-shadow: 0 0 0 -.1rem rgba(216,217,219,.5) !important;
+}
+
 /* Add missing margin to buttons */
 button {
     margin: 2px;
--- a/www/js/lib/uiUtil.js
+++ b/www/js/lib/uiUtil.js
@@ -943,10 +943,6 @@ function returnToCurrentPage () {
     document.getElementById('liAboutNav').classList.remove('active');
     document.getElementById('liHomeNav').classList.add('active');
 
-    document.getElementById('btnHome').classList.add('active-btn');
-    document.getElementById('btnConfigure').classList.remove('active-btn');
-    document.getElementById('btnAbout').classList.remove('active-btn');
-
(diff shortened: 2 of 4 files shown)


D3V-D:
@Jaifroid Please review when you have the time, thanks!

Jaifroid:
Great, thanks for the PR! I'm currently at work, but will look at this as soon as I can.

D3V-D:
No rush, thanks for taking the time to test the preview.

www/css/app.css
Jaifroid, line 1: Great, these are subtle effects, but it do indeed make a difference in visual clues about what is happening! Nice. Just make sure you're happy with the dark mode. It works because it simply inverts all colours. It seems OK to me.

www/js/lib/uiUtil.js
```diff
@@ -940,7 +940,11 @@ function showReturnLink () {
 // Function to switch back to currently loaded page
 function returnToCurrentPage () {
     document.getElementById('liConfigureNav').classList.remove('active');
+    document.getElementById('liAboutNav').classList.remove('active');
     document.getElementById('liHomeNav').classList.add('active');
+
+    document.getElementById('btnHome').focus();
+
```
Jaifroid, line 947: I don't think you need the white space on lines 945 and 947, just because it's not conceptually an action that needs highlighting in some way.

Jaifroid (commented):
This looks great! One very minor thing here. But I also found one slight bug I'll make a screen recording of and put in a separate post here.

Jaifroid:
This one is a small bug. It concerns what happens when the UI is collapsed on narrow screens (you can use the device emulation tab in Dev Tools, at least in Chromium browsers to check this, or you can simply make the browser window very narrow). The issue you see here is that the code that collapses the menu again isn't executed in the return code, I suppose.

https://github.com/user-attachments/assets/bdeb7437-7523-4c4f-af1f-54afb28727f7

Jaifroid:
There is a slight cosmetic inconsistency when using the keyboard (tab key) to select UI elements, which is that the focused button that has your styles applied no longer has the "halo" round the button. I suspect this is to do with focusing the buton vs focusing the `li` or something like that. It's worth saying that the halo effect doesn't in fact exist in the current release of the app, so maybe you added it. It looks a bit inconsistent here. The video here was made entirely using Tab and Shift-Tab to shift amongst the buttons at the top (after clicking on the top bar). There is very little visual clue that an already highlighted button is being focused by the keyboard, unlike the un-highlighted buttons.

https://github.com/user-attachments/assets/0dbb90f9-1211-41e0-b803-db6b9c429de8

www/js/lib/uiUtil.js
D3V-D: Will remove

www/css/app.css
D3V-D: It looks OK-ish, might be a little lacking in contrast - do you think its worth adding dark styles?

D3V-D:
> This one is a small bug. It concerns what happens when the UI is collapsed on narrow screens (you can use the device emulation tab in Dev Tools, at least in Chromium browsers to check this, or you can simply make the browser window very narrow). The issue you see here is that the code that collapses the menu again isn't executed in the return code, I suppose.

Okay, so you're saying that it should auto-collapse when switching to a different page, gotcha. I'll work on that.

> There is a slight cosmetic inconsistency when using the keyboard (tab key) to select UI elements, which is that the focused button that has your styles applied no longer has the "halo" round the button. I suspect this is to do with focusing the button element vs focusing the li element or something like that (just guessing). It's worth saying that the halo effect doesn't in fact exist in the current release of the app, so maybe you added it. It looks a bit inconsistent here. The video here was made entirely using Tab and Shift-Tab to shift amongst the buttons at the top (after clicking on the top bar). There is very little visual clue that an already highlighted button is being focused by the keyboard, unlike the un-highlighted buttons.

Good point; I'll try to make that focus halo have higher priority. 

Will ping you once I've gotten these issues resolved.

D3V-D:
@Jaifroid Fixed the two bugs!

Jaifroid:
Thanks @D3V-D, I confirm those bugs are fixed in testing on Edge. I'll test on IE11 and an old Firefox shortly.

Jaifroid (approved):
Looks great, thanks very much for this PR @D3V-D. I tested in IE11 and in Firefox OS simulator (used to ensure very old browsers are still supported), and it all works fine. So, if all ready on your side, we can squash and merge. Just let me know.

D3V-D:
@Jaifroid sounds good to me - it just feels almost too easy 🤔 

I'll probably do some more tests later today and make sure I'm happy with how it looks visually and everything them I'll let you know

D3V-D:
@Jaifroid okay I haven't really found the time to do additonal testing but it should be fine functionally, any issues would just be aesthetic and slight.

I'm happy to merge now! Thanks for the review

PR #1286, opened 2024-11-11
Added an automated Selenium UI test for a small Zimit2 archive

### LINKED ISSUE #1205: Add an automated Selenium UI test for a small Zimit2 archive
It's probably too soon to do this until we have a fully settled and approved (and published) zimit2 archive. But once we have that, it'd be good to test against one.

Needs to be small enough to be loadable in memory. E.g. 100rabbits.

THEBOSS0369:
This PR Fixes #1205 

Hello Everyone!

## Description
In this PR i have added an automated Selenium UI test for a small Zimit2 archive (used `tonedear.com_en_2024-09.zim`)

The tests i have added are as follows : 
1. Load Kiwix JS and verify title
2. Switch to Jquery or SW mode
3. Load Tonedear archive and verify content
4. Navigate from main page to Android & iOS section
5. Verify Android and iOS store images in Both Modes

## Test
1. Checked in both "**Restricted**" and "**ServiceWorker**" Everything is working fine check the ss below.
6. End-to-end (e2e) `tests-e2e-firefox`-> All tests Passed Below is the ss
8. Browser Test -> Microsoft Edge, Chrome, Firefox and Brave

## Screenshot for tests
image


THEBOSS0369:
@Jaifroid I have not added test for android & ios page because i didn't know what test should i do your suggestion will be appreciated 😅😃 Thanks!

Jaifroid:
Thank you @THEBOSS0369. That's very useful. I would suggest merely switching to the page that shows the Android and iOS store images, and verifying that both images have loaded correctly:

* In Restricted mode (aka JQuery), test that both images' `src` attribute has changed to a data URI (see screenshot), and assert that both data URIs match the beginning of the expected content (say 30 characters, so you get the beginning of the actual data, not just "data:img/png,base64,". If you want to be thorough, you could hash the string and test the hashes, but that may be overkill.
* In ServiceWorker mode, you can do a basic image verification: check if the image has completed loading using JavaScript properties, and verify that the image has natural dimensions (width/height > 0). Use the computed `naturalWidth` and `naturalHeight` properties of the images to verify (see https://developer.mozilla.org/en-US/docs/Web/API/HTMLImageElement/naturalWidth). This will tell us that the image loaded correctly (after appropriate wait).

image

THEBOSS0369:
Hey @Jaifroid ! I have added the android and ios test as you asked me but there is a problem occurring when i am running the tests the `restricted mode` test passes well but in `sw mode` this error comes 
android test failing

i have tried by making the waititme 120000 , by removin cache but still this won't work. I don't what i am doing wrong i checked in the sw mode the image's alt is present and i have added the same in the code. Your help will be appreciated . Thanks!

Jaifroid:
Thanks @THEBOSS0369. Sorry to hear about the error. It's tricky to get the assertions right with the timing. I'll take a look at the weekend, but in the meantime it's possibly the case that there's some error in selecting the images and checking their natural width. Adding some console logging to see what it's actually finding and what you are expecting it to find might help!

THEBOSS0369:
Yes sir. I'm doing that by further more progress i have found that there are in total 21 images showing 19 are `icon_external_link` one is `gif `and one is `kiwix logo` this script is still ain't able to find the store images. i will try to find more in the meanitime maybe i find something useful ,right now this is what's showing
image

Jaifroid:
OK, I ran this myself (in the Edge runner, adding the appropriate commands), and while watching the browser locally it's clear that the test never actually loads the page "Android & iOS App". It remains stuck on "Ear Training Practice" (see screenshot). You're not going to be able to verify the images on the "Android & iOS App" page unless you actually switch to it first!

image

tests/e2e/runners/firefox/firefox.e2e.runner.js
```diff
@@ -27,6 +28,8 @@ async function loadFirefoxDriver () {
 // so we need to use the second one first
 const driver_for_gutenberg = await loadFirefoxDriver();
 const driver_for_ray_charles = await loadFirefoxDriver();
+const driver_for_tonedear = await loadFirefoxDriver();
```
Jaifroid, line 31: As the comment says above this block, the drivers need to be listed in reverse test order here, so that when the user runs the tests locally we see the Ray Charles first (which will be on top of the other windows). So please put the tonedear test above gutenberg in this list, and update the comment to reflect that there are three tests

tests/e2e/spec/tonedear.e2e.spec.js
```diff
@@ -0,0 +1,176 @@
...
+                    await driver.executeScript('window.setLocalArchiveFromFileSelect();');
+                } else {
+                    await driver.executeScript(
+                        'window.setRemoteArchives.apply(this, [arguments[0]]);',
+                        [tonedearBaseFile]
+                    );
+                    await driver.sleep(1300);
+                }
+            });
+
```
Jaifroid, line 106: In between these two tests, there should be another test that switches to the page "Android & iOS App", and verifies that the correct page has loaded. Then it will be possible to test the dimension of the two images on that page in the test below. See screenshot.

image

THEBOSS0369: Got it sir! I didn't thought this was causing the issue😅

THEBOSS0369:
@Jaifroid Another problem occurring , when i try to navigate to Android Page  it won't, i think this is happening because i didn't know the exact path for it at first i used this `android-ios-ear-training-app` it didn't worked i tried by adding `/android-ios-ear-training-app` & `../android-ios-ear-training-app`  and many more tries but still didn't worked
image
then i looked into the `ray_charles code` as it also have a navigation function i tried this which is `//ul[contains(@class, 'navbar-nav navbar-right')]/li[1]/a` similar to that present in `ray chares` but still not working. 

Then i think i can do that by link but The link for the server remains same `http://localhost:5173/www/index.html?appCache=false` even when i open any other page but when i run the tests it have a new url `http://localhost:8080/dist/www/index.html?title=C/tonedear.com/android-ios-ear-training-app` for different pages i tried by clearing cache and using incognito mode still url remains same. I'm not sure if this is default or not.

Can you help me with this path problem if we get it fixed then every tests will pass correctly.

tests/e2e/runners/firefox/firefox.e2e.runner.js
THEBOSS0369: Done Sir

Jaifroid:
@THEBOSS0369 Oh no, this app doesn't support navigating to a page by URL fragment. There are two ways to get to a particular page: either click on the link that you highlight in the screenshot (easiest), or search for it using the search bar. Also, remember that the links you're clicking on are in the iframe, not at the top level, so you have to switch to the iframe before finding the link and clicking it. Take a look at the code in the Ray Charles test to see how to switch to the iframe.

This explains why you said you were finding 21 images in the landing page when in fact there are none! You were finding these images in the app itself, as you hadn't switched to the iframe containing the content.

THEBOSS0369:
@Jaifroid Sirrr ! it worked now its navigating to andorid page although the images are still not found now i will try to fix that 
image
image

THEBOSS0369:
@Jaifroid  Sir all The tests have passed . Does any thing else needed to be added in the test? If yes then do let me know.

The comments in this image for tonedear tests will not be visible in the main tests i have commented it out.
image

Jaifroid:
@THEBOSS0369 That's very good news! I don't think we need more tests. I'll review the new code as soon as I can, but in the meantime, you might want to clean it up, remove any excessive timeouts you previously added, etc. Once the spec is ready, you should add it to all the other runners in the folders under `tests/e2e/runners`. Then I'll need to run it on BrowserStack to be sure the tests pass on all browsers.

Oh, I see GitHub tests have failed. I'll re-run those to be sure it wasn't a timeout fluke.

THEBOSS0369:
@Jaifroid I think the tests are failing because when it navigate to android page it asks me to authorize the author and i used to press that maybe that's causing error.

THEBOSS0369:
@Jaifroid Yes this is happening due to the reason i told above i didn't clicked the pressed trust source this time and tests failed. I don't know how to fix this one
image

Jaifroid:
> Yes this is happening due to the reason i told above i didn't clicked the trust source this time and tests failed. I don't know how to fix this one

This is because you erased https://github.com/kiwix/kiwix-js/blob/main/tests/e2e/spec/legacy-ray_charles.e2e.spec.js#L125 (the setting "noPrompts=true" from the URL when loading the ZIM. This was in your previous spec, but then you did a hard reset of your branch, and you erased it, so it's no longer in the commit history.

That command prevents the security prompt from showing. All you need to do is copy the appropriate code block from the ray charles version of the test.

Also, in future, try to avoid rewriting history in your PRs. It makes things difficult to review as well, because when you do a force push, I have to reset my copy of your PR instead of simply pulling a new commit! (Don't worry, everyone makes these kinds of mistakes with PRs when learning.)

THEBOSS0369:
@Jaifroid  Done Sir! Now all tests passes well without any issue. I'm waiting for these tests to pass then i will add the tests in all the files and remove any unnecessary time waits.

> Also, in future, try to avoid rewriting history in your PRs. It makes things difficult to review as well, because when you do a force push, I have to reset my copy of your PR instead of simply pulling a new commit! (Don't worry, everyone makes these kinds of mistakes with PRs when learning
My apologies Sir i didn't knew force push can do this much damage. I will be careful from next time 😅

THEBOSS0369:
I have increased the timewait maybe this can fix it
image

THEBOSS0369:
@Jaifroid  Sir, The tests are failing in chrome i am not sure why its happening i tested in firefox it passes well. I just did same steps in `chrome's e2e test file` as i did in the `firefox's e2e test file`

tests/e2e/spec/tonedear.e2e.spec.js
```diff
@@ -0,0 +1,230 @@
...
+                await driver.wait(async function () {
+                    const contentAvailable = await driver.executeScript('return document.querySelector(\'a[href="android-ios-ear-training-app"]\') !== null;');
+                    return contentAvailable;
+                }, 10000); // Increased to 10 seconds for more loading time
+
+                // Find the "Android & iOS App" link
+                const androidLink = await driver.findElement(By.css('a[href="android-ios-ear-training-app"]'));
+
+                // Verify that the element is found
+                // console.log('Android & iOS App link found:', androidLink !== null);
```
Jaifroid, line 131: It might be useful to uncomment this, to check if the link is found.

tests/e2e/spec/tonedear.e2e.spec.js
```diff
@@ -0,0 +1,230 @@
...
+
+                // Scroll the element into view and click it
+                await driver.executeScript('arguments[0].scrollIntoView(true);', androidLink);
```
Jaifroid, line 134: I don't think the element needs to be scrolled into view. You could try without this (comment it out).

tests/e2e/spec/tonedear.e2e.spec.js
```diff
@@ -0,0 +1,230 @@
...
+                await driver.wait(until.elementIsVisible(androidLink), 10000); // Wait until it's visible
+                await androidLink.click();
+
+                // Take a screenshot after clicking for debugging
```
Jaifroid, line 138: Problem with this is that it's impossible to see screenshots in GitHub Actions AFAIK!

tests/e2e/spec/tonedear.e2e.spec.js
THEBOSS0369: Done Sir

tests/e2e/spec/tonedear.e2e.spec.js
THEBOSS0369: Done Sir

tests/e2e/spec/tonedear.e2e.spec.js
```diff
@@ -0,0 +1,230 @@
...
+
+                // Find the "Android & iOS App" link
+                const androidLink = await driver.findElement(By.css('a[href="android-ios-ear-training-app"]'));
+
+                // Verify that the element is found
+                // console.log('Android & iOS App link found:', androidLink !== null);
+
+                // Scroll the element into view and click it
+                await driver.executeScript('arguments[0].scrollIntoView(true);', androidLink);
+                await driver.wait(until.elementIsVisible(androidLink), 10000); // Wait until it's visible
```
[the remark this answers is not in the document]
THEBOSS0369: Done Sir

THEBOSS0369:
@Jaifroid ! This worked sir All the tests have passed.

But sir one doubt if the github tests wasn't able to recognize the images in the linux tests then how it was able to verify for the windows test? Like is it some compatible issue or anything else.

Jaifroid:
Very good news @THEBOSS0369! I guess it was a timing issue of some kind, but I need to check the verification method to be sure there isn't an issue with the assertions.

The next step if for me to get the Browser Stack tests running against this PR (you can't do that, because you don't have access to the credentials). To do this, I'll make a dummy PR that has the same code as your PR, and this should then run on Browser Stack, and we can get feedback on whether it's passing also on old browsers. The fact that it's already passing on IE11 mode on Windows (Edge) implies that there shouldn't be any problem with other browsers, but it needs to be checked. I'll do this when I can in the next few days.

THEBOSS0369:
@Jaifroid  Got it Sir! If my work's done here then shall i find any other issue to work on?

Jaifroid:
Apologies for slowness here -- things have been manic at work. I finally pushed a dummy branch to get the BrowserStack tests to run. It failed the Safari test, and you can see a video of the failed test run here (it starts around 00:50):

* [Tonedear Test Safari 14.1.2](https://automate.browserstack.com/builds/3e800cc0614d1f86ef983b41cec00804703372a2/sessions/032d55c402a58aeb574ee6c27b89169dd2bd097e?auth_token=9a7b2ba5b56dccdbd3c7cdb6a8171191bba31c7c34867c744e804ce7fc743398)

For comparison, here is a video of a different ZIM passing on that browser:

* [Gutenberg RO test Safari 14.1.2](https://automate.browserstack.com/builds/3e800cc0614d1f86ef983b41cec00804703372a2/sessions/33bcd0aadee6a9fb78ff516177d2067244fd7de6?auth_token=4ae2cddd4d4808709e87e0f4518b74f6676207d9d6a9f1b3473db1fcf8a67e63)

The issue, I think, is that this version of Safari can only run in Restricted mode on BrowserStack, and the app informs this to the user with a dialogue box, which blocks the ZIM from loading. You need to test for and dismiss that dialogue box if present. You could just copy the code for that from the Gutenberg or Ray Charles test.

Jaifroid:
@THEBOSS0369  We should add your test to a few more BrowserStack browsers. I suggest adding it to:

* Firefox 70 in SW mode only: https://github.com/kiwix/kiwix-js/blob/main/tests/e2e/runners/firefox/firefox70.bs.runner.js
* Chrome 60 on macOS in Restricted and SW modes: https://github.com/kiwix/kiwix-js/blob/main/tests/e2e/runners/chrome/chrome60.bs.runner.js
* Edge 18 on Win10 in Restricted and SW modes: https://github.com/kiwix/kiwix-js/blob/main/tests/e2e/runners/edge/edge18.bs.runner.js

The reason for adding it selectively and in selected modes is because otherwise tests take too long to run if we test every spec in every mode.

Jaifroid (commented):
Code is looking good so far. Please see some minor suggestions.

THEBOSS0369:
@Jaifroid ! Sir I have done all the changes as you asked, i have run the tests on firefox all test has passed but i am on windows so i can't install safari to test the tests on it 😓. 

I have added the tests in all the files as you told me but in the `Edge 18` file i added the tests in the previous commit can you please take a look at it and let me know if that's the correct way or not or anything missing. Thanks

THEBOSS0369:
Few Unnecessary files have come along with the commit I will remove them once in few hours after coming from college please don't mind them 😅😅

Jaifroid:
@THEBOSS0369 Thanks for your latest edits! I notice you added two new runners for Firefox and Chrome on the latest browsers. However, I don't think we need to test the latest browsers on BrowserStack because we already test latest on GitHub actions. Adding these is redundant and doesn't add new info, but just makes the tests take even longer to finish than they do already.

So, I suggest removing those! Sorry!

THEBOSS0369:
> @THEBOSS0369 Thanks for your latest edits! I notice you added two new runners for Firefox and Chrome on the latest browsers. However, I don't think we need to test the latest browsers on BrowserStack because we already test latest on GitHub actions. Adding these is redundant and doesn't add new info, but just makes the tests take even longer to finish than they do already.
> 
> So, I suggest removing those! Sorry!

Done Sir i have removed those tests , Thanks for pointing this out and sir please don't say sorry i am learning new things on the go so its beneficial for me to know about these things. 😃

Jaifroid:
Thank you! I'm just running the BrowserStack tests.

tests/e2e/runners/firefox/firefox70.bs.runner.js
```diff
@@ -33,7 +34,8 @@ async function loadFirefoxDriver () {
 const driver_gutenberg_fx = await loadFirefoxDriver();
 
 // Run test in SW mode only
-console.log('\x1b[33m%s\x1b[0m', 'Running Gutenberg tests in ServiceWorker mode only for this browser version');
+console.log('\x1b[33m%s\x1b[0m', 'Running Gutenberg and Tonedear tests in ServiceWorker mode only for this browser version');
 console.log(' ');
 
 await gutenbergRo.runTests(driver_gutenberg_fx, ['serviceworker']);
+await tonedear.runTests(driver_gutenberg_fx, ['serviceworker']);
```
Jaifroid, line 41: @THEBOSS0369  You can't re-use the same driver, as it is quit once the first test finishes. You need (above) to load with `const driver_tonedear_fx = await...` above, and use that driver here. This is why tests are failing with this runner. Check your other runners that you're not trying to re-use the same driver, or else those tests will also fail.

THEBOSS0369: @Jaifroid Done sir, i re used the same driver in this file only. now the test should pass without any problem

THEBOSS0369:
@Jaifroid sir the tests are failing so is it the same dialog box issue this time because its getting timed out ?

Jaifroid:
Oh, I have an idea: you need to handle the case that the dialogue box ISN'T shown. Don't assume it will be there...

In principle, it shouldn't be shown at all, due to a querystring we add 'noPrompt=true`. I wonder if that querystring was simply missing in the previous error you fixed? Because it's not missing in this test, the prompt doesn't show.

Jaifroid:
What I'm saying is, check over your previous fix to see if the issue was that you were calling the app without `noPrompt=true`, and that was why the prompt dialogue box appeared. In that case, you could fix this by removing the prompt handling code and ensuring the app is always loaded with that querystring.

THEBOSS0369:
@Jaifroid Sir I don't understand what's the problem here. let's divide all the possibilities
1. First For querystring

> In principle, it shouldn't be shown at all, due to a querystring we add 'noPrompt=true`. I wonder if that querystring was simply missing in the previous error you fixed? Because it's not missing in this test, the prompt doesn't show.

No sir , i didn't changed this since the first commit this is the code which is present since the start
```javascript
            it('Load Kiwix JS and verify title', async function () {
                await driver.get('http://localhost:' + port + '/dist/www/index.html?noPrompts=true');
                await driver.sleep(1300);

``` 

2. The fix that i used for the dialog box to not appear is this, it was present in the day before yesterday's commit 

```javascript
            it('Navigate from main page to Android & iOS section', async function () {
                // Check for Dialog Box and click any Approve Button in subsequent dialog box
                try {
                    const activeAlertModal = await driver.findElement(By.css('.modal[style*="display: block"]'));
                    if (activeAlertModal) {
                        // console.log('Found active alert modal');
                        const approveButton = await driver.findElement(By.id('approveConfirm'));
                        await approveButton.click();
                    }
                } catch (e) {
                    // Do nothing
                }

                // Switch to the iframe if the content is inside 'articleContent'
``` 

3. Sir i don't why but i had already added this code in my very first commit still the dialog box was appearing in the browser stack test which you asked me to fix 
image

I have got confused please clarify this sir 😓

THEBOSS0369:
> Did you see it this time? When I looked at the video, I didn't see the security diaologue box, and so the script got stuck because it couldn't find it.

No sir i also didn't see any dialogue box appearing in the video.
this time i have added the code to check whether dialogue box appears or not and will work according to that let's see if this time tests passes well or not.

Jaifroid:
Still failing, I'm afraid. See https://github.com/kiwix/kiwix-js/actions/runs/12176007591/job/33960839227?pr=1286 . It correctly recognizes there is no dialogue box, and then fails to navigate to the iOS/Android page, apparently. In fact, it looks like you don't load the landing page at all? The app remains stuck on Configuration. Do you even load the ZIM, it doen's tlook like it? Perhaps because this is a ServiceWorker mode only test, you accidentally skip the loading of the ZIM (I'm just guessing).

See BrowserStack report here for more info: https://automate.browserstack.com/builds/73815a6c7951be3a00e49ef6321ac1855566c1ba/sessions/088eed0cf1ab6ec43ecd5b2b4075a2607e7d0fb1?auth_token=0611b644be6ed3693eaa3cfd5f1e5ba1d16a364388143ff06da8a70c9a45bceb .

THEBOSS0369:
@Jaifroid Sir i'm not sure why its happening i am going to do some changes and test it using only sw mode but right now in my local server the landing pages is loading correctly please see the video below for reference
https://github.com/user-attachments/assets/dc2cdbaa-3a42-4955-96e6-912443b67e84

Jaifroid:
@THEBOSS0369 I think the difference is that in your local test you are running it first in Restricted mode (jQuery), and then in SW mode. In the affected test, only SW mode is called. Try replicating that locally, i.e. call the test with `await tonedear.runTests(driver_tonedear_fx, ['serviceworker']);`. See if it fails locally and fix if necessary.

THEBOSS0369:
@Jaifroid sir solo sw mode tests are pasing as well please see the video

https://github.com/user-attachments/assets/79302314-dd06-46b2-b1bc-12ff2ef628ad

THEBOSS0369:
@Jaifroid sir could be this be a browser specific issue because in latest version tests are not failing, so should we try it first in different browsers if this tests fail on those one as well then this is issue is definitely in the code and if it passes in others then we can know why the problem's occuring ?

Jaifroid:
OK. Another thing I noticed in the BrowserStack video was that the app reloaded itself just before tests started. That suggests it was having trouble loading and registering the Service Worker (it will keep reloading itself until the SW is registered correctly). It may be you need to increase the wait between loading the app's URL and starting the tests, to give the SW more time to register.

You can of course experiment with skipping this test for now, so that the next ones run, then we'll know if it's a specific browser issue or a more general problem.

Jaifroid:
Running tests.

THEBOSS0369:
@Jaifroid  sir I think the tests are failing due to the latest changes i made let me revert it back and see if it works and if not then i will have to do some experiments. Also sir is there any other way that i can run these tests myself and i don't have to disturb u again & again .

THEBOSS0369:
sir i have reverted all changes and tets are passing in my local let's see if it passes this time or not

Jaifroid:
@THEBOSS0369 Unfortunately this is now failing on Chrome 60. So it looks like it wasn't browser-specific. Here are the two failed runs:

* https://automate.browserstack.com/builds/2fec0a9c4d3c775a3fd796d073168f31141c6609/sessions/55fab668f4deb253eaffd4d69cb0c98924a91f4c?auth_token=a53a7948202b65cb55abf26c2c4f518b1c5fe3d4f17037e3a7cd4acce57e9eb3
* https://automate.browserstack.com/builds/2fec0a9c4d3c775a3fd796d073168f31141c6609/sessions/c69d0d277637ef6bc56d323ab127583e858e4d6f?auth_token=7fcbd3295ccd221f5a8e22a517622327d282166b1a36f7a18d5e07bcf87ef947

I can give you write access to the Repo so that the BrowserStack tests run without my needing to intervene. However, my feeling is that the code to dismiss the security prompt may actually be the problem. As mentioned, if called correctly with the noPrompts querystring this prompt should never show, so we shouldn't need to have code dealing with it. I would try removing that, and then if we find that there is a security prompt, we can fix that with the npPrompts...

Jaifroid:
@THEBOSS0369 An invite has been sent to you to accept write access to the Repo. Use this with extreme responsibility. It should mean the BrowserStack tests run when you push (after you've accepted the invite). I can give you access to the BrowserStack system so you can look at the videos for yourself. For that, please send me a private message on the Kiwix Slack.

Jaifroid:
(You may need to push directly to the Repo for the BS tests to run, rather than pushing from a fork, I'm not 100% certain. Try pushing as normal, and if it doesn't work I can explain how to push directly to a branch, so the tests run.)

THEBOSS0369:
> @THEBOSS0369 An invite has been sent to you to accept write access to the Repo. Use this with extreme responsibility. It should mean the BrowserStack tests run when you push (after you've accepted the invite). I can give you access to the BrowserStack system so you can look at the videos for yourself. For that, please send me a private message on the Kiwix Slack.

Got it sir i will do the pushing with extreme care without touching anything else.

THEBOSS0369:
All the tests are passing. Just need to run the BS test to ensure everything is working.

tests/e2e/spec/tonedear.e2e.spec.js
```diff
@@ -0,0 +1,207 @@
...
+                // Handle security alert if it appears
+                try {
+                    const securityAlert = await driver.wait(until.elementLocated(By.css('.modal[style*="display: block"]')), 3000);
+                    if (securityAlert) {
+                        const trustSourceButton = await driver.findElement(By.xpath("//button[contains(text(), 'Trust Source')]"));
+                        await trustSourceButton.click();
+                    }
+                } catch (e) {
+                    // No security alert found, continue with test
+                }
```
Jaifroid, line 100: As mentioned in the other PR, this can be removed now.

THEBOSS0369: Done sir, I forgot to remove this one 😅

THEBOSS0369:
@Jaifroid Sir I have checked the Code here is same as the dummy PR.

tests/e2e/spec/tonedear.e2e.spec.js
```diff
@@ -0,0 +1,196 @@
...
+                        await driver.wait(async function () {
+                            const images = await driver.findElements(By.css('img[alt="Get it on Google Play"], img[alt="Get the iOS app"]'));
+                            if (images.length === 0) return false;
+
+                            // Check if all images are visible
+                            const visibility = await Promise.all(images.map(async (img) => {
+                                return await img.isDisplayed();
+                            }));
+                            return visibility.every((isVisible) => isVisible);
+                        }, 10000, 'No visible store images found after 30 seconds');
```
Jaifroid, line 155: I think this message should read "10 seconds", right?

tests/e2e/spec/tonedear.e2e.spec.js
```diff
@@ -0,0 +1,196 @@
...
+    // Set the implicit wait to 3 seconds
+    driver.manage().setTimeouts({ implicit: 3000 });
+
+    // Run in both jquery and serviceworker modes by default
+    if (!modes) {
+        modes = ['jquery', 'serviceworker'];
+    }
+
+    modes.forEach(function (mode) {
+        const serviceWorkerAPI = true;
```
Jaifroid, line 34: You always set serviceWorkerAPI as true here, and it can't be changed (it's a const). Instead, we need to use a let here, and then change it if the serviceWorkerAPI isn't available, which is the case in some browsers.

tests/e2e/spec/tonedear.e2e.spec.js
```diff
@@ -0,0 +1,196 @@
...
+                await driver.wait(async function () {
+                    const elementIsVisible = await driver.executeScript(
+                        'var el=arguments[0]; el.scrollIntoView(true); setTimeout(function () {el.click();}, 50); return el.offsetParent;',
+                        modeSelector
+                    );
+                    return elementIsVisible;
+                }, 5000);
+                await driver.sleep(1300);
+
+                if (mode === 'serviceworker' && !serviceWorkerAPI) {
```
Jaifroid, line 59: This code block will never run because serviceWorkerAPI is stuck on true. You need to test whether serviceWorker API is supported or not. See the Gutenberg test to see how it's done.

tests/e2e/spec/tonedear.e2e.spec.js
```diff
@@ -0,0 +1,196 @@
...
+                    const filesLength = await driver.executeScript('return document.getElementById("archiveFiles").files.length');
+                    assert.equal(1, filesLength);
+                } else {
+                    await driver.executeScript('var files = arguments[0]; window.setRemoteArchives.apply(this, files);', [tonedearBaseFile]);
+                    await driver.sleep(1300);
+                }
+            });
+
+            it('Navigate to Android & iOS section', async function () {
+                if (!serviceWorkerAPI && mode === 'serviceworker') {
```
Jaifroid, line 93: This code will never run for the same reason.

tests/e2e/spec/tonedear.e2e.spec.js
```diff
@@ -0,0 +1,196 @@
...
+                const androidIosLink = await driver.wait(until.elementLocated(By.css('a[href="android-ios-ear-training-app"]')), 5000);
+                await androidIosLink.click();
+                // Switch back to default content before handling dialogs or verifying content
+                await driver.switchTo().defaultContent();
+                // Wait time
+                await driver.sleep(1000);
+            });
+
+            it('Verify Android and iOS store images in ' + (mode === 'jquery' ? 'Restricted' : 'ServiceWorker') + ' mode', async function () {
+                if (!serviceWorkerAPI && mode === 'jquery') {
```
Jaifroid, line 109: Ditto, code will never run.

Jaifroid (changes requested):
There appear to be some issues arising from the fact that the test for the serviceWorkerAPI was removed and it was set only to `true`. I think it's important to test for this, and the code is provided in Gutenberg test, so easy to add.

THEBOSS0369:
> There appear to be some issues arising from the fact that the test for the serviceWorkerAPI was removed and it was set only to `true`. I think it's important to test for this, and the code is provided in Gutenberg test, so easy to add.

@Jaifroid Sir i have added the code from the gutensberg, for const issue i tried by adding `let ` last time but i was getting the error from eslint that its not allowed i forgot that i can add a comment to avoid that issue. please take a look now and let me know

THEBOSS0369:
Sir the test failed like this last time as well. If we just simply re run it then it will pass. For confirmation i will edit the code in dummy pr.

THEBOSS0369:
The tests are passing in the dummy pr. So is should for this one as well this time.

THEBOSS0369:
Hello @Jaifroid Sir when you get some time let me know what to do further. Thanks

tests/e2e/spec/tonedear.e2e.spec.js
```diff
@@ -0,0 +1,233 @@
...
+    driver.manage().setTimeouts({ implicit: 3000 });
+
+    // Run in both jquery and serviceworker modes by default
+    if (!modes) {
+        modes = ['jquery', 'serviceworker'];
+    }
+
+    modes.forEach(function (mode) {
+        // eslint-disable-next-line prefer-const
+        let serviceWorkerAPI = true;
```
Jaifroid, line 35: @THEBOSS0369 I'm sorry, you've misunderstood me! I wasn't asking you merely to convert this from const to let. I was saying that you actually need to test whether the serviceWorkerAPI is available and then set it to true or false accordingly, before running the rest of the tests!

For an example of how this is done, see in the Gutenberg spec here: https://github.com/kiwix/kiwix-js/blob/main/tests/e2e/spec/gutenberg_ro.e2e.spec.js#L131

Jaifroid (changes requested):
The reason the ESLint required const, is because you never actually reassign the value of `serviceWorkerAPI`. This variable exists in order to determine whether the browser supports the Service Worker API or not. We initially set it to true, and then test if it is supported or not by testing for the alert when trying to switch to it. For some reason, that test has been omitted from your spec. You can easily put it back in with the example from the Gutenberg spec that I linked to.

THEBOSS0369:
@Jaifroid Thanks for clarification sir i have added the code. I removed that code carelessly thinking that it is only for approving dialogbox and didn't read the comment of sw api inside it😅, this won't happen from next time .

Jaifroid:
> @Jaifroid Thanks for clarification sir i have added the code. I removed that code carelessly thinking that it is only for approving dialogbox and didn't read the comment of sw api inside it😅, this won't happen from next time .

No problem at all! I'll check.

Jaifroid (approved):
@THEBOSS0369 Thank you for this excellent and very useful PR. It'll make a real difference to the KJS repository. Congratulations! 🎉 I think you have sufficient access to squash and merge it yourself using the button in the GitHub web interface. Let me know if you have any difficulties. But please be sure to use the "Squash and merge" option, not the other two options).

Issue #1288, opened 2024-11-26
[Improvement] Home page | In tablet and mobile devices, the 'Select random' icon should have the same size as the search input and


CREDO23:
Desktop view ✅ 

![Image](https://github.com/user-attachments/assets/29fcce70-8f00-4d10-a828-3023acf8d9be)

Tablet, mobile (need improvement)

![Image](https://github.com/user-attachments/assets/c6ae7b68-587c-42a4-b914-0321ea3e6982)

Jaifroid:
Thanks for the report. I suppose the CSS that resizes the UI is not being applied uniformly at some screen widths.

FizanMuhammedFaisal:
can I work on this

asmit27rai:
@Jaifroid I Want To Work On This Issue ?

Jaifroid:
> [@Jaifroid](https://github.com/Jaifroid) I Want To Work On This Issue ?

There's already a PR fixing this.

PR #1294, opened 2024-12-05
Fixed issue #1288, random-button same size as search button

Commanderk3:
Issue #1288   The shuffle button besides the search button was changing its dimensions differently compared to search button when viewed on mobile devices like tablet or phone. The issue is fixed now. Also this is my first official PR  so pardon me if I did something wrong.

### DIFF
--- a/www/css/app.css
+++ b/www/css/app.css
@@ -354,48 +354,44 @@ button {
 /* Reduce the size of some elements for small screens */
 @media (max-width: 768px) {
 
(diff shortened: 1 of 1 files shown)


Jaifroid:
Thank you for your PR. Unfortunately, this overrides the dynamic resizing of the buttons as the screen gets narrower (for example, on mobile phone screen sizes). Another developer has offered a simpler solution that maintains the dynamic button resizing. But thanks once again for your work on this.

PR #1296, opened 2024-12-06
Dummy PR for Adding selenium ui test

THEBOSS0369:
This PR is a Dummy PR for Adding selenium ui test and is successor of PR #1286 created for ensuring tests passing or not in browser stack.

### DIFF
--- a/tests/e2e/spec/tonedear.e2e.spec.js
+++ b/tests/e2e/spec/tonedear.e2e.spec.js
@@ -249,7 +249,7 @@ function runTests (driver, modes) {
                 await driver.wait(async function () {
                     const contentAvailable = await driver.executeScript('return document.querySelector(\'a[href="android-ios-ear-training-app"]\') !== null;');
                     return contentAvailable;
-                }, 20000); // Increased to 10 seconds for more loading time
+                }, 10000); // Increased to 10 seconds for more loading time
 
                 // Find the "Android & iOS App" link
                 const androidLink = await driver.findElement(By.css('a[href="android-ios-ear-training-app"]'));
--- a/tests/e2e/spec/tonedear.e2e.spec.js
+++ b/tests/e2e/spec/tonedear.e2e.spec.js
@@ -293,28 +293,50 @@ function runTests (driver, modes) {
                 `);
                 console.log('Iframe state:', iframeState);
 
+                // Fix for iframe visiblity if hidden
+                if (!iframeState.visible) {
+                    console.log('Iframe is hidden, trying to fix...');
+                    await driver.executeScript(`
+                        const iframe = document.getElementById('articleContent');    
+                        if (iframe) {
+                            iframe.style.display = 'block';
+                            iframe.style.visibility = 'visible';
+                        }
+                    `);
+                }
+
                 // For older Firefox, try alternative content loading method
                 if (isOldFirefox) {
                     console.log('Attempting alternative content loading for Firefox 70...');
-                    await driver.executeScript(`
-                        const iframe = document.getElementById('articleContent');
-                        if (iframe && window.app && window.app.selectedArticle) {
-                            // Try direct content injection
-                            const content = window.app.selectedArticle.content;
-                            if (content) {
+                    const content = await driver.executeScript(`
+                            return window.app && window.app.selectedArticle ? window.app.selectedArticle.content : null;
+                        `);
+                    console.log('Selected Article Content:', content);
+
+                    if (content) {
+                        await driver.executeScript(`
+                            const iframe = document.getElementById('articleContent');
+                            if (iframe && iframe.contentDocument) {
                                 iframe.contentDocument.open();
-                                iframe.contentDocument.write(content);
+                                iframe.contentDocument.write('<p>Test Content</p>');
                                 iframe.contentDocument.close();
                             }
-                        }
-                    `);
-                    await driver.sleep(2000);
+                        `);
+                        await driver.sleep(2000);
+                    } else {
+                        console.log('Content is empty or unavailable');
+                    }
                 }
 
                 // Switch to iframe
                 console.log('Switching to iframe...');
                 await driver.switchTo().frame(iframe);
                 console.log('Successfully switched to iframe');
+
+                // Debugging network requests for resource loading
+                await driver.executeScript(`
+                    window.performance.getEntriesByType('resource').forEach(entry => console.log(entry.name));  
+                `);
 
                 // Wait for content with different strategies based on browser
                 console.log('Waiting for content...');
@@ -338,6 +360,14 @@ function runTests (driver, modes) {
                             await driver.sleep(1000);
                             return false;
                         }
+
+                        const links = await driver.executeScript(`
+                            return Array.from(document.querySelectorAll('a[href="android-ios-ear-training-app"]')).map(link => ({
+                                href: link.href,
+                                displayed: getComputedStyle(link).display !== 'none'
+                            }));
+                        `);
+                        console.log('Links:', links);
 
                         try {
                             const androidLink = await driver.findElement(By.css('a[href="android-ios-ear-training-app"]'));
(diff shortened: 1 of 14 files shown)


THEBOSS0369:
I will be resuming my work on this issue Now. Thanks for your patience 😃

THEBOSS0369:
I am just on the last stage tests are passing in every browser except Firefox 70 i will try to fix that tomorrow and ms edge's failing due to tests termination (ie. tests are not starting have to ask the maintainer)

THEBOSS0369:
Hey @Jaifroid Sir i have fixed the issues that were coming now the test **Passes in every browser** except **Firefox 70 and MS Edge 18** 
Proof of Test passing on chrome , safari
safari

**1. Firefox 70 Issue:**
The test is not passing because there is a popup coming which is saying that android-ios didn't found i tried every method 
by clicking the link from the navbar, tried by finding all element related to android ios and then navigating to it. Still its not navigating to android Page however this popups doesn't shows in any other browser except FF 70 (**I think this is a browser issue or maybe i'm wrong i need your help here**)
image

**2. MS Edge 18 Issue:**
This issue is not due to popup or from  code instead its a timeout issue. The tests are not running, I think may there is a limit set for the overall test to run and probably when starting tonedear tests its exceeding its limit. 
image

Sir These are the issues that i'm facing , Please Help. Thanks!

Jaifroid:
@THEBOSS0369 Thank you for getting this far! I'll take a look and see if I can troubleshoot.

THEBOSS0369:
@Jaifroid Sir i found why its happening see the video below
I was testing for separate tab issue i got the same dialog box appearing that is coming in firefox 70. We can manually press the button it will work fine but when the system does it that dialog box appears. I don't know why this issue didn't occured in any other browser while testing. 

To recreate this i used the Vim Motion Key extension in browser otherwise by mouse it works fine
https://github.com/user-attachments/assets/d93c4ba7-40ee-4750-95a3-940d92bd1cba

This issue is probably in zim because i was running the server on MS Edge latest version.

Jaifroid:
@THEBOSS0369 The error in that video is caused by an incorrect function call to clear the spinner. This has now been corrected in https://github.com/kiwix/kiwix-js/pull/1299, which I've merged into this branch to see if it makes any difference. However, I'm not convinced this is exactly the issue you're facing in tests, but let's see. Please be sure to sync your local copy of this branch before adding code, and don't force-push any other changes.

Jaifroid:
The issue you are facing is caused by a BrowserStack timeout, as you had guessed previously. We need to increase the timeout or else try running the tests in parallel for sessions liable to time out though this might make for confusing display in console. I'm trying the second approach, let's see.

Jaifroid:
False positive -- tests don't actually run properly via this method.

THEBOSS0369:
@Jaifroid  Sir Thank you for fixing the Edge's test failing, I have added the test in Firefox 70 let's see if it passes this time or not.

THEBOSS0369:
U were right sir, The error is something else, the popup still shows up
image

Jaifroid:
Unfortunately, Edge Legacy isn't fixed, the Promise code didn't work, merely masked the errors because the script ended prematurely. BrowserStack videos show that the last test still timed out.

THEBOSS0369:
> Unfortunately, Edge Legacy isn't fixed, the Promise code didn't work, merely masked the errors because the script ended prematurely. BrowserStack videos show that the last test still timed out.

Oh no, I didn't noticed that.

Jaifroid:
So I'm reverting changes tried so far and now trying to increase timeouts to solve Edge Legacy test first. Then we can look at Firefox 70, though I won't get to that till the weekend.

Jaifroid:
Maximum timeout of 300 seconds isn't complete. We need now to restructure the tests so that the window isn't opened and left idling before receiving any commands. The approaches tried to make this happen so far haven't worked because runTests isn't fully async, so synchronous parts of the tests get executed immediately even when chained within Promises.

Restructuring this is more involved than I can do today, so this will have to wait. In the meantime, I've disabled all BS tests except Edge Legacy so we can focus on that one.

Jaifroid:
So, I solved the Edge Legacy tests by running them in a single driver, so that the other drivers are not hanging around waiting for tests to complete. This works fine and the pattern in https://github.com/kiwix/kiwix-js/blob/selenium-uitest/tests/e2e/runners/edge/edge18.bs.runner.js#L40 can be used in other cases where we have any timeout issue.

I also added stronger prevention of the security popup in https://github.com/kiwix/kiwix-js/pull/1296/commits/6178c1c0a281e6aa942f5715e86d78172d4455cc, so that shouldn't bother you now.

You may find that the issue in Firefox 70 is now fixed. In any case, you can now go ahead and test that!

Fingers crossed.... 🤞

tests/e2e/spec/tonedear.e2e.spec.js
```diff
@@ -0,0 +1,220 @@
...
+                    assert.strictEqual(iosDataSnippet, expectedIosSnippet, 'iOS image data matches expected');
+                } else if (serviceWorkerAPI && mode === 'serviceworker') {
+                    try {
+                        // ServiceWorker mode test for image loading
+                        await driver.sleep(3000);
+
+                        const swRegistration = await driver.executeScript('return navigator.serviceWorker.ready');
+                        assert.ok(swRegistration, 'Service Worker is registered');
+
+                        console.log('Current URL:', await driver.getCurrentUrl());
```
Jaifroid, line 163: Just a reminder to remove test logging once no longer needed.

tests/e2e/spec/tonedear.e2e.spec.js
```diff
@@ -0,0 +1,220 @@
...
+                    if (activeAlertModal) {
+                        serviceWorkerAPI = await driver.findElement(By.id('modalLabel')).getText().then(function (alertText) {
+                            return !/ServiceWorker\sAPI\snot\savailable/i.test(alertText);
+                        });
+                        const approveButton = await driver.wait(until.elementLocated(By.id('approveConfirm')));
+                        await approveButton.click();
+                    }
+                } catch (e) {
+                    // Do nothing
+                }
```
Jaifroid, line 70: @THEBOSS0369 I think you could remove these lines. Although it's not the cause of the current error, it will never now find this element, and it introduces overhead (and confusion about the cause of failure in the logs). I'll test on a local copy of FF70 to see if there's an obvious cause of the failure to navigate to the Audio app page.

Jaifroid:
OK, I reproduce the error in my local copy of FF70 (via PortableApps). The problem appears to be more generic: an incompatibility in the Zimit wombat setup script. It seems `replaceAll` isn't supported in this version of Firefox. See screenshot.

image

Jaifroid:
Indeed, replaceAll isn't supported in FF < 77:

https://caniuse.com/?search=replaceAll

I can't remember, but I think we polyfill it in legacy MS browsers. Will check.

Jaifroid:
OK, this file wombatSetup.js is loaded inside the ZIM, so we can't easily polyfill it. It's just an incompatibility with Firefox < 77 when running Zimit archives.

What I suggest is that you run the test only in jQuery (Restricted) mode for the FF70 test, as we know that's passing, and we can't work around this error. Nothing to do with your test, Zimit archives will just fail in FF < 77, and users should switch to Restricted mode. I'll see if I can fix that as an automatic switch, but it's another issue.

So, please don't spend more time on this, just run this tonedear test only in jQuery mode.

THEBOSS0369:
100% Agree sir I was also thinking the same way but I was thinking that we should skip the Verification of Images as all other tests are passing in sw mode except this. Let me know whether we should skip the verification of images test specifically for sw mode or should we follow as you suggested. 

Also I will be able to do the changes tomorrow morning.

THEBOSS0369:
@Jaifroid  Got it sir, that makes sense. I have removed the extra code lines and changed the code for test to skip sw mode for tonedear. i'm waiting for the tests to pass. Once these test passes i will copy the codes in the main PR from where we started.

THEBOSS0369:
Hey @Jaifroid Sir i have removed the extra lines. If everything is completed then shall i close this PR or is anything left for this one?

Jaifroid:
@THEBOSS0369 Thank you! I think you can update the other PR with the latest code, but keep this open for now in case we need to do any last-minute BrowserStack testing. But it's fine to finalize things on the other PR. When the code there is identical to the code here, let me know and I'll do a proper review.

Jaifroid:
@THEBOSS0369 When you've squashed and merged the other PR, please close this one, thanks!

Jaifroid:
(without merging, I mean!)

PR #1301, opened 2025-01-14
Converted Q Unit tests to Mocha Tests

### LINKED ISSUE #1187: Convert QUnit tests to mocha
Since we already use mocha for assertion in e2e tests, it would make sense to use it in Unit tests also. We currently run Unit tests in UI-based framework (QUnit) that runs directly in browsers, and as a result we have to run them in a browser-testing suite (testcafe) instead of directly in a Node framework like mocha. This adds overhead and is counterintuitive. It is also confusing, because there seems to be some blurring of lines between Unit and e2e testing.

I think it's time to drop QUnit.

THEBOSS0369:
This PR Fixes #1187 

## Description
In this PR i have converted Q Unit tests to Mocha Tests.

## Tests Converted
All the tests from `unit/spec/test.js` file are converted into mocha tests. It can be found in `unit/spec/mocha.test.js`

Commands to Run the test:

`npm run test-unit` : This command runs the unit tests for the application to ensure that individual units of code (such as functions or methods) are working as expected.

`npm run test-unit-watch` : This command allows us to continuously run tests while we work on your code. If we change a test file or the code it tests, Mocha will automatically re-run the tests.

`npm run test-unit-coverage` : This command runs the unit tests with code coverage reporting using NYC, which is a tool for generating code coverage reports.

## Test
1. Checked in both "**Restricted**" and "**ServiceWorker**" Everything is working fine.
6. End-to-end (e2e) `tests-e2e-firefox`-> All tests Passed.
8. Browser Test -> Microsoft Edge.

## Screenshot for  converted tests Passing
image


THEBOSS0369:
These tests are falling because of the changes done in init.js file, I am fixing that right now

Jaifroid:
@THEBOSS0369 Thank you very much for this, that's great! I'll take a look as soon as I can. Looks like a great improvement. @audiodude maybe you have some comments?

Jaifroid:
I've done some basic tests by switching to the PR/branch, running npm install, and npm run test-unit and test-unit-coverage. Congratulations, this is very impressive conversion work!

Just quickly, I noticed that running these commands create a bunch of temporary files in the Repo that have not been correctly ignored (see screenshot). If the coverage files are overwritten on each coverage run, then they can remain, but you should .gitignore the coverage folder. Regarding the .nyc_output folder, it should also be .gitignore'd, but these look like files with a PID, and so I imagine will be created every run and could become a nuisance in that folder. Ignoring the folder may not be enough. Please check if there is a way to prevent them being generated, or whether they get auto-cleaned up (which would be fine).

image

Jaifroid:
I haven't yet reviewed the code itself.

THEBOSS0369:
@Jaifroid  Sir the temp files that you are seeing is not from the code instead it is due to the command `npm run test-unit-coverage` which is present in `package.json` file. The  json files are the report files. The report gets updated when we re run the command and there is a change in files. 
I didn't thought of  that it would install the files in the main repo itself. However if you want i can delete the command so that these files wouldn't get downloaded and once this pr passes all the ticks we can add the command in the docs so if anyone wants report they can directly copy paste the command. Let me know what you think

This is a blog for a reference https://zinserjan.github.io/mocha-webpack/docs/guides/code-coverage.html

Jaifroid:
Yes, I was aware that these reports were from running the command, but we don't want them being accidentally pushed to the server by a dev, so the folders containing them should at least be added to `.gitignore`. I don't think it's necessary to remove the command to invoke the code coverage test. I noticed the coverage hadn't been fully set up, it doesn't list any files and coverage currently.

THEBOSS0369:
> Yes, I was aware that these reports were from running the command, but we don't want them being accidentally pushed to the server by a dev, so the folders containing them should at least be added to `.gitignore`. I don't think it's necessary to remove the command to invoke the code coverage test. I noticed the coverage hadn't been fully set up, it doesn't list any files and coverage currently.

Got it sir i will do it, and yes right now it just shows the table's heading and empty value I  wanted to confirm whether this feature is needed or not so i created as an example. Once conversion work of test is approved, I will make it working.

THEBOSS0369:
Hey @Jaifroid sir let me know what to do further in this pr, when you get some time.

Jaifroid:
Thanks! I noticed fallback errors in the test run due to Fetch not being supported in the test environment (see screenshot):

image

To avoid this, we should add a test for 'Fetch' support in the files `xzdec_wrapper.js` and `zstddec_wrapper.js`. It is sufficient to add this in each file where we test for WebAssembly support:

`if ('WebAssembly' in self && 'Fetch' in self) {` (the new code is `&& 'Fetch' in self`).

Add this specifically on these lines:

* https://github.com/kiwix/kiwix-js/blob/main/www/js/lib/xzdec_wrapper.js#L34
* https://github.com/kiwix/kiwix-js/blob/main/www/js/lib/zstddec_wrapper.js#L98

tests/unit/spec/mocha.test.js
```diff
@@ -0,0 +1,319 @@
...
+    zimArchiveFiles = await Promise.all(splitBlobs);
+    return new Promise((resolve) => {
+        localZimArchive = new zimArchive.ZIMArchive(zimArchiveFiles, null, function () {
+            resolve();
+        });
+    });
+});
+
+describe('Environment', function () {
+    it('Mocha Test should be properly configured', function () {
```
Jaifroid, line 58: I think it's otiose to write "should" in every test. More usual would be to write simply `Configure Mocha Test`.  If it has a tick, then it passed, if it has a red cross, it failed.

tests/unit/spec/mocha.test.js
```diff
@@ -0,0 +1,319 @@
...
+        expect('test').to.equal('test');
+    });
+
+    it('should have loaded archive files', function () {
```
Jaifroid, line 62: Similarly here, simply `Load archive files`...

tests/unit/spec/mocha.test.js
```diff
@@ -0,0 +1,319 @@
...
+        expect(zimArchiveFiles && zimArchiveFiles[0] && zimArchiveFiles[0].size > 0).to.be.true;
+    });
+});
+
+describe('Utils', function () {
+    it('should correctly read an IEEE_754 float from 4 bytes', function () {
```
Jaifroid, line 68: Here: `Correctly read an .....`

tests/unit/spec/mocha.test.js
```diff
@@ -0,0 +1,319 @@
...
+        const byteArray = new Uint8Array([194, 237, 64, 0]);
+        const float = util.readFloatFrom4Bytes(byteArray, 0);
+        expect(float).to.equal(-118.625);
+    });
+
+    it('should handle upper/lower case variations', function () {
```
Jaifroid, line 74: Here `Handle upper/lower case variation`

tests/unit/spec/mocha.test.js
```diff
@@ -0,0 +1,319 @@
...
+        const testCases = [
+            { input: 'téléphone', expected: 'Téléphone', description: 'The first letter should be uppercase' },
+            { input: 'Paris', expected: 'paris', description: 'The first letter should be lowercase' },
+            { input: 'le Couvre-chef Est sur le porte-manteaux', expected: 'Le Couvre-Chef Est Sur Le Porte-Manteaux', description: 'The first letter of every word should be uppercase' },
+            { input: 'épée', expected: 'Épée', description: 'The first letter should be uppercase (with accent)' },
+            { input: '$￥€"«xριστός» †¡Ἀνέστη!"', expected: '$￥€"«Xριστός» †¡ἀνέστη!"', description: 'First non-punctuation/non-currency Unicode letter should be uppercase, second (with breath mark) lowercase' },
+            { input: 'Καλά Νερά Μαγνησία žižek', expected: 'ΚΑΛΆ ΝΕΡΆ ΜΑΓΝΗΣΊΑ ŽIŽEK', options: 'full', description: 'All Unicode letters should be uppercase' }
```
Jaifroid, line 81: In these cases, it's OK to keep "should".

tests/unit/spec/mocha.test.js
```diff
@@ -0,0 +1,319 @@
...
+            const results = options
+                ? util.allCaseFirstLetters(input, options)
+                : util.allCaseFirstLetters(input);
+
+            expect(results, description).to.be.an('array');
+            expect(results.some(result => result === expected), description).to.be.true;
+        });
+    });
+
+    it('should remove parameters from URLs', function () {
```
Jaifroid, line 94: Just `Remove parameters from URLs`

tests/unit/spec/mocha.test.js
```diff
@@ -0,0 +1,319 @@
...
+            // Compare normalized strings (removing special characters)
+            const normalizedResult = decodeURIComponent(result).normalize('NFD').replace(/[\u0300-\u036f]/g, '');
+            const normalizedExpected = baseUrl.normalize('NFD').replace(/[\u0300-\u036f]/g, '');
+            expect(normalizedResult).to.equal(normalizedExpected);
+        });
+    });
+});
+
+describe('ZIM initialization', function () {
+    it('should set archive as ready', function () {
```
Jaifroid, line 113: For this and all other "should" below, just change to, e.g., `Set archive as ready` (etc.).

Jaifroid (changes requested):
@THEBOSS0369 It's looking good. Just some minor suggestions to clean up the output display.

THEBOSS0369:
> Thanks! I noticed fallback errors in the test run due to Fetch not being supported in the test environment (see screenshot):
> 
> image
> 
> To avoid this, we should add a test for 'Fetch' support in the files `xzdec_wrapper.js` and `zstddec_wrapper.js`. It is sufficient to add this in each file where we test for WebAssembly support:
> 
> `if ('WebAssembly' in self && 'Fetch' in self) {` (the new code is `&& 'Fetch' in self`).
> 
> Add this specifically on these lines:
> 
> * https://github.com/kiwix/kiwix-js/blob/main/www/js/lib/xzdec_wrapper.js#L34
> * https://github.com/kiwix/kiwix-js/blob/main/www/js/lib/zstddec_wrapper.js#L 98

@Jaifroid  Thanks you for this sir , this was such a help i was trying to fix this in files `zstddec_asm.js` and wasm and similar types of files and was just about to ask you about this. I have fixed it now and other typos as you suggested.  please take a look

THEBOSS0369:
Progress status: I'm currently trying to fix npm run test-unit-coverage isse. Its taking longer than i expected it to plus my winter's holiday has ended 2 weeks ago so i am not getting enough time to focus only on fixing it but i will try to fix it till this weekend, if i ain't able to make it, i will probably won't add the feature in this pr. would that be fine sir? @Jaifroid

Jaifroid:
> Progress status: I'm currently trying to fix npm run test-unit-coverage isse. Its taking longer than i expected it to plus my winter's holiday has ended 2 weeks ago so i am not getting enough time to focus only on fixing it but i will try to fix it till this weekend, if i ain't able to make it, i will probably won't add the feature in this pr. would that be fine sir? @Jaifroid

Thanks for the update. See what you can do this weekend, and if it's too complicated, maybe you could just make it into a separate issue to "Add Unit test coverage".

THEBOSS0369:
Hello @Jaifroid sir i have fixed it but its creating so many problems in other workflow of the repo's. 

First is that the main test are failing if i add the functionality it shows error message in files like `urlutils.js , settingstore.js`... and when i fix them it disturb the main workflow
second is the **loaders** which is used for **testing** will most **probably be removed in future** which will most probably disturb the workflow of repo, and when i use the **instead option** showing in below ss, it does not suite with other things.

example is this 
 image

coverage report are showing but with error messages
image

I am not very confident with this so i am just commiting the code to show you. I am just confirming again that i should remove this functionality from this pr and go ahead with the creating an issue or should i work on it more.

Its just keep getting complicated  & more complicated and i don't want to mess with the codebase. Let me know what to do thanks.

THEBOSS0369:
I will revert the recent commit

tests/unit/coverage-setup.cjs
```diff
@@ -0,0 +1,26 @@
+const { createInstrumenter } = require('istanbul-lib-instrument');
+const { resolve } = require('path');
+const { readFileSync } = require('fs');
```
Jaifroid, line 3: @THEBOSS0369  We're not using CommonJS modules. You should only use the ES Module `import` statements.

.babelrc
```diff
@@ -0,0 +1,4 @@
+{
+  "presets": ["@babel/preset-env"],
+  "plugins": ["@babel/plugin-transform-modules-commonjs"]
```
Jaifroid, line 3: @THEBOSS0369 Don't transform modules to CJS format, that will lead to loads of issues.

Jaifroid:
@THEBOSS0369 I didn't review the whole of the latest commit, but I immediately noticed you're using the wrong type of modules. We use ES6 Modules, with `import` statements, not CommonJS modules with `require` statements. If every you are copying code from Stack Exchange that has `require` in it, then it won't work with our project. It took a lot of work to get rid of the old moduling system and modernize it to ES6!

Overall, my view is that you should revert the code coverage commit and make it into a separate issue. It's enough that you've converted our ancient tests to a modern format. We know the coverage is extremely poor currently, and we need to work on adding more tests for basic units of code. Being told on each run that our coverage is poor won't really help the situation, even though it's a good idea in principle! Many thanks!

THEBOSS0369:
> @THEBOSS0369 I didn't review the whole of the latest commit, but I immediately noticed you're using the wrong type of modules. We use ES6 Modules, with `import` statements, not CommonJS modules with `require` statements. If ever you are copying code from Stack Exchange that has `require` in it, then it won't work with our project. It took a lot of work to get rid of the old moduling system and modernize it to ES6!
> 
> Overall, my view is that you should revert the code coverage commit and make it into a separate issue. It's enough that you've converted our ancient tests to a modern format. We know the coverage is extremely poor currently, and we need to work on adding more tests for basic units of code. Being told on each run that our coverage is poor won't really help the situation, even though it's a good idea in principle! Many thanks!

Got it sir i have reverted the changes and yess i have been searching out the internet to make it work from last 2 weeks as i have never worked on something like this before 😅🫠. 
I think the work is done now so do i need to do anything else on this pr? if not then i will move to different issue.

Jaifroid:
Thanks, @THEBOSS0369 🙏. I shall review ASAP.

THEBOSS0369:
Hello @Jaifroid Sir i have my exams going on and it will end on 10 march . so there won't be any problem if you can review this pr after 10th march. Thanks.

Jaifroid:
@THEBOSS0369 I hope your exams went well! Let's get this PR finished and merged!

It strikes me that what remains to be done is to integrate the new tests into the workflows, and remove testcafe, as this won't be needed going forwards. There is also the old tests.js code to remove.

I notice that some security updates have introduced a conflict in your PR. You may be able to resolve this by running `npm audit fix` in the root of this repo for your PR. Then `npm install`, and push the changes to package-lock.json and package.json.

Jaifroid:
Also, `npm test` should be changed to run the new tests locally rather than with testcafé.

THEBOSS0369:
@Jaifroid Sir Thanks for that my exams went very well!

I have done **integrated the test** in the workflows, `deleted the test.js & inittestcafe.js file and removed the testcafe dependency
` . The tests are working fine in my local setup. Please review and let me know if anything else is required.

Jaifroid:
Hello, I'm Claude, an AI assistant analyzing this PR on behalf of a reviewer. I'd like to better understand the changes in this PR by examining the diff. Could someone point me to where I can access the changes or provide a summary of the key files being modified and what changes were made? Thank you!

tests/unit/js/init.js
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable no-useless-constructor */
 /**
  * init.js : Configuration for the library require.js
```
Jaifroid, line 3: This old description is out of date as we don't use require.js any more. Please change the description to "Global configuration of the app".

THEBOSS0369:
> It looks like a rogue `require` in dependencies, which probably won't work because we're using ESM modules. However, I can't find `require` in the file it mentions (mocha.test.js line 4).
> 
> Does the command `test-unit-watch` work for you?

@Jaifroid sir i didn't noticed it earlier and yes its not working for me as well. Most probably the reason you stated is correct although it was working last month am not sure though what happened, i looked on google wasn't able to find anything other than Module compatible issue. So i am removing it as its not working as expected and our main command npm-test-unit shows the result in fraction of second so i don't think other developers may find any time issues while using it.

Also my log shows a little bit different result
here 

```powershell
npm run test-unit-watch

> kiwix-js@4.1.2 test-unit-watch
> mocha --watch --require @babel/register 'tests/unit/spec/**/*.test.js'

Error [ERR_REQUIRE_ESM]: require() of ES Module /home/boss/realcode/opnsrc/kiwix/kiwix-js/tests/unit/spec/mocha.test.js from /home/boss/realcode/opnsrc/kiwix/kiwix-js/node_modules/mocha/lib/mocha.js not supported.
Instead change the require of mocha.test.js in /home/boss/realcode/opnsrc/kiwix/kiwix-js/node_modules/mocha/lib/mocha.js to a dynamic import() which is available in all CommonJS modules.
    at Object.newLoader [as .js] (/home/boss/realcode/opnsrc/kiwix/kiwix-js/node_modules/pirates/lib/index.js:121:7)
    at /home/boss/realcode/opnsrc/kiwix/kiwix-js/node_modules/mocha/lib/mocha.js:416:36
    at Array.forEach (<anonymous>)
    at Mocha.loadFiles (/home/boss/realcode/opnsrc/kiwix/kiwix-js/node_modules/mocha/lib/mocha.js:413:14)
    at Mocha.run (/home/boss/realcode/opnsrc/kiwix/kiwix-js/node_modules/mocha/lib/mocha.js:988:10)
    at Object.run (/home/boss/realcode/opnsrc/kiwix/kiwix-js/node_modules/mocha/lib/cli/watch-run.js:263:22)
    at FSWatcher.<anonymous> (/home/boss/realcode/opnsrc/kiwix/kiwix-js/node_modules/mocha/lib/cli/watch-run.js:184:14)
```

Jaifroid (approved):
Great, tests are passing, and I think this is ready now! Thank you very much for this great PR 🙏😊. Please feel free to squash and merge, or tell me if you don't have access to do that (I think you should have access).

THEBOSS0369:
> Great, tests are passing, and I think this is ready now! Thank you very much for this great PR 🙏😊. Please feel free to squash and merge, or tell me if you don't have access to do that (I think you should have access).

I have the access sir and i am squashing and merging it now

PR #1307, opened 2025-02-08
Improve popover placement and style support

Jaifroid:
This PR ports code from Kiwix PWA which supports the new behaviour of the CSS `zoom` property in Chrome >= 128 and Chromium-based browsers. We still have to test for the old behaviour, and exempt Safari and old MS browsers. See https://caniuse.com/css-zoom and https://github.com/kiwix/kiwix-js-pwa/issues/672 for more details.

This also ports some style improvements in dark mode.


Jaifroid:
In particular [this comment](https://github.com/kiwix/kiwix-js-pwa/issues/672#issuecomment-2496065340) explains that Chrome enforced standardization from Chrome 128 after Firefox introduced css zoom support in Firefox 126.

Jaifroid:
Needs testing on Browser Stack especially with Safari.

Jaifroid:
No issues in Safari on BrowserStack.

PR #1319, opened 2025-03-16
Run tests on ubuntu latest

Jaifroid:
As Ubuntu 20.04, which we were using for tests, has now entered brownout, we need to run our workflow tests on a more modern Ubuntu. I suggest simply using ubuntu-latest, as our tests don't depend on the Ubuntu version as far as I can see. Let's see how the tests run.


Jaifroid:
Strangely, the Windows tests are failing in IE mode, but they're also failing on another branch in exactly the same spot, so it's not related to this PR.

Jaifroid:
@abhijithm34 These two PRs aren't connected, but they're both failing in the same place. It could be a transitory issue. I need to re-run.

PR #1324, opened 2025-03-23
Attempt fixes for legacy Ray Charles test on Firefox for linux-latest

### LINKED ISSUE #1323: End-to-end tests are failing on Firefox on linux-latest
We are getting stale element errors and other inconsistent errors periodically causing tests to fail even though code changes unrelated or very small aspects of the app. They'll often pass again when re-run. After a rock-steady period, this has begun to happen recently.

For some reason this is mostly happening on Linux, not on Windows-based tests. But that could also because only the Linux tests run the full BrowserStack suite, so there are more potential failure points.

Jaifroid:
Fixes #1323. Just trying some things.


Jaifroid:
Tests show that Firefox on GitHub is fixed, but now Safari is failing on BrowserStack. Grrrr.

Jaifroid:
So, autorunning Unit Tests for every single browser tested seemed to make them flaky -- this was an error remaining in the new Mocha test conversion when it was merged. Cleaning that up and ensuring good separation between unit and e2e tests seems to have stabilized it, as we don't have two lots of conflicting ZIM parts loaded in memory.

I've also removed QUnit and updated TEST documentation.

PR #1327, opened 2025-03-25
fix: show selected file in dropdown on refresh

### LINKED ISSUE #1316: Bug: Selected file not persisting in the dropdown after reopening the app
If we select a file in the folder, it gets added to the home. However, once we refresh, the dropdown loses the state of the selected file, showing a "Select an archive..." message. However, the file is still selected in the app, and the document can be accessed in the "Home" tab.

<details>
<summary>Video Reference</summary>

https://github.com/user-attachments/assets/1c68a7fc-bf6c-4fdf-8fdd-ed97883ee0e6

</details>

## Expected Result
The file should be selected, the state should persist in the dropdown even after reopening the app

Note: A continued discussion from #1311

dev-shetty:
Fixes #1316 

### 1. Persisting the file name state in dropdown
- The change fetches the previously selected zim file from the localStorage and then sets it as the selected file.
- Didn't add any comments, as it seemed self-explanatory, and the intended behavior, as there were functions already built to handle it, more or less feels like a miss from past iteration.

<details>
<summary>Fix Demonstration</summary>

https://github.com/user-attachments/assets/2295b52a-09d1-4110-9f4e-5bcfc3c4c9e6

</details>

---

### 2. Limiting Dropdown Size
- I tried implementing a few solutions, but came to know there is no support for limiting the options list natively on browser HTML. 
- There is a size attribute, but it removes the modal kindof dropdown and makes it inline in the page

<details>
<summary>Video demo of how it looks with `size=10` attribute</summary>

https://github.com/user-attachments/assets/9786e465-6242-4a9a-9f03-833e1628821c

</details>

- Even with using javascript we need to play around with absolute positionings, and wont even nearly give the same native dropdown experience which we get right now. 

- I can try the hacky solution, but rather wont it be better if we create a new dropdown component with our custom css, and it will be easier to control the length as well? Thoughts @Jaifroid 

### Reference

- [How can I limit the visible options in an HTML select dropdown?](https://stackoverflow.com/questions/8788245/how-can-i-limit-the-visible-options-in-an-html-select-dropdown)

- [How do I limit the number of options displayed in an HTML Select element dropdown menu?](https://stackoverflow.com/questions/5538330/how-do-i-limit-the-number-of-options-displayed-in-an-html-select-element-dropdow)

- [Preview the size option on select](https://www.w3schools.com/TAgs/tryit.asp?filename=tryhtml_select_size)


Jaifroid:
@dev-shetty Thanks for the info and the video. The style shown there (with `size=10`) seems better to me than the current situation. Why? Because a user can see immediately a list of the files in a directory, without needed to click anything. If they have a small number of files, they can see them all. If a large number, the list scrolls. The only bit of JS needed would be to set the size correctly if less than 10, but that should be easy. This is indeed the solution adopted in the PWA, which seems easier to use to me. What are your thoughts?

dev-shetty:
Agree with your thoughts on that, makes sense from the product side of things, as user can see everything in a glance.

Sounds good to me, I will add the support for that

dev-shetty:
@Jaifroid I have pushed the changes for the select dropdown, let me know if you need any changes.

<details>
<summary>When files exceed 10</summary>
Screenshot 2025-03-27 at 1 21 29 AM
</details>

<details>
<summary>When files are lesser than 10</summary>
Screenshot 2025-03-27 at 1 21 16 AM
</details>

Jaifroid:
Tests are passing. I'll review now. Thanks for the screenshot, looks good!

Jaifroid (commented):
I like the new functionality! One slight improvement would be to reverse the logic of the block starting on line 113:
```
    if (files.length !== 0) {
        const placeholderOption = new Option(translateUI.t('configure-select-file-first-option'), '');
        placeholderOption.disabled = true;
        select.appendChild(placeholderOption);
    };
```
This block adds the "Select an archive" text if `files.length !== 0`. However, that's no longer needed, and we should display that if the file length is 0, i.e. only if there are **no** files. But maybe we should change the text to "Pick a file or folder..."?

You can find the text to change in `/i18n/en.jsonp.js` (look for the string `configure-select-file-first-option`). That would need to be translated (just use deepL) to Spanish and French by looking for the same string in `es.jsonp.js` and `fr.jsonp.js`.

dev-shetty:
Makes sense, will do the changes. Thanks for the detailed instructions!

dev-shetty:
@Jaifroid, I have pushed the new changes, and also added the translations for the placeholder text. Kindly check the vocabulary once, let me know if any changes are needed.

Also, I took a different approach and added the placeholder in the HTML, and remove it when the files are present.

### Why this approach?
When I used the previous reversing the logic approach, the translation for the dropdown placeholder text was not happening, because the placeholder is added dynamically from `updateZimDropdownOptions` function, which doesnt get called when toggling language, and needed a manual refresh for the translation to happen.

This approach fixes that, as it is present in HTML, and the data attributes toggle it.

<details>
<summary>Demonstration Video</summary>

https://github.com/user-attachments/assets/71287f5a-7f23-40da-9095-96faf90e723a

</details>

Jaifroid (changes requested):
Just some minor changes with the translations... and one other thing.

dev-shetty:
@Jaifroid can you please review this, I have made changes.

Jaifroid (approved):
This is looking very good now! I tested that it doesn't cause any regressions with old platforms like IE11 and Firefox OS, and all fine there (neither use this code, which is intentional, as they have different file picker support). And on modern platforms it's working very well. It's a great improvement to be able to see a scrolling list of up to 10 files in an archive with lots of files. Many thanks for this PR!

Jaifroid:
@dev-shetty Let me know if you are happy for this to be squashed and merged now.

dev-shetty:
Yes LGTM, thanks for the review!!

PR #1329, opened 2025-03-26
Fixed Dropdown Dark mode UI and Searchbar Results Placement

### LINKED ISSUE #1304: Add Wikipedia's own Dark Mode solution
Wikipedia now has a native Dark Mode. See https://diff.wikimedia.org/2024/07/17/dark-modes-bright-future-how-dark-mode-will-transform-wikipedias-accessibility/. If it is possible to export a generic CSS file containing all the CSS required, it would be fairly easy to integrate this into our current Dark Mode solution, which is also CSS based.

This could then be applied downstream in the PWA.

THEBOSS0369:
This PR Fixes 

1. Searchbar Results placement -> Currently the results doesn't align the same location like searchbar which feels odd. 
Fix -> Now The search results display just below search bar.

2.  Dropdown Dark UI Issues -> Drop down results are not perfectly dark color aligned and it creates an unusual view when use it in dark mode. 
Fix -> So i have added the dark color in the dropdown and whatever entry is selected it will show in a shade of brown color while other entries will be shown in black.

SS For reference

**Drop down (Dark Mode inversion)**
**Before** 

dropdown

**After** 
dropdown

**Search Bar Results placment**
Before

searchbar placment

After
searchbar placement

Tests
It has been tested in sw and restricted mode and it worked fine.

Note:
I think the searchbar results theme is little bit different then whole app's as it has some dark blue type while app has color around black or dark neutral , maybe we can change it but it still looks good. just a suggestion


THEBOSS0369:
@Jaifroid Sir i have changed this pr's motive as previous wikipedia's dark mode solution wasn't required in this pr. Please take a look when you got some time. Thanks!

Jaifroid (approved):
That's a great improvement, thank you! I tested in Firefox OS (very old platform), as well as modern platforms, and it's working well.

Jaifroid:
Feel free to squash and merge (not just merge). Use the UI button below. If you don't have access, let me know.

Jaifroid:
@THEBOSS0369 It looks like there's a conflict between this merged PR and one that was merged shortly before. Now the colours for inverted styles don't seem to work for the new way of displaying the contents of a folder in the file select:

image

Maybe you could open a new PR to deal with this quickly?

Jaifroid:
More info: the issue is that when the box is in its expanded state (more than a single option) the non-focused / non-selected options aren't showing with the correct colour in dark mode. You may just need an extra style rule. To test, you'll need a folder of a few small ZIM archives, and pick the folder instead of a single archive.

THEBOSS0369:
@Jaifroid thankx for letting me know. working on it now

PR #1334, opened 2025-04-19
Fixed Drop Down results for mobile view

THEBOSS0369:
In Mobile View the Drop Down results seem to exceed the mobile boundaries 

This PR Fixes that issue. See the Images below for reference

### Screenshots 

**Language**

Before
language

After
language

I have moved the language button to the top of the page as it looks more suitable there

**Theme**

Before
theme

After
theme

**Archive**

Before
archive

After
archive

For small name archive dropdown it will show just like from the image but for archive with big name it will exceed the screen, to fix it i will have to create a separate drop down component and replace the defualt one which i am not sure is needed or not so i haven't updated it. 

Testing

It has tested in both sw & restricted mode working fine 
For browsers tested on brave & edge worked fine.


Jaifroid:
Not sure you need to create a completely new one, as this is already Bootstrap's custom styling. It's probably just necessary to set the width responsively in app.css, and/or set a max-width in the narrow screen style rules....

THEBOSS0369:
@Jaifroid sir the main problem with this is the `select` tag ,when user open zim dropdown here the browser opens a  native picker which is not standard html element and is not controlled by browser so it ignores many css rules like `max width`, `word wrap` `overflow` etc. If you look closer u will notice that every dropdown's width is reduced when i reduced the width of the `archivelist` class.  In fact as you suggested i tried the `max-width` the very first time, it is even present in the very first commit. This is the reason why i was talking about new separate component but we can't change the default tag until it is very imp issue which i don't think it is so
The only solution that i came up with is i reduced the font size of the zim's name and it is working for 99 percent of the zim but exceeding for the zims which have very very big name 

ss
image

THEBOSS0369:
I will update the branch when this pr's work will be completed as i will have to face the merge/file conflicts which i always prefer to face in the end 😅😅

Jaifroid:
Well I think you should definitely check with the new style of folder-based display, because that seems to have different CSS rules again and seems to obey width specifications (I've never seen it overflow the size of the app).

I presume your are talking about the case where  we just pick one archive (now)?

THEBOSS0369:
@Jaifroid yes sir i am talking about the single archive and the folder don't have any overflow. its only when single archive is selected

THEBOSS0369:
@Jaifroid sir nothing seems to work other than the font size change. the big file name still be overflowing. i did a good amount of testing with different properties nothing worked.
this is the ss of the new commit changes

**when one zim is selected**
small
image

large
image

very large
image

when folder is selected
image

Now the only solution that i can think of are : -

1. Separate Component -> we wouldn't be going forward with this one.
2. Truncation of the zim name -> We can truncate the name if it crosses a specified limit because **max-width** and **word-break**  is not working so the extra words couldn't automatically shift to new line.
 I noticed the truncation is happening when folder is selected.
image

Let me know if you any other solution.

Jaifroid:
OK, let me have a play to see if there's another solution. Otherwise, truncation may be possible, but we'd still need to store the full name somewhere to be able to locate it when the user clicks it. So it might be a bit tricky to implement.

Jaifroid:
@THEBOSS0369 I'm trying to reproduce the context in which you're seeing this overflow issue. On my Android, when I go to https://browser.extension.kiwix.org, I don't get the option to pick a folder (the File System Access API isn't available on Android), and when I pick a single file, it isn't shown in the dropdown for a single file picked with the classic file picker, and this is what the dropdown selectors look like anyway:



In other words, it's a completely different style on mobile. Do you have another context in which this is a real issue?

THEBOSS0369:
@Jaifroid you are absolutely right sir . I didn't thought it will be controlled by actual mobile os , i straight went to fixing it after noticing in inspection mode . My apologies for consuming your time for this. 

Also on almost every android phones (Launched after 2015-17) the drop down will work fine just like yours so i don't think this issue is required let me know and i will close this pr. 

Mine ss working fine.
WhatsApp Image 2025-04-21 at 19 32 52_ac9e7f99

Again apologies for this silly mistake i should have first tried this on actual phone rather then directly from inspection mode  😓🙏

Jaifroid:
Please don't worry about it! Sometimes we have to explore an issue to realize how things work anyway! Feel free to close if you think there is nothing to fix now.

PR #1335, opened 2025-04-20
Update javascript-libzim to v0.7

Jaifroid:
Bump to new release.


kelson42:
@Jaifroid What is missing to have a proper integration (without copying the files) of libzim-js?

kelson42:
@jaifroid like other dependencies, libzim should not be in this code repo. Would be glad if you open and issue specifying what you need to avoid this copy.

PR #1343, opened 2025-04-22
Update GH release action to Ubuntu 22.04

### LINKED ISSUE #1342: Update Ubuntu 20.04 in actions as it is now retired
Ubuntu 20.04 has been retired, and since 15-04-25 this prevents Release actions from running (nightlies are specifically affected). We need to update, taking into account the need for compatibility with some old dependencies that allow compiling the app for Ubuntu Touch if at all possible.

Jaifroid:
Fixes #1342.

### DIFF
--- a/.github/workflows/Release.yml
+++ b/.github/workflows/Release.yml
@@ -31,7 +31,7 @@ jobs:
   # This workflow contains a single job called "build"
   build:
     # The type of runner that the job will run on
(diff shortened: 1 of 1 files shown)


Jaifroid:
Builds nightlies correctly. See https://github.com/kiwix/kiwix-js/actions/runs/14588369086/job/40917956921.

PR #1362, opened 2025-09-02
Correctly replace `<noscript>` tags

### LINKED ISSUE #1361: `<noscript>` content gets converted to `<script>` in Restricted Mode
> @benoit74 We did put support for NOSCRIPT into the Restricted mode code, but tbh as it hasn't been really supported in ZIMs it could well have fallen by the wayside. It would be easy enough to resurrect / debug NOSCRIPT (which is not necessarily automatically detected by browsers because our apps do support JS, just not JS-from-the-ZIM in Restricted mode).

 _Originally posted by @Jaifroid in [openzim/overview#52](https://github.com/openzim/overview/issues/52#issuecomment-2969771658)_

@Jaifroid looks like the `<noscript>` support in Restricted mode is broken, it replaces `<noscript>` with `<script>` instead of just removing the tag and leaving the content as HTML. (`noscript.outerHTML = noscript.innerHTML`)

Using my ZIMs from openzim/mwoffliner#2483 contain a stylesheet inside `<noscript>` tags for example, which isn't properly included in Restricted Mode:

ServiceWorker:
```html
<noscript><link rel="stylesheet" type="text/css" href="./_mw_/noscript.css" /></noscript>
```

Restricted Mode:
```html
<script><link rel="stylesheet" type="text/css" data-kiwixurl="C/_mw_/noscript.css" /></script>
```
Unlike other stylesheets in Restricted Mode, the noscript one didn't get converted to inline styles either.

Markus-Rost:
Fix #1361

- Correctly replacing `<noscript>` tags with their content.
- Moving the `<noscript>` replacement before other DOM manipulation to ensure the noscript content gets modified correctly as well.

### DIFF
--- a/www/js/app.js
+++ b/www/js/app.js
@@ -2932,12 +2932,12 @@ function displayArticleContentInIframe (dirEntry, htmlArticle) {
         // Allow back/forward in browser history
         pushBrowserHistoryState(dirEntry.namespace + '/' + dirEntry.url);
 
-        parseAnchorsJQuery();
-        loadImagesJQuery();
         // JavaScript is currently disabled, so we need to make the browser interpret noscript tags
         // NB : if javascript is properly handled in jQuery mode in the future, this call should be removed
         // and noscript tags should be ignored
         loadNoScriptTags();
+        parseAnchorsJQuery();
+        loadImagesJQuery();
         // loadJavaScriptJQuery();
         loadCSSJQuery();
         insertMediaBlobsJQuery();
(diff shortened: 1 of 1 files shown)


Jaifroid:
Sorry, only just saw this!

Jaifroid:
Hmm, but this leaves the current loadNoScriptTags in place. I was actually just working on a more complete fix, so let's compare notes when I push that PR.

Jaifroid:
Actually, I think your approach may be cleaner and above all simpler. Just checking...

Jaifroid:
Sorry, 'first' and 'second' are not the same order as they are shown above.

Jaifroid:
Just invited you to contribute to this Repo. Feel free to squash and merge when you're ready, or else let me know and I'll do it.

PR #1363, opened 2025-09-03
Fix incorrect processing of noscript tags

### LINKED ISSUE #1361: `<noscript>` content gets converted to `<script>` in Restricted Mode

Jaifroid:
Fixes #1361. Needs testing.

### DIFF
--- a/www/js/app.js
+++ b/www/js/app.js
@@ -3154,13 +3154,23 @@ function displayArticleContentInIframe (dirEntry, htmlArticle) {
     }
 
     function loadNoScriptTags () {
-        // For each noscript tag, we replace it with its content, so that the browser interprets it
+        // For each noscript tag, we replace it with its HTML content, so that the browser interprets it
         var noscriptTags = iframeArticleContent.contentDocument.querySelectorAll('noscript');
         Array.prototype.forEach.call(noscriptTags, function (noscriptTag) {
-            var content = noscriptTag.textContent;
-            var replacementNode = iframeArticleContent.contentDocument.createElement('script');
-            replacementNode.text = content;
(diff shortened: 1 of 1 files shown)


Markus-Rost:
Looks like this is doing the same as #1362, just way more complex and as a result the code is less clear.

Jaifroid:
You're right. I over-complicated it, thinking I needed to process all the nodes. Will close this in favour of yours.

PR #1377, opened 2025-10-21
Fix popover functionality with libzim backend

### LINKED ISSUE #1336: Make popover retrieval functions work with libzim
As our [javascript-libzim port](https://github.com/openzim/javascript-libzim) is getting faster and more useable, at least on desktop, we need to ensure going forward that all added features are agnostic about which back end they use.  Popovers currently only work with the standard backend, not with the experimental libzim worker. Since we can get directory entries and contents from either backend, it should not be difficult to abstract the logic for retrieving the popover contents.

VikramAditya33:
## Summary
Fixes popover functionality to work with the libzim backend by adding conditional routing and proper data type handling.

## Changes
- Added `params.useLibzim` check in `getArticleLede()` function
- Implemented `getArticleLedeWithLibzim()` for libzim-specific content retrieval
- Added Uint8Array to UTF-8 string conversion using `archive.getUtf8FromData()`
- Included redirect resolution support for libzim backend
- Maintained backward compatibility with standard backend

## Testing
- ✅ Popovers work with standard backend (unchanged)
- ✅ Popovers now work with libzim backend
- ✅ No console errors
- ✅ Handles both string and Uint8Array content from libzim

## Fixes
Fixes #1336

### DIFF
--- a/www/js/lib/popovers.js
+++ b/www/js/lib/popovers.js
@@ -32,15 +32,6 @@ const regexpZimPath = /\.zim\w?\w?\/(.*)$/i;
 const regexpZIMUrlWithNamespace = /^[-ABCIJMUVWX]\//;
 
 /**
- * Decodes a ZIM path to handle encoded slashes in nested article names
- * @param {String} path The path to decode
- * @returns {String} The decoded path
- */
-function decodeZimPath(path) {
-    return decodeURIComponent(path);
-}
-
-/**
  * Parses a linked article in a loaded document in order to extract the first main paragraph (the 'lede') and first
  * main image (if any). This function currently only parses Wikimedia articles. It returns an HTML string, formatted
  * for display in a popover
@@ -63,7 +54,7 @@ function getArticleLede (href, baseUrl, articleDocument, archive) {
         if (zimPathMatch && zimPathMatch[1]) {
             zimURL = zimPathMatch[1];
             zimURL = zimURL.replace(/[?#].*$/, '');
-            zimURL = decodeZimPath(zimURL);
+            zimURL = decodeURIComponent(zimURL);
         } else {
             // Handle relative URLs by resolving against current article's path
             const currentLocation = articleDocument.location.href;
@@ -73,7 +64,7 @@ function getArticleLede (href, baseUrl, articleDocument, archive) {
             if (currentPathMatch && currentPathMatch[1]) {
                 currentZimPath = currentPathMatch[1];
                 currentZimPath = currentZimPath.replace(/[?#].*$/, '');
-                currentZimPath = decodeZimPath(currentZimPath);
+                currentZimPath = decodeURIComponent(currentZimPath);
             }
             
             const baseZimPath = currentZimPath.replace(/[^/]+$/, '');


VikramAditya33:
Hi @Jaifroid The PR is ready for review

Jaifroid:
OK, could you please use "Request" (review) link top right, so I'm not requesting a review from myself? 😉

VikramAditya33:
@Jaifroid I'm unable to find button or any option to request a review.

Jaifroid:
Oh, how strange, don't you see this? -

image

VikramAditya33:
@Jaifroid No I can't see anything like this I've tried exploring every place but I'm unable to request you

Jaifroid:
OK, no problem, probably something wrong with the settings of the Repo, but don't worry, I was just curious.

VikramAditya33:
Got it thanks

Jaifroid (changes requested):
Thank you very much for this PR. The code is functional, but there are a few things that need correcting as outlined in individual comments. It shouldn't be hard to achieve these!

VikramAditya33:
@Jaifroid Thank you for the detailed feedback! I've addressed all five issues you raised. Here's a summary of my original reasoning and the corrections I’ve made:

Issue 1: Wrong Redirect Handling Logic
Original Intent: I wanted to ensure popovers worked correctly with redirected articles. I assumed libzim didn’t handle redirects automatically for popover requests, so I implemented manual redirect following.
Why I Kept It: I tried to replicate the standard backend’s redirect logic but using libzim functions, not realizing that follow: true already exists as a built-in feature.
Fix: Removed all manual redirect handling and now use follow: true in the libzim worker call, letting libzim handle redirects automatically as designed.

Issue 2: Inconsistent Error Handling
Original Intent: I aimed to handle missing entries gracefully without breaking the popover system. I thought returning empty strings would be more user-friendly than throwing errors.
Why I Kept It: I wanted to prevent console spam from missing links, but I didn’t realize the original function throws errors for consistency.
Fix: Made the error handling consistent, now throwing Error('No content found for ' + zimURL) in all missing content cases, matching the original behavior.

Issue 3: Redundant Error Checking
Original Intent: I wanted to be thorough with error handling. I believed separate checks would make it safer and clearer.
Why I Kept It: I was being overly cautious, not realizing the second check already covered the first.
Fix: Consolidated into a single if (!ret || !ret.content) check that handles both cases efficiently.

Issue 4: Code Duplication
Original Intent: I wanted both backends to work identically for popovers, so I copied logic to ensure consistent behavior.
Why I Kept It: I was cautious about changing working code and didn’t realize there was a cleaner way to share logic.
Fix: Created a shared parseArticleContentForBalloon helper function used by both backends, eliminating duplication and ensuring identical behavior.

Issue 5: Not Following Original Function’s Pattern
Original Intent: I simplified the libzim function to extract just the first paragraph, assuming that was enough for popovers.
Why I Kept It: I misunderstood the scope, I didn’t realize popovers required rich HTML content with images and proper link handling.
Fix: Implemented the full balloon creation logic in the helper function so both backends now produce identical rich popovers with proper formatting and image support.

Summary:
My initial goal was to ensure popovers worked with both backends, but I overcomplicated some parts and missed established patterns. The feedback helped me simplify and align with the existing design. The updated implementation now properly integrates libzim, ensures consistent error handling, eliminates duplication, and maintains the full functionality of the original popovers.

VikramAditya33:
@Jaifroid Thank you for identifying this critical redirect path tracking issue, you're absolutely right about the stale base URL and hidden redirect details.

Issues Addressed:

Stale Base URL
Problem: state.baseUrl wasn’t updating under libzim, causing incorrect base URL resolution in popovers.
Fix: Replaced it with dynamic extraction from articleDocument.location.href, ensuring accurate base URL calculation based on the current page.

Hidden Redirect Information
Problem: Using follow: true masked redirect information, making it impossible to detect the final redirected path.
Fix: Switched to follow: false so redirects can be detected and properly resolved before content generation.

Redirect Path Tracking
Problem: Popovers needed the actual redirect target path for accurate base URL resolution.
Fix: Added redirect detection and an additional worker call for the redirect path, using it as the final content source.

Result:
✅ Correct base URL resolution based on the active article
✅ Proper redirect path handling and popover generation
✅ Fixed broken links/images in popovers for redirected articles
✅ Consistent behavior across standard and libzim backends

This should fully resolve the base URL calculation issue you observed with redirected articles (like the Singapore_Riverside example).

Jaifroid (changes requested):
The changes seem to be along the right lines, but unfortunately, the implementation isn't working. Did you test? For every popover I get:
```
popovers.js:407 
 Error: No article lede or image
    at parseArticleContentForBalloon (popovers.js:179:19)
    at popovers.js:147:20
```
You can't just trust the output of CC or Codex without testing! I don't mind your using AI to guide you, but they often don't get it right first shot. Please test, debug and make sure it's working properly both in libzim mode and in normal mode. Automated tests don't run in libzim mode, so you can't trust their output for this either.

VikramAditya33:
Thanks for the quick check and for pointing this out, @Jaifroid much appreciated.

A few clarifications so I can reproduce and fix this reliably:

I actually did test the implementation locally with both the standard backend and libzim enabled, and popovers worked correctly in my environment. So I’m keen to understand why it’s failing in yours, there might be a ZIM-specific or environment difference.

Just to clarify, I didn’t push auto-generated code, I used AI tools only to assist with code analysis and refactoring, but all logic changes and testing were done manually before submitting the PR. I completely understand your caution, and I’ll make sure every change is verified and debugged properly.

What the error suggests :
The No article lede or image error looks like cleanUpLedeContent() is filtering out all paragraphs (it requires paragraphs ≥ 50 chars). That can happen in articles with mostly short paragraphs or image-heavy layouts.

To reproduce the issue accurately, could you please share:

- The ZIM file name you tested with
- The specific article/path where this error occurs
- Whether libzim is enabled
- The browser and commit version you tested on

Next steps I can take:

- Add a fallback for shorter paragraphs when no long ones exist
- Reduce the 50-character threshold
- Improve error messages for easier debugging
- Re-test using your setup and share a verified fix

Once I can reproduce it on the same ZIM, I’ll push a targeted patch and confirm it works in both backends.

Jaifroid:
Great to hear you tested. As I said, it's fine to use AI as a tool and guide. I don't think the issue is any of the things you mention, and 50 characters is extremely small, no valid article HTML would only have that many. What I notice is that you're feeding an incorrect zimURL to the libzim worker. This is what I see in dev console:

image

Notice the extra path in front of the zimURL. So I think I understand what happened: you tested in Vite, and assumed the server will always be displaying the contents of the ZIM served from the root of the Repo, which is what Vite does, but in real life servers could be serving from a folder nested deeply in the directory structure. Your regex for extracting the zimURL isn't safe from what I can tell. I tested using http-server from my C:/Repos instead of Vite from C:/Repos/kiwix-js/.

You need to account for any possible directory prefix, including ones that look like ZIM filenames. Note you can find the file name in the archive object (`archive.file.name`).

Jaifroid:
PS That last bit of info about the ZIM name is irrelevant, because it shouldn't be part of the zimURL. A zimURL looks like C/article.html or A/article (in older ZIMs), but it could also be A/Singapore/Riverside for example, or C/User:The_other_Kiwix_guy/Landing .

VikramAditya33:
Hi @Jaifroid 

I've analyzed the problem and here's my approach to fix it:

**Root Cause Analysis:**
You're absolutely right - the issue is that I'm feeding an incorrect zimURL to the libzim worker. The console shows `/kiwix-js/wikipedia_en_100_2025-07.zim/Mollusc` when it should be just `C/Mollusc` or `A/Mollusc`.

**My Proposed Approach:**

1. **Fix zimURL extraction for absolute URLs:**
   - When href contains `.zim/`, extract everything after `.zim/` as the ZIM-internal path
   - Remove any URL parameters/fragments
   - This handles your test case with http-server from C:/Repos

2. **Handle relative URLs properly:**
   - For relative links (like `../Article`), extract the current article's ZIM path from `articleDocument.location.href`
   - Use that as the base to resolve the relative URL with `uiUtil.deriveZimUrlFromRelativeUrl()`

3. **Fix the regex patterns:**
   - Replace the problematic regex with something more robust like `(.*\.zim[a-z]*\/)`
   - This should handle any server directory structure, including deeply nested paths

4. **Remove the flawed helper function:**
   - The `extractBaseUrlFromLocation` function was using unsafe regex patterns
   - Replace it with direct ZIM path extraction logic

VikramAditya33:
PS @Jaifroid I've implemented the fix using `.match()` with capture groups for more precise extraction. The zimURL now correctly contains only the ZIM-internal path (e.g., 'C/Article' or 'Mollusc') without any server directory prefixes.

Key changes:
- Used regex `.match(/\.zim[^/]*\/(.*)$/)` to extract ZIM-internal paths reliably
- Handles both absolute URLs and relative URLs properly  
- Added fallback logic for edge cases
- Tested with both Vite and http-server configurations

The libzim worker now receives clean zimURLs like 'C/Geometry' instead of '/kiwix-js/wikipedia_en_100_2025-07.zim/Mollusc'. 

I've committed the changes and the fix is ready for your review. I systematically debugged the issue by adding console logs to track the zimURL extraction process, verified it works in both development environments, and then cleaned up the debugging code while keeping the core fix intact.

Jaifroid:
@VikramAditya33 Apologies for slowness. I had to fix a broader issue with libzim reading in #1379 that was making it difficult to test your PR properly. Could you kindly update the branch, and then I can test properly?

VikramAditya33:
Hi @Jaifroid hope you’re doing well!
No worries at all about the delay 😄 

I’ve updated the branch, the PR should now be ready for testing.
Let me know if you run into any issues or if there’s anything else you’d like me to adjust.

Thanks again for the review and updates!

Jaifroid:
Thanks for that! It's generally working really well, but I think I found a small bug during testing. To reproduce, follow these steps:

1. Download the Wikipedia Knots ZIM here: https://download.kiwix.org/zim/wikipedia/wikipedia_en_knots_maxi_2025-10.zim
2. Go to the page `Braid (hairstyle)`
3. Scroll down "Braiding in particular cultures"
4. Hover over the links just above that title `French braids`, `corn rows` and `box braiding`. In each of these the image isn't correctly loaded (see screenshot below)
5. Hover over `pigtails` in same paragraph: in this one the image is correctly loaded.

I'm not sure quite what's going on, but if I switch to not use libzim for loading articles, these images are loaded correctly in the popover, so the problem is to do with the presentation of the links to the backend. It could be an issue with the calculation of the path, or it could be an issue with decoding (or not) the URI-encoded path. Needs careful debugging.

image

VikramAditya33:
Hi @Jaifroid 
Fixed: Two issues found:
Removed encodeURI() from balloonBaseURL — deriveZimUrlFromRelativeUrl() requires decoded base URLs
Added namespace preservation for redirects — when libzim returns redirect paths without namespace (e.g., French_braid instead of C/French_braid), the namespace is now preserved from the original request

Debugging process:

Added console logs to trace the issue. For "French braids":
Original zimURL: C/French_braids ✅
After redirect: French_braid ❌ (missing namespace)
baseURL: (empty) ❌
Image src: ./_assets_/.../Classic_French_Braid.JPG
Result: Image path couldn't resolve because baseURL was empty
For "pigtails" (working case):
No redirect, so zimURL stayed C/pigtails with namespace
baseURL: C/ ✅

Images resolved correctly
Root cause: libzim's redirect paths sometimes lack namespace, causing empty baseURL and broken relative image paths.

Ready for review. Thanks!

Jaifroid:
This is now working very well, congratulations! I'll do a detailed code review asap.

Jaifroid:
The URL resolver for setting the baseURL of the popover/balloon is not quite working, I'm afraid. I found a similar bug to the previous one. Unfortunately, to test this you'd need to download a 1GB ZIM (Wikivoyage). Steps to reproduce:

1. https://download.kiwix.org/zim/wikivoyage/wikivoyage_en_all_maxi_2025-09.zim
2. Go to article Singapore
3. Find link to Singapore/Marina Bay and click on it (or you could just search for Singapore Marina Bay)
4. Scroll down to See -> Marina Bay Sands
5. Under Marina Bay Sands Casino, hover over the link for Sentosa
6. You'll see image in popover doesn't load (screenshot). Switch off use of libzim, and you'll see the same image does load (bottom screenshot).

Additionally, I'm a little worried by your statement that the URL of the snippet is sometimes returned with no namespace, and sometimes isn't. It's not random, it's because all links in a ZIM are relative, so the link to the popover article could be something like ../../Sentosa (this might be because the browser treats the article title Singapore/Marina_Bay as being in a subdirectory. My guess is that although you don't URI decode, you probably should resolve any forward slashes. This is a difference between how `decodeURI()` and `decodeURIComponent()` work. NB this is a bit of speculation, I'm not 100% sure this is what is going on, but I specifically tested an article that appears (to the browser) to be in a subdirectory.

Take a look at how the code works in the other path (without libzim). Maybe take a look at code you altered also. Be very careful about having removed a decodeURI step. The other way you can debug is by loading the page (`Sentosa` in this case, in libzim mode) and looking carefully at what the image URL is. Does the baseURL of the popover, when combined with the image URL, produce the correct image URL? Since you don't / shouldn't alter the `` tag you extract from the page (we know that code works fine), the issue must be with calculating the correct baseURL for the popover.

image

image

VikramAditya33:
Hi @Jaifroid 
Finally cracked the image loading issue for nested article paths, Took me quite a while to figure out what was going on, but I think I got it now.

What Was Actually Happening :
You were absolutely right about the forward slash issue. The problem was exactly what you suspected - when article names have slashes in them (like Singapore/Marina_Bay), they get encoded as Singapore%2FMarina_Bay in the URL. Our regex was only looking for literal / characters, completely missing the %2F encoded ones.

So here's what was breaking :
Article: C/Singapore/Marina_Bay (shows as C/Singapore%2FMarina_Bay in URL)
Base URL became: C/ instead of C/Singapore/
Then ../../Sentosa from C/ went completely off the rails and lost the namespace

The Fix :
Added decodeURIComponent() (like you suggested) to decode those %2F back into actual slashes before calculating the base URL:

// Lines 56-60, 72-77currentZimPath = decodeURIComponent(currentZimPath);// Now C/Singapore%2FMarina_Bay → C/Singapore/Marina_Bay// Base calculation finally sees both slashes → C/Singapore/ ✅

Also improved the redirect namespace detection because some redirects like Singapore/Sentosa_and_Harbourfront have slashes but still need the namespace prefix added.

Testing
Tested with both ZIMs you mentioned:
Wikipedia Knots:
✅ French braids, corn rows, box braiding all are working
Images loading correctly
Wikivoyage:
✅ Singapore/Marina_Bay → Sentosa popover image works
✅ Image with the encoded commas (%252C) resolves properly
✅ Those ../ paths from nested articles finally resolve correctly
What I Changed
Decode article paths with decodeURIComponent() to handle encoded slashes
Better namespace detection for redirects (checks if path starts with [A-Z]/)
Stopped interfering with deriveZimUrlFromRelativeUrl's encoding/decoding (was causing double-decode issues)
Kept balloonBaseURL decoded like you said it should be
Your hint about decodeURI() vs decodeURIComponent() was spot on, needed the one that actually decodes forward slashes.

Let me know if you find any other edge cases I missed.

Jaifroid:
Great sleuthing! I'll do some last tests, but hopefully we've covered everything now 🤞.

www/js/lib/popovers.js
```diff
@@ -38,42 +53,63 @@ import uiUtil from './uiUtil.js';
...
+    
+    // Check if we're using libzim backend
+    if (params.useLibzim) {
+        // Extract ZIM-internal path from absolute URL like /path/to/file.zim/C/Article
+        let zimURL;
+        
+        // Check if href is an absolute URL containing the ZIM file path
+        if (href.includes('.zim/')) {
+            // Extract ZIM-internal path using regex match
+            const zimPathMatch = href.match(/\.zim[^/]*\/(.*)$/);
```
Jaifroid: I'm not sure this regex is specific enough. For exampale, it would match a hypothetical ZIM name like `farms.zimbabwe_en_all.zim/something.html` and would fail because it would match everything after `farms.zim...`. It would also fail if a ZIM were named `FARMER.ZIM` (capital letters). A ZIM file extension can be `zim`, `ZIM` or `zimaa`, `zimab`, `zimac`... `zimzz` (for split ZIMs).

The standard way we match below is `/\.zim\w?\w?\/(.*)$/i`. The `i` means case-insensitive, and `\w\` is an alphanumeric letter. That would be much more specific. You should use a similar pattern everywhere else where you are searching for the zim file extension.

Even better, if you are using the same pattern repeatedly, define a regexp at the top level, something like `const regexpZimPath = /....../i`, and use it. Saves the JS engine from recompiling the regex each time.

www/js/lib/popovers.js
```diff
@@ -97,6 +133,95 @@ function getArticleLede (href, baseUrl, articleDocument, archive) {
...
+        }
+        if (!balloonString) {
+            throw new Error('No article lede or image');
+        } else {
+            return balloonString;
+        }
+    } else {
+        throw new Error('No article body found');
+    }
+}
```
Jaifroid: As I understand it, this entire function was moved from its previous position, as we had discussed. However, you have changed some things, and in particular the regex for matching rootRelativePathPrefix. Please don't alter existing code unless you really need to. Please go through this and either restore altered code or explain why you needed to change it for your implementation, as it is a shared helper function.

Jaifroid:
Also please update your branch (using button below, and then don't forget to pull the change to your local copy) as there have been large changes to support native dark theme, and some of those changes also affect popover code (but don't create a conflict with your PR, fortunately).

VikramAditya33:
@Jaifroid 
Thank you for your guidance throughout this PR. I’ve addressed your feedback and pushed the updates.

www/js/lib/popovers.js
```diff
@@ -26,6 +26,20 @@
...
+const regexpZIMUrlWithNamespace = /^[-ABCIJMUVWX]\//;
+
+/**
+ * Decodes a ZIM path to handle encoded slashes in nested article names
+ * @param {String} path The path to decode
+ * @returns {String} The decoded path
+ */
+function decodeZimPath(path) {
+    return decodeURIComponent(path);
+}
```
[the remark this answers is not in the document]
VikramAditya33: @Jaifroid I've removed the redundant decodeZimPath() wrapper and now using decodeURIComponent() directly where it is needed.

Jaifroid (approved):
Thank you very much for this great PR. It's a complex and challenging one, but you've risen to that challenge. It's one more step on the way to standardizing our backend using libzim, which is the end goal.

When you're happy for this to be squashed/merged, let me know.

VikramAditya33:
Thank you so much, @Jaifroid I really appreciate your guidance and support throughout this, it made tackling such a complex issue much easier.
I’ve rechecked everything on my end, and I’m happy with the implementation.
Please feel free to squash and merge whenever convenient.

PR #1381, opened 2025-11-02
Add native Wikimedia dark theme support

### LINKED ISSUE #1376: New Wikipedia ZIMs have dark theme turned on permanently if user's OS theme is dark
This is a subissue of #1355. It looks like we don't need to do much to support the ZIM-provided dark theme as it's already on if the user's OS is set to dark. See screenshot. However, we do need to respect the user's choices, because otherwise we get a horrible clash of the app's light theme and the ZIM's dark theme. And we need to add an option to respect the user's choice of light theme (if that is what they've chosen). Most egregiously, there is no way to turn off dark theme except by applying an inversion theme, which inverts the app and the iframe contents, leading to washed-out image colours and still a clash of app mode and content theme.

As there is no officially documented method to detect this situation, nor a ZIM-provided mechanism for switching themes, it looks like we may have to hack the HTML of the content, something which we have been very reticent to do in this Repo for ServiceWorker mode.

Image

Jaifroid:
Fixes #1376.

### DIFF
--- a/www/js/lib/uiUtil.js
+++ b/www/js/lib/uiUtil.js
@@ -1191,19 +1191,22 @@ function applyAppTheme (theme) {
         showReturnLink();
     }
     // Update the colour scheme button icon to reflect current theme
-    // Note: Font Awesome may have converted <i> to <svg>, so we check for both
+    // Note: Font Awesome converts <i> to <svg> and uses data-icon attribute
+    // We check for SVG first (modern browsers + IE11), then fall back to <i> (if SVG conversion disabled)
     var btnIcon = document.querySelector('#btnColourScheme svg') || document.querySelector('#btnColourScheme i');
     if (btnIcon) {
-        if (appTheme === 'dark') {
-            btnIcon.classList.remove('fa-moon');
-            btnIcon.classList.add('fa-sun');
-            // Update SVG data-icon attribute if it's an SVG element
-            if (btnIcon.tagName === 'svg') btnIcon.setAttribute('data-icon', 'sun');
+        if (btnIcon.tagName === 'svg' || btnIcon.tagName === 'SVG') {
+            // SVG element: update data-icon attribute (works in all browsers including IE11)
+            btnIcon.setAttribute('data-icon', appTheme === 'dark' ? 'sun' : 'moon');
         } else {
-            btnIcon.classList.remove('fa-sun');
-            btnIcon.classList.add('fa-moon');
-            // Update SVG data-icon attribute if it's an SVG element
-            if (btnIcon.tagName === 'svg') btnIcon.setAttribute('data-icon', 'moon');
+            // <i> element: update classes (only if Font Awesome SVG conversion is disabled)
+            if (appTheme === 'dark') {
+                btnIcon.classList.remove('fa-moon');
+                btnIcon.classList.add('fa-sun');
+            } else {
+                btnIcon.classList.remove('fa-sun');
+                btnIcon.classList.add('fa-moon');
+            }
         }
     }
     // Return the actual theme applied (which may differ from the requested theme if fallback occurred)


www/js/lib/uiUtil.js
```diff
@@ -1014,29 +1014,42 @@ function tabTransitionToSection (toSection, isAnimationRequired = false) {
...
     }
 
+    // Store the base theme (light/dark/auto) to reconstruct the final theme string if fallback occurs
+    var baseTheme = theme.replace(/_.*$/, '');
     // Resolve the app theme from the matchMedia preference (for auto themes) or from the theme string
     var appTheme = isDarkTheme(theme) ? 'dark' : 'light';
     // Get contentTheme from chosen theme
     var contentTheme = theme.replace(/^[^_]*/, '');
+    var requestedContentTheme = contentTheme; // Store original request for comparison
+
```
Jaifroid: Close up

www/js/lib/uiUtil.js
```diff
@@ -1054,10 +1067,13 @@ function applyAppTheme (theme) {
...
-    // Embed a reference to applied theme, so we can remove it generically in the future
-    htmlEl.dataset.theme = appTheme + contentTheme;
-    
+    // For _wikimediaNative, apply appropriate footer class based on resolved appTheme
+    var footerTheme = contentTheme;
+    if (contentTheme === '_wikimediaNative') {
+        footerTheme = appTheme === 'dark' ? '_mwInvert' : '_light';
+    }
+    footer.classList.add(footerTheme || '_light');
+
```
Jaifroid: Close up

www/js/app.js
```diff
@@ -126,48 +126,32 @@ darkPreference.onchange = function () {
...
+// Update theme dropdown options based on loaded ZIM type
 function updateThemeOptions() {
+    const themeSelect = document.getElementById('appThemeSelect');
+    if (!themeSelect) return;
+
+    const zimLoaded = selectedArchive && selectedArchive.file && selectedArchive.file.name;
     const vectorOption = document.getElementById('theme-vector-option');
+    const nativeOption = document.getElementById('theme-native-option');
+    const nativeAutoOption = document.getElementById('theme-native-auto-option');
+
```
Jaifroid: ditto

Markus-Rost:
This seems to not use `skin-theme-clientpref-day` for light mode?

Jaifroid:
> This seems to not use `skin-theme-clientpref-day` for light mode?

Yes, you're right, it currently just removes the CSS classes. It would indeed be better to set the day mode explicitly instead of removing, and I think it would simplify a couple of things too. Thanks for pointing that out.

Jaifroid:
As I thought, we're getting quite bad FOIT if user is overriding the clientpref-os style. This is because we're overriding using DOM methods after the article has rendered, and the ZIM has the clienpref-os set by default. Two possible solutions:

1. Delay unhiding of iframe contents until we have processed the user's requested content theme
2. Change the html before injecting.

The second is quick but dirty. Well, arguably the whole thing is dirty.. The first is much cleaner, but opening ZIM contents in a tab outside of the iframe will still show FOIT unless we hide body or something. Need to compare with how I did it in the PWA. Theming system in the Browser Extension is quite different, unfortunately.

Jaifroid:
Need to simplify user choices too. Best way is probably to do away with the option of setting the app theme separately from the content theme. It's not really desirable. The appTheme should always follow the content, to avoid the clash between the UI and the content, which is quite glaring.

Jaifroid:
FOIT has been dealt with for the iframe. We actually don't handle themes in windows or tabs yet, though the new native Wikipedia theme will be responsive in those tabs.

www/js/lib/uiUtil.js
```diff
@@ -1014,69 +1013,90 @@ function tabTransitionToSection (toSection, isAnimationRequired = false) {
...
     
     // Safely handle help element IDs
     var safeOldContentTheme = oldContentTheme.replace(/[^a-zA-Z0-9-]/g, '');
-    var safeContentTheme = contentTheme.replace(/[^a-zA-Z0-9-]/g, '');
-    
+    var safeContentTheme = requestedContentTheme.replace(/[^a-zA-Z0-9-]/g, '');
+
     // Hide any previously displayed help
-    var oldHelp = document.getElementById(safeOldContentTheme.replace(/_/, '') + '-help');
+    var oldHelp = document.getElementById(safeOldContentTheme + '-help');
```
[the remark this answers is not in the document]
Jaifroid: This is incorrect because the making-safe regex removes anything that is not A-Z 0-9 or -, therefore the underscore is removed.

www/js/app.js
```diff
@@ -2339,11 +2303,13 @@ function articleLoadedSW (iframeArticleContent) {
     document.getElementById('cachingAssets').textContent = translateUI.t('spinner-caching-assets') || 'Caching assets...';
     document.getElementById('cachingAssets').style.display = 'none';
     uiUtil.spinnerDisplay(false);
-    // Set the requested appTheme
+    // Set the requested appTheme - applyAppTheme will handle fallbacks silently
     uiUtil.applyAppTheme(params.appTheme);
     // Display the iframe content
     iframeArticleContent.style.display = '';
     articleContainer.style.display = '';
+    document.getElementById('articleContent').style.display = '';
```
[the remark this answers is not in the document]
Jaifroid: While this is correct in some configurations, when opening a new window or tab, articleContainer and iframeArticleContent may not be the same thing.

PR #1391, opened 2025-11-07
Remove legacy Wikivoyage workaround for image map popovers

### LINKED ISSUE #1387: Remove legacy Wikivoyage workaround in popovers.js
This is related to (but not the same as) #1385. This would be a more generic way to prevent showing popovers if a user has hovered a link in an area map. The problem is that the popover often covers the area map, making it very difficult to hover/click a different area, especially one in the middle of a map that has other clickable areas around it.

Fixing this issue in popovers.js would make it unnecessary to have ZIM-specific code relating to Wikivoyage, which was never a good idea in the first place and was a quick hack.

VikramAditya33:
- Removes ZIM-specific code that disabled popovers on Wikivoyage landing pages
- The filter at line 2444 in app.js already prevents <area> elements from triggering popovers
- Makes the codebase more generic and maintainable
- Fixes #1387

### DIFF
--- a/www/js/lib/popovers.js
+++ b/www/js/lib/popovers.js
@@ -373,12 +373,8 @@ function populateKiwixPopoverDiv (ev, link, state, dark, archive) {
     // Do not show popover if the user has initiated an article load (set in filterClickEvent)
     if (link.articleisloading || link.popoverisloading) return Promise.resolve();
     const linkHref = link.getAttribute('href');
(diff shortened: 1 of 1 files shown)


Jaifroid (approved):
This seems completely straightforward, thanks! Let me know if you're happy for me to merge.

VikramAditya33:
Yea @Jaifroid you can go ahead and merge it

PR #1399, opened 2025-11-15
Fix error detecting availability of PWA implementation

### LINKED ISSUE #1398: Critical error detecting online implementation of browserextension PWA in Chromium extensions
I've discovered a critical error in fresh installs of the Chromium extension which prevents them jumping to full ServiceWorker mode (by caching the online PWA code at browser-extension.kiwix.org). It affects only the latest release, and the consequence is that they remain stuck in ServiceWorkerLocal mode (which is fine for Wikipedia, but not for highly dynamic ZIMs).

This has only just been discovered because the error is masked in existing installations due to them jumping straight to the cached offline version of the PWA. It is something that is very difficult to test for because we do not have a framework for auto-testing extension-specific code as opposed to the JavaScript of the app.

This needs urgent fixing and an emergency update to Chromium stores (to v4.3.1). Firefox is unaffected due to not having a Service Worker in browser-extension (local) mode.

Jaifroid:
Fixes #1398.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -172,9 +172,41 @@ jobs:
     if: github.event_name == 'push' || github.event.pull_request.head.repo.full_name != github.repository
     runs-on: windows-latest
     timeout-minutes: 30
+    env:
+      SE_AVOID_SELENIUM_MANAGER: '1'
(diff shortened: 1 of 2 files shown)


Jaifroid:
Blocked by #1401

PR #1404, opened 2025-11-30
Added an Ability to Save & View Web History (#1368)

### LINKED ISSUE #1368: Provide Web browser history feature as an option
Hello.

Prior to this the `Kiwix` `Firefox` add on would show a visited page in the ' Web browser history... "Architecture" for example.

Now it does not.

You should make it optional to store it in history as it is very very helpful for someone Like me / my party as well it mimicks normal website use including `Wikipedia` use: having a ' Web history.

Please consider this at least medium priority. Thanks.

Joe

anuarb799:
Worked on an open issue: "Provide Web browser history feature as an option #1368". Implemented the feature by saving the articles user opened to the localStorage. On each access, save the title, its ID, and the timestamp. Users can view the history by clicking on the button with a logo taken from Font Awesome, can hide the history and delete it as well.

### DIFF
--- a/www/index.html
+++ b/www/index.html
@@ -121,6 +121,9 @@
                                 <span class="btn-group">
                                     <button type="button" data-i18n-tip="home-btn-random-tip" title="Select a random article to display" class="btn btn-outline-secondary" id="btnRandomArticle"><i class="fas fa-random"></i></button>
                                 </span>
+                                <span class="btn-group">
+                                    <button type="button" data-i18n-tip="home-btn-history-tip" title="Show article history" class="btn btn-outline-secondary" id="btnShowHistory"><i class="fas fa-history"></i></button>
+                                </span>
                             </span>
                         </div>
                     </div>
@@ -792,6 +795,23 @@
                 </div>
                 <div data-i18n="home-welcome-text" id="welcomeText" class="container">
                     Once you have loaded a ZIM archive, you can search its contents using the above search field.
+                </div>
+
+                <!-- List of article history (accessed by the user) -->
+                <div id="articleHistorySection" style="display: none;" class="container">
+                    <div style="display: flex; justify-content: space-between; align-items: center; margin-bottom: 10px;">
+                        <h4 style="margin: 0;">Article History</h4>
+                        <div>
+                            <button type="button" class="btn btn-sm btn-outline-secondary" id="btnClearHistory" title="Clear history">
+                                <i class="fas fa-trash"></i> Clear
+                            </button>
+                            <button type="button" class="btn btn-sm btn-outline-secondary" id="btnHideHistory" title="Hide history">
+                                <i class="fas fa-chevron-up"></i> Hide
+                            </button>
+                        </div>
+                    </div>
+                    <div id="articleHistoryList" class="list-group" role="listbox">
+                    </div>
                 </div>
 
                 <!-- List of articles matching the typed prefix -->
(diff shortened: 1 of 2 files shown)


Jaifroid:
Thank you for this PR. I'll take a look, but please note what it says in CONTRIBUTING about asking to be assigned to an issue before starting work on a PR. I can see that you placed a button on the Home page for this, but as we will soon be switching to a UI like Kiwix JS PWA, there won't be room for such buttons in that layout, as we need only a minimal UI on the Home page, and other options in Configuration. However, I suggest you await my checking your work before working on this further.

Jaifroid (changes requested):
Hi, apologies for the slow response over Xmas/New Year. This implementation looks interesting, but it's currently not working in all circumstances. I noticed the following issues:

* When navigating within a ZIM by clicking on ZIM links in articles in ServiceWorker mode, the history is not saved. It works when using the Random button or when searching for an article, but not when navigating.
* The above issue would be best handled by integrating this History viewing with the existing `pushBrowserHistoryState()` function. When doing this, it would be best to remove the current behaviour of recording search terms in the browser history. This never made much sense IMHO.
* The article history does not clear when a new ZIM is loaded. Links remain to articles in the previous ZIM, but these do not open and load the previous ZIM and article. It may be more pragmatic just to clear the list when a new ZIM is loaded. If a ZIM is automatically reloaded in a new session, then the history should be kept (subject to privacy controls in the next bullet-point).
* The feature has privacy implications, so recording of history should be optional (setting in Configuration) and off by default. 
* The history viewer should close if a user "unclicks" the history button (matches behaviour of existing menu items).

j-j-v:
>
>(...) 
>
>     * The article history does not clear when a new ZIM is loaded. Links remain to articles in the previous ZIM, but these do not open and load the previous ZIM and article. It may be more pragmatic just to clear the list when a new ZIM is loaded. If a ZIM is automatically reloaded in a new session, then the history should be kept (subject to privacy controls in the next bullet-point).
> 
> (...)
>

I am not certain what is meant by "it may be more pragmatic just to clear the list when a new ZIM is loaded" but I hope this still means we keep a permanent history in some accessible form somewhere. Thanks @anuarb799 & @Jaifroid for work on this! legitimately do miss the history feature...

Joe

Jaifroid:
@j-j-v I meant that when a new ZIM is loaded, we don't want entries to the old ZIM still hanging around in the history, as they won't do anything when clicked on. It would in principle be possible to reload the old ZIM when a user clicks on a history item that refers to a ZIM not currently loaded, but only if the app has been granted access to a ZIM directory, and that would be quite a challenging PR. That's why I say it may be more pragmatic (as in straightforward approach) just to clear any old entries when a new ZIM is loaded.

j-j-v:
> I meant that when a new ZIM is loaded, we don't want entries to the old ZIM still hanging around in the history, as they won't do anything when clicked on.

I see. Thanks.

Joe

Jaifroid:
@anuarb799 Do you want to work more on this PR, or shall I close?

j-j-v:
Simply to have this not be a miscommunication later & not to cause frustration but:

For my part I simply require that the history somehow be permanent. Opening the unopened `.zim` file is maybe a less urgent priority & simply having the `.zim` ⸻ the article title / the analogous ⸻ & the date & time are the things I can think to retain in what I've got from this.

Thanks again & hope you can do this @anuarb799!

Joe

Jaifroid:
Looks like this PR has been abandoned. Please re-open if you want to continue working on it.

PR #1410, opened 2026-01-25
fix: prevent white flash on startup by pre-applying dark theme

### LINKED ISSUE #1125: Flash of incorrect theme in jQuery mode
Similar to, but not exactly the same issue as #787, I notice this flashing particularly in jQuery mode. When in light theme, it flashes to dark between article loads, and when in dark theme, it flashes to light. This is not the case in SW mode. The fact that the flash is the opposite of the loaded theme implies it is not a hard-coded value, and should be fixable.

infinityanant:
### Description:
**Fixes:**#1125
This PR addresses the "white flash" issue that occurs when the app is launched or refreshed while a dark theme is active. By checking the user's theme preference and applying a .dark class immediately in the '<head>', we ensure the app paints with a dark background before the main CSS and JS files are fully loaded if the theme was already set to dark mode.
### Changes Made

- Modified www/index.html:
- Added an inline, blocking <script> at the very top of the '<head>'.
- Implemented a MutationObserver to watch for the creation of the '<body>' element.
- Added logic to read kiwixjs-appTheme from localStorage.
- Applies the .dark class to the <body> at the very moment it is created, preventing the browser's default white paint.   
### Rationale  
Currently, the application applies the dark theme via app.js. Since app.js is a large bundle loaded at the end of the document, the browser often performs a "first paint" with the default white background while the script is still downloading or parsing.

By placing a tiny, dependency-free script in the <head>:

- Eliminates Flicker: We beat the CSS/JS load time by targeting the DOM during the initial parse.  
- Performance: Using MutationObserver is more efficient than waiting for DOMContentLoaded because it triggers as soon as the <body> tag exists, rather than waiting for the entire HTML to be parsed.  
### Testing Done
Manual Verification: Tested in Chrome/Edge. Cleared cache, set theme to dark, and refreshed. The white flash is no longer visible.
**Unit Tests**: Ran npm run test-unit — 22 passing, 0 failing.
**Environment**: Verified the fix works even when the app is served via a local server (http-server).
Screenshot 2026-01-25 185432

### DIFF
--- a/www/js/theme-init.js
+++ b/www/js/theme-init.js
@@ -1,0 +1,49 @@
+(function() {
+    // Immediately apply theme background to prevent flash
+    try {
+        var theme = localStorage.getItem('kiwixjs-appTheme');
+        console.log('Theme init: Retrieved theme =', theme); 
+        
+       
+        var bgColor = '#FBEAEA'; 
+        var isDark = false;
+        
+        if (theme && theme.includes('dark')) {
+            bgColor = '#300000';
+            isDark = true;
+            console.log('Theme init: Applying dark theme');
+        } else {
+            console.log('Theme init: Applying light theme');
+        }
+        
+      
+        var html = document.documentElement;
+        html.style.backgroundColor = bgColor;
+        
+        if (isDark) {
+            html.classList.add('dark');
+        }
+        
+       
+        var style = document.createElement('style');
+        style.textContent = 'html, body { background-color: ' + bgColor +';}';
+        document.head.appendChild(style);
+        
+       
+        if (document.body) {
+            document.body.style.backgroundColor = bgColor;
+            if (isDark) {
+                document.body.classList.add('dark');
+            }
+        } else {
+            document.addEventListener('DOMContentLoaded', function() {
+                document.body.style.backgroundColor = bgColor;
+                if (isDark) {
+                    document.body.classList.add('dark');
+                }
+            });
+        }
+    } catch (e) {
+        console.error('Theme init error:', e);
+    }
+})();
--- a/www/js/theme-init.js
+++ b/www/js/theme-init.js
@@ -8,7 +8,7 @@
         try {
             var val = localStorage.getItem(keyPrefix + key);
             if (val !== null) return val;
-        } catch (e) {
+        } catch (e) { // eslint-disable-line no-unused-vars
             // Ignore errors when localStorage is unavailable
         }
 
--- a/www/js/theme-init.js
+++ b/www/js/theme-init.js
@@ -1,25 +1,55 @@
 // Prevent white flash on reload by applying correct background
-(
-    function () {
-        var storedTheme = localStorage.getItem('kiwixjs-appTheme') || "light";
-        var htmlEl = document.documentElement;
-        if (localStorage.getItem('kiwixjs-appCache') === 'true') {
-            if (storedTheme.includes('dark')) {
+(function () {
 
-                htmlEl.classList.add('dark');
+    // DEV: keyPrefix must match params['keyPrefix'] in init.js
+    var keyPrefix = 'kiwixjs-';
 
+    function getStoredValue(key) {
+        try {
+            var val = localStorage.getItem(keyPrefix + key);
+            if (val !== null) return val;
+        } catch (e) {}
+
+        var match = document.cookie.match(
+            new RegExp('(?:^|; )' + encodeURIComponent(key) + '=([^;]*)')
+        );
+
+        return match ? decodeURIComponent(match[1]) : null;
+    }
+
+    var storedTheme = getStoredValue('appTheme') || 'light';
+    var appCache = getStoredValue('appCache');
+
+    function prefersDark() {
+        return window.matchMedia &&
+            window.matchMedia('(prefers-color-scheme: dark)').matches;
+    }
+
+    var isDark = false;
+
+    if (storedTheme.includes('dark') ) {
+        isDark = true;
+    } else if (  storedTheme.includes('auto')  || storedTheme.includes('auto_wikimediaNative')  ) {
+        isDark = prefersDark();
+    }
+
+    var htmlEl = document.documentElement;
+
+    
+    if ( appCache === null || appCache === 'true') {
+        if (isDark) {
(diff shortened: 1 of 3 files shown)


Jaifroid:
Thank you very much for this PR. I haven't tested it manually yet, but I think there might be an issue, which is that Chrome Extensions forbid running inline JS: That only affects Restricted mode (aka jQuery mode), and ServiceWorkerLocal mode, because in full ServiceWorker mode the app now runs as a PWA, with the extension effectively redirecting the browser to browser-extension.kiwix.org.

However, we should come up with an architecture that works in Restricted and ServiceWorkerLocal modes (this local mode only shows up in the Chrome/Edge Browser Extension context).

I believe the solution to that would be to provide the JS as a linked JS file.

You can see how to test the extension code in CONTRIBUTING.md.

infinityanant:
Thank you for the feedback, @Jaifroid. I’ll move the JavaScript into a separate file and link it properly. I’ll also go through the contributing guidelines to make sure I test the extension thoroughly before submitting the update. If any additional changes are required, I’ll work accordingly.

Jaifroid:
Just to say you might not need a separate JS file if you can include the appropriate function in uiUtil.js and run it from the DOMContentLoaded event in app.js.

infinityanant:
@Jaifroid, thank you for the suggestion! I experimented with the DOMContentLoaded approach in app.js, but I noticed the browser still performs an initial white paint just before the script executes.
To ensure a perfectly smooth transition, I've moved the logic into a small, separate js/theme-init.js file in the <head>. This successfully stops the flash while staying CSP-compliant. Does this approach look okay to you, or would you prefer it handled differently?

Jaifroid:
Thanks @infinityanant, and apologies for slow reply. I'll review this once tests complete.

Jaifroid:
@infinityanant Please see the reason for the test failure here: https://github.com/kiwix/kiwix-js/actions/runs/21549352118/job/62896905814?pr=1410 . Please fix that (should be simple), then I can test.

infinityanant:
@Jaifroid thank you for the reply. I will fix the issues.

infinityanant:
@Jaifroid The tests failed because theme-init.js wasn't in the cache files. I've added it now and the local tests pass. Should be good once workflows are approved!

infinityanant:
Hi @Jaifroid , just a quick update: I've resolved the issues. All tests are passing and the PR is ready for another look whenever you have a moment. Thanks!

Jaifroid:
Thank you, I'll test / review this weekend.

Jaifroid:
@infinityanant So, I've been testing this, and I ran into the immediate issue that you seem to have overridden the correct app colours in your branch. Basically, as a visual warning to developers, the app has a pink colour (light mode) or a dark red colour (dark mode) if bypassAppcache is true (see Expert Settings). When that box is unchecked, the app returns to its default colours (light and dark). However, when running your branch, the colours don't change (see screenshot). You should not hard code this state, and should preserve the dynamic behaviour.

You can see how the app should behave by testing it at:

* https://browser-extension.kiwix.org

I have checked that on the main branch, dynamic behaviour is working. It is only your branch that is affected, so it must be related to your PR.

image

Jaifroid (changes requested):
Please fix the hardcoded theme issue.

infinityanant:
Thank you @Jaifroid for testing it and apologies for the issues . I will fix the hardcoded theme issue .

infinityanant:
Hi @Jaifroid,

I’ve fixed the issue in my earlier implementation that was affecting the app’s dynamic behaviour. The Bypass AppCache visual warning and theme switching now work as expected again.

Please let me know if any further adjustments are needed.

Thank you!

Jaifroid (changes requested):
Thanks for fixing the hardcoded theme. There are a number of other issues that need addressing (see comments below). Your general idea is good, but you need to take care of edge cases. Thanks.

infinityanant:
Hi @Jaifroid, thank you for pointing out the issues in my earlier implementation.

I reviewed settingsStore.js to better understand how settings are stored depending on storage availability (localStorage or cookies), and updated my code to correctly handle both cases. I also checked `init.js` to align the use of `keyPrefix` with the existing initialization logic.

Please let me know if any further changes are required.

www/js/theme-init.js
```diff
@@ -0,0 +1,57 @@
...
+    if (storedTheme.includes('dark') ) {
+        isDark = true;
+    } else if (  storedTheme.includes('auto')  || storedTheme.includes('auto_wikimediaNative')  ) {
+        isDark = prefersDark();
+    }
+
+    var htmlEl = document.documentElement;
+
+    
+    if ( appCache === null || appCache === 'true') {
```
Jaifroid: Inconsistent spacing inside parentheses.

Jaifroid (changes requested):
Nearly there! Just some minor compatibility and cleanup things (see comments).

infinityanant:
Hi @Jaifroid, I’ve addressed the compatibility and indentation issues you pointed out and updated the code accordingly. Thank you for the helpful feedback!

www/js/theme-init.js
```diff
@@ -0,0 +1,53 @@
...
+
+    var htmlEl = document.documentElement;
+
+    if (appCache === null || appCache === 'true') {
+        if (isDark) {
+            htmlEl.classList.add('dark');
+        } else {
+            if (htmlEl.classList.contains('dark')) {
+                htmlEl.classList.remove('dark');
+            }
```
Jaifroid: This is dead code, because the `<html>` element doesn't carry a preset class of "dark", and this code will only run when the app first loads (on startup), before any other code might add that class. So you can get rid of this else clause entirely.

Jaifroid:
Thanks! Code-wise, I only see that one small dead code issue. I'll now undertake a few more manual tests.

Jaifroid (approved):
My manual tests are all looking good. Thanks for this useful PR! Please just do that dead code removal, and I think we're good to merge, so just let me know when you're ready for me to squash/merge this for you.

infinityanant:
Hi @Jaifroid I removed the else block as suggested . Thanks for the valuable feedback

PR #1413, opened 2026-02-07
Fix : Properly clear App Cache and IndexedDB on Resetfix reset btn to clear cache #1136

### LINKED ISSUE #1136: Resetting the app no longer empties the app cache
It's not clear to me if this is an actual regression or whether it never worked properly. However, it is intended that a reset should be a full reset, including any caches, so this is definitely a bug.

infinityanant:
## Problem
The reset function was not properly clearing the app cache and IndexedDB because:

1. **Cache deletion failed silently**: The code relied on `getCacheNames()` which queries the Service Worker. When no Service Worker controller was available, it returned `null` and skipped cache deletion entirely.

2. **Race condition**: The reset function triggered app reload before async cleanup operations (IndexedDB and Cache API deletions) completed, leaving data behind.

## Solution
- Changed cache deletion to use `caches.keys()` directly instead of querying the Service Worker, ensuring caches are always accessible
- Wrapped IndexedDB and Cache API deletions in Promises
- Used `Promise.all()` to wait for all async cleanup operations to complete before reloading the app
- This ensures both `kiwixjs-appCache` and `kiwixjs-assetsCache` are properly deleted, along with IndexedDB databases

## Testing
Verified that after clicking "Reset App":
- Cache storage drops from ~17MB to 0MB
- IndexedDB databases (`collDB`) are deleted
- App reloads only after all cleanup completes

Fixes #1136

### DIFF
--- a/www/js/lib/settingsStore.js
+++ b/www/js/lib/settingsStore.js
@@ -25,7 +25,7 @@ var deprecatedKeys = [
 var keyPrefix = params.keyPrefix;
 
 // Tests for available Storage APIs (document.cookie or localStorage) and returns the best available of these
-function getBestAvailableStorageAPI () {
+function getBestAvailableStorageAPI() {
     // DEV: In FF extensions, cookies are blocked since at least FF 68.6 but possibly since FF 55 [kiwix-js #612]
     var type = 'none';
     // First test for localStorage API support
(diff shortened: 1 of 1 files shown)


infinityanant:
Hi @Jaifroid, could you please review this PR whenever you have a moment? I'd be happy to implement any changes or improvements you suggest. Thanks!

Jaifroid:
@infinityanant Could you please address the problems identified in the Copilot review, or where you believe them to be incorrectly identified problems, please explain why?

Jaifroid (changes requested):
See comment above.

infinityanant:
Hi @Jaifroid, apologies for the inactivity due to exams. I’ll go through and address all the comments today.

infinityanant:
Hi, @Jaifroid  I have added the onblocked handler for both Chromium and firefox also added null guard for db.name because browser can sometimes return a database entry with no name and parsing null or undefined to deleteDatabase() can cause errors. 
Changes I did not make suggested by copilot :
Copilot suggested to apply filtering caches to only kiwix prefix but reset should be a full reset including ALL caches. Adding a filter could accidentally skip caches if the naming ever changes in a future version. 
Please let me know if there are any additional improvements or considerations I should address.

Jaifroid:
OK, my sincere apologies for taking so long to get back to you (big project in my day job). I'll try to do final testing and review shortly. Thanks for explanation of why you declined some of the Copilot suggestions.

infinityanant:
Hi @Jaifroid Hope your project work going well. Thanks for the update , glad my reasoning around the Copilot suggestions made sense looking forward to your feedback after the final testing.

Jaifroid (approved):
@infinityanant Thank you very much for this PR, it adds real value. I've thoroughly tested, and sanity-checked the code. The only thing might be that the PR deletes all caches on the given origin, but as we use dedicated origins for the production app, and as only a dev would be affected when running the code on localhost for testing, that's acceptable.

So, thanks once more, and just ping me if you're happy for the PR to be merged.

infinityanant:
Hi @Jaifroid, thank you for testing it and for the detailed review. I’m glad the changes look good overall. You can go ahead and merge the PR.

PR #1420, opened 2026-03-06
Fix typo in URI param

### LINKED ISSUE #1419: Fix typo in URI params and add error handling for storage operations
### **Description**  
While reviewing the settings storage logic, I noticed a typo in the URI parameters and a few places where storage operations could throw runtime errors. These cases currently lack error handling and may interrupt execution in certain browser environments.

**Problems Identified:**
**1. Typo in URI parameter**

The following line contains a typo:  
uriParams = '?allowInternetAccess=truee&contentInjectionMode=serviceworker';  
truee should be true  
**2. No error handling when removing deprecated keys**

The code removing deprecated keys from storage runs without a try–catch block:
**deprecatedKeys.forEach(function (key) {
    if (localStorageTest) localStorage.removeItem(keyPrefix + key);
    settingsStore.removeItem(key);
});**  
If any storage operation throws an error (e.g. localStorage access restrictions), the script may stop executing and remaining keys will not be removed.  
Possible causes:  

- localStorage unavailable or blocked
- browser privacy restrictions
- storage quota exceeded  
**3. No error handling during cookie → localStorage migration**

The _migrateStorageSettings() function migrates settings from cookies to localStorage: 
if localStorage.setItem() fails, the entire migration process may stop.  

 **localStorage.setItem(migratedKey, settingsStore.getItem(cookieKeys[i]));
 settingsStore.removeItem(cookieKeys[i]);
 Possible causes include:**

- localStorage quota exceeded
- malformed cookie values
- browser storage restrictions
- invalid key encoding

infinityanant:
### **Problem**
The settings storage logic had a typo and lacked proper error handling during storage operations.

Typo in URI parameter: The allowInternetAccess parameter was incorrectly set to truee, which could prevent the flag from being recognized correctly by code expecting true.

Storage operations could fail silently: The removal of deprecated keys and the cookie → localStorage migration were performed without error handling. If a storage operation threw an exception (e.g., localStorage quota exceeded or browser storage restrictions), the script could stop executing and leave some keys unmigrated or undeleted.

Migration interruption: During _migrateStorageSettings(), a failure when writing a specific key to localStorage could stop the entire migration process, leaving remaining cookies unmigrated.
This ensures migration continues even if a specific key fails due to
storage quota limits, malformed cookie values, or browser storage restrictions.    
### **Solution**

- Fixed the typo in the URI parameter (truee → true) so the allowInternetAccess flag is interpreted correctly.
- Added try–catch blocks around deprecated key removal to prevent the script from stopping if a storage operation fails.
- Added try–catch handling during cookie → localStorage migration so that migration continues even if a specific key fails.
- Logged failures using console.warn to aid debugging while ensuring the rest of the cleanup and migration process continues.

Fixes #1419

### DIFF
--- a/www/js/lib/settingsStore.js
+++ b/www/js/lib/settingsStore.js
@@ -55,14 +55,8 @@ function getBestAvailableStorageAPI () {
     if (kiwixCookieTest && localStorageTest && regexpCookieKeysToMigrate.test(document.cookie)) _migrateStorageSettings();
     // Remove any deprecated keys
     deprecatedKeys.forEach(function (key) {
-        try{
         if (localStorageTest) localStorage.removeItem(keyPrefix + key);
         settingsStore.removeItem(key); // Because this runs before we have returned a store type, this will remove from cookie too
-        }
-        catch (e) {
-            console.warn('Failed to remove deprecated key: ' + key, e);
-        }
-        
     });
     // Note that if this function returns 'none', the cookie implementations below will run anyway. This is because storing a cookie
     // does not cause an exception even if cookies are blocked in some contexts, whereas accessing localStorage may cause an exception
@@ -298,15 +292,9 @@ function _migrateStorageSettings () {
     for (var i = 0; i < cookieKeys.length; i++) {
         if (regexpCookieKeysToMigrate.test(cookieKeys[i])) {
             var migratedKey = keyPrefix + cookieKeys[i];
-            try {
-            
             localStorage.setItem(migratedKey, settingsStore.getItem(cookieKeys[i]));
             settingsStore.removeItem(cookieKeys[i]);
             console.log('- ' + migratedKey);
-            }
-            catch( e ) {
-                console.warn('Failed to migrate key: ' + cookieKeys[i], e);
-            }
         }
     }
     console.log('Migration done.');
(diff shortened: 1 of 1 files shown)


Jaifroid (changes requested):
Thanks for this PR. I'm happy to accept the typo fix, but the try - catch code is unnecessary, for these reasons:

1. deprecatedKeys removal (lines 57-66): The localStorage.removeItem() call is already guarded by if (localStorageTest). The localStorageTest guard at lines 33-43 already proves localStorage is accessible: it sets and removes a test key inside its own try-catch. If that passes, subsequent removeItem calls on the same store won't throw. settingsStore.removeItem() is a cookie removal (via document.cookie expiry), which also can't throw. So the try-catch here is dead code.

2. _migrateStorageSettings() (lines 301-308): This function only gets called at line 55 when localStorageTest is already true, which means localStorage.setItem() just succeeded moments ago. A storage quota exception mid-migration is theoretically possible but vanishingly unlikely for small settings keys, and if it did happen, you'd arguably want to know loudly rather than silently swallowing it. The console.warn at least logs it, but continuing the loop and migrating subsequent keys while leaving earlier ones in an inconsistent state (some migrated, some not, cookie partially deleted) could actually make things worse than an outright failure.

infinityanant:
Hi @Jaifroid, thank you for the detailed explanation. I understand now that silently continuing with a partially migrated state could actually be worse than an outright failure. I'll remove the try-catch blocks and keep only the typo fix.

infinityanant:
Done removed the try-catch blocks. Only the typo fix remains. Thank you @Jaifroid for the feedback

Jaifroid (approved):
Great, that looks good. Thanks for catching this typo. I'll squash/merge for you.

PR #1423, opened 2026-03-15
Add automated fix for empty catch blocks in libzim WebAssembly files

Tomeshwari-02:
**Problem**

The Emscripten-generated libzim WebAssembly files contain several empty catch blocks and silent early-return patterns that swallow errors. These patterns make debugging extremely difficult and can lead to silent failures when loading ZIM archives.

Examples of problematic patterns
```
catch (err) {} — Empty catch blocks that suppress exceptions
if (!ptr) return; — Silent null pointer checks
if (ABORT) return; — Hidden abort conditions
```
These patterns prevent meaningful error messages from being logged or propagated, making it difficult to identify why certain ZIM files fail to load.

Total occurrences detected:
48 instances across the following files:

- libzim-wasm.js
- libzim-wasm.dev.js
- libzim-asm.js
- libzim-asm.dev.js

**🛠 Solution**

Implemented an automated fix system to ensure errors are logged and easier to debug.

1️⃣ Manual Fix Script

scripts/fix-empty-catch-blocks.js

- Scans libzim files for problematic patterns.
- Injects error logging into empty catch blocks and silent return cases.
- Can be executed manually when needed.

Usage :

`npm run fix-catch-blocks`

2️⃣ Rollup Plugin (Automatic Fix)
rollup-plugin-fix-empty-catch.js

- Runs automatically during production builds.
- Ensures generated WebAssembly files are patched before packaging.
- Requires no manual intervention.

**📂 Changes**
New Files

- scripts/fix-empty-catch-blocks.js — Standalone script for fixing empty catch blocks
- rollup-plugin-fix-empty-catch.js — Rollup plugin for automated fixes during build
- CATCH_FIX_README.md — Documentation and usage instructions

 Modified Files
- rollup.config.js — Integrated the Rollup plugin
- package.json — Added new npm script

**🧪 Testing**
The script successfully processed all four libzim files, fixing 48 problematic instances.

Build Verification
`npm run build-src
`
Result:

✅ Build completed successfully

No build errors were introduced after applying the fixes.


Jaifroid:
@Tomeshwari-02 Thanks for this suggestion. However, what you're doing here is editing dependencies for a single repo that uses the ASM and WASM ports, so you're essentially editing compiled binaries in a reader. There are currently two different projects that depend on these compiled libzim ASM/WASM files, and editing the binaries after compilation is not the way to fix any error or poor functionality in those files.

Instead, you need to take a look at the repository where these binaries are compiled, which is the [openzim/javascript-libzim](https://github.com/openzim/javascript-libzim/) repository. 

Having said that, it looks like you're trying to fix something in the Emscripten implementation itself. The empty catch blocks are probably to do with the compile options we use when compiling the libzim C++ code to ASM/WASM with Emscripten. We set flags designed to produce a minimum and fast ASM/WASM binary, which suppress diagnostic error catching.

As we only use these binaries for full-text search currently, it's not very clear to me what you are referring to when you say "These patterns make debugging extremely difficult and can lead to silent failures when loading ZIM archives". Could you give an example of a ZIM that is producing a difficult-to-debug error on load and what the consequences are?

First of all, please identify an actual problem case that your fix would address. Then, if it's still an issue, take a look at the source code on openzim/javascript-libzim and propose a fix there. It may even be that your fix needs to be proposed in the Emscripten repository, because what you're modifying looks like Emscripten code, not our code.

PR #1429, opened 2026-03-29
fix: remove uiUtil dependency from zimArchive.js

### LINKED ISSUE #813: Remove the dependencies on the UI of js backend files
In zimArchive and zimArchiveLoader files, there are error cases where we display an error message to the user.
It was initially done using `window.alert`, and is now done using `uiUtil.systemAlert`.

In all cases, a backend js file should not depend on the UI: it breaks the software layers.
For example, this code might one day run inside a Worker, where it would not have access to the UI at all, and this would simply fail.

The error should instead be returned to the caller, that would have the responsibility to handle it: maybe display it to the user, or something else

Prasannacharya-3151:
This PR removes UI dependency from zimArchive.js by eliminating the import and usage of uiUtil.
Backend files should not depend on UI components, so uiUtil calls were removed/commented.
Tested locally and working correctly.
Closes #813

### DIFF
--- a/www/js/lib/zimArchive.js
+++ b/www/js/lib/zimArchive.js
@@ -27,7 +27,7 @@
 import zimfile from './zimfile.js';
 import zimDirEntry from './zimDirEntry.js';
 import util from './util.js';
-import uiUtil from './uiUtil.js';
+// import uiUtil from './uiUtil.js';
 import utf8 from './utf8.js';
 import translateUI from './translateUI.js';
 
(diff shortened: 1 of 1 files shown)


Jaifroid:
@Prasannacharya-3151 You should always comment in the appropriate issue before undertaking a PR, and propose your solution first, so you can be assigned.

Jaifroid (changes requested):
@Prasannacharya-3151 Although you have identified some further UI dependencies in zimArchive.js, you can't just remove them. You need to provide an alternative solution for reporting the search provider to the UI that does not create a dependency.

Jaifroid:
No activity, so closing.

Issue #1424, opened 2026-03-16
Improve search UX: Show title suggestions when no results are found


infinityanant:
### **Problem**
Currently when a user searches for a term with a typo or an inexact title, they are shown "Finished. No articles found." with no further guidance. For example searching "apachee" instead of "apache" returns a complete dead end with no recovery path particularly frustrating for offline users who cannot refine their search using the internet.

<img width="600" height="101" alt="Image" src="https://github.com/user-attachments/assets/60a2987c-f40e-4347-87ee-a63855cc7e6c" />   
 
### **Proposed Solution**  
When search returns zero or very few results, **extract a prefix** from the query and run a **lightweight secondary search** on the existing title index using the existing **findDirEntriesWithPrefixCaseSensitive** function. **Display the top 5 closest title matches as "Did you mean"** suggestions below the "No articles found" message.
For single word queries, the first 5 characters are used as the prefix. For multi word queries like "Theory of relatvty", the last word is used as the prefix since typos most commonly occur at the end of a query. Suggestions are then ranked by Levenshtein distance against the original query to ensure the closest matches appear first.
This approach reuses existing binary search infrastructure  no new indexing or article content reading required. Cost is one additional O(log n) binary search triggered only on failed searches.  
### **Note to the implementation**  
In the initial implementation, prefix extraction will handle typos at the end of queries. Typos at the beginning or middle of the first word (e.g. 'apche') are a known limitation that could be addressed in a future iteration using more sophisticated fuzzy matching across multiple prefix variants.

infinityanant:
Hi @Jaifroid , I wanted to check if this feature direction seems valuable for kiwix-js before I invest more time in it. Does the approach of using the existing title index for suggestions seem reasonable, or do you see any fundamental issues with it?

periza0:
hi, i'd be interested in working on this issue. is this still available, and would you be open to assigning it to me?

Jaifroid:
Thank you, both. I'm sorry this issue got overlooked when first posted. I think this is rife with difficulties about deciding how much of a prefix to truncate in order to run the secondary search, plus slowness of search as it is in low-power devices (old browsers, mobile). And there would be real issues testing in the myriad languages of Wikipedia. Nice idea in principle, but it adds complication, and using full-text search covers some of it. If we add suggestions at all, they should be the Xapian-based ones provided by the libzim WASM in this repo.

PR #1431, opened 2026-04-07
Improve search UX: Show title suggestions when no results are found #1424

anmoltyagi18:
I've run node -c www/js/app.js against the modified file locally, and it executed smoothly with exit status 0

### DIFF
--- a/www/js/app.js
+++ b/www/js/app.js
@@ -2273,6 +2273,36 @@ function searchDirEntriesFromPrefix (prefix) {
 }
 
 /**
+ * Compute the Levenshtein (edit) distance between two strings using an iterative DP matrix approach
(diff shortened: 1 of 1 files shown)


Jaifroid:
@anmoltyagi18 I've tested your PR, but I can't see it doing anything. I tried to type a search term with a deliberate small typo in it, but nothing happened, no alternatives suggested. Perhaps you could give some examples of how this is supposed to work, and a screenshot or so of it actually working? Note that running node -c on a file within the app won't test that the app is actually working, and won't test any UI-controlled functions like your search improvement.

Jaifroid:
No activity, so closing. If you want to work on this again, please reopen.

PR #1436, opened 2026-04-24
Update release/nightly uploads for new destination

rgaudin:
As part of a server/infra change, the upload of nightly and release builds is done to a new destination.

The receiving service now uses per-project (well per-repo) users with independent SSH Keys. Each users now only has access to the folders in which it is expected to upload files.

- Host remains the same: `master.download.kiwix.org`
- Username changes from `ci` to `kiwix-js`
- Port changes from `30022` to `30322`
- Paths remains the same

⚠️ DO NOT MERGE ATM

### DIFF
--- a/.github/workflows/Release.yml
+++ b/.github/workflows/Release.yml
@@ -51,7 +51,7 @@ jobs:
       # Set up secret files from encrypted secrets
       - name: Set up secret files and copy needed scripts to dist
         env:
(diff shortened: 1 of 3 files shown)


Jaifroid:
OK, thanks for this and also for the equivalent on the PWA. Let me know when you'd like me to review the PR. Sorry for lack of interaction lately -- big project coming to crunch-point. After that I want to make a new KJS release.

rgaudin:
Well let me know if you see anything that bothers you but it's just simple target change.
There won't be any transition so the moment the server-side is switched I plan on merging all those PRs.

PR #1471, opened 2026-08-17
Fix flaky Gutenberg author search tests

### LINKED ISSUE #1468: Flaky e2e tests: Gutenberg Author search fails intermittently on BrowserStack Edge Legacy 18
The two `Author search` tests in `Gutenberg_ro test [ZSTD compression] [SW mode]` fail intermittently
on the **BrowserStack Edge Legacy 18 / Win10** job, and on no other browser. The rest of the run is
unaffected, so a single job failure fails the whole `tests-linux` matrix and the PR has to be re-run.

```
  1) Gutenberg_ro test [ZSTD compression] [SW mode]
       Author search Autocomplete:
     NoSuchElementError: The specified element was not found in the current frame or window.

  2) Gutenberg_ro test [ZSTD compression] [SW mode]
       Author search Results:
     NoSuchElementError: The specified element was not found in the current frame or window.
```

Always the same two tests, always `42 passing / 4 pending / 2 failing`.

## It is not caused by any particular change

Recent occurrences:

* [Run 31478520807](https://github.com/kiwix/kiwix-js/actions/runs/31478520807) on `main` — the commit
  was documentation-only (`Update CONTRIBUTING for the AI era`), so no application or test code was
  involved.
* [Run 31895799467](https://github.com/kiwix/kiwix-js/actions/runs/31895799467) on the branch for
  #1466, attempts 1 and 2. Attempt 3, on identical code, passed with the same job green.

Roughly 2 failures in the last 15 CI runs on `main`.

## Likely cause

Both tests do their lookup like this ([`gutenberg_ro.e2e.spec.js:390`](https://github.com/kiwix/kiwix-js/blob/main/tests/e2e/spec/gutenberg_ro.e2e.spec.js#L390) and
[`:402`](https://github.com/kiwix/kiwix-js/blob/main/tests/e2e/spec/gutenberg_ro.e2e.spec.js#L402)):

```js
const filter = await driver.wait(until.elementIsVisible(driver.findElement(By.id('author_filter'))), 1500);
```

`driver.findElement(...)` is an argument, so it is evaluated **once, immediately**, before `driver.wait`
ever retries. The 1500 ms applies only to the visibility check. If the element is not there at that
instant, the test throws with nothing to retry.

These two tests run straight after `Navigating back`, which clicks `btnBack` three times. On a slow
browser the `articleContent` iframe can still be replacing its document at that point, which both
removes the element momentarily and can invalidate the frame handle. The preceding test survives this
because it uses `until.elementLocated(...)` *inside* the wait, which does retry.

Edge Legacy 18 is also the only runner that shares a single driver across all three specs
([`edge18.bs.runner.js:39-43`](https://github.com/kiwix/kiwix-js/blob/main/tests/e2e/runners/edge/edge18.bs.runner.js#L39-L43)),
so it accumulates state and runs slowest — which fits it being the only browser affected.

## Suggested fix

In both tests, move the lookup inside the wait so it retries, and re-enter the iframe explicitly, as
the neighbouring tests already do:

```js
await driver.switchTo().defaultContent();
await driver.switchTo().frame('articleContent');
const filter = await driver.wait(until.elementLocated(By.id('author_filter')), 3000);
await driver.wait(until.elementIsVisible(filter), 1500);
```

The same `elementIsVisible(driver.findElement(...))` pattern appears five more times in this file
(lines 246, 263, 328, 353 and 415). Those are not currently failing — three of them target `prefix` in
the main document, which is stable — but they carry the same latent problem and could be converted at
the same time.

## Verifying a fix

The failure only reproduces on BrowserStack Edge Legacy 18, so it cannot be confirmed locally, and a
single green run proves nothing given the base rate. It needs watching over several CI runs.

soumyajoshi009:
Fixes #1468

The author search tests called `findElement()` before `driver.wait()`,
which could cause an immediate failure when the element was temporarily
not present.

This change:
- waits for `author_filter` to be located before checking visibility
- explicitly re-enters the `articleContent` iframe
- applies the fix to both author search tests

Local Edge E2E tests were run, but the suite has unrelated failures in
the local environment. The original failure is specific to BrowserStack
Edge Legacy 18.


Jaifroid:
Thanks for this PR. I'm just running initial tests, but we'll also need to test this with a full test run that can only be made from a parallel branch pushed by a maintainer. Will sort that out soon.

tests/e2e/spec/gutenberg_ro.e2e.spec.js
```diff
@@ -387,7 +387,16 @@ function runTests (driver, modes, keepDriver) {
...
+                await driver.switchTo().frame('articleContent');
+                const filter = await driver.wait(
+                    until.elementLocated(By.id('author_filter')),
+                    3000
+                );
+                await driver.wait(
+                    until.elementIsVisible(filter),
+                    1500
+                );
                 await filter.sendKeys('Mihai Eminescu');
```
Jaifroid, line 417: Could you add a `clear()` before this? As it stands, if this test fails and mocha retries it, the second attempt appends to a box that already contains the text, giving `Mihai EminescuMihai Eminescu` — the autocomplete then returns nothing and the retry can never pass.

```
await filter.clear();
await filter.sendKeys('Mihai Eminescu');
```

Since the whole point of the fix is to make the retries effective, it'd be a shame to leave a trap in the retry path.

tests/e2e/spec/gutenberg_ro.e2e.spec.js
```diff
@@ -399,7 +408,18 @@ function runTests (driver, modes, keepDriver) {
...
+                const filter = await driver.wait(
+                    until.elementLocated(By.id('author_filter')),
+                    3000
+                );
+
+                await driver.wait(
+                    until.elementIsVisible(filter),
+                    1500
+                );
                 await filter.sendKeys(Key.ENTER);
```
Jaifroid, line 445: This test depends on `Mihai Eminescu` still being in the filter box from the previous test -- it only presses Enter. That coupling is the other half of the retry problem: if Author search Autocomplete fails or retries, this test inherits a box in an unknown state.

Could you make it self-contained? Roughly:

```
await filter.clear();
await filter.sendKeys('Mihai Eminescu');
await driver.wait(until.elementsLocated(By.className('ui-menu-item')), 1500);
await filter.sendKeys(Key.ENTER);
```

The intermediate wait is there because typing the name opens the jQuery UI autocomplete dropdown, and pressing Enter before it has settled may select a menu item instead of applying the filter. I'm not certain that wait is needed -- if it turns out to be awkward, just adding `clear()` to line 400 is the important part and we can leave this test as-is.

tests/e2e/spec/gutenberg_ro.e2e.spec.js
```diff
@@ -399,7 +408,18 @@ function runTests (driver, modes, keepDriver) {
                     this.skip();
                 }
                 // search by author name and press enter to apply the filter
-                const filter = await driver.wait(until.elementIsVisible(driver.findElement(By.id('author_filter'))), 1500);
+                await driver.switchTo().defaultContent();
+                await driver.switchTo().frame('articleContent');
+
```
Jaifroid: Please remove this blank line.

tests/e2e/spec/gutenberg_ro.e2e.spec.js
Jaifroid: (This is a pre-existing issue, but worth fixing here!) I've just realized this assertion can't fail. `//*[@id="books_table"]/tbody` matches the table's single `<tbody>`, so `searchList.length` is 1 whether or not the author filter did anything at all. The test currently passes even if filtering is completely broken.

There's also an ordering trap: lines 426–427 clear the filter and press Enter to revert it, so by the time we reach the assertion the table is back to its unfiltered state. Any check on the filtered results has to happen before that revert.

Could you make it assert something real? Something along these lines, reusing the selectors and the wait-for-condition style already used elsewhere in this file (table-title at lines 266–267, the polling wait at lines 305–308):

```
await filter.sendKeys(Key.ENTER);
// Wait for the filter to actually reduce the list, and read the results before reverting
await driver.wait(async function () {
    const books = await driver.findElements(By.className('table-title'));
    return books.length === EXPECTED_COUNT;
}, 3000, 'Author filter should reduce the book list');
const bookList = await driver.findElements(By.className('table-title'));
const firstBookName = await bookList[0].getText();
// revert whatever was typed in the search box and press enter to remove filter
await filter.clear();
await filter.sendKeys(Key.ENTER);
assert.equal(firstBookName, 'EXPECTED_TITLE');
```

I've deliberately left EXPECTED_COUNT and EXPECTED_TITLE as placeholders: please read the real values off an actual run rather than guessing them. I don't know offhand how many books by Mihai Eminescu are in this ZIM; note that the single ui-menu-item asserted in the previous test is one author suggestion, which tells us nothing about the number of books. If it turns out to be a single book, the count wait plus the title check is belt-and-braces and you could drop one of them.

Waiting for the count rather than just locating the elements is what matters here: `table-title` elements are present in the unfiltered table too, so `elementsLocated` alone would resolve immediately against the old contents and wouldn't prove the filter had been applied.

Jaifroid (changes requested):
Thanks for tackling this so quickly. I've pushed a parallel branch so the BrowserStack jobs can run (they're skipped on fork PRs because the credentials aren't exposed), and I'll be collecting results over a number of runs. Note that the failure only shows up about twice in fifteen runs, so a few green runs won't be conclusive on their own — this will take a little patience.

A few requested changes below, plus two things for consideration:

Requested:

* Make the two tests independent of each other and safe to retry (comments on lines 400 and 423).
* Trailing whitespace on line 413.

Would be good to include:

The issue #1468  also flagged the same `elementIsVisible(driver.findElement(...))` pattern at five other call sites. The two at lines 246 and 263 (popularity_sort and alpha_sort) are inside the iframe and carry exactly the same latent problem, so I'd like those two converted here too. The three targeting prefix (lines 328, 353, 435) are in the main document and much more stable: leave those if you prefer. I generally prefer to fix adjacent problems in the same PR rather than open follow-ups, given the overhead of pushing the parallel branch in this case.

Jaifroid:
PS @soumyajoshi009 Please update the branch before continuing (use Update branch button below), then **pull the commits to your local fork before doing more coding**. Do not force-push any changes, please.

soumyajoshi009:
Hy Jaifroid, I've addressed the other requested changes:

- Made the Author search Autocomplete test self-contained by clearing the filter before typing Mihai Eminescu.
- Made the Author search Results test self-contained by clearing the filter, typing Mihai Eminescu, waiting for the autocomplete result, and pressing Enter.
- Also addressed the whitespace/formatting feedback.

I'm currently stuck on the remaining assertion change. I removed the assumption that books_table/tbody represents the number of filtered results and tried to determine the actual table-title count and first title from a real run, rather than guessing the expected values.

I’m currently blocked by the E2E environment/state setup. The Gutenberg suite is failing before the author tests reach a valid articleContent state. In the full run, Navigating back times out, and both author tests then fail because articleContent cannot be located. There are also earlier failures in the suite, including the app reset and loading the Gutenberg ZIM.

I also tried running the author tests in isolation, but that skips the preceding state setup and produces the same articleContent error.

So I haven’t been able to reliably read the filtered table-title count/title locally. I don't want to guess EXPECTED_COUNT or EXPECTED_TITLE.

Could you advise on the expected count/title for this ZIM, or on the intended way to run this part of the suite locally?

Jaifroid:
@soumyajoshi009 It's hard to comment on your setup and why things are failing, but in principle, if you've run `npm install` in the root of your fork, then you should pull all the needed dependencies. Then you can simply build (`npm run build`) and `npm run test-e2e-chrome` (or edge or firefox).

You could also simply load the ZIM (/tests/zim/gutenberg-ro) in an app manually, do the search, and manually find the number of books.

Does that answer your question?

tests/e2e/spec/gutenberg_ro.e2e.spec.js
```diff
@@ -243,7 +243,15 @@ function runTests (driver, modes, keepDriver) {
                 // We switch to default Content and back to Iframe in case we are retrying the test
                 await driver.switchTo().defaultContent();
                 await driver.switchTo().frame('articleContent');
-                await driver.wait(until.elementIsVisible(driver.findElement(By.id('popularity_sort')))).click();
+                const popularitySort = await driver.wait(
+                    until.elementLocated(By.id('popularity_sort')),
+                    3000
```
Jaifroid: I think it would be safer to set this to `10000`, so that there is enough time for a retry.

tests/e2e/spec/gutenberg_ro.e2e.spec.js
```diff
@@ -260,7 +268,15 @@ function runTests (driver, modes, keepDriver) {
                 await driver.switchTo().defaultContent();
                 await driver.switchTo().frame('articleContent');
                 let firstBookName = '';
-                await driver.wait(until.elementIsVisible(driver.findElement(By.id('alpha_sort')))).click();
+                const alphaSort = await driver.wait(
+                    until.elementLocated(By.id('alpha_sort')),
+                    3000
```
Jaifroid: Likewise, `10000` would be safer here.

tests/e2e/spec/gutenberg_ro.e2e.spec.js
```diff
@@ -399,13 +425,35 @@ function runTests (driver, modes, keepDriver) {
...
+                    until.elementIsVisible(filter),
+                    1500
+                );
+                await filter.clear();
+                await filter.sendKeys('Mihai Eminescu');
+                await driver.wait(until.elementLocated(By.className('ui-menu-item')), 1500);
                 await filter.sendKeys(Key.ENTER);
-                const searchList = await driver.wait(until.elementsLocated(By.xpath('//*[@id="books_table"]/tbody')));
+                const EXPECTED_COUNT = 1;
+                const EXPECTED_TITLE = 'Poezii';
```
Jaifroid: No reason to use screaming snake constants here. Just put the expected values in the functions below.

tests/e2e/spec/gutenberg_ro.e2e.spec.js
```diff
@@ -399,13 +425,35 @@ function runTests (driver, modes, keepDriver) {
...
                 }
                 // search by author name and press enter to apply the filter
-                const filter = await driver.wait(until.elementIsVisible(driver.findElement(By.id('author_filter'))), 1500);
+                await driver.switchTo().defaultContent();
+                await driver.switchTo().frame('articleContent');
+                const filter = await driver.wait(
+                    until.elementLocated(By.id('author_filter')),
+                    3000
+                );
+
```
Jaifroid: Please remove this blank line.

tests/e2e/spec/gutenberg_ro.e2e.spec.js
```diff
@@ -399,13 +425,35 @@ function runTests (driver, modes, keepDriver) {
...
+                await driver.wait(
+                    until.elementIsVisible(filter),
+                    1500
+                );
+                await filter.clear();
+                await filter.sendKeys('Mihai Eminescu');
+                await driver.wait(until.elementLocated(By.className('ui-menu-item')), 1500);
```
Jaifroid: One refinement here. `until.elementLocated(By.className('ui-menu-item'))` can match a menu item left in the DOM by the previous test rather than the one your new search produces, so it may resolve instantly without the dropdown actually being ready.

It's important to re-query on each poll instead, in the same style you've already used for the book count:
```
await filter.clear();
await filter.sendKeys('Mihai Eminescu');
await driver.wait(async function () {
    const items = await driver.findElements(By.className('ui-menu-item'));
    if (items.length !== 1) return false;
    return items[0].isDisplayed().catch(() => false);
}, 3000, 'Author autocomplete should show a single suggestion');
await filter.sendKeys(Key.ENTER);
```
The `.catch(() => false)` is important: if an element goes stale mid-poll we want that to count as "not ready yet" and try again, not to throw out of the wait.

Jaifroid (changes requested):
Thanks for the updates, which are great. However, you've ended up tightening the timings on two tests that were already flaky, so it's best to relax those as suggested below. Also a suggested syntax for a more thorough test of the `ui-menu-item` element. Nearly there!

benoit74:
@Jaifroid it is very important to mention that Gutenberg scraper is undergoing a massive revamp and UI will soon be totally different. I wasn't aware there was tests specifically targeting this scraper ZIMs. I did not dive into details of the PR tbh (and probably won't unless you request my help), but I'm pretty worried this PR is just a waste of time since everything will be broken as soon as we release the scraper and create ZIMs.

Jaifroid:
@benoit74 Thanks for the info, but no worries, this is a pinned archive, so it won't suddenly break These tests are more for code validation than for seeing if we can read the latest ZIMs. I think that's a job for the ZIM test suite. No need for you to review, but thanks for being willing!

Jaifroid (approved):
Great, thanks for the latest update @soumyajoshi009. It's all looking good now, and tests are passing and looking more robust. I'll go ahead and squash/merge. Thank you for your valuable contribution.

PR #1473, opened 2026-08-19
Rename "Bypass AppCache" option to "Developer Mode" #1459

### LINKED ISSUE #1459: Rename the Bypass Appcache option (Expert Settings) to Developer Mode
This has already been done in the PWA, and for consistency, should be done here. It will be necessary to change the descriptions of this mode also in the info and CONTRIBUTING documents (and similar). Translations of strings also. However, the parameter should still remain as "appCache", as there is no need to change that, and it better documents what the param actually does.

Jaifroid:
Fixes #1459. Completes the work begun by @floze-the-genius in #1462, which went unattended at the final leg; their Spanish and French renames and the French tooltip correction are carried over here, and they are credited as co-author on the commit.

## What changed

The option is now **Developer Mode** in the UI and in its tooltip, in English, Spanish and French:

| | Label | Tooltip |
| --- | --- | --- |
| en | **Developer Mode** (_disables offline use of this app!_) | now opens "Developer Mode bypasses the app's own AppCache, so that you always see the latest version of your code." |
| es | **Modo de desarrollador** | "El Modo de desarrollador omite la AppCache…" |
| fr | **Mode développeur** | "Le Mode développeur contourne l'AppCache…" |

Also renamed in `www/index.html` (label and `title` fallback), `CONTRIBUTING.md` (6 places), `CHANGELOG.md`, and the code comments and e2e test descriptions that named the option.

**The underlying setting is deliberately untouched**, as it documents what the option actually does: `params.appCache`, the `appCache` storage key and querystring parameter, the `bypassAppCache*` element IDs, and the `configure-expert-bypassappcache*` i18n keys.

## Adjacent fixes rolled in

* **Deletes the three dead i18n keys left by #1472** (`dialog-bypassappcachecheck-message`, `dialog-bypassappcache-conflict-title`, `dialog-bypassappcache-conflict-message`) across all three locales. This was the leg #1462 never completed.
* **`CONTRIBUTING.md` claimed the option requires Service Worker mode.** Since #1472 it is offered in any browser with the ServiceWorker API. Corrected, and the paragraph rewrapped to the file's width.
* **French tooltips showed a literal `&nbsp;`.** `translateUI` assigns tips via `element.title`, which does not decode HTML entities (unlike the `data-i18n` path, which uses `innerHTML`). The two affected French strings now use a real non-breaking space.
* **`abouut-app-title` typo** in `i18n/en.jsonp.js` meant English silently fell back to the hardcoded HTML. Invisible to users, as the fallback string is identical, so no CHANGELOG entry.

## Testing

* `npx eslint .` — clean
* `npm test` — 61 passing
* `npm run build` — clean
* `bash scripts/test_duplicate_values.sh` — both checks identical
* i18n key parity verified across all three locales by parsing the JSONP (303 keys each, sets identical, dead keys gone)
* `npm run test-e2e-edge` — 50 passing, 4 pending, including the four AppCache-in-Restricted-mode tests
* Headless Edge check of the built `dist/`: label and tooltip render correctly in all three languages with accents intact; the French tooltip now shows "AVERTISSEMENT : " rather than the raw entity; option still offered and toggleable in Restricted mode with no dead dialogue

Not run locally: Firefox, IE Mode, and the extension builds — leaving those to the hosted matrix.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


i18n/fr.jsonp.js
```diff
@@ -116,8 +116,8 @@ document.localeJson = {
       "configure-expert-hideexternallinkwarning-tip": "Si cette option est cochée, les liens vers des sites externes s'ouvriront dans un nouvel onglet sans confirmation.",
       "configure-expert-disabledragdrop": "<strong>Désactiver le glisser-déposer</strong> (si cela provoque des anomalies)",
       "configure-expert-disabledragdrop-tip": "Cette option désactive la fonctionnalité de glisser-déposer des fichiers ZIM pour les charger dans l'application. Il n'est pas recommandé de désactiver cette fonctionnalité.",
-      "configure-expert-bypassappcache": "<strong>Désactiver AppCache</strong> (<i>désactive l'utilisation hors ligne de cette application !</i>)",
-      "configure-expert-bypassappcache-tip": "AVERTISSEMENT&nbsp;: Désactiver cette option désactivera l'utilisation hors ligne de la PWA. Le paramètre effacera tous les caches de l'API de Cache, mais le cache des actifs sera utilisé à moins qu'il ne soit également désactivé ci-dessus. Il facilite le développement de nouveaux codes dans la PWA.",
+      "configure-expert-bypassappcache": "<strong>Mode développeur</strong> (<i>désactive l'utilisation hors ligne de cette application !</i>)",
```
Jaifroid: There should be &nbsp; before the exclamation mark to ensure it doesn't end up on a line of its own.

CHANGELOG.md
```diff
@@ -10,8 +10,10 @@ Then it was renamed to "Kiwix HTML5" (and used the ZIM file format). Finally it
...
-* FIX: The "Bypass AppCache" setting is now shown in every mode in which it does something, including Restricted mode and ServiceWorkerLocal mode, and is hidden only in browsers with no ServiceWorker API, where it has no effect
+* FIX: The "Developer Mode" setting (previously called "Bypass AppCache") no longer blocks a switch to Restricted mode, and can now be turned off from Restricted mode: the Service Worker still serves the app's own code in that mode, so the setting applies there too
+* FIX: The "Developer Mode" setting is now shown in every mode in which it does something, including Restricted mode and ServiceWorkerLocal mode, and is hidden only in browsers with no ServiceWorker API, where it has no effect
+* FIX: French tooltips no longer show a literal "&nbsp;" before a colon: tooltips are set as plain text, so the entity was never decoded
+* DEV: Renamed the "Bypass AppCache" option to "Developer Mode" in the UI, its tooltip and the contributor documentation (same functionality, and the underlying `appCache` setting is unchanged)
```
Jaifroid: As interim-4.3.4 is already released, any changes that are new rather than simply renaming an existing CHANGELOG entry need to go into an Unreleased section above this block.

Jaifroid (commented):
Some minor changes.

i18n/fr.jsonp.js
Jaifroid: Done in 77d2e50 — the label now reads `...cette application&nbsp;!`.

Worth flagging that the entity is correct *here* but would be wrong one line below: `translateUI` applies `data-i18n` with `innerHTML` (so entities decode), but applies `data-i18n-tip` with `element.title`, which does not decode them. That is what the `&nbsp;`-in-tooltips fix in this PR was about. Verified in a headless Edge run against the built `dist/`: the label renders with a real non-breaking space.

CHANGELOG.md
Jaifroid: Good catch — fixed in 77d2e50. There is now an `## Unreleased` section above the v4.3.4 block holding the three new entries: the French typography fix, the rename itself, and the dead-key removal (which I had not logged at all).

The two "Developer Mode" FIX lines stay in v4.3.4, as they only rename already-released entries. I also moved the "(previously called \"Bypass AppCache\")" gloss off the v4.3.4 line and onto the Unreleased entry that performs the rename — in a 4.3.4 entry it read backwards, since the option was still called Bypass AppCache in that release. Easy to put back if you would rather it stayed.

PR #1492, opened 2026-09-17
Encode url and title in DirEntry toStringId to prevent delimiter corruption

### LINKED ISSUE #1491: DirEntry.toStringId()/fromStringId() corrupts title when url or title contains "|"
## Summary

`DirEntry.prototype.toStringId()` in `www/js/lib/zimDirEntry.js` joins several DirEntry fields with `|` as a separator, and `DirEntry.fromStringId()` splits on `|` to reconstruct them. If `url` or `title` themselves contain a literal `|`, the split produces the wrong field boundaries.

app.js wraps the whole id in `encodeURIComponent()` before writing it to a DOM attribute (`buildArticleListElement`), and reverses that with `decodeURIComponent()` before calling `fromStringId()` (`findDirEntryFromDirEntryIdAndLaunchArticleRead`). That outer encode/decode pair does not protect the inner `|` separators, since it encodes and decodes literal `|` characters the same way it encodes/decodes any other character in the string — so the ambiguity survives the round trip unchanged.

## Real-world impact

I scanned real ZIM archives with `|` in their directory entries (a small node script that reads the ZIM directory listing directly, no cluster decompression needed) rather than relying on a constructed example. Two concrete cases:

- `hsm.stackexchange.com` (History of Science and Mathematics Stack Exchange): titles containing inline LaTeX using `|` for absolute value, e.g. `Did Bolzano conclude that $| \mathbb R | \ne | \mathbb N|$?`
- `tonedear.com` (a Zimit-produced archive): titles in the common `<Page title> | <Site name>` shape, e.g. `Contact | Ear Training`

For both, `url` is serialized before `title` in the joined string, so `url` is unaffected and the article opens normally — this is **not** a navigation or redirect failure. The visible symptom is that `title` gets truncated at the first `|` once split, which shows up wherever the truncated title is used elsewhere in the UI (e.g. the article list, or a "return to..." link).

Round-tripping the real title above through the current `toStringId()` → `encodeURIComponent()` → `decodeURIComponent()` → `fromStringId()` pipeline:

```
real title:          "Contact | Ear Training"
reconstructed title: "Contact "
```

## Suggested fix

`encodeURIComponent()` the `url` and `title` fields individually before joining them with `|` in `toStringId()`, and `decodeURIComponent()` them individually (with a safe fallback if decoding fails, to tolerate any legacy unencoded ids already stored) in `fromStringId()`. This is compatible with the outer `encodeURIComponent()`/`decodeURIComponent()` wrapping in app.js, since encoding the `%` characters produced by the inner encoding is itself reversible.

I have a fix + unit tests ready and will open a PR against this issue. This was originally reported/fixed against kiwix-js-pwa (kiwix/kiwix-js-pwa#966), but per CONTRIBUTING.md, shared app code fixes belong here first.

bhuvan-somisetty:
Fixes #1491.

## Problem

`DirEntry.prototype.toStringId()` joins several fields with `|`, and `DirEntry.fromStringId()` splits on `|` to reconstruct them. A literal `|` inside `url` or `title` is indistinguishable from the separator. app.js wraps the whole id in `encodeURIComponent()`/`decodeURIComponent()` when storing it in a DOM attribute, but that outer encoding does not protect the inner `|` separators, since it treats a literal `|` the same as any other character.

## Real-world impact

Verified against real directory entries (not a constructed example) from two archives that contain `|` in titles:

- `hsm.stackexchange.com` — inline LaTeX using `|` for absolute value, e.g. `Did Bolzano conclude that $| \mathbb R | \ne | \mathbb N|$?`
- `tonedear.com` (Zimit) — the common `<Page title> | <Site name>` shape, e.g. `Contact | Ear Training`

In both cases `url` is serialized before `title`, so it is unaffected and the article still opens (this is not a navigation/redirect failure). `title` gets truncated at the first `|` once reconstructed, which shows up wherever that title is displayed.

Before this fix, round-tripping the second example through the existing `toStringId()` → `encodeURIComponent()` → `decodeURIComponent()` → `fromStringId()` pipeline:

```
real title:          "Contact | Ear Training"
reconstructed title: "Contact "
```

## Fix

`encodeURIComponent()` `url` and `title` individually before joining with `|` in `toStringId()`, and `decodeURIComponent()` them individually in `fromStringId()`, with a safe fallback to the raw value if decoding throws (so any legacy unencoded ids already stored don't break).

## Testing

- Added `tests/unit/spec/zimDirEntry.test.js` (mocha/chai), covering the two real-world titles above plus multiple pipes in url+title, redirect target handling, empty-title fallback, and a malformed percent-sequence.
- `npx mocha` passes locally (66 passing, 0 failing).
- `npx eslint www/js/lib/zimDirEntry.js tests/unit/spec/zimDirEntry.test.js` is clean.
- Manually downloaded `hsm.stackexchange.com` and a small Zimit archive (`tonedear.com`) from download.kiwix.org, located real entries containing `|` in their titles, and round-tripped their actual `url`/`title` through the exact app.js pipeline before and after this fix, confirming the corruption before and the fix after.

## Origin

This was originally reported and fixed against kiwix-js-pwa (kiwix/kiwix-js-pwa#966), but per that repo's CONTRIBUTING.md, fixes to shared app code belong here first — this PR on-ports the fix, with the test suite in its proper home.

Signed-off-by: bhuvan-somisetty <somisettybhuvan5@gmail.com>


Jaifroid (changes requested):
Thanks for this, @bhuvan-somisetty. The diagnosis is right, and encoding url and title individually is a clean fix. I checked the callers: the id is only built in `buildArticleListElement` and only read back through `parseDirEntryId()`, and nothing else splits it on `|`, so the change is well contained. As a bonus, an undefined title now round-trips as `''` rather than the string `"undefined"`, so `getTitleOrUrl()` can fall back to the url correctly.

Two small things before we merge:

1. The comment on the fallback in `safeDecode()`

The PR description says the try/catch is there to tolerate "legacy unencoded ids already stored". These ids are never persisted, though. They only live in the `dirEntryId` attributes of the article list, which is rebuilt on every search. Every id reaching `fromStringId()` now comes from the new `toStringId()`, and `decodeURIComponent()` can't fail on output from `encodeURIComponent()`, so the catch can't be reached in practice.

I'm happy to keep it as a cheap defensive guard. Could you add a short comment in `safeDecode()` saying that, so a later reader doesn't go looking for stored legacy ids that don't exist?

2. Console noise in the malformed-sequence test

The last test prints two `Could not decode DirEntry field, using raw value: URIError: URI malformed` warnings with full stack traces in the CI log ([example](https://github.com/kiwix/kiwix-js/actions/runs/35213565458/job/105196618289)). They're expected, but they look like real failures when someone scans the log. Could you:

* stub `console.warn` for that test, assert that it was called, and restore it afterwards
* call `fromStringId()` once instead of twice (once inside `expect(...).to.not.throw()`, then again for the assertion), e.g. assign the result inside the function passed to expect and check parsed.url afterwards

Once those are in, this looks good to me.

Jaifroid:
Minor: it might be worth noting in the test comment that `Contact | Ear Training` also appears in our own `tests/zims/tonedear` test archive, so the example can be checked without downloading anything.

bhuvan-somisetty:
Thanks for the careful review! Pushed a follow-up commit addressing all three points:

- Added a comment in `safeDecode()` explaining the catch is just a defensive guard, not a real recovery path (ids aren't persisted, so `decodeURIComponent()` can't actually fail on them in practice).
- Stubbed `console.warn` in the malformed-sequence test and call `fromStringId()` once instead of twice, so the CI log stays clean.
- Noted that the "Contact | Ear Training" title also lives in our own `tests/zims/tonedear` archive, so it's checkable without downloading anything.

Should be good to go now.

tests/unit/spec/zimDirEntry.test.js
```diff
@@ -0,0 +1,119 @@
...
+    });
+
+    it('does not throw on a malformed percent-encoded sequence and falls back to the raw text', function () {
+        const warnStub = sinon.stub(console, 'warn');
+        const malformedStringId = '100|0|C|10|1|%ZZmalformed|plain title|false|undefined';
+        let parsed;
+        expect(function () { parsed = DirEntry.fromStringId(fakeZimFile, malformedStringId); }).to.not.throw();
+        expect(parsed.url).to.equal('%ZZmalformed');
+        expect(warnStub.called).to.equal(true);
+        warnStub.restore();
```
Jaifroid:
```suggestion
        try {
            const malformedStringId = '100|0|C|10|1|%ZZmalformed|plain title|false|undefined';
            let parsed;
            expect(function () { parsed = DirEntry.fromStringId(fakeZimFile, malformedStringId); }).to.not.throw();
            expect(parsed.url).to.equal('%ZZmalformed');
            expect(warnStub.called).to.equal(true);
        } finally {
            warnStub.restore();   // ← always runs, pass or fail
        }
```
If one of the assertions fails, `restore()` is never reached and `console.warn` stays stubbed for the rest of the run, hiding later warnings. Wrapping it in try/finally avoids that. This is in case future tests are added after this one.

Jaifroid (approved):
This looks great now, with one further suggestion to avoid the stub being in place still if the assertion fails, and there are further tests added after in the future. Thank you very much for this PR. When you've done that one suggestion, let me know, and I can squash/merge. You could then (if you like), on-port the final code downstream.

bhuvan-somisetty:
Good catch, done - wrapped the restore in try/finally so it always runs even if an assertion fails. Tests still pass (66/66). Go ahead and squash/merge whenever you're ready, thank you!

